PR #109, opened 2017-04-07
Added verbiage to add action re encryption

joescars:
added in verbiage to the add settings action letting the user know that the setting will be encrypted and the file will not be portable.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/AddSettingAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/AddSettingAction.cs
@@ -9,7 +9,7 @@ using Azure.Functions.Cli.Common;
 
 namespace Azure.Functions.Cli.Actions.LocalActions
 {
-    [Action(Name = "add", Context = Context.Settings, HelpText = "Add new local app setting to appsettings.json")]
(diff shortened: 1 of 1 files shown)


ahmelsayed:
Maybe say something like "**If** settings are encrypted, they won't be portable"? @lindydonna what do you think?

lindydonna:
@ahmelsayed Agreed, and I would also add a little more detail: **If settings are encrypted, they can only be decrypted on the current machine**.

joescars:
I like that it is more concise but do we want to tell the user they will be encrypted by default?

lindydonna:
@joescars Agreed. Maybe, "Settings are encrypted by default. If encrypted, they can only be decrypted on the current machine."

joescars:
Perfect. I'll update it and push a new commit

lindydonna:
Awesome!  Please also squash your commits (and do a force push) so the history is clean. If your git-fu fails you, we can fix it up for you.

SteveALee:
@lindydonna  did you know GitHub lets you squash when you merge the PR?

joescars:
@lindydonna all set. squashed commits as well.

lindydonna:
@SteveALee Yes, but it doesn't let you rebase and squash at the same time, which is how @ahmelsayed likes to do it.

PR #120, opened 2017-04-17
Resolves Null Ref Error on settings. Closes 113

joescars:
Added in description for settings -ShowValue to resolve null reference exception. Closes 113


ahmelsayed:
Thanks @joescars!

Issue #112, opened 2017-04-11
func azure command tries to log in to azure


lindydonna:
Func azure used to just print out the help text. Now it tries to log in. If I close the login window, I get the following exception:

> An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = ArmManager (ReflectionActivator), Services = [Azure.Functions.Cli.Arm.IArmManager], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime, Sharing = None, Ownership = OwnedByLifetimeScope ---> An exception was thrown while invoking the constructor 'Void .ctor(ARMClient.Authentication.IAuthHelper, ARMClient.Library.IAzureClient, Azure.Functions.Cli.Interfaces.ISettings)' on type 'ArmManager'. ---> One or more errors occurred. (See inner exception for details.) (See inner exception for details.)

ahmelsayed:
crap, right... This is a regression from 5f3978f1dadcc7052e693168dac226ee375e9f65. I'll have to think about it.

PR #161, opened 2017-06-14
allow for explicit login credentials

### LINKED ISSUE #160: Supply login credentials through the CLI
It doesn't look like it is possible to specify login credentials as parameters to the CLI instead of popping the OAuth sign in page. I would like to be able to deploy function apps via CI, but if interactivity is required then that won't be possible. Something like:

```
$ func azure login -u myUsername -w myPassword
```

jaredcnance:
Closes #160 

If you guys are cool with this, let me know and I'll get some tests written.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -1,7 +1,5 @@
-﻿using System;
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
 using Azure.Functions.Cli.Arm;
-using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Interfaces;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
@@ -16,15 +14,6 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         }
 
         public async Task Initialize()
-        {
-            try
-            {
-                await _armManager.Initialize();
-            }
-            catch (Exception e)
-            {
-                throw new CliException("Error during login to azure. Please verify your internet connection or try again later.", e);
-            }
-        }
+        { }
     }
 }
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -7,13 +7,20 @@ namespace Azure.Functions.Cli.Actions.AzureActions
     abstract class BaseAzureAction : BaseAction, IInitializableAction
     {
         protected readonly IArmManager _armManager;
+        private readonly bool _requiresLogin;
 
-        public BaseAzureAction(IArmManager armManager)
+        protected BaseAzureAction(
+            IArmManager armManager,
+            bool requiresLogin = true)
         {
             _armManager = armManager;
+            _requiresLogin = requiresLogin;
         }
 
         public async Task Initialize()
-        { }
+        {
+            if(_requiresLogin)
+                await _armManager.Initialize();
+        }
     }
 }
--- a/src/Azure.Functions.Cli/Actions/AzureActions/LoginAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/LoginAction.cs
@@ -2,6 +2,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Arm;
+using Azure.Functions.Cli.Helpers;
 using Azure.Functions.Cli.Interfaces;
 using Fclp;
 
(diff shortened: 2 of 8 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -16,15 +14,6 @@ public BaseAzureAction(IArmManager armManager)
         }
 
         public async Task Initialize()
-        {
-            try
-            {
-                await _armManager.Initialize();
```
jaredcnance: Under the hood, this calls `ArmManager.SelectTenantAsync`, which calls `_authHelper.GetToken` and `_authHelper.AcquireTokens()` which I don't think makes much sense if there is a `login` command. this should also resolve #112

src/Azure.Functions.Cli/Actions/AzureActions/LoginAction.cs
```diff
@@ -1,22 +1,37 @@
...
+            return base.ParseArgs(args);
+        }
+
         public override async Task RunAsync()
         {
-            await _armManager.LoginAsync();
+            if (string.IsNullOrWhiteSpace(_username))
+                await _armManager.LoginAsync();
+            else
+                await _armManager.LoginAsync(_username, _password);
```
jaredcnance: may want to prompt if the username is provided but the password is empty?

ahmelsayed:
I'm very sorry, some how I missed the notification for this PR, and just saw it now. I'll take a look at it today.

src/Azure.Functions.Cli/Actions/AzureActions/LoginAction.cs
```diff
@@ -1,23 +1,63 @@
...
+            var password = new StringBuilder();
+            while (true)
+            {
+                var key = Console.ReadKey(true);
+                if (key.Key == ConsoleKey.Enter)
+                    break;
+                password.Append(key.KeyChar);
+            }
+            Console.Write(Environment.NewLine);
+            return password.ToString();
```
ahmelsayed: There is [`SecurityHelpers.ReadPassword()`](https://github.com/Azure/azure-functions-cli/blob/master/src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs#L73) which is similar to this method.

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
ahmelsayed: the reason for having this is to prompt to login if the user never actually called `func azure login`, or if the cached token has expired. otherwise everytime you run any other command it'll fail until you run `login` again.

src/Azure.Functions.Cli/Actions/AzureActions/LoginAction.cs
```diff
@@ -1,23 +1,63 @@
...
 
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser.Setup<string>('u')
+                .WithDescription("username")
+                .Callback(username => _username = username);
+
+            Parser.Setup<string>('w')
+                .WithDescription("password")
+                .Callback(password => _password = password);
```
ahmelsayed: Can you add a long option for these as well for consistency with other flags?

Maybe the description can be `Username to use for non-interactive login. Note that accounts with 2 factor-auth require interactive login. Default: interactive` and `Password to use for non-interactive login only in conjunction with -u. Default: prompt`

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
jaredcnance: added an option `requiresLogin` param to the constructor.

jaredcnance:
@ahmelsayed any thoughts on the updates?

ahmelsayed:
@jaredcnance sorry about the delay. I fixed my notifications for this repo which weren't making it to my email for some reason.

Sorry I don't mean to keep pushing back, but why not just make `LoginAction` not drive from `BaseAzureAccountAction` but drive from `BaseAction` directly? The only thing `BaseAzureAccountAction` is buying you is also being an `IInitializableAction` so that we can call login, and that's what `LoginAction` is avoiding with the `requireLogin` flag, right?

jaredcnance:
pushback is good 😄 

the only reason is so it can call `BaseAzureAccountAction.PrintAccountsAsync()`

ahmelsayed:
ah, you're right. I forgot about that. Your change actually makes more sense now as login is just one form of initialization.

ahmelsayed:
Merged - Thank you very much for your contributions, @jaredcnance :)

ahmelsayed:
https://github.com/Azure/azure-functions-cli/releases/tag/0.100

Issue #143, opened 2017-05-24
Remove 'func functionapp' command


lindydonna:
It has only init, so we can remove it.

PR #169, opened 2017-06-29
Removed 2 failing tests that were not needed 

dariuszparzygnat:
I found tests that were failing because of removing func functionapp context from InitAction #143.


ahmelsayed:
Thanks @dariuszparzygnat

PR #235, opened 2017-09-11
Add worker logs to cli output

mamaso:
@ahmelsayed I want to get better worker diagnostics in the console, this can help in the interim.

### DIFF
--- a/src/Azure.Functions.Cli/Diagnostics/ConsoleLoggerFactoryBuilder.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/ConsoleLoggerFactoryBuilder.cs
@@ -22,7 +22,7 @@ namespace Azure.Functions.Cli.Diagnostics
             }
 
             // TODO: Get trace level, add console logger
-            // scriptConfig.HostConfig.LoggerFactory.AddColoredConsole((n, l) => l >= LogLevel.Information, false);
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Diagnostics/ConsoleLoggerFactoryBuilder.cs
```diff
@@ -22,7 +22,7 @@ public void AddLoggerProviders(ILoggerFactory factory, ScriptHostConfiguration s
             }
 
             // TODO: Get trace level, add console logger
-            // scriptConfig.HostConfig.LoggerFactory.AddColoredConsole((n, l) => l >= LogLevel.Information, false);
+            scriptConfig.HostConfig.LoggerFactory.AddColoredConsole((cat, lvl) => cat.StartsWith("Worker."), false);
```
ahmelsayed: shouldn't it be an `||` of the 2 conditions? or is `.startsWith("Worker.")` includes the other one as well?

mamaso: That sounds reasonable. I was whitelisting only worker stuff, but more logs of all kinds should be helpful at this stage

ahmelsayed: oh I just realized that the whole line was commented. I thought you just changed the condition. I merged it as is and we can add that latter. I'll make a release shortly

PR #244, opened 2017-09-21
Enable worker targets on osx/linux

mamaso:
If a .targets file is in the worker directory, we'll execute the logic on install.

This enables consistent install logic between VS project & cli.

Example worker target which installs the correct grpc native module for platform: https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/Microsoft.Azure.Functions.NodeJsWorker.targets


ahmelsayed:
:shipit:

PR #246, opened 2017-09-22
Initial checkin for java debugging.

jeffreyCline:


mamaso:
Thanks @jeffreyCline for the good work, really appreciate you taking a look at this.

Removed most of the switch statement code to try to keep the cli code language agnostic. Added both launch configurations as defaults. Passed IConfiguration style arguments to the IConfigurationBuilder, which allows the `LanguageWorkerConfigs` in the host to consume the configuration.

In order to specify debug, you now need to call:

`func host start /worker:java:debug=(true || port#)`

`func host start /workers:node:inspect=(true || port#)`

We could normalize these options (perhaps add `debug` to node as well as `inspect` with the same behavior).

It might be nice to provide backwards compatibility with the previous node debug experience.

I'm working on a host update to consume IConfiguration in the language workers, see here:
https://github.com/mamaso/azure-webjobs-sdk-script/commit/ce2936c4b8d82cc1c59f2367decd228beaa5748d

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -105,11 +99,7 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .WithDescription("to use with --cert. Either the password, or a file that contains the password for the pfx file")
                 .Callback(p => CertPassword = p);
 
-            Parser
-                .Setup<DebuggerType>("debug")
-                .WithDescription("Default is None. Options are VSCode and VS")
-                .SetDefault(DebuggerType.None)
-                .Callback(d => Debugger = d);
+            IConfigurationArguments = args.Where(arg => arg.StartsWith('/')).ToArray();
```
ahmelsayed: @mamaso was there a merge issue with your change maybe? I think --debug is needed and it seems that @jeffreyCline had updated it in his commit, but it's not here for some reason.

ahmelsayed: ignore. I'm not too smar

mamaso:
closed in #247

PR #259, opened 2017-09-27
Reenable launch.json creation

mamaso:
Could determine debug ports if specified in `--workers` for launch.json, but I think that's out of scope for now, created https://github.com/Azure/azure-functions-cli/issues/258 to track

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -56,8 +56,8 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         public IDictionary<string, string> IConfigurationArguments { get; set; } = new Dictionary<string, string>()
         {
-            ["node:debug"] = "true",
-            ["java:debug"] = "true"
+            ["node:debug"] = "9229",
+            ["java:debug"] = "5005"
         };
 
         public StartHostAction(ISecretsManager secretsManager)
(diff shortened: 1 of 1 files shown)


nelak:
@mamaso I saw this yesterday while looking at the code but I didn't understand what it was supposed to be doing, does this allow to debug the cli, the worker or the functions?
From what I saw each function gets assigned a random inspect port, but it would be nice if they could be setup through the host.json or function.json so that they don't change between runs.

mamaso:
This creates launch.jsons which allow debugging user / language worker code (in vscode it's as simple as setting a breakpoint)

The inspect ports are not random - for node the default is `5858`, but you can specify a different port via `--workers node:debug=<port>`

nelak:
You are right, I confused the debugger port with the `nodejsWorker.js --port` which is indeed random.
Is it possible to change the default port to 9229? 
node --inspect defaults to `9229`, `5858` is the port for the legacy debugger.

mamaso:
Yep, good call - should be a quick fix.

nelak:
I was looking at this as I have some concerns when trying to debug multiple functions running in the same host with VSCode as I haven't seen any options to pass the uuid part of the address when launching the debugger. From the node inspect documentation. : https://nodejs.org/en/docs/guides/debugging-getting-started/
`Inspector clients must know and specify host address, port, and UUID to connect to the WebSocket interface. The full URL is ws://127.0.0.1:9229/0f2c936f-b1cd-4ac9-aab3-f63b0f33d55e, of course dependent on actual host and port and with the correct UUID for the instance.`
I'll see what I can find related to this and correct me if I'm wrong but it looks likely that for this scenario end up needing to use Chrome DevTools

mamaso:
We don't start a node process in debug mode per function - instead, we load multiple functions in a single node worker process. If that process is started with debugging, you should be able to debug both functions as they are executed.

Is there something else that the UUID enables?

nelak:
@mamaso The UUID is as a security feature of `node --inspect` it's highly likely that vscode is resolving this by using the endpoint listed in the node inspect documentation.
I now have a better understanding of what caused my confusion regarding this, I'll create a new in issue in the worker tracker to follow up.

PR #263, opened 2017-09-29
Add --script-root option to host start

### LINKED ISSUE #264: Add option to specify ScriptRoot for func host start action
This will allow running the functions host from other directories than cwd, which is useful for tooling scenarios where artifacts do not live in the same directory as the source code.

mamaso:
resolves #264

@ahmelsayed unfortunately this is more hacky than I want it to be.

The main issue is that ISecretsManager is created via the Autofac container registry, but I need to configure it with the proper script root path to search for AppSettingsFilePath.

With the way the actions are set up, I don't know that there's a good way to use autofac's lambdas or provider patterns in order to supply the scriptRoot argument to the secrets manager - I'd love any input you have on that.

One possibility is that we could ignore ISecretsManager altogether and use local.settings.json that exists on the cwd, not the ScriptRoot. This might be more reasonable, will revisit tomorrow.

TODO: reapply defaults from hostSettings

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/AddStorageAccountSettingAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/AddStorageAccountSettingAction.cs
@@ -57,7 +57,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 var name = $"{storageAccount.StorageAccountName}_STORAGE";
                 _secretsManager.SetSecret(name, storageAccount.GetConnectionString());
                 ColoredConsole
-                    .WriteLine($"Secret saved locally in {SecretsManager.AppSettingsFileName} under name {ExampleColor(name)}.")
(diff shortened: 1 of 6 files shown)


ahmelsayed:
yeah I forgot about that :/

I can think of 2 potential ways, but I'm not sure how viable they are. one way is to intercept that argument early in [here](https://github.com/Azure/azure-functions-cli/blob/master/src/Azure.Functions.Cli/ConsoleApp.cs#L37) similar to how `--pause-on-error` is handled. Then actions would just ignore it as the parser ignores arguments that haven't been setup. (_you may need to make the `container` an `Action<IContainer>` if autofac gives you too much grief_)

the other option I could think of is to maybe have an event that secretsManager (and other objects that care about workingDir) subscribes to, then any action that wants to change the workingDirectory should fire it before it does. however since events are asynchronous, not sure how to avoid potential race conditions.


I'm leaning more towards the first approach, as you'd be handling that before any of the complicated logic of `ConsoleApp` kicks in, so everything else should continue to work as is. What do you think?

mamaso:
closed in favor of #266

PR #274, opened 2017-10-05
dependency added

johnpapa:
link to the .net core tools

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ To install globally:
 npm i -g azure-functions-core-tools
 ```
 
+### Dependencies
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,10 @@ To install globally:
 npm i -g azure-functions-core-tools
 ```
 
+### Dependencies
+
+There is a dependency on the .NET Core tools for the cross platofrm support. You can [install these here](https://www.microsoft.com/net/core).
```
timheuer: Typo: "platofrm" should be "platform"

johnpapa:
fixed typo as per @timheuer

Issue #255, opened 2017-09-26
Allow to install azure-functions-core-tools as dependency


nelak:
It would be nice if it were possible to install azure-functions-core-tools globally and as dependency.

Currenly it's not possible to install azure-functions-core-tools as dependency since in main.js
https://github.com/Azure/azure-functions-cli/blob/core/src/Azure.Functions.Cli/npm/lib/main.js#L14
binary lookups are always performed on `path.join(os.homedir(), '.azurefunctions', 'bin')`

Additionally if you add the package as dependency and then remove when the uninstall script runs it will remove the global binaries breaking the global install.

SteveALee:
This is a must have. Has there been a recent breaking change as ive been running as dev dep for sometime without funding any issues. Have never uninstalled though.

https://github.com/OpenDirective/brian/blob/master/package.json#L80

SteveALee:
Yes a breaking change. https://github.com/Azure/azure-functions-cli/commit/9a2d80a9b88d2faf6f9902b734e98d3b5e684f5c

SteveALee:
Sorry I missed the email was refering to the core / branch. 
Un helpfully GitHub issues to not specify a branch :(

Still, local and global installs must be supported as is npm standard. Plus I have policy of never globally installing any dep - otherwise `npm install` has to be supplimented with instructions in a readme.

janaagaard75:
I can added that this is still an issue in version `2.0.1-beta.22` when using the `publish` command. I get errors similar to the one below, when I run `yarn func azure functionapp publish powc`. It varies a bit what dll is causing the error. I get the error regardless of using npm or Yarn, macOS or Windows. Using Node version 8.9.4.

Installing `azure-functions-core-tools@core` globally fixes this issue.

> The process cannot access the file 'C:\Projects\poweroffice\powc\node_modules\azure-functions-core-tools\bin\AccentedCommandLineParser.dll' because it is being used by another process.
Retry: 1 of 2

jgwinner:
I know this is old, but I'd like to bump this. I had a developer who under-documented prereq's, left, and I'm trying to jump in. Was rather baffled by the error:

```
$ yarn prestart
yarn run v1.22.17
$ yarn run build && func extensions install
$ ./node_modules/.bin/ttsc -p ./
/bin/sh: 1: func: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

[...]

PR #284, opened 2017-10-19
Making cli install relative to node_modules

nelak:
This PR should address #255 and make the install relative to the `node_modules` folder either globally or locally. The only caveat when using as a dependency and calling through yarn or npx is that `process.cwd()` will default to the `package.json`  root folder.
The uninstall script has been removed as binaries will be handled by npm and `.azurefunctions` should only hold config after this change.


ahmelsayed:
@nelak sorry for the long delay, I missed the notification on this PR and never checked. 
Can you change the target branch to `core` instead of `master`?

nelak:
@ahmelsayed Next time I'll tag you then
I'm thinking the next change would be to add some kind of progress indicator as installing may take a while on slower connections and you are left without any feedback of what's going on

PR #306, opened 2017-11-06
Add missing $ in the template string

jdneo:
Add missing $ in the template string


ahmelsayed:
Thank you @jdneo for your contribution :)

PR #311, opened 2017-11-09
Update to latest runtime version of WebJobs.Script

safihamid:
Update to latest runtime version of WebJobs.Script which includes the proxy dlls for running locally


ahmelsayed:
the change looks good; however, can you make sure to run [redirectCompare](https://github.com/Azure/azure-functions-cli/tree/v1.x/tools/redirectCompare) to make all the redirection in app.config is correct? 
It's a weird step, but the runtime packages don't have proper nuget dependency list and whenever a new one is added there, it'll break here without this

PR #333, opened 2017-12-05
Update readme to reflect installation for non-windows platforms

sdras:
Per Functions/CDA team meeting today to make this command more discoverable for Macs

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To install globally for non-Windows platforms:
 ```
 npm i -g azure-functions-core-tools@core
 ```
-This installs a higher beta version of the cli, so it is highly recommended that while using the core tag, you enable the `beta` runtime in function app settings, otherwise you may not see the same results as running locally.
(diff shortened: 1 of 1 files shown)


sdras:
cc @jeffhollan @eduardolaureano @simonaco

nelak:
@sdras My 2cents the `core` tag installs the Azure functions cli 2.0.1-beta.22, while for windows it'll use the `latest` tag installing version 1.0.7. 
For users using the `core` tag it should be recommend that they enable the `beta` runtime in function app settings, or they may not see the same results as running locally.

sdras:
@nelak ah, interesting, I wasn't aware of this, thank you for the clarification. I'll update it now.

sdras:
Ok, updated. Let me know if you think that clarifies it. Thanks

nelak:
Looks good to me :)

jeffhollan:
Thanks @sdras - looks good to me as well. Can you also just add a link to the app setting tag to go to [this document](https://docs.microsoft.com/en-us/azure/azure-functions/functions-versions#target-the-version-20-runtime) which explains how to opt-in to beta version?  

Also @ahmelsayed will this update the [npm landing](https://www.npmjs.com/package/azure-functions-core-tools) page as well automatically?

sdras:
Good call, @jeffhollan! Done. Please let me know if you need any further changes, or ready to merge if not.

Issue #298, opened 2017-10-31
Mark all preview/experiment languages as such in `func new` experience


jeffhollan:
Python, PowerShell, Batch, Bash, Typescript (anything that's not C#, F#, or Node) should have (Experimental) next to it on `func new` experience

mikhailshilkov:
Seems to work fine now in v1. Should we close?

PR #349, opened 2018-01-13
Mark experimental languages and templates in `func new` dialog

mikhailshilkov:
Implements #298
- Displays " (Experimental)" next to each template which is assigned to category experimental
- Displays " (Experimental)" next to each language for which all templates are experimental

The change is for 1.x branch, since 2.0 branches don't have experimental templates at all.

Note that PHP and TypeScript are not classified as experimental at the moment due to https://github.com/Azure/azure-webjobs-sdk-templates/issues/619. If/when that fix is merged, these languages will be shown as experimental with no further change required in CLI.


ahmelsayed:
thanks @mikhailshilkov !

PR #356, opened 2018-01-17
updating based on ILogger changes to Script

brettsam:
PR https://github.com/Azure/azure-webjobs-sdk-script/pull/2182 removed all internal usage of `TraceWriter` in v2 and changed some internal types. This absorbs those changes in the CLI.


src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private readonly Func<string, LogLevel, bool> _filter;
+        private readonly string _category;
+
+        public ColoredConsoleLogger(string category, Func<string, LogLevel, bool> filter)
+        {
+            _category = category;
+            _filter = filter;
+            ColoredConsole.Out = new ColoredConsoleWriter(Console.Out);
+        }
```
ahmelsayed: Is this needed? the other class was doing this to then null out the default Console.Out to prevent anyone else from writing to it.

ahmelsayed:
btw, I updated the PR to merge into master since master is now even with core.

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
brettsam: No, it's not needed.

brettsam:
I was just about to ask that -- I wasn't sure which branch I was supposed to push this to. I'll remove that line, squash, then get this in. Thanks!

ahmelsayed:
I just set master head to core head yesterday :)

PR #376, opened 2018-02-11
Fix install command for non windows platform

fumiya-kume:
before command cannot install in macOS.
so fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,13 @@ To install globally for non-Windows platforms:
 ```
 npm i -g azure-functions-core-tools@core --unsafe-perm true
 ```
+
+>When installing on macOS and Linux, you may need to include the unsafe-perm flag, as follows:
+
+[Code and test Azure Functions locally](https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local)
+
+**NOTE**: That's due to npm behavior of post install script. if you sudo the npm command, which is the default on macs and linux, you most likely need --unsafe-perm, unless you have also set the npm user.
+
 This installs a higher beta version of the cli, so it is highly recommended that while using the core tag, you [enable the `beta` runtime](https://docs.microsoft.com/en-us/azure/azure-functions/functions-versions#target-the-version-20-runtime) in function app settings, otherwise you may not see the same results as running locally.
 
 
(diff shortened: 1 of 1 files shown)


wilsonge:
I can use the existing command to install for mac and do not need this

fumiya-kume:
Thanks for review.
In this document write it.
https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local

>When installing on macOS and Linux, you may need to include the `unsafe-perm` flag, as follows:
>```bash
>sudo npm install -g azure-functions-core-tools@core --unsafe-perm true
>```

which one is correct?

wilsonge:
It does say `may` :P I'm not sure what the condition is (maybe a microsoft employee can clarify) - but it's definitely not a requirement for me. I should add I guess that maybe the difference is I don't run it with sudo - I just use 

`npm install -g azure-functions-core-tools@core`

ahmelsayed:
That's due to npm behavior of post install script. if you `sudo` the npm command, which is the default on macs and linux, you most likely need `--unsafe-perm`, unless you have also set the npm user. 

Basically by default when you `sudo npm`, it only runs its own code in sudo, and any postinstall script with `nobody` user by default. You can either change npm to not require sudo for global packages, or change the default user it uses after install to something other than `nobody` to allow postinstall script to have an id. 

[`--unsafe-perm`](https://docs.npmjs.com/misc/config#unsafe-perm) tells npm to run the postinstall script as sudo as well. Something you probably don't wanna do lightly if you're installing a package you don't completely trust. 

Since this cli supports running javascript functions, but is not itself javascript, and can be a bit large, I host the actual content on a CDN and [download it a postinstall script](https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/npm/lib/install.js#L22) and for that script to be able to put the files in the default npm dir it needs to have access to it. 
I hope that makes sense. It took me a while to figure it out myself since it works differently on Windows. 

Thanks @fumiya-kume for your contribution. Would you mind also adding the bit from the docs also on top of the command as part of this PR 

>When installing on macOS and Linux, you may need to include the `unsafe-perm` flag, as follows:

fumiya-kume:
@ahmelsayed Thanks comments.
I added bit from the doc.

ahmelsayed:
thanks @fumiya-kume!

PR #384, opened 2018-02-16
updating DefaultConnectionLimit

brettsam:


ahmelsayed:
do we also need this in v2? or does that work differently?

brettsam:
It'll work differently in v2. I believe the new API is [HttpClientHandler.MaxConnectionsPerServer](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.maxconnectionsperserver?view=netcore-2.0#System_Net_Http_HttpClientHandler_MaxConnectionsPerServer). You shouldn't have to increase it yourself anymore as the default value is already `Int32.MaxValue`.

aderderian:
@brettsam if we used to use ServicePointManager.DefaultConnectionLimit = Int32.MaxValue; in webjobs, do we need to do anything in Functions V2 to throttle this up as well? I am using an App Tier plan of Standard. We have a socket listener with a good amount of data coming in that we then batch messages to Event Hubs, I want to make sure I am not maxing out anything and limiting my scale here. Thx!

PR #359, opened 2018-01-24
First pass at fixing connection strings

wilsonge:
Some sort of attempted fix for #348 and #214 (basically connection strings are not imported with the same environment variable names as the Azure Function Apps)

This naming convention seems to be derived from the one described in the following blog post: https://blogs.msdn.microsoft.com/silverlining/2012/10/23/getting-database-connection-information-in-windows-azure-web-sites/

I've never coded in csharp so this is slightly speculative coding (so please code review/test hard)

Running tests gave me:

```
$ dotnet test
/<root_path>/azure-functions-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : warning NU1701: Package 'Microsoft.AspNet.WebApi.Client 5.2.2' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. [/<root_path>/azure-functions-cli/Azure.Functions.Cli.sln]
/<root_path>/azure-functions-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : warning NU1701: Package 'Microsoft.Azure.AppService.Proxy.Client.Contract 0.3.1.1' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. [/<root_path>/azure-functions-cli/Azure.Functions.Cli.sln]
Build started, please wait...
/<root_path>/azure-functions-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : warning NU1701: Package 'Microsoft.AspNet.WebApi.Client 5.2.2' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
/<root_path>/azure-functions-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : warning NU1701: Package 'Microsoft.Azure.AppService.Proxy.Client.Contract 0.3.1.1' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
Build completed.

Test run for /<root_path>/azure-functions-cli/src/Azure.Functions.Cli/bin/Debug/netcoreapp2.0/Azure.Functions.Cli.dll(.NETCoreApp,Version=v2.0)
[...]

Issue #348, opened 2018-01-12
Inconsistent handling of connection strings


anoff:
It seems that the `local.settings.json` generated by `func azure functionapp fetch-app-settings` creates inconsistency when running the function local or on azure. I defined a Redis Cache String on the Azure UI
![image](https://user-images.githubusercontent.com/7142618/34864128-03894322-f773-11e7-9df5-e8e94f811633.png)

When I run my function using `func start` locally and inspect the (node.js) environment variables I find **ConnectionStrings:RWD_CACHE**. I also note that in the local settings it shows with _System.Data.SqlClient_ as _ProviderName_.

When inspecting `process.env` in Azure I see the connection string as **REDISCACHECONNSTR_RWD_CACHE**

Is this the current behavior or am I missing something? Would a valid workaround be using just application settings instead?

```
Azure Functions Core Tools (2.0.1-beta.22)
Function Runtime Version: 2.0.11415.0
node v8.9.3
```

liliankasem:
Closing as stale.

Issue #214, opened 2017-08-20
ConnectionStrings are not imported into local environment from local.settings.json


gunzip:
According to the docs here: 
https://docs.microsoft.com/en-us/azure/app-service-web/web-sites-configure

> For PHP, Python, Java and Node applications, [ConnectionStrings] settings will be available as environment variables at runtime, prefixed with the connection type.  [...]
> For example, if a MySql connection string were named connectionstring1, it would be accessed through the environment variable MYSQLCONNSTR_connectionString1.
> [...]

But when running locally, *ConnectionStrings* variables are **not** imported into environment 
using local.settings.json:

```
UpdateEnvironmentVariables(getSecrets(), ...)

getSecrets() ->
var appSettingsFile = new AppSettingsFile(AppSettingsFilePath);
            return appSettingsFile.GetValues();
```

only the AppSettings -> Values part gets imported so using NodeJS you cannot access
```
process.env.<my connection strings>
```
from application code.

ahmedelnably:
Closing as stale

PR #385, opened 2018-02-16
Fix Connection Strings Logic

wilsonge:
Some sort of attempted fix for #348 and #214 (basically connection strings are not imported with the same environment variable names as the Azure Function Apps)

This naming convention seems to be derived from the one described in the following blog post: https://blogs.msdn.microsoft.com/silverlining/2012/10/23/getting-database-connection-information-in-windows-azure-web-sites/

I've never seriously coded in csharp so this is slightly speculative coding - but I've tested it and in itself it works :)

I'm very confident the Type of connection is a string rather than integer. But wasn't very sure if the provider name should map to this property or something else.

```
$ az webapp config connection-string list --name <func_name> --resource-group <rs_group>
[
  {
    "name": "ConnectionStringName",
    "slotSetting": false,
    "value": {
      "type": "SQLAzure",
      "value": "<connectionString>;"
    }
  }
]
```

This was a redo of #359 in an attempt to force appveyor to run tests on it


khkh-ms:
The PR has been open for some time and has lots of conflicts. Closing it now. Please open it if needed.

PR #390, opened 2018-02-24
New paket bootstrapper that works around github TLS disaster

forki:
Since Github disabled SSL 1.0 we need to update the paket bootstrappper


ahmelsayed:
Yep, just hit this now :) Thanks @forki for the heads up!!. 

I verified the SHA2 of the file in the PR to be `546f32e6d1887716e133cb06fb9c0e95687372b0a27b6217caad72d2ac87fadc`  same as the the one in https://github.com/fsprojects/Paket/releases/tag/5.145.1

PR #403, opened 2018-03-13
Fix chmod in post install logic

snobu:
Octal notation required, otherwise results in bad permissions, no error is thrown -- https://nodejs.org/api/fs.html#fs_file_modes


ahmelsayed:
Thanks @snobu !!

PR #411, opened 2018-03-19
Add node_modules to .gitignore

burkeholland:
It looks like our .gitignore files are not project specific (i.e. C# vs JavaScript), but we can add `node_modules` to the generated .gitignore to make sure that it's there no matter what.


ahmelsayed:
Thanks @burkeholland

PR #437, opened 2018-04-16
adding console output for ILogger

### LINKED ISSUE #130: Logging written to the ILogger does not appear in the CLI
Logging written to the ILogger does not appear in the CLI

brettsam:
Fixes #130.

Had to handle some special cases as simply adding a ConsoleLoggerProvider wouldn't work:
1. Since we set `Console.Out` to null, no console output could come from elsewhere.
2. The standard console logger doesn't look anything like our existing logging, so I wanted to use ConsoleTraceWriter internally so everything looked nice.
3. TraceWriter traces are forwarded to ILoggers to make App Insights work as expected, but this can cause duplicate logs. Making sure this doesn't happen.
4. We log to both TraceWriter and ILogger everywhere in the host, so I've restricted this output to be strictly for function logs.

### DIFF
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -739,6 +739,9 @@
     <Compile Include="Common\Utilities.cs" />
     <Compile Include="ConsoleApp.cs" />
     <Compile Include="Context.cs" />
+    <Compile Include="Diagnostics\CliLoggerFactoryBuilder.cs" />
+    <Compile Include="Diagnostics\ColoredConsoleLogger.cs" />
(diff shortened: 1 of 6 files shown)


pragnagopa:
@brettsam - with this change can we eliminate setting [host:logger:consoleLoggingMode](https://github.com/Azure/azure-functions-docker-python-sample/blob/master/Dockerfile#L3)?

brettsam:
No, this change is only for v1. The v2 runtime has some different requirements where we want to selectively enable the console logger. For the CLI, it is already enabled, but some scenarios need it during runtime. We'll come up with a nicer way to do that soon.

MuhammadBilalYar:
@ahmelsayed in which CLI version these changes are included.

ahmelsayed:
It should be in 1.0.12

Issue #433, opened 2018-04-12
Support func --version


asavaritayal:

ahmelsayed:
done in cf4f7b808488bf3eba07047d17b8e16508db5285

PR #440, opened 2018-04-18
add --version, #433

watashiSHUN:
image
the change is a little raw...though


ahmelsayed:
@watashiSHUN  actually can you fix this test case https://github.com/Azure/azure-functions-core-tools/blob/cf4f7b808488bf3eba07047d17b8e16508db5285/test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs#L47-L49 

It's expecting to get `HelpAction` from the parse function

PR #448, opened 2018-04-28
Update Nuget feeds to consume appropriate packages 

mhoeger:
Used in proxy change


paulbatum:
So it looks like the windows build failed due to this issue with a missing package source, but the windows build is not running for PRs right now. So I'm going to merge this so we can confirm this resolves the issue with the windows build.

PR #461, opened 2018-05-09
make postinstall script work behind proxy, fixes #251

watashiSHUN:
basically followed this [reference](https://www.npmjs.com/package/https-proxy-agent#examples)
might not be the best way to do it, but should get people unblocked
here is the test result from a simple local proxy setup
image

PR #465, opened 2018-05-12
make postinstall script work behind proxy, fixes #251

### LINKED ISSUE #251: Impossible to install the npm package behind a corporate proxy
Hello,

I tried to install the package via the command

`npm i -g azure-functions-core-tools`

But the issue is that I'm behind a corporate proxy and the postinstall step of the package (the one downloading "https://functionscdn.azureedge.net/public/1.0.4/Azure.Functions.Cli.zip") is doing a get request that can't be proxied, so I'm unable to download the package. As a result, the whole installation fails and I'm not able to use the package.

watashiSHUN:
v1 version of https://github.com/Azure/azure-functions-core-tools/pull/461
image
again, simple test result: before the blueline, the proxy server was not up

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ Publish
 /TestResults
 
 /tools/NuGet.exe
-/node_modules
+node_modules/
 
 *.ncrunchproject
 *.ncrunchsolution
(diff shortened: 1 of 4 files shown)


watashiSHUN:
why are we having two `.gitignore` ([root](https://github.com/Azure/azure-functions-core-tools/blob/master/.gitignore), [subfolder](https://github.com/watashiSHUN/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/.gitignore)) for v2? but one `.gitignore` for v1?
not a big deal...but its hard maintain

ahmelsayed:
I think the gitignore in the subfolder in v2 is just a mistake, feel free to remove it

Issue #289, opened 2017-10-24
Extensions installation action should create `extensions.csproj` file on the root folder


fabiocav:
Currently the `extensions install` action will create a the `extensions.csproj` file in a sub-folder named `functions-extensions`, that file should be located on the root, which is where the runtime expects it and where it will be built during deployment, bringing the required artifacts into the bin folder.

PR #471, opened 2018-05-14
Changing default extensions project location to app root

fabiocav:
This had been fixed in #289


ahmelsayed:
not sure how it was reverted. Maybe it was a bad merge or something

fabiocav:
That's my guess. I'm just glad I'm not crazy, as I was pretty sure this had been fixed :)

PR #498, opened 2018-05-29
Update runtime to 11820 and cli version to 1.0.13

mathewc:


src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -831,12 +831,8 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="azurefunctions\http\request.js">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="azurefunctions\http\response.js">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
+    <Content Include="azurefunctions\http\request.js" />
+    <Content Include="azurefunctions\http\response.js" />
```
ahmelsayed: undo this change. It's annoying, but these files need to be marked as always copy. Every time VS updated the references it does this for some reason that I don't know.

PR #512, opened 2018-06-06
Updating runtime and language worker references

fabiocav:


src/Azure.Functions.Cli/Common/SelfHostWebHostSettingsFactory.cs
```diff
@@ -12,8 +12,7 @@ public static WebHostSettings Create(string scriptPath)
                 IsSelfHost = true,
                 ScriptPath = scriptPath,
                 LogPath = Path.Combine(Path.GetTempPath(), @"LogFiles\Application\Functions"),
-                SecretsPath = Path.Combine(Path.GetTempPath(), "secrets", "functions", "secrets"),
-                IsAuthDisabled = true
```
fabiocav: This property was removed from the latest runtime

PR #522, opened 2018-06-07
Updating WebHost reference

fabiocav:


fabiocav:
@ahmelsayed , this updates the runtime version to the one we'll be releasing. Which has a couple of additional fixes.

PR #533, opened 2018-06-14
Update node debug key

### LINKED ISSUE #529: Update setting debug options
Functions host has been updated to parse environment variables/app settings for language worker command line arguments by settings laguageWorkers:{language}:arguments in 
PR https://github.com/Azure/azure-functions-host/pull/2975 

Set Environment varaible key languageWorkers:node:arguments value= "--inspect={portnumber}"
in [StartHostAction.cs#L81](https://github.com/Azure/azure-functions-core-tools/blob/859e64707e54c0c387118e0b441a47da4544f818/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs#L81
)

pragnagopa:
Fixes #529

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -67,7 +67,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         private IDictionary<string, string> InitializeConfigurationArguments()
         {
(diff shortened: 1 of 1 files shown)


ahmelsayed:
@pragnagopa the `nodeDebugKey = nodeDebugKey.Replace(":", "__");` conversion for non-Windows is still valid, right?

pragnagopa:
Yes. Correct. As we are setting environment variables on non windows we need to use __ instead of :

PR #546, opened 2018-06-21
Set support language-worker settings in the host start command line. …

### LINKED ISSUE #521: Enable Python debug port
For VS Code to use /cc @nturinski @fiveisprime

ahmelsayed:
…Closes #521

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -57,30 +57,12 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         public DebuggerType Debugger { get; set; }
 
-        public IDictionary<string, string> IConfigurationArguments { get; set; }
+        public string LanguageWorkerSetting { get; set; }
+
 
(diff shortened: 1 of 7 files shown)


ahmelsayed:
@pragnagopa but java is `debug`, right?

pragnagopa:
Correct. Java still uses different configuration model https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Rpc/Configuration/JavaWorkerProvider.cs haven’t update java yet

ahmelsayed:
Thanks Pragna. I updated the map.

vfedonkin:
I have some issue with this code on my dev machine:
return new Dictionary<string, string>
                {
                    { _map[workerRuntime], value }
                };

 _map[workerRuntime] can be equal to string.Empty ("") and this will fail since Environment.SetEnvironmentVariable can't set empty variable.

ahmelsayed:
Thanks @vfedonkin 
Fixed in https://github.com/Azure/azure-functions-core-tools/commit/cad3ac5ac6da524902a2a45fddea5dc4c6a9f1c3

PR #549, opened 2018-06-22
display correct proxy endpoint

vfedonkin:
display correct proxy endpoint


ahmelsayed:
can you rebase on master?

PR #583, opened 2018-07-17
added kubernetes deployment support

yaron2:
This PR allows for a new deploy action to a custom platform. (currently, kubernetes).


yaron2:
@ahmelsayed done

ahmelsayed:
I just gave it a try with a js function and it worked pretty well! I think C# ones need an update because a build step is now required, but I can fix that. 

Thanks @yaron2!

PR #629, opened 2018-08-15
Changes to adopt the new WebHost/SDK updates (DI/hosting/configuration/logging

fabiocav:
@ahmelsayed this is still a bit of a work in progress as there are a couple of changes (particularly around logging) that are still pending. @brettsam is working on those and they'll be landing today. I'll update the PR once they do.


fabiocav:
@ahmelsayed looks like I broke something with my changes. Apparently the exe for the CLI is not being correctly set/resolved.

ahmelsayed:
yeah, the E2E tests don't work on Linux yet. They only work on Windows.

fabiocav:
@ahmelsayed addressed all the pending items. We should be good to merge this.

src/Azure.Functions.Cli/Diagnostics/ConsoleTelemetryClientFactory.cs
```diff
@@ -1,27 +0,0 @@
...
-        public ConsoleTelemetryClientFactory(string instrumentationKey, SamplingPercentageEstimatorSettings samplingSettings, Func<string, LogLevel, bool> filter)
-            : base(instrumentationKey, samplingSettings, filter)
-        {
-        }
-
-        public override TelemetryClient Create()
-        {
-            TelemetryClient client = base.Create();
-
-            client.Context.GetInternalContext().SdkVersion = $"azurefunctionscoretools: {Constants.CliVersion}";
```
brettsam: You're missing this logic now. You can do this by registering a TelemetryClient (factory) with DI. See how we do this in Script here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L164-L173

fabiocav: Thanks! Planned to have that added after your PR and didn't.  Will have it updated.

PR #634, opened 2018-08-17
Fixes #614

### LINKED ISSUE #614: Improve func new success message
Currently, when `func new` completes successfully, this message is displayed:

    The template "QueueTrigger" was created successfully.

This is somewhat confusing/incorrect. It would be better to change instead to this:

    The function was created successfully from the "QueueTrigger" template.

Or even better:

    The function "MyQueueTrigger" was created successfully from the "QueueTrigger" template.

ankitkumarr:
Added a success message for creating Functions using Templates.


ahmelsayed:
Can you also add an End2End test case that checks for that string in the output? 

Here is an example of an E2E test 
https://github.com/Azure/azure-functions-core-tools/blob/2af842f59d22696ad0984737b4aa416e33246ce0/test/Azure.Functions.Cli.Tests/E2E/SettingsTests.cs#L13-L38

But for you, create a new `NewTests.cs` file and instead of `CheckFiles` you'd use `OutputContains` assertion. Check other E2E test cases and let me know if you have any questions

PR #646, opened 2018-08-20
fixes #641

### LINKED ISSUE #641: 'func new' creating function that isn't buildable
I have v2.0.1-beta.35 of the core tools installed in MacOS. 

Ran 'func new'.
Selected #4 (TimerTrigger) template
For function name entered 'txn-puller'

This caused txn-puller.cs to be generated (good)
However the namespace used for the function is also 'txn-puller' (which is invalid and causes build error - not so good).

Looks like the class name and function name (in the FunctionName attribute) were correctly converted to txn_puller (which is valid/buildable) . Just not the namespace.

ankitkumarr:
Added sanitization when creating new dotnet functions using templates.

### DIFF
--- a/src/Azure.Functions.Cli/Common/Utilities.cs
+++ b/src/Azure.Functions.Cli/Common/Utilities.cs
@@ -2,6 +2,9 @@
 using Colors.Net;
 using Colors.Net.StringColorExtensions;
 using Microsoft.Azure.WebJobs.Script;
+using System;
+using System.Linq;
+using System.Text;
 
 namespace Azure.Functions.Cli
 {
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -47,5 +50,33 @@ private static RichString AlternateLogoColor(string str, int firstColorCount = -
...
+        }
+
+        internal static string SanitizeClassName(string className)
+        {
+            return SanitizeLiteral(className);
+        }
+
+        internal static string SanitizeLiteral(string unsanitized, string allowed = "")
+        {
+            var sanitized = !char.IsLetter(unsanitized[0]) && !new[] { '_', '@' }.Contains(unsanitized[0]) ? new StringBuilder("_" + unsanitized.Substring(0, 1)) : new StringBuilder(unsanitized.Substring(0,1));
```
ahmelsayed: nit: for long `?:` lines I usually split them just for readability. 

```csharp
var value = condition
    ? value1
    : value2;
```

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -47,5 +50,33 @@ private static RichString AlternateLogoColor(string str, int firstColorCount = -
...
+            return SanitizeLiteral(nameSpace, allowed : ".");
+        }
+
+        internal static string SanitizeClassName(string className)
+        {
+            return SanitizeLiteral(className);
+        }
+
+        internal static string SanitizeLiteral(string unsanitized, string allowed = "")
+        {
```
ahmelsayed: maybe add a check here for 

```csharp
!string.isNullOrEmpty(unsanitized)
```

PR #649, opened 2018-08-21
Fixes #481 - Support language alias when creating templates

### LINKED ISSUE #481: Normalize --language in func new to accept language name aliases
Trying to create a new Javascript function is not working.

It asks for a Template, but no option is given. Trying to type the template name does not work either, making it unusable.

image

Trying to create the same function in C# actually works fine:

image

ankitkumarr:
Created a mapping for getting the template names from the worker, for now, assuming each worker supports only one language for the template. Let me know if I should change the implementation to not rely on the worker runtime. Thanks!

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
@@ -32,6 +32,13 @@ namespace Azure.Functions.Cli.Helpers
             .SelectMany(i => i)
             .ToDictionary(k => k.key, v => v.value, StringComparer.OrdinalIgnoreCase);
 
+        private static readonly IDictionary<WorkerRuntime, string> workerToLanguageMap = new Dictionary<WorkerRuntime, string>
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -32,6 +32,14 @@ public static class WorkerRuntimeLanguageHelper
             .SelectMany(i => i)
             .ToDictionary(k => k.key, v => v.value, StringComparer.OrdinalIgnoreCase);
 
+        private static readonly IDictionary<WorkerRuntime, string> workerToLanguageMap = new Dictionary<WorkerRuntime, string>
+        {
+            { WorkerRuntime.dotnet, "c#" },
+            { WorkerRuntime.node, "javascript" },
+            { WorkerRuntime.python, "python" },
+            { WorkerRuntime.java, "" }
```
ahmelsayed: since we don't have any java templates anyway, I think you can remove this

PR #703, opened 2018-09-06
Add ability to create certs locally

ankitkumarr:
Added ability to create local certificates for https using openssl

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
@@ -9,6 +9,7 @@ using Azure.Functions.Cli.NativeMethods;
 using Colors.Net;
 using static Colors.Net.StringStaticMethods;
 using System.Text;
+using System.Threading.Tasks;
 
 namespace Azure.Functions.Cli.Helpers
 {
@@ -69,7 +70,7 @@ namespace Azure.Functions.Cli.Helpers
             }
         }
 
-        internal static X509Certificate2 GetOrCreateCertificate(string certPath, string certPassword)
+        internal static async Task<X509Certificate2> GetOrCreateCertificate(string certPath, string certPassword)
         {
             if (!string.IsNullOrEmpty(certPath) && !string.IsNullOrEmpty(certPassword))
             {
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
```diff
@@ -109,5 +113,28 @@ internal static X509Certificate2 GetOrCreateCertificate(string certPath, string
 
             throw new CliException("Auto cert generation is currently not working on the .NET Core build.");
         }
+
+        internal static X509Certificate2 CreateCertificateOpenSSL()
+        {
+            const string DEFAULT_PASSWORD = "localcert";
+
+            var cert_file_names = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName().Replace(".", String.Empty));
```
ahmelsayed: nit: for consistency, in C# code we usually use `camelCasing` for variable names. So `certFileName`, etc

src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
```diff
@@ -109,5 +113,28 @@ internal static X509Certificate2 GetOrCreateCertificate(string certPath, string
...
+
+            var openssl_key = new Executable("openssl", $"req -new -x509 -newkey rsa:2048 -nodes -keyout {cert_file_names}localhost.key -out {cert_file_names}localhost.cer -days 365 -subj /CN=localhost");
+            var exitCode = openssl_key.RunAsync(l => ColoredConsole.WriteLine(l)).Result;
```
ahmelsayed: I'd just make all the chain async to avoid calling `.Result`. It's not a big deal, but it just makes the code cleaner.

src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
```diff
@@ -109,5 +113,28 @@ internal static X509Certificate2 GetOrCreateCertificate(string certPath, string
...
+            if (exitCode != 0)
+            {
+                throw new CliException("Could not create a key pair required for an openssl certificate.");
+            }
+
+            Executable openssl_cert = new Executable("openssl", $"pkcs12 -export -out {cert_file_names}certificate.pfx -inkey {cert_file_names}localhost.key -in {cert_file_names}localhost.cer -passout pass:{DEFAULT_PASSWORD}");
+            exitCode = openssl_cert.RunAsync(l => ColoredConsole.WriteLine(l)).Result;
```
ahmelsayed: instead of writing all of openssl output to the console, write it to a StringBuilder instead, including stderr, and only display it if there was an error calling into openssl. You can add a `ColoredConsole.WriteLine("Generating a self signed certificate using openssl")` line to indicate what is happening if you want.

src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
ankitkumarr: Ah, Right! Sorry, force of habit. 🤓

PR #706, opened 2018-09-10
add PowerShell worker to core tools

tylerl0706:
This PR adds the PowerShell worker to the core tools. The worker's files will end up in `workers/powershell` which is consistent with the other workers.

In addition to adding the worker, this also adds a `build.ps1` build script that works on any platform. I had trouble getting the `build.sh` working on mac because it specified only linux to build for... so I wrote a powershell build script that does the OS check.

~NOTE: The worker has not been published to the feed yet so please do not merge this in until that is published. I will update this PR when it's done.~

fixes https://github.com/Azure/azure-functions-powershell-worker/issues/45


tylerl0706:
cc @SteveL-MSFT @daxian-dbw @pragnagopa @asavaritayal

asavaritayal:
@ahmelsayed could you please review this PR? We'd like PowerShell to be hidden under the `--worker-runtime` flag similar to how we did for Python.

tylerl0706:
I've addressed the merge conflict and updated with the correct version of the worker.

tylerl0706:
@ahmelsayed I think this was the PR that @asavaritayal was referring to 😄

PR #707, opened 2018-09-10
Fixes #678 - show func info after publish

### LINKED ISSUE #678: Add ability to get HTTP-triggered Function's URL
after using `func azure functionapp publish` it'd be great if I could then interrogate the functions that were published and get back the trigger URLs for the HTTP-Triggered ones.

I'm thinking something like:
~~~
C:\Users\hurlburb\Projects\GitHub\serverless-big-data-map-reduce [brandonh/redux ≡]> func azure functionapp list-functions myfuncappv2
Functions in myfuncappv2:
    Begin - [orchestrationTrigger]
    GetFileListAsync - [activityTrigger]
    StartAsync - [httpTrigger - https://myfuncappv2.azurewebsites.net... ]
    WriteToBlob - [activityTrigger]
~~~

or even target a specific function and ask for it with a  new CLI command like `func azure functionapp get-url myfuncappv2 StartAsync`

I played with both Functions CLI and Azure CLI and couldn't readily see a way to get this.

ankitkumarr:
Print the details of all the functions after publish is completed (including the key).

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -236,37 +236,34 @@ namespace Azure.Functions.Cli.Helpers
             }
         }
 
-        public static async Task PrintFunctionsInfo(Site functionApp, string accessToken)
+        public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, bool showKeys)
         {
             var functions = await GetFunctions(functionApp, accessToken);
+            ColoredConsole.WriteLine(TitleColor($"Functions in {functionApp.SiteName}:"));
+            foreach (var function in functions)
             {
+                var trigger = function
+                    .Config?["bindings"]
+                    ?.FirstOrDefault(o => o["type"]?.ToString().IndexOf("Trigger", StringComparison.OrdinalIgnoreCase) != -1)
+                    ?["type"];
 
-                ColoredConsole.WriteLine(TitleColor($"Functions in {functionApp.SiteName}:"));
-                foreach (var function in functions)
+                trigger = trigger ?? "No Trigger Found";
+
+                ColoredConsole.WriteLine($"    {function.Name} - [{VerboseColor(trigger.ToString())}]");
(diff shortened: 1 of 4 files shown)


src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -211,5 +235,39 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
                     : new HttpResult<Dictionary<string, string>, string>(null, errorMessage);
             }
         }
+
+        public static async Task PrintFunctionsInfo(Site functionApp, string accessToken)
+        {
+            var functions = await GetFunctions(functionApp, accessToken);
+            {
```
ahmelsayed: nit: remove extra `{`s

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -211,5 +235,39 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+                        ?["type"];
+
+                    trigger = trigger ?? "No Trigger Found";
+
+                    ColoredConsole.WriteLine($"    {function.Name} - [{VerboseColor(trigger.ToString())}]");
+                    if (!string.IsNullOrEmpty(function.InvokeUrlTemplate))
+                    {
+                        // If there's a key available, add it to the url
+                        var key = await GetFunctionKey(function.Href.AbsoluteUri, functionApp.ScmUri, accessToken);
+                        if (!string.IsNullOrEmpty(key))
```
ahmelsayed: lets not display the key be default. Add a `--show-keys` flag and only show keys if that's specified

ankitkumarr:
I added a flag to show the keys in the Http-triggered URLs. When the flag is not set, URLs do not show the keys. I am not displaying the `*`s yet, because to know if there's keys to be retrieved, we have to make two network calls for each function. Figured, if we aren't displaying the keys, it might be better to just notify the users that they can get URLs with keys, and display the generic URL thereby saving time.

Let me know if you prefer a different behavior, I would be happy to make the change.

PR #709, opened 2018-09-11
Fixes #553 - detect if built project available

### LINKED ISSUE #553: Improve 'func start' behavior with compiled apps
Currently, you have to pass `--build` all the time with a compiled project, even if it's already built. Ideally, it should be smart enough to detect a compiled project and run from the artifact folder (e.g. `bin\output`).

ankitkumarr:
Added a small logic to check if there is a pre-built project available in `bin/output`

Note- I am currently checking if the current directory does not have a build AND there is one available in `bin/output`, only then would it use the `bin/output` directory. Just so, if the user actually just wants to run the command from the built project directory. 

Ideally (if we don't have one already), I think it would be nice to have a flag (--build-path or something), that can also provide the path we want to use the project from.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -238,17 +238,48 @@ namespace Azure.Functions.Cli.Actions.HostActions
             {
                 PythonHelpers.VerifyVirtualEnvironment();
             }
-            else if (workerRuntime == WorkerRuntime.dotnet && Build)
+            else if (workerRuntime == WorkerRuntime.dotnet)
             {
-                const string outputPath = "bin/output";
-                await DotnetHelpers.BuildDotnetProject(outputPath, string.Empty);
-                Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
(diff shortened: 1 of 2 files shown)


ahmelsayed:
Instead of this change, can we look in the working directory for a csproj that is not `ExtensionsProj.csproj`, and if we found at least one, we assume `--build`. I'd like to have an environment variable that disables that behavior. Ideally we would ask both VS and VSCode to pass a `--no-build` when running and then we can always assume `--build` for dotnet projects. Does this make sense?

ankitkumarr:
@ahmelsayed, Yes, that does make sense. 
I thought that the original issue was that if there was a built project available, we should use that instead of building it again.
Just so I understand correctly, you are suggesting that instead of doing that, we just assume `--build` for all dotnet projects and rebuild the project everytime `func host start` is called. I like this suggestion, and I think that will also close #765.

ahmelsayed:
yeah I agree having `--no-build` should be enough. But I'm trying to minimize the changes our partners (VS, VS for mac, and VSCode) will have to do. So the check for a `.csproj` in the root should catch most of these scenarios

ankitkumarr:
Closing this PR as the requested changes would be easier and cleaner to simply start over. Please check the new the PR #811

PR #710, opened 2018-09-11
Fixes #554 - remove trigger sync when using zipdeploy

### LINKED ISSUE #554: func CLI should not sync triggers when using zipdeploy
It's redundant since zipdeploy takes care of it.

ankitkumarr:


ankitkumarr:
@ahmelsayed Looks like we don't use zipDeploy when we `PublishRunFromZip`. So, I thought we should only sync triggers when we use `PublishRunFromZip`. 

We use `PublishRunFromZip` when `functionApp.IsLinux && functionApp.IsDynamic) || RunFromZipDeploy`. So, we should only syncTriggers when `functionApp.IsLinux && functionApp.IsDynamic) || RunFromZipDeploy`.

The wording of it may be a little weird. Or perhaps, I understood it wrong?

ahmelsayed:
looks like you'll need a rebase to resolve some conflicts, but feel free to merge after that.

PR #715, opened 2018-09-13
updating templates and metadata generator version

soninaren:


soninaren:
- [ ] pending webhost package update

PR #735, opened 2018-09-19
Removing beta from cli version, adding optional force parameter

soninaren:


src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -25,11 +25,13 @@ internal static class Constants
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string StorageEmulatorConnectionString = "UseDevelopmentStorage=true";
         public const string AzureWebJobsStorage = "AzureWebJobsStorage";
+        public const string PackageReferenceElementName = "PackageReference";
+        public const string PackageReferenceVersionElementName = "Version";
 
         public static string CliVersion => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.ToString(3);
         public static string CliBetaRevision => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.MinorRevision.ToString();
```
ahmelsayed: is this still valid?

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -75,9 +81,14 @@ public async override Task RunAsync()
 
                 if (string.IsNullOrEmpty(Package) && string.IsNullOrEmpty(Version))
                 {
+                    var project = ProjectHelpers.GetProject(extensionsProj);
                     foreach (var extensionPackage in ExtensionsHelper.GetExtensionPackages())
                     {
-                        await AddPackage(extensionsProj, extensionPackage.Name, extensionPackage.Version);
+                        // Only add / update package referece if it does not exist or forced update is enabled
```
ahmelsayed: I don't quite understand this comment. Why only add/update if it doesn't exist? wouldn't `dotnet` just takes care of doing what ever is right?

soninaren: When you run 'dotnet add' it always updates the extension version in the csproj, even if the package reference is already present.
This change would only add extension package if it is already not present.

src/Azure.Functions.Cli/Common/Constants.cs
soninaren: only `PackageReferenceElementName` is needed, `PackageReferenceVersionElementName` is not used

PR #739, opened 2018-09-20
Fixing the version string

soninaren:


src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -30,7 +30,7 @@ internal static class Constants
         public static string CliVersion => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.ToString(3);
         public static string CliBetaRevision => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.MinorRevision.ToString();
 
-        public static string CliDisplayVersion => $"{Constants.CliVersion}.{Constants.CliBetaRevision}";
+        public static string CliDisplayVersion => CliVersion + ((CliBetaRevision == "0") ? string.Empty : $"-beta.{CliBetaRevision}");
```
ahmelsayed: why not just remove `CliBetaRevision`. Why would we go back to beta from rtm?

soninaren: Just thought we may have beta versions for internal testing. Would prevent it to break again for java plugin

PR #746, opened 2018-09-21
Fixes #738 - show functionapp if it's in the next page

### LINKED ISSUE #738: "Can't find app with name ..." if there are many resources
When I try to make call: 
> func azure functionapp fetch-app-settings myfuncapp

I receive: **Can't find app with name myfuncapp**. :(
After some investigation: 
**az functionapp list --query "[].name"** shows **myfuncapp**.
There are more than 999 resources in subscription.
To publish function application this library tries to find apropriate place so it`s doing request like this:

> https://management.azure.com/subscriptions/{subscriptionid}/resources?api-version=2014-04-01&$filter=(name eq 'myfuncapp' and resourceType eq 'Microsoft.Web/sites').

But such request makes search only in first page (999 items). To query all rest resources it is needed to consistently make calls on url which is in previous response in *nextLink* value. In my situation **myfuncapp** appeared in response of 2nd request:.
>{"value":[{"id":"/subscriptions/{id}/resourceGroups/{rg}/providers/Microsoft.Web/sites/myfuncapp","name":"myfuncapp","type":"Microsoft.Web/sites","kind":"functionapp","location":"westus2"}],"nextLink":"https://management.azure.com/subscriptions/{id}/resources?api-version=2014-04-01&%24filter=(name+eq+%27myfuncapp%27+and+resourceType+eq+%27Microsoft.Web%2fsites%27)&%24skiptoken={token}"}

Unfortunately I do not see such code [here](https://github.com/Azure/azure-functions-core-tools/blob/d776e404643776ce823e95f6fa35d48252f3d223/src/Azure.Functions.Cli/Helpers/AzureHelper.cs#L21-L45) and [here](https://github.com/Azure/azure-functions-core-tools/blob/d776e404643776ce823e95f6fa35d48252f3d223/src/Azure.Functions.Cli/Arm/ArmClient.cs)
Also it would be great to limit request by resource group (publish can take an optional parameter) as **az** do. Here is example of **az** request:
 > https://management.azure.com/subscriptions/{subscriptionid}/resourceGroups/rg-dmp-k8s-mmp-dev-westus2/providers/Microsoft.Web/sites/myfuncapp?api-version=2016-08-01

ankitkumarr:
I made the change to go to retrieve all the functionapps.

I didn't add the suggested `resource-group` flag yet. I was unable to make a successful request to the ARM with a resource group that would give me my functionapp using the link suggested in the issue. It there's a specific format of the link and if we want this flag, please let me know, I will update the PR.

### DIFF
--- a/src/Azure.Functions.Cli/Arm/Models/ArmArrayWrapper.cs
+++ b/src/Azure.Functions.Cli/Arm/Models/ArmArrayWrapper.cs
@@ -5,5 +5,7 @@ namespace Azure.Functions.Cli.Arm.Models
     internal class ArmArrayWrapper<T>
     {
         public IEnumerable<ArmWrapper<T>> value { get; set; }
+
+        public string nextLink { get; set; }
     }
 }
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -33,6 +33,20 @@ namespace Azure.Functions.Cli.Helpers
                 }),
                 accessToken);
 
+                // If we haven't found the functionapp, and there is a next page, keep going
+                while (!functionApps.value.Any() && !string.IsNullOrEmpty(functionApps.nextLink))
+                {
+                    try
+                    {
+                        functionApps = await ArmHttpAsync<ArmArrayWrapper<ArmGenericResource>>(HttpMethod.Get, new Uri(functionApps.nextLink), accessToken);
+                    }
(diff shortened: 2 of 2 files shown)


ahmelsayed:
let's create another issue to track adding a resourceGroup switch

ahmelsayed:
I think your branch might not be rebased on master, which is why that test case is failing. It should be fixed in master

ankitkumarr:
Oh, I see. I will be sure to rebase and merge.

PR #761, opened 2018-09-26
Add installation instructions for other Linux distros

bbrowning:
This adds generic installation instructions that should work on most Linux distributions. They were specifically tested to work against Fedora 27.


ahmelsayed:
Thank you for your contribution @bbrowning !

Issue #759, opened 2018-09-26
CLI should start host in Development mode


mathewc:
The Functions Runtime can be put in Development mode to optimize debugging scenarios. E.g. polling intervals will be reduced, etc. to facilitate immediate feedback.

The portal puts the runtime into this mode when it interacts with it, ensuring portal debugging is responsive.

The CLI should always start the host in Development mode I think. While users can set the AZURE_FUNCTIONS_ENVIRONMENT environment variable locally to "Development" to get this behavior, I think it should be the default.

See https://github.com/Azure/azure-functions-core-tools/issues/755

Issue #755, opened 2018-09-25
Azure Storage Queue polling configuration not working in VS 2017 tools


pablocastillaquintas:
Azure Functions Tools version: 2.8.1

Hi!

When I am working using VS with AF binding against azure storage queues the polling configuration doesn't work and it takes the default time to get the messages (30 seconds or even more).

When I upload the functions into Azure it works perfectly (1 second).

Regards.

pablocastillaquintas:
Perfect! thanks! 

As a recommendation maybe the debug behaviour could be the default one when you use functions under VS.

pablocastillaquintas:
Ok, I have set the environment variable and it doesn't work.

I have set:
AZURE_FUNCTIONS_ENVIRONMENT to Development in the debug tab of the project.

When it starts it says:
Starting Host (HostId=qesevpc159-610740046, InstanceId=5d0f8bf2-f2da-455b-b39b-820df544b9b8, Version=2.0.12115.0, ProcessId=28664, AppDomainId=1, Debug=False, FunctionsExtensionVersion=)

Thanks

PR #763, opened 2018-09-26
Default local environment to Development (#759)

mathewc:
Addresses https://github.com/Azure/azure-functions-core-tools/issues/759.

As discussed in https://github.com/Azure/azure-functions-core-tools/issues/755, I verified that setting this in local.settings.json puts the host in dev mode correctly. See host code [here](https://github.com/Azure/azure-functions-host/blob/a2d2308ac22fd4c87b62d4e3ca1dc36f470c699e/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L182) which is where this setting is consumed.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -160,22 +160,25 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         private async Task<IDictionary<string, string>> GetConfigurationSettings(string scriptPath, Uri uri)
         {
-            var secrets = _secretsManager.GetSecrets();
+            var settings = _secretsManager.GetSecrets();
+            settings.Add(Constants.WebsiteHostname, uri.Authority);
+
+            // Add our connection strings
+            var connectionStrings = _secretsManager.GetConnectionStrings();
+            settings.AddRange(connectionStrings.ToDictionary(c => $"ConnectionStrings:{c.Name}", c => c.Value));
+            settings.Add(EnvironmentSettingNames.AzureWebJobsScriptRoot, scriptPath);
+
             var environment = Environment
                     .GetEnvironmentVariables()
                     .Cast<DictionaryEntry>()
                     .ToDictionary(k => k.Key.ToString(), v => v.Value.ToString());
-            var connectionStrings = _secretsManager.GetConnectionStrings();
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/StaticResources/local.settings.json.template
```diff
@@ -2,6 +2,7 @@
   "IsEncrypted": false,
   "Values": {
     "FUNCTIONS_WORKER_RUNTIME": "{FUNCTIONS_WORKER_RUNTIME}",
-    "AzureWebJobsStorage": "{AzureWebJobsStorage}"
+    "AzureWebJobsStorage": "{AzureWebJobsStorage}",
+    "AZURE_FUNCTIONS_ENVIRONMENT": "Development"
```
mathewc: However, this will only help for new apps. Maybe we should also force dev mode imperatively when the cli builds the host?

ahmelsayed: I'm wondering if it'll be better to do that in the cli code before we invoke the webhost. That way you'll always get it if it's not defined. The problem is that during publishing, this file gets turned into App Settings. I'd think most people won't want that by default when they publish. We can exclude it, but then will need to add a flag is someone actually wants it. What do you think?

mathewc: Ah, based on the file name I assumed this was ONLY for running locally. Yes, we don't want this pushed. I'll set this imperatively instead.

mathewc: I thought the settings in local.settings.json are ONLY for running locally. That's what our documentation says here: https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local. 

> "Settings in the local.settings.json file are only used by Functions tools when running locally. By default, these settings are not migrated automatically when the project is published to Azure."

ahmelsayed: the file is only loaded locally. That's used when people are working locally with the cli and need to set various App Settings etc. on publish you can pass a `--publish-local-settings` which will parse this file and update the Azure App Settings with the values from here.

mathewc:
Merged https://github.com/Azure/azure-functions-core-tools/commit/6e190f9f2a6c11e8d2d959968f41f1af5930c1c6

PR #773, opened 2018-10-02
Fixes #695 - Run from package by default

### LINKED ISSUE #695: Use Run From Package by default
Current deployment behavior allows you to specify --zip to deploy and use run from package. We'd like to make this the default behavior, and add a flag to opt out (maybe --nozip)

ankitkumarr:


ankitkumarr:
@ahmelsayed Please let me know if I shouldn't do ` Callback(f => RunFromZipDeploy = !f);` , but instead change the `RunFromZipDeploy` variable to be `ExtractZipRunDeploy` or something and then change the logic everywhere accordingly.

PR #781, opened 2018-10-04
Displaying HTTP methods for functions triggered with [HttpTrigger]

kamil-mrzyglod:
Hi,

While working with Azure Functions CLI and localhost, I often find it difficult to immediately know, how a function triggered via HTTP request should be called(which HTTP method should I use). This PR proposes a simple solution by prepending e.g. [POST, GET] to the displayed function URL when a host runs.

Regards,
Kamil


ahmelsayed:
Thanks @kamil-mrzyglod!

Issue #741, opened 2018-09-21
Deploy to Kubernetes doesn't allow upper/mixed cased function names


krancour:
Using version 2.0.3...

This seems to be because the function name is reused as the image name, but Docker image names must be lowercase only.

Azure Functions names can legitimately be upper/mixed case, so I propose that deployment to k8s should account for this in some way.

```
$ func deploy --platform kubernetes --name GetMessages --registry krancour
Building Docker image...
Running 'docker build -t krancour/GetMessages /Users/kent/kentssecondserverlessapp'..done
Error running docker build -t krancour/GetMessages /Users/kent/kentssecondserverlessapp.
output: 

invalid argument "krancour/GetMessages" for "-t, --tag" flag: invalid reference format: repository name must be lowercase
See 'docker build --help'.
```

PR #784, opened 2018-10-08
Fix docker image name generation

acesyde:
* Fix docker image generation #741 
* Add unit tests


ahmelsayed:
Thank you very much @acesyde for your contribution.

PR #786, opened 2018-10-09
Removing powershell from runtime worker options

soninaren:


soninaren:
Fixing in reference to https://github.com/Azure/azure-functions-core-tools/pull/706#issuecomment-425669241

NavaneethaDev:
Can you please let me know why PowerShell is removed from worker options? Currently we are seeing only below 3 options,

dotnet
node
python

We wanted a PowerShell also. Can you please tell me why this was removed? Thanks!

Issue #367, opened 2018-02-01
[V2] remove dependency on ./dotnet from the CLI


ahmelsayed:
The only scenario that requires `dotnet` executable to be on the machine is that of installing extensions because we just launch [`dotnet add extensions.csproj package <extension> --version <version>`](https://github.com/Azure/azure-functions-cli/blob/89ada273881bc42da7a1f32f496f8e6a552a5c3c/src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs#L55)

this is the only command that would prevent building a self-contained cli

mpseidel:
Any ETA on when this dependency can be removed? 

Running azure functions via node container and need to install extensions so I ran into this.

ahmedelnably:
closing as resolved by extension bundles

PR #789, opened 2018-10-10
Enable Easy Auth Local Dev

glennamanns:
**Main goal**: Enable local development with Easy Auth / Middleware (rebranding in process). 

This is based on Easy Auth Linux, and cannot be incorporated into this repo until the EA Linux + Windows merge is complete. However, I'd like to get feedback on it now.

**One new command, used two different ways:**

1. _func auth create-aad --aad-name {AADAppName}_
This one is exclusively for local testing. It creates a new AAD application with localhost:7071 in the reply URLs, a homepage of localhost:7071, and a URI with localhost in the name. The client ID, client secret, and other settings that Easy Auth needs to run locally are saved to a new file local.middleware.json. This file can later be consumed by the Easy Auth webhost.

2. _func auth create-aad --aad-name {AADAppName} --app-name {ExistingAzureAppName}_
This is designed to link an existing Function/Site with a new AAD application. It creates a new AAD application, but there is nothing related to localhost in its settings. The settings are the same as if you were to create a new app from the Portal's Easy Auth flow. This command also does not create a new .json file (to avoid writing secrets to disk). The existing Function/Site's authsettings (WEBSITE_AUTH_ENABLED, etc) are updated via an ARM call.

**Still to do:**
- Easy Auth Linux + Windows merge. Need to inject dependency upon EasyAuth/Middleware Nuget.


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -184,6 +186,42 @@ private async Task<IWebHost> BuildWebHost(ScriptApplicationHostOptions hostOptio
...
+                    query += $"  --{Constants.MiddlewareCertPathSetting} {certPath} --{Constants.MiddlewareCertPasswordSetting} {certPassword}";
+                }
+                var process = Process.Start("CMD.exe", $"/C dotnet {dllPath} {query}");
+                var cancellationToken = new CancellationTokenSource();
+                cancellationToken.Token.Register(() => process.Kill());
+                await process.WaitForExitAsync();
+            }
+            else
+            {
+                throw new FileNotFoundException("Must have dotnet installed in order to use local authentication.");
```
ahmelsayed: I haven't looked at the full PR yet, but this will be a problem. We only have one code path that depends on `dotnet` right now (other than .NET development scenarios of course) and we are tracking removing that dependency here #367. Ideally we wouldn't add more dependencies on `dotnet` for other non-.NET specific scenarios. Any plans to have self-contained builds of that console app?

glennamanns: We can certainly make it a priority to have a self-contained build if removing the dotnet dependency is important. I'll look into it.

src/Azure.Functions.Cli/Actions/AuthActions/BaseAuthAction.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Linq;
+using Azure.Functions.Cli.Actions.AzureActions;
+using Fclp;
+
+namespace Azure.Functions.Cli.Actions.AuthActions
+{
+    abstract class BaseAuthAction : BaseAzureAction
```
ahmelsayed: You don't really need this class since it's empty, right? you can just drive from `BaseAzureAction` if you need the access token, or were you planning on adding some common code here?

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Threading.Tasks;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Interfaces;
+using Fclp;
+namespace Azure.Functions.Cli.Actions.AzureActions
+{
+    // Invoke via `func azure auth create-aad -aad-name {displayNameOfAAD} --app-name {displayNameOfApp}`
+    [Action(Name = "create-aad", Context = Context.Azure, SubContext = Context.Auth, HelpText = "Creates a production Azure Active Directory application with given name. Links it to specified Azure Application")]
+    class CreateAADApplication : BaseAzureAction
```
ahmelsayed: so the difference between `CreateAADApplication.CreateAADApplication` and `AuthActions.CreateAADApplication` is that the latter assumes `localhost` while the former sets it up for your `app.azurewebsites.net`? If so, you probably could combine them into one Action and have `AppName` default to `localhost` if not specified, right?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -184,6 +186,42 @@ private async Task<IWebHost> BuildWebHost(ScriptApplicationHostOptions hostOptio
...
+                string query = $"--{Constants.MiddlewareListenUrlSetting} {listenUrl} --{Constants.MiddlewareHostUrlSetting} {hostUrl} " +
+                    $"--{Constants.MiddlewareLocalSettingsSetting} {SecretsManager.MiddlewareAuthSettingsFileName}";
+                if (certPath != null && certPassword != null)
+                {
+                    query += $"  --{Constants.MiddlewareCertPathSetting} {certPath} --{Constants.MiddlewareCertPasswordSetting} {certPassword}";
+                }
+                var process = Process.Start("CMD.exe", $"/C dotnet {dllPath} {query}");
+                var cancellationToken = new CancellationTokenSource();
+                cancellationToken.Token.Register(() => process.Kill());
+                await process.WaitForExitAsync();
```
ahmelsayed: I'm a bit unsure about the use of the `cancellationToken` here. It doesn't look like you really need it. You can also use the `Executable` class I have defined to help with this. The usage is fairly straight forward. Check [here ](https://github.com/Azure/azure-functions-core-tools/blob/299805e0edad5cb895c5f0d3d432ad847d8d721a/src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs#L139)  or [here](https://github.com/Azure/azure-functions-core-tools/blob/299805e0edad5cb895c5f0d3d432ad847d8d721a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs#L65) for samples

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -217,7 +255,33 @@ public override async Task RunAsync()
...
 
-            (var listenUri, var baseUri, var certificate) = await Setup();
+            // Determine if middleware (Easy Auth) is enabled
+            var middlewareAuthSettings = _secretsManager.GetMiddlewareAuthSettings();
+            bool authenticationEnabled = middlewareAuthSettings.ContainsKeyCaseInsensitive(Constants.MiddlewareAuthEnabledSetting) &&
+                middlewareAuthSettings[Constants.MiddlewareAuthEnabledSetting].ToLower().Equals("true");
+
+            (var listenUri, var baseUri, var certificate, string certPath, string certPassword) = await Setup();
+
+            int originalPort = Port;
```
ahmelsayed: maybe call it `hostPort` for clarity.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -217,7 +255,33 @@ public override async Task RunAsync()
...
+            if (authenticationEnabled)
+            {
+                // If it is enabled, then the Functions Host needs a different port
+                Port = Port + 2;
+            }
+
+            // Regardless of whether or not auth is enabled, clients should send requests here
+            var originalListenUri = listenUri.SetPort(originalPort);
+            var originalBaseUri = baseUri.SetPort(originalPort);
```
ahmelsayed: same thing here, I think it'll be clearer if you called them `hostListenUri` and `authListenUri`

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+
+                int exitCode = await az.RunAsync(o => stdout.AppendLine(o), e => stderr.AppendLine(e));               
+
+                // Clean up file we created to pass data in proper format to az CLI
+                File.Delete($"{tempFile}");
+
+                if (exitCode != 0)
+                {
+                    ColoredConsole.WriteLine(Red(stderr.ToString().Trim(' ', '\n', '\r', '"')));
+                    return;
```
ahmelsayed: do you wanna throw here? if you don't throw on an error, the exit code of the cli process won't indicate an error. It's useful when people are scripting the CLI and then try to check exit code for any errors before proceeding.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+            authSettingsToPublish.Add("allowedAudiences", serializedArray);
+            authSettingsToPublish.Add("isAadAutoProvisioned", "True");
+            authSettingsToPublish.Add("clientId", clientId);
+            authSettingsToPublish.Add("clientSecret", clientSecret);
+            authSettingsToPublish.Add("defaultProvider", "0"); // 0 corresponds to AzureActiveDirectory
+            authSettingsToPublish.Add("enabled", "True");
+            authSettingsToPublish.Add("issuer", "https://sts.windows.net/" + tenant + "/");
+            authSettingsToPublish.Add("runtimeVersion", "1.0.0");
+            authSettingsToPublish.Add("tokenStoreEnabled", "True");
+            authSettingsToPublish.Add("unauthenticatedClientAction", "1"); // Corresponds to AllowAnonymous
```
ahmelsayed: You could use the dictionary initialization syntax, it makes it a bit more readable, i.e:

```csharp
var authSettingsToPublish = new Dictionary<string, string>
{
    { "allowedAudiences", serializedArray },
    { "isAadAutoProvisioned", "True" },
    ...
}
```

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+        {
+            functionApp.AzureAuthSettings = authSettings;
+            var result = await AzureHelper.UpdateFunctionAppAuthSettings(functionApp, accessToken);
+            if (!result.IsSuccessful)
+            {
+                ColoredConsole
+                    .Error
+                    .WriteLine(ErrorColor("Error updating app settings:"))
+                    .WriteLine(ErrorColor(result.ErrorResult));
+                return false;
```
ahmelsayed: same thing here regarding the error. If this is an error that you would like to communicate as a cli error, then consider either throwing here, or where you call this like I mentioned in the other comment above

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+            }
+            return true;
+        }
+
+        static string ComputeSha256Hash(string rawData)
```
ahmelsayed: You can add this method to [`StringExtensions.cs`](https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/Extensions/StringExtensions.cs)

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+                StringBuilder builder = new StringBuilder();
+                for (int i = 0; i < bytes.Length; i++)
+                {
+                    builder.Append(bytes[i].ToString("x2"));
+                }
+                return builder.ToString();
+            }
+        }
+
+        public static string GeneratePassword(int length)
```
ahmelsayed: can you explain a bit what this password is for and what its format is like?

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,312 @@
...
+        {
+            return !string.IsNullOrEmpty(password) &&
+                password.Any(c => char.IsUpper(c)) &&
+                password.Any(c => char.IsLower(c)) &&
+                password.Any(c => char.IsDigit(c)) &&
+                password.Any(c => !char.IsLetterOrDigit(c));
+        }
+    }
+
+    static class AADConstants
```
ahmelsayed: move this class to the `Constants.cs` class as a nested class there.

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
glennamanns: Yes, you're correct. It's currently two separate commands based on a meeting I had with @mattchenderson, but combining them would be trivial. I'm happy to change it if it seems like that would provide a better UX :)

src/Azure.Functions.Cli/Common/AuthManager.cs
glennamanns: I based this on PublishAppSettings / UpdateFunctionAppAppSettings, which modifies /config/AppSettings (mine modifies /config/authsettings). Is there a reason those do not throw?

src/Azure.Functions.Cli/Common/AuthManager.cs
glennamanns: This generates an AAD application client secret. An example of what it generates is: "77S2ca0Gwf8mTr7s85ZEqe7dcxKnF4DdyMAF5gNFaTBY6gBEDtcPwQWlTP3KWNElvld9eFyXZdvFfmEnqmrHQA7ki0$LRQ8TL3uLrj6S$F0PRtRYnZ8Dp9xReKNQK9z"

fabiocav:
@glennamanns just saw the last commit land. Would you remove the WIP/let us know when this is ready for review? Just want to make sure we don't ignore your changes once this is ready to go :)

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Threading.Tasks;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Interfaces;
+using Fclp;
+namespace Azure.Functions.Cli.Actions.AzureActions
```
cgillum: nit: add a newline in between the `using` and `namespace` statements.

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    // Invoke via `func azure auth create-aad -aad-name {displayNameOfAAD} --app-name {displayNameOfApp}`
+    [Action(Name = "create-aad", Context = Context.Azure, SubContext = Context.Auth, HelpText = "Creates a production Azure Active Directory application with given name. Can be linked to specified Azure Application")]
+    class CreateAADApplication : BaseAzureAction
+    {
+        private readonly IAuthManager _authManager;
+
+        public string AADName { get; set; }
```
cgillum: Naming nit: "AAD" is not really a thing, so `AADName` sounds unnatural. I'm also not sure what the difference is between an "AADName" and an "AppName". Should they be the same? Or should "AADName" be renamed to "AppRegistrationName"?

cgillum: Same comment for the parameter syntax below.

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
```diff
@@ -0,0 +1,43 @@
...
+        public override async Task RunAsync()
+        {
+            await _authManager.CreateAADApplication(AccessToken, AADName, AppName);
+        }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<string>("aad-name")
+                .WithDescription("Name of AD application to create")
```
cgillum: Product naming nit: "Name of the **Azure Active Directory app registration** to create."

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
```diff
@@ -0,0 +1,43 @@
...
+                .Callback(f => AADName = f);
+
+            Parser
+                .Setup<string>("app-name")
+                .WithDescription("Name of Azure Websites Application/Function to link AAD application to")
```
cgillum: Grammar and product naming nits: "Name of the Azure App Service or Azure Functions app which corresponds to the Azure AD app registration."

cgillum: I get nit-picky about this since these strings are public-facing and therefore need to be considered somewhat official.

src/Azure.Functions.Cli/Common/AuthSettingsFile.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+namespace Azure.Functions.Cli.Common
+{
+    class AuthSettingsFile
+    {
+        public bool IsEncrypted { get; set; }
+        public Dictionary<string, string> Values { get; set; } = new Dictionary<string, string>();
```
cgillum: Make this a case-insensitive dictionary: new `Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)`. That way you don't have to remember to use case-sensitive lookups every time you reference this.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -217,9 +225,14 @@ public override async Task RunAsync()
 
             var settings = SelfHostWebHostSettingsFactory.Create(Environment.CurrentDirectory);
 
-            (var listenUri, var baseUri, var certificate) = await Setup();
+            // Determine if middleware (Easy Auth) is enabled
+            var middlewareAuthSettings = _secretsManager.GetMiddlewareAuthSettings();
+            bool authenticationEnabled = middlewareAuthSettings.ContainsKeyCaseInsensitive(Constants.MiddlewareAuthEnabledSetting) &&
+                middlewareAuthSettings[Constants.MiddlewareAuthEnabledSetting].ToLower().Equals("true");
```
cgillum: This would be a bit cleaner:
```csharp
bool authenticationEnabled = 
    middlewareAuthSettings.TryGetValue(Constants.MiddlewareAuthEnabledSetting, out string enabledValue) &&
    bool.TryParse(enabledValue, out bool isEnabled) &&
    isEnabled;
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -395,25 +408,53 @@ internal static async Task CheckNonOptionalSettings(IEnumerable<KeyValuePair<str
...
+        private async Task<(Uri listenUri, Uri baseUri, X509Certificate2 cert, string path, string password)> Setup()
         {
             var protocol = UseHttps ? "https" : "http";
-            X509Certificate2 cert = UseHttps
-                ? await SecurityHelpers.GetOrCreateCertificate(CertPath, CertPassword)
-                : null;
-            return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
+            if (UseHttps)
+            {
+                (X509Certificate2 cert, string certPath, string certPassword) = await SecurityHelpers.GetOrCreateCertificate(CertPath, CertPassword);
```
cgillum: It feels unnatural to fetch the certificate and both its path and password. Is this because the Easy Auth code you're calling expects these? Could the Easy Auth code be refactored to take an `X509Certificate2` for embedded scenarios like this?

glennamanns: Removing all of this altogether. Easy Auth, when truly middleware (which it is now), doesn't need its own reference to the certificate.

src/Azure.Functions.Cli/Common/AuthSettingsFile.cs
```diff
@@ -0,0 +1,81 @@
...
+                string path = Path.Combine(Environment.CurrentDirectory, _filePath);
+                var content = FileSystemHelpers.ReadAllTextFromFile(path);
+                var authSettings = JObject.Parse(content);
+                Values = new Dictionary<string, string>(authSettings.ToObject<Dictionary<string, string>>(), StringComparer.OrdinalIgnoreCase);
+            }
+            catch
+            {
+                Values = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                IsEncrypted = false;
+            }
```
cgillum: If something goes wrong, we need to provide some kind of signal so that users can either fix the issue themselves or report the details to us. For that reason, I suggest the following:
1. Check to see if the file exists before trying to load it (`File.Exists(filePath)`). If it doesn't exist, we can silently move on like you're currently doing.
2. If there is a JSON parsing exception, log the message of the exception so that the user can figure out why their JSON file was invalid.
3. If it's a file permission error, log just the exception message.
4. Any other exception type, log the full exception details.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+{
+    internal class AuthManager : IAuthManager
+    {
+        public AuthManager() { }
+
+        private AuthSettingsFile MiddlewareAuthSettings;
+
+        private WorkerRuntime _workerRuntime;
+
+        public async Task CreateAADApplication(string accessToken, string AADName, WorkerRuntime workerRuntime, string appName)
```
cgillum: Let's renamed `AADName` to `appRegistrationName`.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+        {
+            if (string.IsNullOrEmpty(AADName))
+            {
+                throw new CliArgumentsException("Must specify name of new Azure Active Directory registration with --appRegistrationName parameter.",
+                    new CliArgument { Name = "appRegistrationName", Description = "Name of new Azure Active Directory registration" });
```
cgillum: "Azure Active Directory registration" --> "Azure Active Directory **app** registration" (two places)

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+            }
+
+            if (CommandChecker.CommandExists("az"))
+            {
+                _workerRuntime = workerRuntime;
+
+                List<string> replyUrls;
+                string tempFile, clientSecret, hostName, query = CreateQuery(AADName, appName, out tempFile, out clientSecret, out replyUrls, out hostName);
+
+                ColoredConsole.WriteLine("Query successfully constructed. Creating new Azure AD Application now..");
```
cgillum: I suggest we print out the command-line we're running.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+            {
+                throw new CliArgumentsException("Must specify name of new Azure Active Directory registration with --appRegistrationName parameter.",
+                    new CliArgument { Name = "appRegistrationName", Description = "Name of new Azure Active Directory registration" });
+            }
+
+            if (CommandChecker.CommandExists("az"))
+            {
+                _workerRuntime = workerRuntime;
+
+                List<string> replyUrls;
```
cgillum: Do you need this to be a `List<string>` or can you instead use `string[]`. `string[]` array is preferred since it is less overhead and it is a fixed size, making it easier to reason about how it is used throughout the code (i.e. nobody is going to add to it or remove from it).

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Create the query to send to az ad app create
+        /// </summary>
+        /// <param name="AADName">Name of the new AAD application</param>
+        /// <param name="appName">Name of an existing Azure Application to link to this AAD application</param>
+        /// <returns></returns>
+        private string CreateQuery(string AADName, string appName, out string tempFile, out string clientSecret, out List<string> replyUrls, out string hostName)
```
cgillum: Should this be named `GetCommandLineArguments`?

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+                if (functionDirProjFiles.Count == 1)
+                {
+                    return functionDirProjFiles.First();
+                }
+                else
+                {
+                    return null;
+                }
+            }
+            catch (DirectoryNotFoundException e)
```
cgillum: Rather than relying on exceptions, it's best practice to explicitly check to see if the directory exists - e.g. `Directory.Exists(...)`.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+            project.Add(newItemGroup);
+            xmlFile.Save(csProj);
+
+            ColoredConsole.WriteLine(Yellow($"Modified {csProj} to include {SecretsManager.MiddlewareAuthSettingsFileName} in output directory."));
+        }
+
+        private Dictionary<string, string> CreateAuthSettingsToPublish(string clientId, string clientSecret, string tenant, List<string> replyUrls)
+        {
+            // The 'allowedAudiences' setting of /config/authsettings is of the form ["{replyURL1}", "{replyURL2}"]
+            string serializedArray = JsonConvert.SerializeObject(replyUrls, Formatting.Indented);
```
cgillum: It's important that variable names give us an idea of what the data actually is and/or what it will be used for. In this case, `replyUrlArray` or something similar seems more appropriate.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,406 @@
...
+
+            var authSettingsToPublish = new Dictionary<string, string>
+            {
+                { "allowedAudiences", serializedArray },
```
cgillum: `allowedAudiences` should not be the same as the reply URLs. It should just be the base URL - e.g. "https://myapp.azurewebsites.net" or "http://myapp.localhost".

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
glennamanns: What are your thoughts on AADRegistrationName for both the variable and what goes on the command line? And for Azure App Service apps/functions, appName?

src/Azure.Functions.Cli/Common/SecretsManager.cs
```diff
@@ -28,6 +28,21 @@ public static string AppSettingsFilePath
             }
         }
 
+        public static string MiddlewareAuthSettingsFilePath
+        {
+            get
+            {
+                var authFile = "local.middleware.json";
+                var rootPath = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory, new List<string>
```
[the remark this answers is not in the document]
glennamanns: Is the reason for this that arrays are more lightweight than lists?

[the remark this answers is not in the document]
cgillum: Yes, and they are more immutable (you cannot add or remove elements, making them easier to reason about). Generally speaking you should try to use the most restrictive data types.

src/Azure.Functions.Cli/Actions/AzureActions/CreateAADApplication.cs
cgillum: Maybe `AADAppRegistrationName` instead of AADRegistrationName? `appName` is fine for the function app name.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -250,6 +250,26 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
             }
         }
 
+        public static async Task<HttpResult<Dictionary<string, string>, string>> UpdateFunctionAppAuthSettings(Site site, string accessToken)
+        {
+            var url = new Uri($"{ArmUriTemplates.ArmUrl}{site.SiteId}/config/authsettings?api-version={_storageApiVersion}");
```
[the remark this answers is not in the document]
glennamanns: GetStorageAccount, an existing method/ARM call, uses it. Authsettings requires a later API version than appsettings, because I got the following error with the older version: "The current api-version doesn't support RBAC users.  Please use a newer version instead"

I could either change the name of the variable to something more generic, or add a second variable like _authSettingsApi and hardcode it to 2018-02-01, which is what the storage api version is.

src/Azure.Functions.Cli/Helpers/ExtensionsHelper.cs
```diff
@@ -52,6 +52,23 @@ private static IEnumerable<string> GetBindings()
             return bindings;
         }
 
+        public static IEnumerable<(string, BindingDirection)> GetBindingsWithDirection()
```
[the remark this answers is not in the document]
glennamanns: The create-aad command. In order to know which permissions to add, I need to know the bindings and their direction

src/Azure.Functions.Cli/Extensions/StringExtensions.cs
```diff
@@ -41,5 +43,22 @@ public static string SanitizeImageName(this string imageName)
 
             return cleanImageName.ToLowerInvariant().Substring(0, Math.Min(cleanImageName.Length, 128)).Trim();
         }
+
+        public static string ComputeSha256Hash(string rawData)
```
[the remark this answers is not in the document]
glennamanns: create-aad to generate an encryption and signing key

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -88,5 +88,12 @@ internal static bool EqualsIgnoreCaseAndSpace(string str, string another)
         {
             return str.Replace(" ", string.Empty).Equals(another.Replace(" ", string.Empty), StringComparison.OrdinalIgnoreCase);
         }
+
+        public static Uri SetPort(this Uri uri, int newPort)
```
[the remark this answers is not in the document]
glennamanns: old - will remove

src/Azure.Functions.Cli/Common/AuthManager.cs
glennamanns: It's a List<string> in order to get the serialized array (_string serializedArray = JsonConvert.SerializeObject(replyUrls, Formatting.Indented_) that we use to update /authsettings in the proper format. The other parts of the code could use string[], but not this one. I can either change all of them to string[] and then convert to List in the publish to /authsettings scenario, or eat the overhead elsewhere.

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -38,6 +47,85 @@ public static class Errors
...
+                public const string User = "Scope";
+            }
+
+            public static Dictionary<(string, BindingDirection), List<Guid>> PermissionMap = new Dictionary<(string, BindingDirection), List<Guid>>
+            {
+               { ("graphwebhooktrigger", BindingDirection.In), new List<Guid>
+                   {
+                       MicrosoftGraphReadPermissions.FilesRead,
+                       MicrosoftGraphReadPermissions.MailRead,
+                       MicrosoftGraphReadPermissions.UserRead // TODO !! check if this is the case
```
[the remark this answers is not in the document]
glennamanns: ehhh...kind of. We can't really know how the user will decide to use the Graph webhooktrigger, so I just added the 3 most common permissions, at the risk of requiring too much access.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,405 @@
...
+
+            if (_workerRuntime == WorkerRuntime.dotnet)
+            {
+                // If this is a dotnet function, we also need to add this file to the .csproj 
+                // so that it copies to \bin\debug\netstandard2.x when the Function builds
+                string csProjFile = GetCSProjFilePath();
+
+                if (csProjFile == null)
+                {
+                    throw new CliException($"Auth settings file {SecretsManager.MiddlewareAuthSettingsFileName} could not be added to a .csproj and will not be present in the the bin or output directories.");
```
cgillum: C# scripts (.csx) are .NET, but they don't have a .csproj file. Will this exception exclude those users? I wonder if it's safer to just skip that step and write an informational trace if we can't find the .csproj file.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,405 @@
...
+            MiddlewareAuthSettings.Commit();
+
+            ColoredConsole.WriteLine(Yellow($"Created {SecretsManager.MiddlewareAuthSettingsFileName} with authentication settings necessary for local development.\n" +
+                $"Running this function locally will only work with the Function Host default port of {StartHostAction.DefaultPort}"));
+
+            if (_workerRuntime == WorkerRuntime.dotnet)
+            {
+                // If this is a dotnet function, we also need to add this file to the .csproj 
+                // so that it copies to \bin\debug\netstandard2.x when the Function builds
+                string csProjFile = GetCSProjFilePath();
```
cgillum: What about `.fsproj` for F# developers?

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,405 @@
...
+
+        private void CreateAndCommitAuthSettings(string hostName, string clientId, string clientSecret, string tenant, List<string> replyUrls)
+        {
+            // The WEBSITE_AUTH_ALLOWED_AUDIENCES setting is of the form "{replyURL1} {replyURL2}"
+            string serializedReplyUrls = string.Join(" ", replyUrls);
+
+            // Create a local auth .json file that will be used by the middleware
+            var middlewareAuthSettingsFile = SecretsManager.MiddlewareAuthSettingsFileName;
+            MiddlewareAuthSettings = new AuthSettingsFile(middlewareAuthSettingsFile);
+            MiddlewareAuthSettings.SetAuthSetting("WEBSITE_AUTH_ALLOWED_AUDIENCES", serializedReplyUrls);
```
cgillum: GitHub isn't letting me reply to our existing comment thread on this, but the portal behavior you observed is wrong, and a bug which can cause CRIs. The allowed_audiences should NOT contain /.auth/login/aad/callback.

src/Azure.Functions.Cli/Common/AuthManager.cs
```diff
@@ -0,0 +1,405 @@
...
+        }
+
+        private Dictionary<string, string> CreateAuthSettingsToPublish(string clientId, string clientSecret, string tenant, List<string> replyUrls)
+        {
+            // The 'allowedAudiences' setting of /config/authsettings is of the form ["{replyURL1}", "{replyURL2}"]
+            string serializedReplyUrls = JsonConvert.SerializeObject(replyUrls, Formatting.Indented);
+
+            var authSettingsToPublish = new Dictionary<string, string>
+            {
+                { "allowedAudiences", serializedReplyUrls },
```
cgillum: (Duplicating my comment from elsewhere)

GitHub isn't letting me reply to our existing comment thread on this, but the portal behavior you observed is wrong, and a bug which can cause CRIs. The allowed_audiences should NOT contain /.auth/login/aad/callback.

src/Azure.Functions.Cli/Extensions/StringExtensions.cs
cgillum: nit: Since this is a class intended to create extension methods, it's better to be consistent and make this method also an extension method:

```suggestion
        public static string ComputeSha256Hash(this string rawData)
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -395,25 +414,53 @@ internal static async Task CheckNonOptionalSettings(IEnumerable<KeyValuePair<str
...
+
         private async Task<(Uri listenUri, Uri baseUri, X509Certificate2 cert)> Setup()
         {
             var protocol = UseHttps ? "https" : "http";
-            X509Certificate2 cert = UseHttps
-                ? await SecurityHelpers.GetOrCreateCertificate(CertPath, CertPassword)
-                : null;
-            return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
+            if (UseHttps)
+            {
```
glennamanns: todo: revert this to the original logic

jabbera:
Any chance of this making it any time soon? This breaks so many things. (Such as not being able to debug authenticated serverless Azure Functions locally at all)

ahmelsayed:
Closing old PR. Please reopen if this still needs to be looked at

Issue #780, opened 2018-10-03
Add --platform ACI to Functions CLI 


odaibert:
For Dev Test scenario it will be simple to deploy the containerized Azure Function to ACI. Something like this: 

**func** **deploy** --**platform** **_**aci**_** --**name** myfunction2 --**registry** odaibert --**resourcegroupname** FunctionsACI --**containergroupname** FunctionsACI --**subscriptionid** SUBSCRIPTIONID --**location** eastus --**port** 80 --**containermemory** 1 --**containercpu** 2 --**ostype** linux 

Make sense?

ahmelsayed:
as I mentioned on the PR, I'm not in favor of this change. I'd rather not do azure management operations in this cli. All management operations for Azure should be in the az cli instead. This only requires building a docker image and pushing it to the right registry, then all the commands to deploy to ACI should be the same as in `az`.

PR #790, opened 2018-10-10
Add --platform ACI to Functions CLI

odaibert:
Added "aci" option to --platform deployment. 

This option creates a Azure Container Instance Group and deploy the container from container repo to there.

This PR addresses Add --platform ACI to Functions CLI #780 feature.

https://github.com/Azure/azure-functions-core-tools/issues/780


ahmelsayed:
I'm not in favor of this change. I'd rather not do azure management operations in this cli. All management operations for Azure should be in the `az` cli instead.

PR #792, opened 2018-10-10
Add ability to bundle python modules with worker

ankitkumarr:
**Note: Please do not merge yet, still working on testing different scenarios.**

Add ability to bundle the python worker with the modules from `requirements.txt` using pyinstaller.
In addition, please review- https://github.com/Azure/azure-functions-docker/pull/34

Additional work to be done (in case we take this path):
1) Make sure venv is not included in the zip created by pack
2) Make `--build-native-deps` the default, as we will need to take the docker path if we use pyinstaller.

### DIFF
--- a/src/Azure.Functions.Cli/Common/Constants.cs
+++ b/src/Azure.Functions.Cli/Common/Constants.cs
@@ -55,6 +55,7 @@ namespace Azure.Functions.Cli.Common
         public static class StaticResourcesNames
         {
             public const string PythonDockerBuild = "python_docker_build.sh";
+            public const string PythonBundleScript = "python_bundle_script.py";
         }
 
         public static ExtensionPackage ExtensionsMetadataGeneratorPackage => new ExtensionPackage { Name = "Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator", Version = "1.0.1" };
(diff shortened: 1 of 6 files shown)


src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -49,12 +49,13 @@ public static class ArmConstants
 
         public static class DockerImages
         {
-            public const string LinuxPythonImageAmd64 = "mcr.microsoft.com/azure-functions/python:2.0";
+            public const string LinuxPythonImageAmd64 = "mcr.microsoft.com/azure-functions/python:2.0.12134";
```
ahmelsayed: Leave this `2.0`, I'll make sure to update that tag with `2.0.12134` before the CLI release. It's fine to keep it with this specific tag on your machine for testing, but the final commit should be `2.0`

ankitkumarr: That sounds good.

ahmelsayed:
I think the before and after numbers for just the package do look very promising. here is what I'm seeing for the shark-prediction sample:

|scenario|before|after|
|--------|-------|-----|
| compressed| 119 MB| 65 MB|
| extracted| 447 MB | 237 MB|
| Number of files | 9,307 files | 107 files|

PR #805, opened 2018-10-12
Fixes #794 - ignore venv when packing

### LINKED ISSUE #794: Do not include venv file in the package
Currently, if we have a venv inside the functions project directory, it gets packaged when trying to pack or publish. It would be nice to exclude any virtual environment from the project package as it will greatly impact the size of the package

ankitkumarr:
A few questions/considerations regarding this PR-
1) In the change, we are detecting if venv directory is present in the current directory. So, this will not work if venv is submerged in some deep sub-directories, but that seems very unlikely.
2) The change takes in account that for some reason `.funcignore` may already be present. To keep things simple, it adds an empty line and then appends the `venv` to it. So, when the `.funcignore` does not exist, it has an empty line and then the venv. This is not breaking anything, but if it looks bad, I can make changes.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -24,6 +24,7 @@ namespace Azure.Functions.Cli.Helpers
             await InstallPipWheel();
             await InstallPythonAzureFunctionPackage();
             await PipFreeze();
+            await EnsureVirtualEnvrionmentIgnored();
         }
 
         public static void VerifyVirtualEnvironment()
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -24,6 +24,14 @@ public static async Task InstallPackage()
             await InstallPipWheel();
             await InstallPythonAzureFunctionPackage();
             await PipFreeze();
+            try
+            {
+                await EnsureVirtualEnvrionmentIgnored();
+            }
+            catch (Exception)
+            {
+                // Safe execution, we aren't harmed by failures here
```
ahmelsayed: can you move the try/catch to the `EnsureVirtualEnvrionmentIgnored()` then? it will make this one a bit more readable.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -34,6 +42,32 @@ public static void VerifyVirtualEnvironment()
...
+                }
+                // Write the current env to .funcignore
+                ColoredConsole.WriteLine($"Writing {Constants.FuncIgnoreFile}");
+                using (var fileStream = FileSystemHelpers.OpenFile(funcIgnorePath, FileMode.Append, FileAccess.Write))
+                using (var streamWriter = new StreamWriter(fileStream))
+                {
+                    await streamWriter.WriteAsync(Environment.NewLine + virtualEnvName);
+                    await streamWriter.FlushAsync();
+                }
+            }
```
ahmelsayed: There is `FileSystemHelpers.WriteAllTextToFile()` that simplifies that pattern.

Issue #765, opened 2018-09-28
"dotnet" functions template start not intuitive


mhoeger:
The normal flow of getting a function to start doesn't work for the "dotnet" create experience. The normal flow of events is `func init` => `func new` => `func host start`. It feels unintuitive to have to do `dotnet build` and `func host start` from the build output.

**Repro steps:**
 - `func init`
    - Choose `dotnet`
 - `func new`
    - Create any function (example: HttpTrigger)
 - `func host start`

Output is `No job functions found. Try making your job classes and methods public. If you're using binding extensions (e.g. ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. config.UseServiceBus(), config.UseTimers(), etc.).`

**To get function to run:**
 - Run `dotnet build`
 - Change directory to be debug directory
    - `cd bin/Debug/netstandard2.0`
 - `func host start`

benc-uk:
You have to run `func host start --build` with Dotnet, which I do agree is really not intuitive at all, and it caught me out too

PR #811, opened 2018-10-17
Fixes #553, #765 - dotnet build by default

### LINKED ISSUE #553: Improve 'func start' behavior with compiled apps

ankitkumarr:
One additional thing to note is:

Before this change, if you were to try and do `azure functionapp publish <name>` on a dotnet runtime project without a `.csproj`, it would error out complaining that it couldn't build as 0 `.csproj` was found. 
But, after this, it would still continue informing the user that the build step was skipped. I think it makes sense to keep it this way if someone wants to just publish their built project directly. I can't imagine a case where this would be a breaking change.

Let me know if you prefer me to keep the publishing the original way, I can update the PR.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -121,9 +121,16 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
             if (workerRuntime == WorkerRuntime.dotnet && !Csx && !NoBuild)
             {
-                const string outputPath = "bin/publish";
-                await DotnetHelpers.BuildDotnetProject(outputPath, DotnetCliParameters);
-                Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                if (DotnetHelpers.CanDotnetBuild())
+                {
+                    var outputPath = Path.Combine("bin", "publish");
+                    await DotnetHelpers.BuildDotnetProject(outputPath, DotnetCliParameters);
+                    Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                }
+                else if (StaticSettings.IsDebug)
+                {
+                    ColoredConsole.WriteLine("Could not find a valid .csproj file. Skipping the build.");
+                }
             }
 
             if (ListIncludedFiles)
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -59,7 +59,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         public string LanguageWorkerSetting { get; set; }
 
-        public bool Build { get; set; }
+        public bool NoBuild { get; set; }
 
 
         public StartHostAction(ISecretsManager secretsManager)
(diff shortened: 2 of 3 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -121,9 +121,16 @@ public override async Task RunAsync()
...
-                Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                if (DotnetHelpers.CanDotnetBuild())
+                {
+                    var outputPath = Path.Combine("bin", "publish");
+                    await DotnetHelpers.BuildDotnetProject(outputPath, DotnetCliParameters);
+                    Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                }
+                else
+                {
+                    ColoredConsole.WriteLine("Could not find a valid .csproj file. Skipping the build.");
```
ahmelsayed: can you log this only in debug mode? `if (StaticSettings.IsDebug)`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -11,6 +11,7 @@
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Diagnostics;
 using Azure.Functions.Cli.Extensions;
+// using Azure.Functions.Cli.Extensions;
```
ahmelsayed: nit: remove if no longer used.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -244,11 +246,18 @@ private async Task PreRunConditions(WorkerRuntime workerRuntime)
...
-                Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                if (DotnetHelpers.CanDotnetBuild())
+                {
+                    var outputPath = Path.Combine("bin", "output");
+                    await DotnetHelpers.BuildDotnetProject(outputPath, string.Empty);
+                    Environment.CurrentDirectory = Path.Combine(Environment.CurrentDirectory, outputPath);
+                }
+                else
+                {
+                    ColoredConsole.WriteLine("Could not find a valid .csproj file. Skipping the build.");
```
ahmelsayed: same here. Log only in debug mode.

src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs
```diff
@@ -70,28 +70,35 @@ internal static IEnumerable<string> GetTemplates()
...
         }
 
-        public static async Task BuildDotnetProject(string outputPath, string dotnetCliParams)
+        public static bool CanDotnetBuild()
         {
             EnsureDotnet();
             var csProjFiles = FileSystemHelpers.GetFiles(Environment.CurrentDirectory, searchPattern: "*.csproj").ToList();
-            if (csProjFiles.Count == 1)
+            // If the project name is extensions only then is extensions.csproj a valid csproj file
+            if (!Path.GetFileName(Environment.CurrentDirectory).Equals("extensions"))
```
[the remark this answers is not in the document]
ankitkumarr: So I thought of a case when the functions project name is `extensions` for some reason, we should then allow `extensions.csproj`. 
The `Path.GetFileName(Environment.CurrentDirectory)` of a functions project called `extensions` would result in the string `extensions`.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
ankitkumarr: Oh, my bad, that wasn't supposed to be there.

src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs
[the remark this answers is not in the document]
ahmelsayed: got it

PR #815, opened 2018-10-19
Bug: Token extension installs wrong nuget

glennamanns:
Token extension should depend on AuthTokens nuget, not MicrosoftGraph nuget.


soninaren:
The bug is fixed now. This should be safe to merge. 
The core issue was that we could not install multiple extensions on the same app, and that was blocking graph scenarios. Since Microsoft Graph package brought both packages in (directly and via transitive dependency) it unblocked those scenarios.

glennamanns:
Got it. Thanks for the clarification :) I'll merge it in now

PR #821, opened 2018-10-22
Fixes #762 - Add verification for linuxFxVersion while publishing

### LINKED ISSUE #762: Verify LinuxFxVersion while publishing
When publishing to dedicated Linux apps, we could verify that the image matches the desired framework. 

Currently the core tools only verifies FUNCTIONS_WORKER_RUNTIME, but that still leaves the image to potentially not match. 

Our images are 
```
- microsoft/azure-functions-base
- microsoft/azure-functions-dotnet2.0
- microsoft/azure-functions-node8
- microsoft/azure-functions-python3.6

- mcr.microsoft.com/azure-functions/base
- mcr.microsoft.com/azure-functions/dotnet
- mcr.microsoft.com/azure-functions/python
- mcr.microsoft.com/azure-functions/node
```

ankitkumarr:
The check will only fail if they have a linuxFxVersion set to something in the portal. Let me know, if I should also fail if they don't have a version configured on linux dedicated.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -117,7 +117,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
             // Check for any additional conditions or app settings that need to change
             // before starting any of the publish activity.
-            var additionalAppSettings = ValidateFunctionAppPublish(functionApp, workerRuntime);
+            var additionalAppSettings = await ValidateFunctionAppPublish(functionApp, workerRuntime);
 
             if (workerRuntime == WorkerRuntime.dotnet && !Csx && !NoBuild)
             {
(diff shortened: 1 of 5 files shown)


ankitkumarr:
Ahh, I see, that does make sense. Didn't realize that we had more images and that custom images were allowed. Sounds good to me! I will update the PR.

ankitkumarr:
@ahmelsayed 

> If the name matches one of our images, but doesn't match the current local framework, we should allow the user to force update the image name on Azure to the correct one.

In case of force update, is there a preference on which type of image to push? Example- `microsoft/azure-functions-node` vs `mcr.microsoft.com/azure-functions/node`

ahmelsayed:
Let's go with the mcr image. They have a better naming structure.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -225,6 +225,40 @@ public override async Task RunAsync()
                 }
             }
 
+            if (functionApp.IsLinux && !functionApp.IsDynamic && !string.IsNullOrEmpty(functionApp.LinuxFxVersion))
+            {
+                var allImages = Constants.WorkerRuntimeImages.Values.SelectMany(image => image).ToList();
+                if (string.IsNullOrEmpty(allImages.FirstOrDefault(image => functionApp.LinuxFxVersion.Contains(image))))
```
ahmelsayed: use `IndexOf(image, StringComparison.OrdinalIgnoreCase) != -1)` instead to get case insensitive matching. You can also use `.Any(` instead of `FirstOrDefault` and `IsNullOrEmpty`

PR #836, opened 2018-10-31
update PowerShell Worker Version

tylerl0706:
Tons of good stuff inside :)

cc @pragnagopa @SteveL-MSFT @daxian-dbw @asavaritayal @soninaren


ahmelsayed:
This doesn't require templates updates, right?

tylerl0706:
My understanding is that the template deployment and the dockerfile deployment are separate. Is that not the case? 

We just had a few templates merged into the azure-functions-templates repo.

tylerl0706:
Ah! Understood. Yes old templates (well, there's only one 😊) are compatible.

No coordination is needed between templates and worker.

tylerl0706:
@ahmelsayed are there any blockers to merge this in?

PR #607, opened 2018-07-30
Durable CLI

cmkerner20:
WIP: Initial CLI support for Durable Functions instance management. 

Supports:
-start new
-delete history store
-show status
-raise event
-terminate
-rewind [not yet supported on durabletask framework]

PR #838, opened 2018-10-31
New commands for testing Durable Functions locally

glennamanns:
This is a continuation of Catherine's PR (https://github.com/Azure/azure-functions-core-tools/pull/607/commits). I don't have write access to her fork of the repo, so I opted to make a new branch on the main repo. Mostly just code clean up, particularly the customer facing aspects.

It's important to note that this PR means the CLI now has a direct dependency on DurableTask.AzureStorage


src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,221 @@
...
+            await _orchestrationService.DeleteAsync();
+            ColoredConsole.Write(Green("History and instance tables and queues successfully deleted."));
+        }
+
+        public async Task GetHistory(string instanceId)
+        {
+            if (string.IsNullOrEmpty(instanceId))
+            {
+                throw new CliArgumentsException("Must specify the id of the orchestration instance you wish to retrieve the history for.",
+                    new CliArgument { Name = "id", Description = "ID of the orchestration instance to retrieve the history of." });
```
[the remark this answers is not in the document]
glennamanns: That makes sense. Thanks for pointing that out :)

src/Azure.Functions.Cli/Actions/DurableActions/DurableGetRuntimeStatus.cs
```diff
@@ -0,0 +1,36 @@
...
+            return base.ParseArgs(args);
+        }
+
+        public override async Task RunAsync()
+        {
+            Parser
+             .Setup<bool>("getAllExecutions")
+             .WithDescription("If true, the status of all executions is retrieved. If false, only the status of the most recent execution is retrieved.")
+             .SetDefault(false)
+             .Callback(n => GetAllExecutions = n);
```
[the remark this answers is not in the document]
glennamanns: whoops - that's what I get for making a last minute change

src/Azure.Functions.Cli/Actions/DurableActions/DurableStartNew.cs
```diff
@@ -0,0 +1,57 @@
...
+            Parser
+               .Setup<string>("input")
+               .WithDescription("Input to the orchestrator function")
+               .SetDefault(null)
+               .Callback(p => Input = p);
+            Parser
+               .Setup<string>("version")
+               .WithDescription("Version of the orchestrator function")
+               .SetDefault(null)
+               .Callback(v => Version = v);
```
cgillum: Let's remove this. We don't support explicit versions in Durable Functions. Instead, we always use `string.Empty`.

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -74,6 +74,7 @@
     <PackageReference Include="KubeClient" Version="2.1.5" />
     <PackageReference Include="KubeClient.Extensions.KubeConfig" Version="2.1.5" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="2.1.1" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.3.1" />
```
cgillum: For now, let's target `1.3.2-rc` since that contains all the new operations we recently added. Before we actually merge this, we should update it again to `1.3.2` (the final released version).

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+    {
+        private readonly AzureStorageOrchestrationService _orchestrationService;
+
+        private readonly TaskHubClient _client;
+
+        private const string TaskHubName = "DurableFunctionsHub";
+
+        public DurableManager(ISecretsManager secretsManager)
+        {
+            var connectionString = secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals("AzureWebJobsStorage", StringComparison.OrdinalIgnoreCase)).Value;
```
cgillum: Some customers will want to use non-default connection strings since we advertise that as a best practice. Can we update our commands to support this? My thinking is that we should look at the `host.json` file for the correct connection string name. See [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-bindings#host-json) for documentation. (WARNING: Make sure we're using the v2 syntax where `"durableTask"` lives inside the `"extensions"` property. This is not shown in the preceding link at the time of writing).

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+
+namespace Azure.Functions.Cli.Common
+{
+    internal class DurableManager : IDurableManager
+    {
+        private readonly AzureStorageOrchestrationService _orchestrationService;
+
+        private readonly TaskHubClient _client;
+
+        private const string TaskHubName = "DurableFunctionsHub";
```
cgillum: Many customers will need to use a non-hardcoded task hub name. By default we should discover this from **host.json**.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            };
+
+            _orchestrationService = new AzureStorageOrchestrationService(settings);
+            _client = new TaskHubClient(_orchestrationService);
+        }
+
+        public async Task DeleteHistory()
+        {
+            await _orchestrationService.DeleteAsync();
+            ColoredConsole.Write(Green("History and instance tables and queues successfully deleted."));
```
cgillum:
```suggestion
            ColoredConsole.Write(Green("Task hub successfully deleted."));
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+        public async Task GetHistory(string instanceId)
+        {
+            var historyString = await _orchestrationService.GetOrchestrationHistoryAsync(instanceId, null);
+
+            JArray history = JArray.Parse(historyString);
+
+            JArray chronological_history = new JArray(history.OrderBy(obj => (string)obj["TimeStamp"]));
+            foreach (JObject jobj in chronological_history)
+            {
+                var parsed = Enum.TryParse(jobj["EventType"].ToString(), out EventType eventName);
```
cgillum: Let's add a comment explaining that we're converting the numeric enum values to the equivalent string values.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                jobj["EventType"] = eventName.ToString();
+            }
+
+            ColoredConsole.Write($"History: {chronological_history.ToString(Formatting.Indented)}");
+        }
+
+        public async Task GetRuntimeStatus(string instanceId, bool getAllExecutions)
```
cgillum: Let's get rid of the `getAllExecutions` parameter. It's not useful for Durable Functions.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                ColoredConsole.WriteLine($"Name: {status.Name}");
+                ColoredConsole.WriteLine($"Instance: {status.OrchestrationInstance}");
+                ColoredConsole.WriteLine($"Version: {status.Version}");
+                ColoredConsole.WriteLine($"TimeCreated: {status.CreatedTime}");
+                ColoredConsole.WriteLine($"CompletedTime: {status.CompletedTime}");
+                ColoredConsole.WriteLine($"LastUpdatedTime: {status.LastUpdatedTime}");
+                ColoredConsole.WriteLine($"Input: {status.Input}");
+                ColoredConsole.WriteLine($"Output: {status.Output}");
+                ColoredConsole.WriteLine($"Status: {status.OrchestrationStatus}");
```
cgillum: I'm thinking we should format this differently, either as a table or as formatted JSON. JSON is nice because then people can output it to a file and process it.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+        public async Task RaiseEvent(string instanceId, string eventName, object eventData)
+        {
+            var orchestrationInstance = new OrchestrationInstance
+            {
+                InstanceId = instanceId
+            };
+
+            await _client.RaiseEventAsync(orchestrationInstance, eventName, eventData);
+
+            ColoredConsole.WriteLine(Green($"Raised event {eventName} to instance {instanceId} with data {eventData}"));
```
cgillum: Let's remove the _with data_ part since the event data could be large.
```suggestion
            ColoredConsole.WriteLine(Green($"Raised event {eventName} to instance {instanceId}."));
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                await _orchestrationService.RewindTaskOrchestrationAsync(instanceId, reason);
+            }
+            catch (ArgumentOutOfRangeException e)
+            {
+                // DurableTask.AzureStorage throws this error when it cannot find an orchestration instance matching the query
+                throw new CliException("Orchestration instance not rewound. Must have a status of 'Failed', or an EventType of 'TaskFailed' or 'SubOrchestrationInstanceFailed' to be rewound.");
+            }
+            
+            ColoredConsole.WriteLine($"Rewind message sent to instance {instanceId}. Retrieving new status now..");
```
cgillum:
```suggestion
            ColoredConsole.WriteLine($"Rewind message sent to instance {instanceId}. Retrieving new status now...");
```

src/Azure.Functions.Cli/Actions/DurableActions/DurableStartNew.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public override async Task RunAsync()
+        {
+            if (string.IsNullOrEmpty(FunctionName))
+            {
+                throw new CliArgumentsException("Must specify the name of of the orchestration function to start.",
+                    new CliArgument { Name = "function-name", Description = "Name of the orchestration function to start." });
+            }
+
+            await _durableManager.StartNew(FunctionName, Version, ID, Input);
```
cgillum: I believe there is a subtle bug here because `Input` is of type string but the method actually requires `object`. What I'm afraid will happen is that the someone will pass in a serialized version of the input they want and then the Durable client will serialize it again from `object` to `string`, resulting in double-serialization. To fix this, we should deserialize `Input` into `object` (displaying friendly error messages if the input is not valid JSON) and then pass the deserialized `object` to `_durableManager.StartNew(...)`.

src/Azure.Functions.Cli/Actions/DurableActions/DurableRaiseEvent.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public override async Task RunAsync()
+        {
+            if (string.IsNullOrEmpty(ID))
+            {
+                throw new CliArgumentsException("Must specify the id of the orchestration instance you wish to raise an event for.",
+                    new CliArgument { Name = "id", Description = "ID of the orchestration instance to raise an event for." });
+            }
+
+            await _durableManager.RaiseEvent(ID, EventName, EventData);
```
cgillum: Same comment I raised in the `StartNew` method about `Input` and double-serialization. In this case, we need to worry about `EventData`.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                instanceId = $"{Guid.NewGuid():N}";
+            }
+
+            await _client.CreateOrchestrationInstanceAsync(functionName, version, instanceId, input);
+
+            var status = await _client.GetOrchestrationStateAsync(instanceId, false);
+
+            if (status != null && status.Count > 0)
+            {
+                ColoredConsole.WriteLine(Green($"Started {status[0].Name} with new instance {status[0].OrchestrationInstance.InstanceId} at {status[0].CreatedTime}."));
```
cgillum: General feedback: string inputs from the user should be surrounded by quotes. This is mainly because the string could contain spaces, and then the full sentence becomes harder to read because you can't tell the input (e.g. the function name or the instance ID) from the sentence itself.

```suggestion
                ColoredConsole.WriteLine(Green($"Started '{status[0].Name}' with new instance '{status[0].OrchestrationInstance.InstanceId}' at {status[0].CreatedTime}."));
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+            else
+            {
+                throw new CliException($"Could not start new instance {instanceId}.");
```
cgillum:
```suggestion
                throw new CliException($"Could not start new instance '{instanceId}'.");
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            await _client.TerminateInstanceAsync(orchestrationInstance, reason);
+
+            var status = await _client.GetOrchestrationStateAsync(instanceId, false);
+
+            if (status != null && status.Count > 0)
+            {
+                if (status[0].OrchestrationStatus == OrchestrationStatus.Running)
+                {
+                    ColoredConsole.WriteLine($"Found & terminated instance {instanceId}. Waiting 10 seconds for 'Terminated' status..");
```
cgillum:
```suggestion
                    ColoredConsole.WriteLine($"Sent a termination message to instance '{instanceId}'. Waiting up to 30 seconds for the orchestration to actually terminate...");
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                    // If it's still marked as running, wait a little bit and then poll again
+                    await Task.Delay(10000);
+                    status = await _client.GetOrchestrationStateAsync(instanceId, false);
```
cgillum: Do this instead (and remove the comment):
```csharp
status = await _client.WaitForOrchestrationAsync(
    instanceId,
    executionId: null,
    timeout: TimeSpan.FromSeconds(30),
    cancellationToken: CancellationToken.None)
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,180 @@
...
+                }
+
+                if (status?[0]?.OrchestrationStatus == OrchestrationStatus.Terminated)
+                {
+                    ColoredConsole.WriteLine(Green($"Successfully terminated {instanceId}"));
+                }
+                else
+                {
+                    throw new CliException($"Instance did not terminate within 10 seconds. Current status: {status[0].OrchestrationStatus}");
```
cgillum:
```suggestion
                    throw new CliException($"Instance did not terminate within the given timeout. Current status: {status[0].OrchestrationStatus}.");
```

src/Azure.Functions.Cli/Context.cs
```diff
@@ -34,7 +34,10 @@ internal enum Context
         Templates,
 
         [Description("Commands for installing extensions")]
-        Extensions
+        Extensions,
+
+        [Description("Commands for working with Durable Functions locally")]
```
cgillum:
```suggestion
        [Description("Commands for working with Durable Functions")]
```

src/Azure.Functions.Cli/Interfaces/IDurableManager.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Azure.Functions.Cli.Interfaces
+{
+    internal interface IDurableManager
+    {
+        Task DeleteHistory();
+
+        Task GetHistory(string instanceId);
+
+        Task GetRuntimeStatus(string instanceId, bool getAllExecutions);
```
cgillum: Let's remove the `getAllExecutions` parameter since that is not relevant to Durable Functions.

src/Azure.Functions.Cli/Interfaces/IDurableManager.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading.Tasks;
+
+namespace Azure.Functions.Cli.Interfaces
+{
+    internal interface IDurableManager
+    {
+        Task DeleteHistory();
```
cgillum: `DeleteHistory()` should be renamed to `DeleteTaskHub()`.

cgillum: Separately, let's add a new command called `PurgeHistory` which internally calls `PurgeOrchestrationHistoryAsync` (new in v1.3.2).

src/Azure.Functions.Cli/Interfaces/IDurableManager.cs
```diff
@@ -0,0 +1,21 @@
...
+        Task GetRuntimeStatus(string instanceId, bool getAllExecutions);
+
+        Task RaiseEvent(string instanceId, string eventName, object eventData);
+
+        Task Rewind(string instanceId, string reason);
+
+        Task StartNew(string functionName, string version, string instanceId, object input);
+
+        Task Terminate(string instanceId, string reason);                
+    }
```
cgillum: Two commands I think we should add, both of which are new in **DurableTask.AzureStorage v1.3.2**):
* **GetInstances**: Internally gets the status of all instances with paging (the longest overload of `AzureStorageOrchestrationService.GetOrchestrationStateAsync`). By default, we should get all instances (from `DateTime.MinValue` to `DateTime.MaxValue`) and specify `top` as `10`.
* **PurgeHistory**: Internally calls `AzureStorageOrchestrationService.PurgeOrchestrationHistoryAsync`.

cgillum:
@jeffhollan I suggest you take a look at this as well from a public surface area perspective.

ahmelsayed:
The only comment from my side is that you seem to have similar `ID` not null validation in few classes and all of them drive from `BaseDurableAction`. I'd think you can probably move that into the `ParseArgs` method of these classes. usually if you mark an arg as `.Required()` that will cause an error saying `arg X is required` but if you want to customize the error like you have you can do 

```cs
var parseResult = base.ParseArgs(args);
// Your string.IsNullOrEmpty(ID) validation code
return parseResult;
```

note that the actual parsing won't happen until you call `base.ParseArgs()` so make sure to call that first then check ID and return the result after that.

ahmelsayed:
@glennamanns can you please also add few test cases for this? It should be straight forward to add a [`ResolveCommandLineCorrectly`](https://github.com/Azure/azure-functions-core-tools/blob/master/test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs#L20-L55) test case. You just need to add an attribute with what you expect your commands to be and the type you expect them to resolve to. This will help catch any regressions if we change the resolution logic.


Also it would be great if you can add an end to end test, some examples [are here](https://github.com/Azure/azure-functions-core-tools/blob/master/test/Azure.Functions.Cli.Tests/E2E/SettingsTests.cs). If your commands require a storage account however (I think `DurableManager` does, right?) You may wanna make them [skippable](https://github.com/Azure/azure-functions-core-tools/blob/5631ec74c205f0b716689347e783a48ddf3254a4/test/Azure.Functions.Cli.Tests/ExtensionsTests/ProcessExtensionsTests.cs#L14-L18) if there is no storage account env var defined, then we will need to add a test storage account on the CI to actually run through them. Let me know if you have any questions there.

src/Azure.Functions.Cli/Actions/DurableActions/DurableRaiseEvent.cs
```diff
@@ -0,0 +1,45 @@
...
+               .SetDefault(null)
+               .Callback(d => EventData = d);
+
+            return base.ParseArgs(args);
+        }
+
+        public override async Task RunAsync()
+        {
+            dynamic input = DurableManager.DeserializeInstanceInput(EventData);
+            await _durableManager.RaiseEvent(Id, EventName, EventData);
```
cgillum:
```suggestion
            await _durableManager.RaiseEvent(Id, EventName, input);
```

src/Azure.Functions.Cli/Actions/DurableActions/DurableStartNew.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Azure.Functions.Cli.Actions.DurableActions
+{
+    [Action(Name = "start-new", Context = Context.Durable, HelpText = "Start a new instance of the specified orchestrator function")]
+    class DurableStartNew : BaseAction
+    {
+        private string FunctionName { get; set; }
+
+        private string Input { get; set; }
+
+        private string Version { get; set; }
```
cgillum: Can this `Version` property be removed?

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,294 @@
...
+            {
+                TaskHubName = TaskHubName,
+                StorageConnectionString = ConnectionString,
+            };
+
+            _orchestrationService = new AzureStorageOrchestrationService(settings);
+            _client = new TaskHubClient(_orchestrationService);
+        }
+
+        // HELP WANTED: Is there a better way to do this?? (parse host.json for durable settings)
```
cgillum: @ahmelsayed is Glenna's approach for parsing host.json below the right thing to do?

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,294 @@
...
+{
+    internal class DurableManager : IDurableManager
+    {
+        private readonly AzureStorageOrchestrationService _orchestrationService;
+
+        private readonly TaskHubClient _client;
+
+        private string TaskHubName;
+
+        private string ConnectionString;
```
cgillum: It would be better to make these `readonly` and to follow existing naming conventions, e.g.
```csharp
private readonly string _taskHubName;
private readonly string _connectionString;
```
And then I think you can do this to assign them.
```csharp
this.SetConnectionStringAndTaskHubName(secretsManager, ref _taskHubName, ref _connectionString);
```

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,294 @@
...
+                    connectionStringKey = hostSettings.extensions.durableTask.AzureStorageConnectionStringName ?? connectionStringKey;
+                    TaskHubName = hostSettings.extensions.durableTask.HubName ?? TaskHubName;
+                }
+                else
+                {
+                    connectionStringKey = hostSettings.durableTask.AzureStorageConnectionStringName ?? connectionStringKey;
+                    TaskHubName = hostSettings.durableTask.HubName ?? TaskHubName;
+                }
+            }
+            catch { }
```
cgillum: I recommend at least printing the `Message` property of the exception as a warning (yellow) so that the user can know that something went wrong and report it to us.

src/Azure.Functions.Cli/Actions/DurableActions/DurableDeleteTaskHub.cs
```diff
@@ -0,0 +1,43 @@
...
+            Parser
+                .Setup<string>("connection-string")
+                .WithDescription("(Optional) Storage connection string to use.")
+                .SetDefault(null)
+                .Callback(n => ConnectionString = n);
+            Parser
+                .Setup<bool>("delete-instance-store")
+                .WithDescription("If set to false, the instance store will not be deleted. The default value is true.")
+                .SetDefault(true)
+                .Callback(n => DeleteInstanceStore = n);
```
cgillum: Let's remove this parameter. It should always be `true`.

src/Azure.Functions.Cli/Actions/DurableActions/DurableRaiseEvent.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                 .Setup<string>("event-name")
+                 .WithDescription("Name of the event to raise")
+                 .SetDefault($"Event_{(Guid.NewGuid().ToString("N"))}")
+                 .Callback(n => EventName = n);
+            Parser
+               .Setup<string>("event-data")
+               .WithDescription("Data to pass to the event, either in-line or via a JSON file. Prefix the path to the file with @ (@path/to/file.json).")
```
cgillum:
```suggestion
               .WithDescription("Data to pass to the event, either in-line or via a JSON file. For files, prefix the path to the file with @ (e.g. "@path/to/file.json").")
```

glennamanns:
@ahmelsayed With regard to adding new E2E, I have been unable to run any of the existing E2E tests for the CLI. 

All of them give me the same error: "[**xUnit**.net 00:01:24.86]         stderr:          Unhandled Exception: System.IO.FileNotFoundException: **Could not load file or assembly 'System.Runtime,** Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified."

This is the case on both my laptop and my desktop with fresh clones. Are you able to run the tests on your machine? Did you have to make any changes to do so?

ahmelsayed:
Sorry I should have explained how to run them locally. 
The simpleist way to run all the tests is to do

```
cd build
dotnet run Test
```

This will build and run all tests. if you want to run individual tests yourself you'll need to set an env var `FUNC_PATH=D:\where\you\have\func.exe` 

you can do that by 

```bash
cd build
dotnet run AddTemplatesNupkgs
cd ..

# if you're using CMD
set FUNC_PATH="D:\Path\To\Your\Cli\Repo\artifacts\win-x86\func.exe"

# if you're using Powershell
$Env:FUNC_PATH="D:\Path\To\Your\Cli\Repo\artifacts\win-x86\func.exe"

dotnet test .\test\Azure.Functions.Cli.Tests\Azure.Functions.Cli.Tests.csproj --filter E2E # or you can replace E2E with your test name
```

Basically you need to build `func.exe` to be tested, set that env var and then run the tests. Or you can change it here https://github.com/Azure/azure-functions-core-tools/blob/master/test/Azure.Functions.Cli.Tests/E2E/Helpers/CliTester.cs#L16

test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
```diff
@@ -52,6 +53,15 @@ public class ResolveActionTests
         [InlineData("azure functionapp --help", typeof(HelpAction))]
         [InlineData("azure --help", typeof(HelpAction))]
         [InlineData("--help", typeof(HelpAction))]
+        [InlineData("durable delete-task-hub --connection-string connection", typeof(DurableDeleteTaskHub))]
+        [InlineData("durable get-history --id 1234 --show-input true", typeof(DurableGetHistory))]
```
cgillum: Instead of `1234` as the example, I suggest using a real GUID.

cgillum: It's perfectly fine to use a hardcoded one. I just want the reader to recognize that most often these will be GUID values.

glennamanns:
@ahmelsayed Hey Ahmed, I added end to end tests for all of the actions. 

Some things to note: 
- Some of the actions/tests require the Functions host to be running, so I added the ability for tests to have the host running in the background during them. 
- I added two new environment variables to the build project.
- I added a new function app test folder, DurableTestFolder, that the host will start/run from. I'm not sure this is the best approach. The start host tests use the `init` command, but we don't have any durable templates right now.

ahmelsayed:
Thank you very much Glenna for adding those test cases! They look good! 

One final remark, it'd be better if you can move that `DurableTestFolder` somewhere under the test project, like under `test\Azure.Functions.Cli.Tests\Resources\DurableTestFolder` or something like that. But other than that the change looks good to me. If everything looks good to @cgillum as well, we can merge it

test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
```diff
@@ -54,14 +54,14 @@ public class ResolveActionTests
...
-        [InlineData("durable terminate --id 1234", typeof(DurableTerminate))]
+        [InlineData("durable get-history --id ab1b5de4c5a9450eafa5a4cf249b3169 --show-input true --connection-string connection", typeof(DurableGetHistory))]
+        [InlineData("durable get-instances --created-before 11/10/2018 --connection-string connection", typeof(DurableGetInstances))]
+        [InlineData("durable get-runtime-status --id ab1b5de4c5a9450eafa5a4cf249b3169 --connection-string connection", typeof(DurableGetRuntimeStatus))]
+        [InlineData("durable purge-history --created-after 11/5/2018 --connection-string connection", typeof(DurablePurgeHistory))]
+        [InlineData("durable raise-event --id ab1b5de4c5a9450eafa5a4cf249b3169 --connection-string connection", typeof(DurableRaiseEvent))]
+        [InlineData("durable rewind --id ab1b5de4c5a9450eafa5a4cf249b3169 --connection-string connection", typeof(DurableRewind))]
+        [InlineData("durable start-new --function-name functostart --connection-string connection", typeof(DurableStartNew))]
+        [InlineData("durable terminate --id ab1b5de4c5a9450eafa5a4cf249b3169 --connection-string connection", typeof(DurableTerminate))]
```
cgillum: Instead of `connection` use `connectionName` as the example to help prevent users from mistakenly put a real connection string here.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -101,19 +106,51 @@ private void SetStorageServiceAndTaskHubClient(ref AzureStorageOrchestrationServ
...
+        private void Initialize(ref AzureStorageOrchestrationService orchestrationService, ref TaskHubClient taskHubClient, string connectionStringKey = null, string taskHubName = null)
+        {
+            CheckAssemblies();
+            SetStorageServiceAndTaskHubClient(ref _orchestrationService, ref _client, connectionStringKey, taskHubName);
+        }
+
+        private void CheckAssemblies()
+        {
+            // Retrieve list of DurableTask.AzureStorage DLL files
+            var assemblyFilePaths = Directory.GetFiles(Environment.CurrentDirectory, $"{DurableAzureStorageExtensionName}", SearchOption.AllDirectories).ToList<string>();
```
cgillum: Two things:
1. You don't need the `$"{DurableAzureStorageExtensionName}"`, just use `DurableAzureStorageExtensionName`.
2. Don't call `.ToList<string>()`, that's just going to copy one array to another. Use the `string[]` result directly.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -101,19 +106,51 @@ private void SetStorageServiceAndTaskHubClient(ref AzureStorageOrchestrationServ
...
+            
+            if (assemblyFilePaths.Count() == 0)
+            {
+                ColoredConsole.WriteLine(Yellow($"Could not find {DurableAzureStorageExtensionName}. The functions host must be running a" +
+                    $" Durable Functions app in order for Durable Functions CLI commands to work."));
+                return;
+            }
+
+            // Convert them to a sorted list of FileInfo objects. Sort by the last modified time of the enclosing bin folder
+            var assemblyFiles = assemblyFilePaths.Select(assemblyFilePath => new FileInfo(assemblyFilePath)).ToList().OrderBy(file => file.Directory.LastWriteTime);
```
cgillum: Remove `.ToList()`. That's unnecessary overhead.

Also, instead of returning a list of assembly files, why not just put the `.Last()` at the end? It seems having the list in a local variable is not useful.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -101,19 +106,51 @@ private void SetStorageServiceAndTaskHubClient(ref AzureStorageOrchestrationServ
...
+
+            var minimumExtensionVersion = new Version(MinimumDurableAzureStorageExtensionVersion);
+
+            var mostRecentAssembly = FileVersionInfo.GetVersionInfo(assemblyFiles.LastOrDefault().FullName);
```
cgillum: You don't want `.LastOrDefault()` because that implies that you expect there could be a `null` value (in which case you would hit a `NullReferenceException`). Instead you should just use `Last()` to make it clear that you expect there to be a list item in the list (and if that's not the case, the error message is more clear).

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -101,19 +106,51 @@ private void SetStorageServiceAndTaskHubClient(ref AzureStorageOrchestrationServ
...
+            var mostRecentAssemblyVersion = new Version(mostRecentAssembly.FileVersion);
+
+            if (mostRecentAssemblyVersion < minimumExtensionVersion)
+            {
+                 throw new CliException($"Durable Functions CLI commands must be used with {DurableAzureStorageExtensionName} versions greater than or equal to 1.4.0." +
```
cgillum: Use your `MinimumDurableAzureStorageExtensionVersion` constant here.

Also, you should include the file path in this exception method in case that the CLI picks up a file that the user wasn't expecting it to pick up.

test/Azure.Functions.Cli.Tests/E2E/DurableTests.cs
```diff
@@ -0,0 +1,297 @@
...
+            string taskHubName = "raiseEventTest";
+            DurableHelper.SetTaskHubName(WorkingDirPath, taskHubName);
+            Environment.SetEnvironmentVariable(DurableManager.DefaultConnectionStringKey, StorageConnectionString);
+
+            await CliTester.Run(new RunConfiguration
+            {
+                Commands = new[]
+                {
+                    $"durable start-new --function-name Counter --input 3 --id {instanceId} --task-hub-name {taskHubName}",
+                    $"durable raise-event --id {instanceId} --event-name operation --event-data add --task-hub-name {taskHubName}"
```
cgillum: Can we also have a version of this test which takes a file input?

test/Azure.Functions.Cli.Tests/E2E/DurableTests.cs
```diff
@@ -0,0 +1,297 @@
...
+
+            string taskHubName = "startNewTest";
+            DurableHelper.SetTaskHubName(WorkingDirPath, taskHubName);
+            Environment.SetEnvironmentVariable(DurableManager.DefaultConnectionStringKey, StorageConnectionString);
+
+            await CliTester.Run(new RunConfiguration
+            {
+                Commands = new[]
+                {
+                    $"durable start-new --function-name Counter --input 3 --task-hub-name {taskHubName}"
```
cgillum: Can we also have a version of this test which takes a file input?

cgillum:
@fabiocav @ahmelsayed let us know if there's anything else you'd like for this PR. It's otherwise ready to go.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -317,7 +317,7 @@ public async Task Terminate(string connectionStringKey, string taskHubName, stri
             }
         }
 
-        public static dynamic DeserializeInstanceInput(string input)
+        public static string RetrieveCommandInputData(string input)
```
cgillum: Why was this change needed? I think the previous version was actually correct (parsing the contents of the file and returning it as either a `dynamic` or an `object`). Otherwise this CLI command will only support string inputs but not numbers, arrays, or complex types.

test/Azure.Functions.Cli.Tests/Resources/DurableTestFolder/Counter.cs
```diff
@@ -71,5 +73,25 @@ public static async Task Increment([OrchestrationTrigger] DurableOrchestrationCo
...
         }
+
+        [FunctionName("JsonInput")]
+        public static async Task JsonInput([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            string input = ctx.GetInput<string>();
+
+            if (!string.IsNullOrEmpty(input))
+            {
+                JObject testParsingInput = JObject.Parse(input);
```
cgillum: As written, it seems the CLI command requires users to take strings as inputs and deserialize them into the object they want in their function code. But this is a hassle and the `DurableOrchestrationContext` is supposed to do that for them. I suggest we change the orchestration code to look more like your test code, e.g.:
```csharp
[FunctionName("HelloWorldOrchestrator")]
public static string HelloWorldOrchestrator([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    var input = ctx.GetInput<dynamic>();
    return $"{input.Hello}, {input.Name}";
}
```
This will require you to make the change to your input parsing method, which I mentioned earlier.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        public async Task GetHistory(string connectionStringKey, string taskHubName, string instanceId)
+        {
+            Initialize(ref _orchestrationService, ref _client, connectionStringKey, taskHubName);
+
+            var historyString = await _orchestrationService.GetOrchestrationHistoryAsync(instanceId, null);
+
+            JArray history = JArray.Parse(historyString);
+
+            JArray chronological_history = new JArray(history.OrderBy(obj => (string)obj["TimeStamp"]));
```
fabiocav: nit: naming... `chronologicalHistory` maybe?

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,362 @@
...
+            foreach (JObject jobj in chronological_history)
+            {
+                // Convert EventType enum values to their equivalent string value
+                var parsed = Enum.TryParse(jobj["EventType"].ToString(), out EventType eventName);
```
fabiocav: `parsed` doesn't seem to be used.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,362 @@
...
+                status.Output = (showOutput) ? status.Output : null;
+                ColoredConsole.WriteLine(JsonConvert.SerializeObject(status, Formatting.Indented));
+            }            
+        }
+
+         public async Task PurgeHistory(string connectionStringKey, string taskHubName, DateTime createdAfter, DateTime createdBefore, IEnumerable<OrchestrationStatus> runtimeStatuses)
+        {
+            Initialize(ref _orchestrationService, ref _client, connectionStringKey, taskHubName);
+
+            var stats = await _orchestrationService.PurgeInstanceHistoryAsync(createdAfter, createdBefore, runtimeStatuses);
```
fabiocav: Run through the IEnumerable passed in early (i.e. have a local and call `ToArray()` on it) and pass that to both methods using it below (`PurgeInstanceHistoryAsync` and `string.Join`)

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,362 @@
...
+                }
+            }
+            catch (Exception e)
+            {
+                ColoredConsole.WriteLine(Yellow($"Exception thrown while attempting to parse override connection string and task hub name from '{ScriptConstants.HostMetadataFileName}':"));
+                ColoredConsole.WriteLine(Yellow(e.Message));
+            }     
+        }
+
+        private void SetStorageServiceAndTaskHubClient(ref AzureStorageOrchestrationService orchestrationService, ref TaskHubClient taskHubClient, string connectionStringKey = null, string taskHubName = null)
```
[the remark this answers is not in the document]
glennamanns: This method is executed before any of the action methods in order to initialize the `AzureStorageOrchestrationService` and `TaskHubClient`. They're `ref` parameters because they modify DurableManager's `_orchestrationService` and `_client` fields. Those fields are used later by all of the action methods.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -0,0 +1,362 @@
...
+                _orchestrationService = new AzureStorageOrchestrationService(settings);
+                _client = new TaskHubClient(orchestrationService);
+            }
+            else
+            {
+                throw new CliException("No storage connection string found.");
+            }
+        }
+
+        private void Initialize(ref AzureStorageOrchestrationService orchestrationService, ref TaskHubClient taskHubClient, string connectionStringKey = null, string taskHubName = null)
```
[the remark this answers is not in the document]
glennamanns: see comment on `SetStorageServiceAndTaskHubClient`

src/Azure.Functions.Cli/Common/DurableManager.cs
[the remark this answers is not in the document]
fabiocav: With the changes in the state I was reviewing, this was never assigned to within the scope of this method. So if the expectation is that this would be set somehow, then there's a bug in the method. We can follow up in person if needed.

src/Azure.Functions.Cli/Common/DurableManager.cs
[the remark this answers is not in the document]
fabiocav: Not sure I understand. If they're not used/assigned, this wouldn't make a difference here.

src/Azure.Functions.Cli/Common/DurableManager.cs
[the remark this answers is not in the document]
fabiocav: I see you've updated the code to assign to that value and change this to an `out` to avoid the issue that existed before. So this looks good now.

src/Azure.Functions.Cli/Common/DurableManager.cs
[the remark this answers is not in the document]
fabiocav: Same as above, issue was addressed with the last commit.

ahmelsayed:
Thank you @cmkerner20 and @glennamanns!

PR #843, opened 2018-11-02
Add ability to allow using cached dependencies for python

ankitkumarr:
Do not merge yet as this still needs to be tested thoroughly.

To test this, if you use `--build-native-deps`, be sure to also use `--no-bundler` because we cannot use package caching when bundling the application.

The link to the `cached_requirements.txt` will not work because the file is not published yet. So, I recommend hosting [this](https://github.com/Azure/azure-functions-docker/pull/50/files#diff-8cc79e5d7ae45669e82f7c75401e43e7) elsewhere, changing the code and then testing it.

Reference - https://github.com/Azure/azure-functions-docker/pull/50

Next Steps:
- Add messages to let users know that if they use different versions of certain packages, we can use cached ones and hence save the size.
- Take care of `package` vs `package==<version>` in `requirements.txt`


ankitkumarr:
@ahmelsayed, Closing this PR and the associated docker PR as we probably will not go this path any further.

Issue #696, opened 2018-09-05
Bug when deploying large files


ColbyTresness:
![deploybug](https://user-images.githubusercontent.com/7645687/45116604-1e3d7900-b108-11e8-8bf4-459c64941670.png)

ahmelsayed:
There are 2 distinct issues here:

1. I'm loading the full zip into memory before sending it. This can cause the out of memory exception if the file is large than 2GB with a 32bit func.exe.

2. reusing a disposed stream.  This is fixed by d563a2c

I still need to look at #1

PR #844, opened 2018-11-02
Fixes #839 - Remove .python_packages directory after packaging

### LINKED ISSUE #839: Remove .python_packages after packaging

ankitkumarr:


src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -161,7 +161,9 @@ internal static async Task<Stream> GetPythonDeploymentPackage(IEnumerable<string
         private static async Task<Stream> InternalPreparePythonDeployment(IEnumerable<string> files, string functionAppRoot)
         {
             var packagesPath = await RestorePythonRequirements(functionAppRoot);
-            return ZipHelper.CreateZip(files.Concat(FileSystemHelpers.GetFiles(packagesPath)), functionAppRoot);
+            Stream zipStream = ZipHelper.CreateZip(files.Concat(FileSystemHelpers.GetFiles(packagesPath)), functionAppRoot);
+            FileSystemHelpers.DeleteDirectorySafe(packagesPath);
```
ahmelsayed: I'm not sure this would work. I thought the zip stream is a lazy stream. I think it'd be safer to delete that folder in `RestorePythonRequirements` if it already exists before creating it

ankitkumarr: I was initially wondering the same thing. But, I actually tested it multiple times and it seems to always work fine, so I went ahead with it. Maybe it's because how we are using a memorystream and copying it to the zip stream for each file that makes it store in the memory?

But, if you would rather want me to be safe, I can remove it elsewhere. I would say it might be better to do it in `GetPythonDeploymentPackage`. Because if the user did `publish --build-native-deps`, we don't go through `RestorePythonRequirements`. 
So, if the user ran `func pack` and then `func pack --build-native-deps`, they would have the large .`python_packages` in the zip file for no reason.

ahmelsayed: I see. I think you might be right about loading all the files in a MemoryStream because I think that's the underlying issue of this https://github.com/Azure/azure-functions-core-tools/issues/696 

I'm fine with whatever makes sense to you

ankitkumarr: On a second thought, is there a reason why we create `.python_packages` directory in the functionapp directory over doing so in a temp location, similar to what we do for `--build-native-deps`?

ankitkumarr: I see. For now,  I just updated the PR to delete if it's present anytime `pack` or `publish` is called on a python project.

That should cover most of the scenarios.

PR #861, opened 2018-11-09
Add telemetry

ankitkumarr:
First take at adding basic telemetry to the CLI. 
I will be performing tests to make sure we log most of the scenarios possible and that we don't log any PII.

To test, please add the Instrumentation Key for you App Insights app. 

UPDATE - 
Here's how the Telemetry information looks in a basic command prompt-
telemetryoptoutlook

UPDATE 2- 
Here's the other PRs for the build systems we have-
https://github.com/Azure/homebrew-functions/pull/15
https://github.com/ahmelsayed/publish-script/pull/1

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/TelemetryHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/TelemetryHelpers.cs
@@ -1,0 +1,91 @@
+﻿using Azure.Functions.Cli.Common;
+using Fclp.Internals;
+using Microsoft.ApplicationInsights;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Azure.Functions.Cli.Helpers
+{
+    internal static class TelemetryHelpers
+    {
+        internal class ConsoleAppLogEvent
+        {
+            public string CommandName { get; set; }
+            public string IActionName { get; set; }
+            public IEnumerable<string> Parameters { get; set; }
+            public bool PrefixOrScriptRoot { get; set; }
+            public bool IsSuccessful { get; set; }
+            public bool ParseError { get; set; }
+            public long TimeTaken { get; set; }
+        }
+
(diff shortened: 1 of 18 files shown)


src/Azure.Functions.Cli/Helpers/TelemetryHelpers.cs
```diff
@@ -0,0 +1,93 @@
...
+            StringBuilder sb = new StringBuilder();
+            // Append the bytes in hexadecimal form to the string builder
+            hashedID.ToList().ForEach(bt => sb.Append(bt.ToString("x2")));
+            return sb.ToString();
+        }
+
+        public static void LogEventIfAllowed(ConsoleAppLogEvent consoleEvent)
+        {
+            // This should be injected during build
+            var instrumentationKey = Guid.Empty;
```
ahmelsayed: It'll be easier to inject the value if you set it to

```csharp
var instrumentationKey  = Guid.Parse("00000000-0000-0000-0000-000000000000");
```

ahmelsayed: or the actual string value can also come from a const in the Constants class

ahmelsayed:
Couple of more things to do sorry :) 

Can you update the [npm install script](https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/npm/lib/install.js#L66-L70) to display information about the opt-out? Check with @ahmedelnably about the exact wording. We can probably use similar wording to dotnet cli. 

and also update the build script by adding a new task before [`DotnetPublish`](https://github.com/Azure/azure-functions-core-tools/blob/master/build/Program.cs#L16) that does a string replace on that file with the right env var.

ankitkumarr:
Here's how the Telemetry information looks in a basic command prompt-

telemetryoptoutlook

ankitkumarr:
It might be nice to include the version of the core-tools with each telemetry entry.

riverar:
Opt-in telemetry like this isn't GDPR compliant, beware.

ankitkumarr:
Closing in favor of https://github.com/Azure/azure-functions-core-tools/pull/1532

Issue #840, opened 2018-11-01
Don't rely on Azure CLI to get access token


TylerLeonhardt:
I was looking at how Core Tools depends on the Azure CLI and I noticed that the only thing it's using it for is to get the access token:

https://github.com/Azure/azure-functions-core-tools/blob/60ea24fd4408ea876882324a9505c2d43bb37726/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs#L72-L96

`az account get-access-token`

A big and quick win for Core Tools would be to get off this dependency so that the entire Azure CLI isn't needed. This will pave the way for when Azure Functions has PowerShell cmdlets for Azure PowerShell.

tylerl0706:
cc @asavaritayal @pragnagopa @joeyaiello @stevel-MSFT

monteledwards:
personally,  i am in favor of this approach and admire it.  let az cli be the central point for authentication. people in azure typically have multiple subscriptions and if i have to set my context on each tool that's not fun,

tylerl0706:
@monteledwards I understand where you're coming from... But az cli and Azure PowerShell don't share the same context. 

There are a large set of users that use Azure PowerShell instead of Azure CLI and to force those users to download another tool just to get an Auth token to use isn't a good practice.

The func cli can follow the standard OAuth 2.0 flow to retrieve the access token from Azure.

ahmelsayed:
That used to be how it worked but there were too many issues trying to do my own authentication and supporting all the various methods and setups Azure has for login so I changed it in d776e404643776ce823e95f6fa35d48252f3d223. Not to mention as @monteledwards mentioned, if I'm doing my own login I have to allow the user to select the account, tenant, subscription, etc which is confusing because it's yet another tool that you'll have to configure that in.  I don't think the core tools shouldn't be in the business of managing Azure resources or login/logout. If you don't have az or don't wanna use az you can always 

```
func ... --access-token {accessToken}

# OR 

echo $token | func ... --access-token-stdin
```

I meant to add support for azure powershell if it's installed to get the token from there as well, but I wasn't clear on how to implement that so I didn't

tylerl0706:
I still don't like the idea of relying on an external tool to retrieve an access token... but I was messing with doing it in Azure PowerShell last week and think this works:

```powershell
$context = Get-AzContext; $context.TokenCache.ReadItems() | ? TenantId -eq $context.TenantId | % { $_.AccessToken }
```

[...]

PR #873, opened 2018-11-14
Inital Azure Powershell support

TylerLeonhardt:
This would allow users to have Azure PowerShell installed and not az CLI.

I've verified this on linux. Need to do proper testing and see if there's an opportunity for tests to be added.

cc @ahmelsayed @stevel-msft @daxian-dbw @asavaritayal @joeyaiello @Francisco-Gamino 

fixes https://github.com/Azure/azure-functions-core-tools/issues/840

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -5,7 +5,9 @@ using System.Text;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Arm;
 using Azure.Functions.Cli.Common;
+using static Azure.Functions.Cli.Common.OutputTheme;
 using Azure.Functions.Cli.Interfaces;
+using Colors.Net;
 using Fclp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -102,6 +104,13 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 {
                     return (true, stdout.ToString().Trim(' ', '\n', '\r', '"'));
                 }
+                else
+                {
+                    if (StaticSettings.IsDebug)
+                    {
+                        ColoredConsole.WriteLine(VerboseColor($"Unable to fetch access token from az cli. Error: {stderr.ToString().Trim(' ', '\n', '\r')}"));
+                    }
+                }
             }
             return (false, null);
         }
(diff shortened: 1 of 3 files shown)


TylerLeonhardt:
Alright I think this is worth looking at now...

This will try to grab the access token using Az or AzureRM if it's available. The one limitation that I didn't want to deal with was...

If Azure CLI fails for some reason (other than not existing) - try Azure PowerShell

if you think this is important, I can refactor but at least this will give those that ONLY have Azure PowerShell an opportunity to use func cli to publish

TylerLeonhardt:
Also, I'm not against NOT adding AzureRM support if someone suggests it. Having to deal with Az OR AzureRM made the code a bit complex.

ahmelsayed:
I think it's fine in it's current shape, I guess it leaves the scenario of someone having both on their machine, but only really using powershell. You can handle that scenario if you refactor the 2 places that return the token into something like

```csharp
string token = null;
if (TryGetAzToken(out token) || TryGetPowershellToken(out token))
{
    return token;
}

// throw error to make sure either az or powershell is installed and logged in;
```

You can do additional logging for the output of both if `StaticSettings.IsDebug` to help debug any potential issues.

ahmelsayed:
But thank you for the PR :) I originally wanted to [support both](https://github.com/Azure/azure-functions-core-tools/blob/d776e404643776ce823e95f6fa35d48252f3d223/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs#L67-L120) but couldn't figure out how to do it for powershell

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -14,6 +14,31 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 {
     abstract class BaseAzureAction : BaseAction, IInitializableAction
     {
+        private const string _getAccessTokenPowerShellScript = @"
```
daxian-dbw: This variable is not used anywhere now.

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -89,10 +114,89 @@ private async Task<string> AzureCliGetToken()
...
+            {
+                throw new ArgumentException($"Expected module to be '{_azProfileModuleName}' or '{_azureRmProfileModuleName}'");
+            }
+
+            // This PowerShell script first grabs the Azure context, fetches the profile client and requests an accesstoken.
+            // This entirely done using the Az.Profile module or AzureRM.Profile
+            return $@"
+$currentAzureContext = Get-{prefix}Context;
+$azureRmProfile = [Microsoft.Azure.Commands.Common.Authentication.Abstractions.AzureRmProfileProvider]::Instance.Profile;
+$profileClient = New-Object Microsoft.Azure.Commands.ResourceManager.Common.RMProfileClient($azureRmProfile);
```
daxian-dbw: Please use `New-Object <typename> $azureRmProfile` instead of `New-Object <typename>($azureRmProfile)`.

TylerLeonhardt:
@ahmelsayed I love the idea of `out` parameters... but you can't use `out` parameters with `async` functions :(

TIL.

Anyway...  I took that idea and leveraged Named Tuples instead.

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
TylerLeonhardt: fixed!

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
TylerLeonhardt: good catch... omnisharp on linux should have warned me about that...

TylerLeonhardt: (fixed)

TylerLeonhardt:
@ahmelsayed can you give me some details on how I could log? I didn't see any logger passed in

ahmelsayed:
>  but you can't use out parameters with async functions 

good point. 

> can you give me some details on how I could log? I didn't see any logger passed in

I just use `ColoredConsole.WriteLine()`, if you want to write stuff only for debug you can make it conditional on `StaticSettings.IsDebug`. There is no logger per se because I hate having to keep passing it all over the place as requirements keep shifting.

ahmelsayed:
oops wrong button.

TylerLeonhardt:
Oh ok - I'll give ColoredConsole a go. And how do I enable debug logging in the func cli?

TylerLeonhardt:
figured it out `$env:CLI_DEBUG = 1`

TylerLeonhardt:
ok logging added! :+1:

TylerLeonhardt:
I've tested this with the Az.Profile and AzureRM.Profile modules on both PowerShell Core and Windows PowerShell.

We should be good to go - unless you can point me to some place I can add tests

TylerLeonhardt:
Cool! I'll let @daxian-dbw and @SteveL-MSFT take a look if they want, otherwise, feel free to merge

PR #880, opened 2018-11-17
Fixes an issue with a false exception and AzureWebJobsStorage config.…

ltwlf:
Throwing an error when a Trigger is not HTTP and AzureWebJobsStorage is not configured is not correct. You can run other triggers without AzureWebJobsStorage for example: [TimerTrigger("0 */5 * * * *", UseMonitor=false)] or custom triggers. 
I replaced the exception with a warning.
Thanks and regards
Christian

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -341,8 +341,8 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
                 if (string.IsNullOrWhiteSpace(azureWebJobsStorage) && !allHttpTrigger)
                 {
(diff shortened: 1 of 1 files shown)


ahmelsayed:
I'd rather add a `--ignore-storage-check` parameter if you know that your trigger doesn't rely on storage than allowing people to proceed to then see an odd error from the runtime. This was added mainly for new users who try to use triggers other than Http (mainly timer with monitor enabled, which is the default) and then get odd errors later on and it can be hard for them to figure out what's happening.

chrisfowles:
A warning would potentially be more useful for new users using VS Code to develop functions, as the current behaviour when debugging is an unhelpful "The terminal process terminated with exit code: 1".

kshyju:
@ltwlf Is this PR still valid/relevant? If so, could you please fix the conflicts?

soninaren:
Closing this is a stale PR. Please open a new against v4.x if the change is still relevant

PR #884, opened 2018-11-20
Updating the webhost version to 2.0.0-12192 and java worker to 1.2.2

soninaren:


ahmelsayed:
update the minor version to 3 https://github.com/Azure/azure-functions-core-tools/blob/master/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj#L9

PR #888, opened 2018-11-26
Fixes #870 - Do not worry about Azure files in python

### LINKED ISSUE #870: Error with Azure Files
I thought we got rid of this error by force publishing by default -

```
Your app is configured with Azure Files for editing from Azure Portal.
To force publish use --force. This will remove Azure Files from your app.
```

ankitkumarr:


ahmelsayed:
so we no longer care about these 2 environment variables (`WEBSITE_CONTENTAZUREFILECONNECTIONSTRING` and `WEBSITE_CONTENTSHARE`) it's fine to have them or not have them. I think the code could be simplified based on that.

ankitkumarr:
@ahmelsayed I thought this change was only for Python. Or is it for all dynamic Linux apps? Because if it's only for Python, I am guessing we still want to check for the variables.

ahmelsayed:
I think Asavari only had python in mind with the issue, but it applies to all languages in dynamic linux apps.

Basically the logic used to be:

```
if dynamic linux:
    if has those 2 settings: 
        if --force:
            remove them // by setting them to null
         else:
            display some error
```


The reason was that "publishing" for dynamic linux apps is basically just setting RUN_FROM_ZIP app setting, but if all 3 app settings are present (the 2 above and this one) the platform used to favor the 2 above, so we had to remove to make publishing work. 



Basically we don't have to check or special case these settings any more. We can just delete all the code that handles them specially, and --force will be just a no-op in that scenario.

PR #894, opened 2018-11-27
Python public preview cli update

maiqbal11:
Updating CLI to latest versions of Python worker and library


maiqbal11:
@ahmelsayed Yes, we'll need one for docker as well. Will submit that PR as well

ahmelsayed:
I pushed an update to the docker repo before seeing your comment https://github.com/Azure/azure-functions-docker/commit/2a3a2836b7dd7cf34cb304a13bf9708152a54240

PR #898, opened 2018-11-28
Fixes 886 - Switch default for bundling

ankitkumarr:


ahmelsayed:
I'd rather we don't do this. Until we have a more efficient way of bundling these dependencies, I prefer we stick with the bundling logic we have, and work through individual reports of issues.

PR #905, opened 2018-11-30
Fixes #887 - Add preview flag to python

### LINKED ISSUE #887: Add (preview) flag to Python
Add **(preview)** flag to Python on func init.

ankitkumarr:
Adding `(preview)` while displaying the selection wizard on `func init` without affecting the parameter `--worker-runtime`. `func init --worker-runtime python` should still be valid.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
@@ -52,6 +52,24 @@ namespace Azure.Functions.Cli.Helpers
             .Where(k => k != WorkerRuntime.java)
             .Where(k => k != WorkerRuntime.powershell);
 
+        public static IDictionary<WorkerRuntime, string> GetWorkerToDisplayStrings()
+        {
+            IDictionary<WorkerRuntime, string> workerToDisplayStrings = new Dictionary<WorkerRuntime, string>();
+            foreach (WorkerRuntime wr in availableWorkersRuntime.Keys)
+            {
+                switch (wr)
+                {
+                    case WorkerRuntime.python:
+                        workerToDisplayStrings[wr] = "python (preview)";
+                        break;
+                    default:
+                        workerToDisplayStrings[wr] = wr.ToString();
+                        break;
+                }
+            }
+            return workerToDisplayStrings;
+        }
+
         public static WorkerRuntime NormalizeWorkerRuntime(string workerRuntime)
         {
             if (string.IsNullOrWhiteSpace(workerRuntime))
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -105,7 +105,9 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             else if (string.IsNullOrEmpty(WorkerRuntime))
             {
                 ColoredConsole.Write("Select a worker runtime: ");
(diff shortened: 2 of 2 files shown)


src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -42,6 +42,17 @@ public static class WorkerRuntimeLanguageHelper
             { WorkerRuntime.powershell, "powershell" }
         };
 
+        public static IDictionary<WorkerRuntime, string> WorkerToDisplayStrings => availableWorkersRuntime.Keys.ToDictionary(wr => wr, wr =>
```
fabiocav: nit: because of the lambdas, I think the expression bodied member here is impacting readability a bit, can we just switch and have a body instead?

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -42,6 +42,17 @@ public static class WorkerRuntimeLanguageHelper
...
+        {
+            switch (wr)
+            {
+                case WorkerRuntime.python:
+                    return "python (preview)";
```
fabiocav: Wondering if we should have a dictionary (WorkerRuntime/string) with runtime and display strings so we could properly format them (including proper casing and runtime name, like `.NET` instead of `dotnet`)

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
ankitkumarr: Sounds good, will update

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
ankitkumarr: That is a good call. I wanted to be least invasive and did not want to add another dictionary to maintain if a new WorkerRuntime is added or changed.
But, I think it should be ok as it's all in the same place and if you think we should change the display strings.

fabiocav: Yeah, I get where you were coming from, just wanted to share that thought as I think it will help us provide a better experience than just the enum `ToString()` result

ankitkumarr: The only other thing I could think of is we will either have to
1) have an inconsistency between `--worker-runtime` and the display strings. (`--worker-runtime dotnet` vs display string being `.NET`)
2) make changes to ensure params `--worker-runtime .NET` work.

fabiocav: That's a good point. I was looking at this as just plain human readable runtime names, but I guess people could use what's displayed for the parameter/setting value as well. That's actually a good argument for leaving this as is.

ankitkumarr: Yeah, that makes sense. Sounds good. I could open a separate PR that changes the display strings and accepts both, the display string and what we have now to be passed as `--worker-runtime`

fabiocav: Sounds good

ankitkumarr:
@fabiocav, I addressed your comment yesterday and force pushed. After removing the lambda, it didn't make sense to keep it a Getter property so I just made it to a function. 
Let me know if I misunderstood or didn't address it properly.

fabiocav:
Looks good. Feel free to merge

PR #924, opened 2018-12-11
Updating to 2.0.12246

pragnagopa:


pragnagopa:
@TylerLeonhardt - Please ping here if powershell worker needs to be updated as well.

PR #929, opened 2018-12-12
Fixes #926 - incorrect description for --nozip

### LINKED ISSUE #926: Incorrect description for publish command --nozip
As per the commit [bae202e](https://github.com/Azure/azure-functions-core-tools/commit/bae202eca284e0587f63d097042b1680ff64356c), `Run-From-Package` is the default deployment now and `--zip` was renamed to `--nozip`, which now disables this feature instead but the description still mentions otherwise.

ankitkumarr:
Part 2 of this PR is to update the documentation as well.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -73,7 +73,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 .Callback(f => ListIncludedFiles = f);
             Parser
                 .Setup<bool>("nozip")
-                .WithDescription("Publish in Run-From-Zip package. Requires the app to have AzureWebJobsStorage setting defined.")
+                .WithDescription("Turns the default Run-From-Package mode off.")
                 .SetDefault(false)
                 .Callback(f => RunFromZipDeploy = !f);
             Parser


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -73,7 +73,7 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .Callback(f => ListIncludedFiles = f);
             Parser
                 .Setup<bool>("nozip")
-                .WithDescription("Publish in Run-From-Zip package. Requires the app to have AzureWebJobsStorage setting defined.")
+                .WithDescription("Turns the default Run-From-Zip mode off")
```
fabiocav: nit: period

Question (perhaps outside of the scope of this change), but should the verbiage/flag here be `run-from-package` ?

ahmelsayed: yes I think it should be updated to say `Run-From-Package`

ankitkumarr: Sounds good. So, the flag would be `nopackage`? Or just the description and internal variable? Because if we change the flag to `--nopackage`, should we still continue to support `--nozip` (with a message) for a bit to not break?

ahmelsayed: I'd just change the description. Leave the flag as is because otherwise like you said we will have to support both.

ankitkumarr: Updated just the description

Issue #937, opened 2018-12-17
Cannot see Trace logs when loglevel set to Trace


paulbatum:
When debugging locally with core tools, I can't seem to get trace messages to appear in the console. Not sure if this is a host bug, a doc bug, or a core tools bug.

Steps to reproduce:
1. Create a new V2 function app in VS
2. Select HTTP trigger, paste in the following code:

```csharp
log.LogTrace("This is a trace log");
log.LogDebug("This is a debug log");
log.LogInformation("This is an information log");
log.LogWarning("This is a warning log");
log.LogError("This is an error log");
log.LogCritical("This is a critical log");

return new OkResult();
```

3. Go to host.json, configure as follows:

```
{
  "version": "2.0",
  "logging": {    
    "logLevel": {      
      "default": "Trace"
    }    
  }
}
```

4. Run the host, trigger the HTTP function

Expected result - trace message appears in the output

Actual result:
```
[12/17/2018 7:30:41 PM] Executing HTTP request: {
[12/17/2018 7:30:41 PM]   "requestId": "2f3c6515-8eb5-4d2d-b39b-3904e9b92baa",
[12/17/2018 7:30:41 PM]   "method": "GET",
[12/17/2018 7:30:41 PM]   "uri": "/api/Function1"
[12/17/2018 7:30:41 PM] }
[12/17/2018 7:30:41 PM] Executing 'Function1' (Reason='This function was programmatically called via the host APIs.', Id=b49bb61b-fc04-4193-86a8-84760a9672a2)
[12/17/2018 7:30:41 PM] This is an information log
[12/17/2018 7:30:41 PM] This is a warning log
[12/17/2018 7:30:41 PM] This is an error log
[12/17/2018 7:30:41 PM] This is a critical log
[12/17/2018 7:30:41 PM] Executed 'Function1' (Succeeded, Id=b49bb61b-fc04-4193-86a8-84760a9672a2)
```

paulbatum:
cc @ahmelsayed, @ankitkumarr

ahmelsayed:
@brettsam is this different from https://github.com/Azure/azure-functions-core-tools/issues/841 ?

brettsam:
Nope, that's the underlying issue. I'll close this one and assign myself to fix the original next sprint.

PR #940, opened 2018-12-18
Fix #841

### LINKED ISSUE #841: Console logging is hardcoded to only write Information and above
This should be using the host.json config but right now there's an explicit filter being pushed into the ColoredConsoleLoggerProvider. Logging this so I can come back and fix for next release.

brandonh-msft:
Remove hardcoded filter. This results in the expected output when testing with the code mentioned in #937

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,4 +1,9 @@
 #!/usr/bin/env pwsh
+param(
+    # The subscription id where the template will be deployed.
+    [Parameter()]
+    [string]$config = "Release"
+)
 dotnet build Azure.Functions.Cli.sln
 
 $outDir = "$([System.IO.Path]::GetTempPath())cli"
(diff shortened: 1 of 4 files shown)


brandonh-msft:
⏲

build.ps1
```diff
@@ -9,7 +14,7 @@ if ($IsMacOS) { $runtime = 'osx-x64' }
 elseif ($IsLinux) { $runtime = 'linux-x64' }
 else { $runtime = 'win-x64' }
 
-dotnet publish src/Azure.Functions.Cli/Azure.Functions.Cli.csproj --runtime $runtime --output $outDir
+dotnet publish src/Azure.Functions.Cli/Azure.Functions.Cli.csproj --runtime $runtime --output $outDir -c $config
```
brettsam: Are the changes to this file related? They don't seem to be. But if @ahmelsayed is okay with these, go for it.

ahmelsayed: it's a dev build script, so it's fine :)

ramanmiddha90: Hello , i am still facing this issue. not able to see Traces from azure function to application insights

brandonh-msft: > Hello , i am still facing this issue. not able to see Traces from azure function to application insights

If you're not seeing them in appinsights it's likely because you don't [have monitoring configured correctly](https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring?tabs=cmd). This PR did not affect logging to AI; just to console.

PR #945, opened 2018-12-19
Allow for regional date formats in Durable Test

### LINKED ISSUE #944: DurablePurgeHistoryTest fails on non mm/dd/yyyy date formats
When running the tests on a Windows locale that doesn't use the mm/dd/yyyy date format, the DurablePurgeHistoryTest fails with the following output:

```
[xUnit.net 00:02:35.73]     Azure.Functions.Cli.Tests.E2E.DurableTests.DurablePurgeHistoryTest [FAIL]
Failed   Azure.Functions.Cli.Tests.E2E.DurableTests.DurablePurgeHistoryTest
Error Message:
 Expected output "[19/12/2018 09:13:21] Pausing to let the Functions host handle previous requests.
Microsoft (R) Build Engine version 15.8.166+gd4e8d81a88 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

<log output skipped for brevity>

Running: > C:\source\azure-functions-core-tools\artifacts\win-x86\func.exe durable purge-history --task-hub-name purgeHistoryTest
Purged orchestration history for all instances created between '01/01/0001 00:00:00' and '30/12/9999 23:59:59'
Instances deleted: 0
Rows deleted: 0

[19/12/2018 09:13:42] Terminating the Functions host." to contain the strings: {"Purged orchestration history for all instances created between '1/1/0001 12:00:00 AM' and '12/30/9999 11:59:59 PM'"}.
```

The actual output contains 

```
Purged orchestration history for all instances created between '01/01/0001 00:00:00' and '30/12/9999 23:59:59
```

Whereas the assertion is for

```
Purged orchestration history for all instances created between '1/1/0001 12:00:00 AM' and '12/30/9999 11:59:59 PM
```

stuartleeks:
Fixes #944 

I considered forcing the output in the purge command to use the invariant culture so that the test could retain non-computed values for the assertion, but there seems to be a precedent for locale-based date output in [ColoredConsoleLogger](https://github.com/Azure/azure-functions-core-tools/blob/cd919dc1f4e993a631a9fdc43bd0946660a4ca86/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs#L47)

### DIFF
--- a/src/Azure.Functions.Cli/Actions/DurableActions/DurablePurgeHistory.cs
+++ b/src/Azure.Functions.Cli/Actions/DurableActions/DurablePurgeHistory.cs
@@ -9,6 +9,9 @@ namespace Azure.Functions.Cli.Actions.DurableActions
     [Action(Name = "purge-history", Context = Context.Durable, HelpText = "Purge orchestration instance state, history, and blob storage for orchestrations older than the specified threshold time.")]
     class DurablePurgeHistory : BaseDurableAction
     {
+        public readonly static DateTime CreatedAfterDefault = DateTime.MinValue;
+        public readonly static DateTime CreatedBeforeDefault = DateTime.MaxValue.AddDays(-1); // subtract one to avoid overflow/timezone error
+
         private readonly IDurableManager _durableManager;
 
         private DateTime CreatedTimeFrom { get; set; }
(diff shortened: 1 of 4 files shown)


glennamanns:
This looks fine to me. We could also make the min/max values static properties of Durable Manager in order to keep the minus one day logic consistent and reduce the number of times the min/max have to be computed.

stuartleeks:
@glennamanns I've just pushed another commit that moves the calculated values to static fields

src/Azure.Functions.Cli/Actions/DurableActions/DurablePurgeHistory.cs
```diff
@@ -9,6 +9,9 @@ namespace Azure.Functions.Cli.Actions.DurableActions
     [Action(Name = "purge-history", Context = Context.Durable, HelpText = "Purge orchestration instance state, history, and blob storage for orchestrations older than the specified threshold time.")]
     class DurablePurgeHistory : BaseDurableAction
     {
+        public readonly static DateTime CreatedAfterDefault = DateTime.MinValue;
```
glennamanns: I think you should add these to DurableManager instead of the DurablePurgeHistory action. Then both DurablePurgeHistory and DurableGetInstances can reference them :)

stuartleeks: Ah, I hadn't spotted that the values were used in DurableGetInstances as well!

glennamanns:
Looks good!

stuartleeks:
@glennamanns  - great! CLA signed now as well :-)

ahmelsayed:
Thank you @stuartleeks and @glennamanns

PR #958, opened 2019-01-04
Adding changes to document running on AKS and ACR

anirudhgarg:


anirudhgarg:
> Maybe we can omit the create aks/acr parts as they are already documented and instead link to the already existing docs?

Yes will do, will remove any commands to show creation of the aks or acr.

PR #959, opened 2019-01-07
Fixes #920 - Verify md5

### LINKED ISSUE #920: zip deploy should validate md5 hash after upload.
similar to https://github.com/Microsoft/vscode-azuretools/issues/354

ankitkumarr:
1. Should I retry if md5 mismatch fails?
2. Do we also need this for `/api/zipdeploy` or is that handled properly?


ahmelsayed:
I don't think we can get this for `/api/zipdeploy`. We will need to add something on Kudu that returns the hash of the zip there. I think it's good enough just for storage for now.

PR #960, opened 2019-01-08
Enable building func cli as a dotnet global tool

adityapatwardhan:
@joeyaiello mentioned there was a discussion about func cli as a dotnet global tool. 

I did a quick change and tested this on my fork. To build and use the global tool:

1. `dotnet pack` under `src\Azure.Functions.Cli`
2. `dotnet tool install -g --add-source <path to nupkg> func`
3. `func`

`dotnet pack` produces a `.nupkg` which needs to be released to `nuget.org`.

### DIFF
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -5,7 +5,6 @@
     <AssemblyName>func</AssemblyName>
     <RuntimeIdentifiers>win-x64;win-x86;linux-x64;osx-x64</RuntimeIdentifiers>
     <PackAsTool>true</PackAsTool>
-    <ToolCommandName>azurefunc</ToolCommandName>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
     <MajorMinorProductVersion>2.3</MajorMinorProductVersion>
     <Version>$(MajorMinorProductVersion).$(BuildNumber)</Version>
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -4,7 +4,8 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <AssemblyName>func</AssemblyName>
     <RuntimeIdentifiers>win-x64;win-x86;linux-x64;osx-x64</RuntimeIdentifiers>
-
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>azurefunc</ToolCommandName>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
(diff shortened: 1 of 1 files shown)


adityapatwardhan:
@SamirFarhat This is a `dotnet` global tool and not a `npm` global tool. You can read about it over here: https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools

SamirFarhat:
Will this be a replacement for the npm tool?

joeyaiello:
Adding @paulbatum here: I sit on the PowerShell team, we discussed with him the possibility of supporting .NET global tools so that we don't have to ask PowerShell customers to download Node/npm (large and unrelated to any of the work they'd be doing) or Chocolatey (technically unsupported). 

@paulbatum said he would look into it if he had time, but @adityapatwardhan was doing some work on a .NET global tool for PowerShell and figured he would see how easy it was to make `func` into one. Turns out it was a couple lines in the csproj. :)

I understand that this PR probably entails more (doc additions, CI/CD or deployment changes to push to Nuget as well as npm, etc.), but we wanted to help out on the code side to show how easy it is (though I suspect you'd probably want the tool to just be called `func` instead of `azurefunc`  as that's the entrypoint name for executing the command).

adityapatwardhan:
Updated the PR to remove the custom name for the tool. It should default to assembly name which is `func`. Also, updated example above.

ahmelsayed:
There is some additional work that's performed by the build script that's not captured in the csproj, namely:

- Download templates nupkgs into `templates/`
- Download python worker

We will also need to sign this nupkg to publish it under Microsoft.* namespace as well as come up with a name for it. `Microsoft.Azure.Functions.Core.Tools`?

fabiocav:
A think we need the following to complete this:
 - The build needs to be updated to produce the package
 - Since this will now be published as a NuGet, we need to ensure the expected metadata is present. You can find the required metadata here: https://github.com/Azure/azure-functions-host/blob/cc15b264ba353a2bf71e9eec3b165083bf9bdac3/build/common.props#L13-L22
   - The title should be `Azure Functions Core Tools`

ahmelsayed:
@fabiocav 

The current build script does [`dotnet publish`](https://github.com/Azure/azure-functions-core-tools/blob/master/build/Program.cs#L16-L21), then adds the python worker and downloads the latest templates nupkgs to output of the dotnet publish, and then packs the whole thing as the cli. 

I'm guessing there is probably a way to do all that using msbuild targets, but it's not happening currently. @adityapatwardhan I think that cli will not be able to perform `func new` since that assumes nupkgs for templates being there. It also won't be able to run python functions.

fabiocav:
Yeah, those are some of the build things we'd need to address in order to have a functional tool.

For the additional files (worker and templates), we should be able to bring them in before packing and have them included as part of the project.

joeyaiello:
Any update on this? Just had to reboot due to some weird file lock on my npm `func` package, and I had to reboot to update. Would love to move off of npm to manage my `func` install 😊

ahmelsayed:
Closing stale pr

PR #973, opened 2019-01-11
Fixes #964 - Update dotnet error message

### LINKED ISSUE #964: .NET SDK Installation Error Message Does Not Provide Installation Details
_From @raahmed on January 8, 2019 20:17_

Context: Running Python on Functions in Linux (Preview)
Severity: Moderately Confusing.
Reproduction Steps:
0) Execute 'func host start' without fulfilling the .NET SDK pre-requisite.
1) Notice this error message:
image
2) Notice that the error message does not contain details of which of the options the users should choose in the download link provided:
image

Mitigation: 
1) In the error message, provide information as to which of the options should be clicked on the installation page.

(Related to https://github.com/Azure/Azure-Functions/issues/1087)

_Copied from original issue: Azure/Azure-Functions#1101_

ankitkumarr:


ahmelsayed:
It looks like the error message between `InstallExtensionAction.cs` and `SyncExtensionsAction.cs` is the same, can you move it to the constants class. Also the error in the `ExtensionsHelper.cs` can get updated to be `The {template.Metadata.Name} template has extensions.` + the same message from above.

ahmelsayed:
also the one test that's failing is failing with a weird error. 

```
stdout: HttpTrigger.cs(11,21): error CS0116: A namespace cannot directly contain members such as fields or methods [C:\Users\appveyor\AppData\Local\Temp\1\tdu4m1hc.0up\tdu4m1hc_0up.csproj]
stdout: HttpTrigger.cs(11,19): error CS1514: { expected [C:\Users\appveyor\AppData\Local\Temp\1\tdu4m1hc.0up\tdu4m1hc_0up.csproj]
stdout: HttpTrigger.cs(12,1): error CS1022: Type or namespace definition, or end-of-file expected [C:\Users\appveyor\AppData\Local\Temp\1\tdu4m1hc.0up\tdu4m1hc_0up.csproj]
stdout: 
stdout: Build FAILED.
stdout: 
stdout: HttpTrigger.cs(11,21): error CS0116: A namespace cannot directly contain members such as fields or methods [C:\Users\appveyor\AppData\Local\Temp\1\tdu4m1hc.0up\tdu4m1hc_0up.csproj]
stdout: HttpTrigger.cs(11,19): error CS1514: { expected [C:\Users\appveyor\AppData\Local\Temp\1\tdu4m1hc.0up\tdu4m1hc_0up.csproj]
stdout: HttpTrigger.cs(12,1): error CS1022: Type or namespace definition, or end-of-file expect
```

I know that it has been flaky, but this is the first time I see that error. can you try running the same commands in the log with the same values and see if it fails for you as well?

ankitkumarr:
@ahmelsayed, right, I will update the PR and make it a constant. Sorry about that.

Good point on the appveyor. I think what happened is that the temporary name of the project path it picked was `tdu4m1hc.0up`.
Inside `HttpTrigger.cs`, `namespace tdu4m1hc.0up` is declared which becomes a problem because a number cannot directly follow `.` while naming namespaces.

ahmelsayed:
>  but would that be a problem(?), because the directory name may not necessarily match the namespace.

I think that's fine. We already do cleanup on that name, right? not all valid directories are valid namespaces

ankitkumarr:
@ahmelsayed, yeah that sounds good. I can make a separate PR for that.

PR #983, opened 2019-01-16
Case insensitive hub name from host.json

glennamanns:
The Durable Functions documentation for [host.json](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings#host-json) indicates that `HubName` should be capitalized. As seen in [this issue](https://github.com/Azure/azure-functions-durable-extension/issues/448#issuecomment-454887043), Visual Studio indicates it should be `hubName`. This PR allows for either capitalization.

### DIFF
--- a/src/Azure.Functions.Cli/Common/DurableManager.cs
+++ b/src/Azure.Functions.Cli/Common/DurableManager.cs
@@ -59,24 +59,26 @@ namespace Azure.Functions.Cli.Common
                 {
                     // Attempt to retrieve Durable override settings from host.json
                     dynamic hostSettings = JObject.Parse(File.ReadAllText(ScriptConstants.HostMetadataFileName));
+                    JObject durableTask = null;
 
                     string version = hostSettings["version"];
                     if (version?.Equals("2.0") == true)
                     {
                         // If the version is (explicitly) 2.0, prepend path to 'durableTask' with 'extensions'
-                        _connectionStringKey = hostSettings?.extensions?.durableTask?.AzureStorageConnectionStringName ?? _connectionStringKey;
-
-                        // Allow uppercase or lowercase hub name in host.json
-                        _taskHubName = hostSettings?.extensions?.durableTask?.HubName ?? _taskHubName;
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -65,12 +65,18 @@ private void SetConnectionStringAndTaskHubName()
...
                         _taskHubName = hostSettings?.extensions?.durableTask?.HubName ?? _taskHubName;
+                        _taskHubName = hostSettings?.extensions?.durableTask?.hubName ?? _taskHubName;
                     }
                     else
                     {
                         _connectionStringKey = hostSettings?.durableTask?.AzureStorageConnectionStringName ?? _connectionStringKey;
+
+                        // Allow uppercase or lowercase hub name in host.json
                         _taskHubName = hostSettings?.durableTask?.HubName ?? _taskHubName;
+                        _taskHubName = hostSettings?.durableTask?.hubName ?? _taskHubName;
```
cgillum: That's an interesting way to solve it. I think it's fine, but a little bit hard to understand. An alternative approach might be to use `JObject.GetValue(string, StringComparison)` as described here: https://stackoverflow.com/a/19205083/2069

ahmelsayed: how does the runtime/durable extension read that value?

cgillum: The runtime reads it in as a `JObject` and populates a custom .NET type defined by the extension. I don't remember the exact details, but it does this in a way that is case-insensitive.

ahmelsayed: That's right, I remember [this](https://github.com/Azure/azure-functions-host/blob/fd6cde34dfa103949d9fb003915969528eaf810c/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs#L289-L318)

glennamanns:
@cgillum thanks for the suggestion :) I'll check it in now - the failing test is a flaky test I see failing in other test runs.

PR #1001, opened 2019-01-24
Update runtime to 2.0.12285

pragnagopa:


pragnagopa:
@maiqbal11  - Python updates look ok?

ahmelsayed:
can you also update https://github.com/Azure/azure-functions-core-tools/blob/dev/build/Settings.cs#L44-L46

PR #1049, opened 2019-02-14
Fixes #1038 - Sign the Core tools binaries

### LINKED ISSUE #1038: Core Tools V2 binaries must be signed
V2 binaries are not currently signed. We need to match the behavior we had in V1 and ensure those are signed with the same process.

ankitkumarr:
The signing job is only for master branch right now.

~~@ahmelsayed, we also need to add the connection string in Appveyor.~~ Thanks Ahmed!

### DIFF
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -31,6 +31,7 @@ namespace Build
         public static readonly string TestProjectFile = Path.Combine(TestProjectPath, "Azure.Functions.Cli.Tests.csproj");
 
         public static readonly string DurableFolder = Path.Combine(TestProjectPath, "Resources", "DurableTestFolder");
+
         public static readonly string[] TargetRuntimes = new[] { "win-x86", "win-x64", "linux-x64", "osx-x64", "no-runtime", "min.win-x86", "min.win-x64" };
 
         public static readonly string[] LanguageWorkers = new[] { "Java", "Powershell", "Node" };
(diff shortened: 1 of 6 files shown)


build/Settings.cs
```diff
@@ -56,5 +57,76 @@ private static string config(string @default = null, [CallerMemberName] string k
...
+
+            public static readonly string[] authentiCodeBinaries = new[] {
+                "DurableTask.AzureStorage.Internal.dll",
+                "DurableTask.Core.Internal.dll",
+                "func.deps.json",
+                "func.dll",
+                "func.dll.config",
+                "func.exe",
+                "func.pdb",
+                "func.runtimeconfig.json",
```
ahmelsayed: non-executables like `.json`, `.config` etc don't have to be signed. executable scripts like `.js` and `.py` do need to be signed.

build/Settings.cs
```diff
@@ -56,5 +57,76 @@ private static string config(string @default = null, [CallerMemberName] string k
...
+                "Microsoft.Azure.WebJobs.Extensions.Http.dll",
+                "Microsoft.Azure.WebJobs.Host.dll",
+                "Microsoft.Azure.WebJobs.Host.Storage.dll",
+                "Microsoft.Azure.WebJobs.Logging.ApplicationInsights.dll",
+                "Microsoft.Azure.WebJobs.Logging.dll",
+                "Microsoft.Azure.WebJobs.Script.dll",
+                "Microsoft.Azure.WebJobs.Script.Grpc.dll",
+                "Microsoft.Azure.WebJobs.Script.WebHost.dll",
+                "Microsoft.Azure.WebSites.DataProtection.dll",
+                Path.Combine("templates", "itemTemplates.2.0.10328.nupkg"),
```
ahmelsayed: @mhoeger  is the signing process able to sign nupkgs in the zip file directly?

mhoeger: Depends on the signing... for host, we strong name sign all of the .dll's in the nupkg and then nuget sign the entire package. I don't know how Authenticode signing works with nupkgs. @ahmelsayed, i think you or fabio would know more on that! 

So to answer your question: not sure if nupkgs are able to be authenticode signed, but we do have a process for nuget signing and/or strong name signing the .dll's

build/Settings.cs
ankitkumarr: The `.py` files in distlib were failing with the third party signing scripts, and it sounded like we neede d the `.dll`s to be signed asap, so for now, I ended us ignoring `.py` files.

build/Settings.cs
ankitkumarr: @ahmelsayed, I made these lists of the unsigned things and then filtered out things that we weren't able to sign from the scripts. Just so we have the lists properly and we are taking into account all the files.

build/Settings.cs
ahmelsayed: Just thought it was a bit misleading for .json and .config files to be in the list, but not a big deal.

build/Settings.cs
ahmelsayed: oh I see, sorry I missed the filtering below.

build/Settings.cs
ankitkumarr: I think you are right, I will remove those files from the list

ankitkumarr: I updated the PR and removed those files.

build/Settings.cs
fabiocav: For clarity, only the contents need to be authenticode signed. The nupkg should be signed like the others.

PR #1060, opened 2019-02-20
Update to `-Identity`

TylerLeonhardt:
which has been fixed in 1.2 of Azure PowerShell which I've confirmed in West US 2.


TylerLeonhardt:
cc @pragnagopa @ahmelsayed @anirudhgarg @joeyaiello

PR #1061, opened 2019-02-20
Update to host version 2.0.12332

pragnagopa:


pragnagopa:
We need to revert this change because core tools are not updated yet for
Current runtime that is deploying now. @mhoeger will send a PR updating to right version.

PR #1064, opened 2019-02-20
Fixes #1038 - Sign the Core tools binaries

### LINKED ISSUE #1038: Core Tools V2 binaries must be signed

ankitkumarr:
Made a base branch v2.4.299 just so we can look at this diff and review.

The PR initiates signing job in all branches which is different than original commit. This is because we needed the signed job to run on this branch.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -26,4 +26,11 @@ else {
     Set-Location ".\build"
 }
 
-Invoke-Expression -Command  "dotnet run"
+if ($env:APPVEYOR_REPO_BRANCH -eq "master") {
+    Invoke-Expression -Command  "dotnet run --sign"
+    if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
+}
+else {
+    Invoke-Expression -Command  "dotnet run --sign"
+    if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
+}
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -2,13 +2,16 @@ using Colors.Net;
 using Colors.Net.StringColorExtensions;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.WindowsAzure.Storage.Queue;
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
+using System.Threading.Tasks;
 
 namespace Build
 {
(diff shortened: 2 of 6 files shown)


build/Settings.cs
```diff
@@ -56,5 +57,73 @@ private static string config(string @default = null, [CallerMemberName] string k
...
+                "Microsoft.Azure.WebJobs.Extensions.Http.dll",
+                "Microsoft.Azure.WebJobs.Host.dll",
+                "Microsoft.Azure.WebJobs.Host.Storage.dll",
+                "Microsoft.Azure.WebJobs.Logging.ApplicationInsights.dll",
+                "Microsoft.Azure.WebJobs.Logging.dll",
+                "Microsoft.Azure.WebJobs.Script.dll",
+                "Microsoft.Azure.WebJobs.Script.Grpc.dll",
+                "Microsoft.Azure.WebJobs.Script.WebHost.dll",
+                "Microsoft.Azure.WebSites.DataProtection.dll",
+                Path.Combine("templates", "itemTemplates.2.0.10328.nupkg"),
```
soninaren: templates would keep changing every release. This would break on every release.

soninaren: Also templates are already signed.

build/Settings.cs
```diff
@@ -56,5 +57,73 @@ private static string config(string @default = null, [CallerMemberName] string k
...
+            public static readonly string ToSignThirdPartyName = $"{BuildNumber}-third.zip";
+            public static readonly string ToSignDir = "unsigned";
+            public static readonly string SignedDir = "signed";
+            public static readonly string Authenticode = "SignAuthenticode";
+            public static readonly string ThirdParty = "Sign3rdParty";
+            public static readonly string[] RuntimesToSign = new[] {"min.win-x86"};
+            public static readonly string[] filterExtenstionsSign = new[] { ".json", ".spec", ".cfg", ".pdb", ".config", ".nupkg", ".py" };
+
+            public static readonly string[] authentiCodeBinaries = new[] {
+                "DurableTask.AzureStorage.Internal.dll",
```
soninaren: Are we referring the Durable package directly. This could cause issues if we have breaking changes in durable extensions.

build/BuildSteps.cs
```diff
@@ -152,6 +155,129 @@ public static void Test()
...
+
+                var toSignPaths = Settings.SignInfo.authentiCodeBinaries.Select(el => Path.Combine(targetDir, el));
+                // Grab all the files and filter the extensions not to be signed
+                var toSignFiles = FileHelpers.GetAllFilesFromFilesAndDirs(toSignPaths).Where(file => !Settings.SignInfo.filterExtenstionsSign.Any(ext => file.EndsWith(ext)));
+                FileHelpers.CreateZipFile(toSignFiles, targetDir, Path.Combine(targetDir, Settings.SignInfo.ToSignDir, Settings.SignInfo.ToSignZipName));
+
+                var toSignThirdPartyPaths = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el));
+                // Grab all the files and filter the extensions not to be signed
+                var toSignThirdPartyFiles = FileHelpers.GetAllFilesFromFilesAndDirs(toSignThirdPartyPaths).Where(file => !Settings.SignInfo.filterExtenstionsSign.Any(ext => file.EndsWith(ext)));
+                FileHelpers.CreateZipFile(toSignThirdPartyFiles, targetDir, Path.Combine(targetDir, Settings.SignInfo.ToSignDir, Settings.SignInfo.ToSignThirdPartyName));
```
soninaren: What happens to the rest of the files. You get the files to sign and then third party binaries. But it can probably miss unknown or new binaries

build.ps1
```diff
@@ -26,4 +26,11 @@ else {
     Set-Location ".\build"
 }
 
-Invoke-Expression -Command  "dotnet run"
\ No newline at end of file
+if ($env:APPVEYOR_REPO_BRANCH -eq "master") {
+    Invoke-Expression -Command  "dotnet run --sign"
```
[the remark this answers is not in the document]
ankitkumarr: This was just a test. This will not be merged.

build/Settings.cs
```diff
@@ -56,5 +57,73 @@ private static string config(string @default = null, [CallerMemberName] string k
...
+            public static readonly string AzureToSignContainerName = "azure-functions-cli";
+            public static readonly string AzureSignedContainerName = "azure-functions-cli-signed";
+            public static readonly string AzureSigningJobName = "signing-jobs";
+            public static readonly string ToSignZipName = $"{BuildNumber}-authenticode.zip";
+            public static readonly string ToSignThirdPartyName = $"{BuildNumber}-third.zip";
+            public static readonly string ToSignDir = "unsigned";
+            public static readonly string SignedDir = "signed";
+            public static readonly string Authenticode = "SignAuthenticode";
+            public static readonly string ThirdParty = "Sign3rdParty";
+            public static readonly string[] RuntimesToSign = new[] {"min.win-x86"};
```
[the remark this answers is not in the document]
ankitkumarr: Is that needed for VS right now? I can make a separate PR to add that. This is a quick update as they needed it ASAP

build/Settings.cs
ankitkumarr: Ok, right now, any files from the list that are not present in the Filesystem is ignored. I can remove this in the PR. Unless this is blocking, we should go ahead with the update.

Additionally, I am not signing nupkgs yes, I filter out certain extensions defined in this Settings file.

build/Settings.cs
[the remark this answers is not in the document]
soninaren: Not needed right away

ankitkumarr:
Closing because this PR was created to verify the diff and get a signed CLI patch build.

PR #1066, opened 2019-02-20
Update to worker 1.0.0b4

maiqbal11:


ankitkumarr:
@maiqbal11 did we update the docker image to use b4 yet?

maiqbal11:
> @maiqbal11 did we update the docker image to use b4 yet?

Yes, that change will need to be added here once the Docker image is released: https://github.com/Azure/azure-functions-docker/pull/77

ankitkumarr:
Once the docker image is updated and we have a new image version, we need to make sure to update this - https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Common/Constants.cs#L62
before we release `azure-functions-core-tools`.

ahmelsayed:
`mcr.microsoft.com/azure-functions/python:2.0.12309` is available now.

PR #1070, opened 2019-02-21
Add typescript options

### LINKED ISSUE #1037: Add TypeScript option for func init and func new
We want to add the ability to `func init` a TypeScript template with a pre-populated package.json that contains npm scripts. We also want to be able to `func new` TypeScript templates from the templates feed.

@mhoeger to provide templates through feed [see here](https://github.com/Azure/azure-functions-templates/issues/804)

@ColbyTresness to provide more details on how the user should be prompted.

ankitkumarr:
~~I still need to test properly and add a couple of e2e.~~
UPDATE: Added E2E tests. This should be ready to go.

But, wanted to get feedback on the changes while I do that.

Closes #1037

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
@@ -23,7 +23,7 @@ namespace Azure.Functions.Cli.Helpers
         private static readonly IDictionary<WorkerRuntime, IEnumerable<string>> availableWorkersRuntime = new Dictionary<WorkerRuntime, IEnumerable<string>>
         {
             { WorkerRuntime.dotnet, new [] { "c#", "csharp", "f#", "fsharp" } },
-            { WorkerRuntime.node, new [] { "js", "javascript" } },
+            { WorkerRuntime.node, new [] { "js", "javascript", "typescript", "ts" } },
             { WorkerRuntime.python, new []  { "py" } },
             { WorkerRuntime.java, new string[] { } },
             { WorkerRuntime.powershell, new [] { "pwsh" } }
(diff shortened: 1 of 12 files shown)


src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -86,6 +107,22 @@ public static WorkerRuntime NormalizeWorkerRuntime(string workerRuntime)
             }
         }
 
+        public static string NormalizeLanguage(string languageString)
+        {
+            if (string.IsNullOrWhiteSpace(languageString))
+            {
+                return null;
```
ahmelsayed: any callers actually expecting and handling a null? I wonder if this should also be an exception

ankitkumarr: [This guy](https://github.com/Azure/azure-functions-core-tools/pull/1070/files#diff-26a1bb4213c73dc23de6444d3ffc7ab3R133) expects it to be null right now. Just so, it can get the default for that worker. But I can change to catch the exception and then do the default if you prefer that?

ankitkumarr: I think you are right, because that codepath should never really be hit. And something must have gone wrong if it did. So, probably throwing an exception is better. I have updated the PR.

src/Azure.Functions.Cli/StaticResources/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "",
+  "scripts": {
+    "start": "npm run build && npm run start:host & npm run watch",
+    "start:host": "func start",
+    "build": "tsc && func extensions install",
+    "build:production": "npm run clean && npm run build && npm prune --production",
+    "watch": "tsc -w",
+    "clean": "rimraf dist",
+    "test": "echo \"No tests yet...\""
+  },
```
mhoeger: "scripts" : {
    "build": "tsc",
    "build:production": "npm run prestart && npm prune --production",
    "watch": "tsc --w",
    "prestart": "npm run build && func extensions install",
    "start:host": "func start",
    "start": "npm run start:host & npm run watch",
    "test": "echo \"No tests yet...\""
}

src/Azure.Functions.Cli/StaticResources/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "",
+  "devDependencies": {
+    "@azure/functions": "^1.0.1-beta1",
+    "rimraf": "^2.6.3",
```
mhoeger: Remove "rimraf" dependency please :)

PR #1071, opened 2019-02-21
Updating package reference for v1 release 1.0.12299

soninaren:


ahmelsayed:
looks like we missed the test project. That needs to be updated as well or the build will fail.

PR #1079, opened 2019-02-25
Fixes #1034 - use linux selection view for mingw

### LINKED ISSUE #1034: System.IO.IOException: The handle is invalid.
When I run this command in git bash for Windows, which I do almost anything git or repo related in, it gives me this error and then gives a new command prompt.

When I run this in PowerShell it then gives me an option to select from dotnet, node, etc.

Is there any way you can work on how this displays in git bash?

gitbashinvalidhandle

ankitkumarr:


ahmelsayed:
I wonder if `try...catch` for the API in addition to that would also make sure it handles any other consoles where CursorTop isn't available.

ankitkumarr:
@ahmelsayed, ooh, that's a good call. Let me update that.

PR #1083, opened 2019-02-27
Added flag to enable full host auth handling pipeline.

kashimiz:
Required for testing auth scenarios locally, both in our end-to-end testing on language workers and by users.

@paulbatum 🍰


ahmelsayed:
How do you get the keys?

kashimiz:
Today, users can call `func host start` without auth enabled, call the API to get their keys, then call `func host start --enableAuth`. It could be worthwhile to provide a smoother experience in the future, but offline discussion determined that out of the scope of this particular PR.

klyse:
@kashimiz @ahmelsayed I would like to use the Azure OpenID Authentication but struggle debugging this locally / finding the docs for this. Can you please explain what needs to happen locally besides adding the `--enableAuth` flag. Do I need to setup the OpenID config locally too? How?

PR #1088, opened 2019-02-28
Add files to sign

ankitkumarr:


build/FileHelpers.cs
```diff
@@ -42,6 +42,23 @@ public static void RecursiveCopy(string sourcePath, string destinationPath)
...
+        {
+            var allEntries = new List<string>();
+            foreach (var entry in filesAndDirs)
+            {
+                if (entry.Contains("*"))
+                {
+                    var files = Directory.GetFiles(Path.GetDirectoryName(entry), Path.GetFileName(entry), SearchOption.AllDirectories);
+                    allEntries.AddRange(files);
+                }
+                else {
```
ahmelsayed: nit: formatting. Put { on a new line

ahmelsayed:
One thing that can make this more robust is to run sigcheck.exe on the output that you expect to all be signed, and fail the build if there is a file that's not signed. Though probably not very urgent to do.

build/FileHelpers.cs
ankitkumarr: Oops, my bad

ankitkumarr:
@ahmelsayed, I have create this issue https://github.com/Azure/azure-functions-core-tools/issues/1089 and assigned it to me.

PR #1091, opened 2019-03-01
Add SignalR Service extension to package map

anthonychu:
Automatically install SignalR Service bindings


ahmelsayed:
@soninaren Would you mind reviewing this change?

PR #1095, opened 2019-03-02
Remove venv from basic publishing path

### LINKED ISSUE #1023: Remove venv dependency for Python
Here's a brain dump from our discussion: 

- The user shouldn't need to have a virtual environment anymore since the worker and system dependencies are packaged into a nuget. 
- Basically, they should be able to publish an app with a blank requirements file and the app should just work. 
- Keep in mind not to run a `pip freeze` since that has the potential to load up the requirements file with everything on the user's system

ankitkumarr:
DO NOT MERGE
Need to run some tests on the artifacts by appveyor. And this is dependent on python worker to be consumed as nuget package to be finished. Closes #1023

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -248,7 +248,6 @@ namespace Azure.Functions.Cli.Actions.HostActions
         {
             if (workerRuntime == WorkerRuntime.python)
             {
-                PythonHelpers.VerifyVirtualEnvironment();
                 // We need to update the PYTHONPATH to add worker's dependencies
                 var pythonPath = Environment.GetEnvironmentVariable("PYTHONPATH") ?? string.Empty;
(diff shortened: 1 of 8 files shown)


ankitkumarr:
Once we merge https://github.com/Azure/azure-functions-core-tools/pull/1098, I will update the PR, test and this should be ready to merge after that.

ankitkumarr:
After this is in, it is important that we update [this](https://github.com/Azure/azure-functions-core-tools/pull/1095/files#diff-f0fbe728234301bbbfa101ed4e53e5a8R46) on every python worker release.

@maiqbal11, we should add that to our python worker release wiki.

ahmelsayed:
Looks like the CI has python 2

ankitkumarr:
@ahmelsayed, I will remove the init test for Python for now in that case.

maiqbal11:
> After this is in, it is important that we update [this](https://github.com/Azure/azure-functions-core-tools/pull/1095/files#diff-f0fbe728234301bbbfa101ed4e53e5a8R46) on every python worker release.
> 
> @maiqbal11, we should add that to our python worker release wiki.

@ankitkumarr, here's the release instructions page: https://github.com/Azure/azure-functions-python-worker/wiki/Release-Instructions. You should be able to edit step 8 to include the relevant instructions once you've checked in.

PR #1098, opened 2019-03-04
Consume python worker from a nuget package

ankitkumarr:
Removing the venv and related changes will come in as a separate PR once this is in.

### DIFF
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -107,17 +107,14 @@ namespace Build
 
         public static void AddPythonWorker()
         {
-            var pythonWorkerPath = Path.Combine(Settings.OutputDir, "python-worker");
-            Directory.CreateDirectory(pythonWorkerPath);
-            using (var client = new WebClient())
-            {
-                client.DownloadFile(Settings.PythonWorkerUrl, Path.Combine(pythonWorkerPath, "worker.py"));
-                client.DownloadFile(Settings.PythonWorkerConfigUrl, Path.Combine(pythonWorkerPath, "worker.config.json"));
-            }
-
             foreach (var runtime in Settings.TargetRuntimes)
             {
-                FileHelpers.RecursiveCopy(pythonWorkerPath, Path.Combine(Settings.OutputDir, runtime, "workers", "python"));
+                // Python worker's dependencies are platform dependent and need to be copied accordingly
+                var pythonDir = Path.Combine(Settings.OutputDir, runtime, "workers", "python");
+                var allOsDirectories = Directory.GetDirectories(pythonDir);
+                var pythonPlatformSpecificWorker = Path.Combine(pythonDir, Settings.RuntimesToOS[runtime]);
+                FileHelpers.RecursiveCopy(pythonPlatformSpecificWorker, pythonDir);
+                allOsDirectories.ToList().ForEach(dir => Directory.Delete(dir, recursive: true));
             }
         }
 
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Actions.HostActions.WebHost.Security;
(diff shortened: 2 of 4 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -248,6 +249,13 @@ private async Task PreRunConditions(WorkerRuntime workerRuntime)
             if (workerRuntime == WorkerRuntime.python)
             {
                 PythonHelpers.VerifyVirtualEnvironment();
+                // We need to update the PYTHONPATH to add worker's dependencies
+                var pythonPath = Environment.GetEnvironmentVariable("PYTHONPATH") ?? string.Empty;
+                var pythonWorkerDeps = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "workers", "python", "deps");
+                if (!pythonPath.Contains(pythonWorkerDeps))
+                {
+                    Environment.SetEnvironmentVariable("PYTHONPATH", $"{pythonWorkerDeps};{pythonPath}", EnvironmentVariableTarget.Process);
```
ahmelsayed: use `Path.PathSeparator` instead of `;`. On Windows it's `;` but on Mac/Linux it's `:`

build/BuildSteps.cs
```diff
@@ -107,17 +107,14 @@ public static void AddDistLib()
...
-                client.DownloadFile(Settings.PythonWorkerConfigUrl, Path.Combine(pythonWorkerPath, "worker.config.json"));
-            }
-
             foreach (var runtime in Settings.TargetRuntimes)
             {
-                FileHelpers.RecursiveCopy(pythonWorkerPath, Path.Combine(Settings.OutputDir, runtime, "workers", "python"));
+                // Python worker's dependencies are platform dependent and need to be copied accordingly
+                var pythonDir = Path.Combine(Settings.OutputDir, runtime, "workers", "python");
+                var allOsDirectories = Directory.GetDirectories(pythonDir);
+                var pythonPlatformSpecificWorker = Path.Combine(pythonDir, Settings.RuntimesToOS[runtime]);
```
ahmelsayed: small nit, but I'd prefer if the path was all lowercase, so `Settings.RuntimeToOS[].ToLower()`

build/BuildSteps.cs
```diff
@@ -107,17 +107,14 @@ public static void AddDistLib()
...
+                FileHelpers.RecursiveCopy(pythonPlatformSpecificWorker, pythonDir);
+                allOsDirectories.ToList().ForEach(dir => Directory.Delete(dir, recursive: true));
```
ahmelsayed: did you mean to have these 2 in the reverse order? It seems that the last line will just delete all the workers including the one you just copied.

ahmelsayed: Actually I think I may have misunderstood the flow here. The `Windows, Linus, Mac` folders come from the nuget package, right? then you copy all the content of the right one outside it's platform specific folder, then delete all the platform specific folder?

build/BuildSteps.cs
```diff
@@ -107,17 +107,14 @@ public static void AddDistLib()
 
         public static void AddPythonWorker()
         {
-            var pythonWorkerPath = Path.Combine(Settings.OutputDir, "python-worker");
-            Directory.CreateDirectory(pythonWorkerPath);
-            using (var client = new WebClient())
-            {
-                client.DownloadFile(Settings.PythonWorkerUrl, Path.Combine(pythonWorkerPath, "worker.py"));
-                client.DownloadFile(Settings.PythonWorkerConfigUrl, Path.Combine(pythonWorkerPath, "worker.config.json"));
```
ahmelsayed: delete `Settings.PythonWorkerUrl` and `Settings.PythonWorkerConfigUrl` from settings

build/BuildSteps.cs
ankitkumarr: @ahmelsayed, actually I meant it the way I have. The directory `workers/python` has 3 runtime directories (`Linux`, `Mac` and `Windows`). I am copying everything from `workers/python/<runtime>/` to `workers/python/` and then deleting all the three runtime directories.

So, I end up with `workers/python/deps`, `workers/python/worker.py` and `workers/python/worker.config.json`.

build/BuildSteps.cs
ankitkumarr: @ahmelsayed, I think that would have to be updated in the nuget package itself (@maiqbal11). The runtimes `Windows, Linux, Mac` comes from the nuget package. 
After the build, they don't exist.

build/BuildSteps.cs
ankitkumarr: @ahmelsayed sorry, didn't see your reply. Yes, you are right. That's what I am doing.

build/BuildSteps.cs
ahmelsayed: That makes sense. Sorry I misread it.

ahmelsayed:
With this PR, you can also change the init part for python apps to remove all the pip install stuff, right?

ankitkumarr:
@ahmelsayed, yes, I had actually put that stuff in the PR https://github.com/Azure/azure-functions-core-tools/pull/1095 to remove venv. Once this is merged in, I figured I could update that PR, test and merge. 

Or if you prefer I can combine the two PRs. I made that one before we had the nuget package ready.

ahmelsayed:
2 PRs are fine.

ankitkumarr:
Updated the python worker to release version.

PR #1099, opened 2019-03-05
Add User-Agent to zipdeploy calls

### LINKED ISSUE #1096: Add a user-agent to zipDeploy path
We are working to correlate our logs to understand who is publishing from the CLI tools.  Today we can't see when the CLI is publishing via zipDeploy as it doesn't appear any user agent is sent.  Would be good if we could include a user-agent similar to the ArmClient.  Ideally we would include both the fact that it's the CLI and the version of the CLI.  

/cc @dariagrigoriu

ahmelsayed:
Closes #1096


ankitkumarr:
@ahmelsayed should we make https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Arm/ArmClient.cs#L83 the same agent as the one in PR as well? Or does ARM need it to be `"functions-cli/2.0"`

ahmelsayed:
Updated it as well.

Issue #1031, opened 2019-02-08
[packapp] Support conda install


asavaritayal:
If `environment.yaml` is present, `conda install` the dependencies instead of `pip install`.

asavaritayal:
@elprans can you look into this?

asavaritayal:
@elprans correct me if I'm wrong but the Conda repository only offers pre-built wheels. So there is no requirement for `--build-native-deps` path on this one right?

daguito81:
Any news on this issue?

yubozhao:
Would love to see any update/news on this feature.

anthonychu:
@anirudhgarg @vrdmr Do you have the latest guidance on this?

georghildebrand:
I would also be very much interested in this!

martikevin:
Hey @anirudhgarg @vrdmr we are also waiting for conda support because some dependencies like geopandas cannot be installed over pip install command. 

Please an update!?

yputter:
Also waiting for this, as turbodbc fails with pip

martinjoshua:
Would also love to see an update on this feature request.

Servot:
Also waiting for this, as scikit-garden is failing with pip

yubozhao:
We end up using custom Dockerfile.  You can find what we did here: https://github.com/bentoml/BentoML/blob/master/docker/azure-functions/Dockerfile

Basically, build on top this and add miniconda

owenlamont:
I'm also holding out for this - would love Azure functions to have the same support for conda environments as Azure ML.

stefanushinardi:
Hi @owenlamont, we are planning to publish an image with conda dependencies that you could use in custom container on a premium plan.  Would that work for you? or you're planning to use consumption plan?

owenlamont:
Hi @stefanushinardi - I'm ignorant of what plan my workplace has. I'll have to enquire with the with our data engineering team.

I was aware conda environments could be used by specifying a custom container Docker image but I had hoped I could let Azure Functions abstract Docker away and it would work as written on the description of this ticket - that I could just replace the requirements.txt with an environment.yml and everything else would auto-magically work.

stefanushinardi:
I see, got it. I dont have a timeline for that yet, but appreciate the nudge. Will discuss internally about the support. cc: @vrdmr @Hazhzeng @AnatoliB

kyle-pena-nlp:
[...]

PR #1115, opened 2019-03-08
Add initial support for Conda environments

elprans:
This adds the necessary bits to enable the publishing of conda-based
function apps.  This support is incomplete without the corresponding
server-side changes.

Conda environments are fundamentally different from regular Python
virtual environments.  Every environment contains a complete set of
binaries and shared libraries, including Python itself, required to 
run the application.  As such, the generated function package is
substantial: ~800MiB for a minimal environment.  

The difference in organization also necessiates a different launch
protocol, as a Conda environment is a mini-buildroot, "activating" 
it is a lot more involved.

Issue: #1031

### DIFF
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -45,6 +45,9 @@
     <EmbeddedResource Include="StaticResources\local.settings.json.template">
       <LogicalName>$(AssemblyName).local.settings.json</LogicalName>
     </EmbeddedResource>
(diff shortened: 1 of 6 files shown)


ahmelsayed:
@ankitkumarr can you please take a look at this PR. The change LGTM but I think you have more context around this area.

ankitkumarr:
@ahmelsayed , @asavaritayal had advised to hold off on conda work. Asavari, can you confirm?

ahmelsayed:
oh, sorry I didn't know that. It's @asavaritayal's call.

ahmelsayed:
Closing stale PR

PR #1126, opened 2019-03-14
Moving to runtime version 2.0.12353

mathewc:


ahmelsayed:
there seem to be some conflicts around powershell and python worker versions

PR #1127, opened 2019-03-15
Update publish function app to do run from package = 1

mhoeger:
Updated logic:
````
if (Consumption Linux):
    publish run from package = sas
else if (Dedicated Linux OR "--no-zip"): 
    publish zip deploy
else:
    publish run from package = 1 and zip deploy
````
Resolves https://github.com/Azure/azure-functions-core-tools/issues/1086

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -351,46 +351,37 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
         private async Task PublishRunFromPackage(Site functionApp, Stream zipFile)
         {
+            // Upload zip to blob storage
             ColoredConsole.WriteLine("Preparing archive...");
-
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs [hunk 1]
```diff
@@ -366,6 +372,27 @@ private async Task PublishRunFromPackage(Site functionApp, Stream zipFile)
...
+            functionApp.AzureAppSettings["WEBSITE_RUN_FROM_PACKAGE"] = "1";
+
+            var result = await AzureHelper.UpdateFunctionAppAppSettings(functionApp, AccessToken);
+            
+            if (!result.IsSuccessful)
+            {
+                ColoredConsole
+                    .Error
+                    .WriteLine(ErrorColor("Error updating app settings:"))
+                    .WriteLine(ErrorColor(result.ErrorResult));
```
ahmelsayed: I think you might wanna throw here instead?

mhoeger: @ahmelsayed - updated! mind taking another look?

PR #1154, opened 2019-03-26
Fixes #1015 - Consistent way to provide user dependencies

### LINKED ISSUE #1015: Standardize virtual environment between --build-native-deps and packapp
Right now, depending on the path taken, user dependencies can be found in two places-

1. `.python_packages`
2. `worker_venv/lib/python36/site-packages/`

We should standardize this to just `.python_packages`.

ankitkumarr:
Will run my e2e tests and update with results.

After this change, we can make it very clear to users about our deployment path. They can see `.python_packages` that will be created with or without `--build-native-deps`. In case if they don't want to restore custom dependencies in subsequent publish commands, they will now be able to just do `publish --no-build` and it will work.

@asavaritayal, once merged, we should update our docs as well.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -107,11 +107,14 @@ namespace Azure.Functions.Cli.Helpers
                 FileSystemHelpers.DeleteDirectorySafe(Path.Combine(functionAppRoot, Constants.ExternalPythonPackages));
             }
 
+            var packagesLocation = Path.Combine(functionAppRoot, Constants.ExternalPythonPackages);
+            FileSystemHelpers.EnsureDirectory(packagesLocation);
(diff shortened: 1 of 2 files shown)


ankitkumarr:
Note: I used `--install-option="--install-scripts=/.python_packages/bin"` [here](https://github.com/Azure/azure-functions-core-tools/pull/1154/files#diff-8d75a8a43d70590e9885877a145f0841R6) to get the scripts that may be downloaded as part of a python package. However, on testing, I noted that this command causes wheels install to be disabled and uses `setup.py` install (see https://github.com/pypa/pip/issues/3934). This heavily slows down pip install for certain packages and sometimes even errors out. 

I realized, looking at our docker images, that right now we don't support using such package brought script. So, I am removing this ability during packaging as well. We may revisit it later once the tagged pip issue is closed and pip has support to install scripts in a target location.

PR #1157, opened 2019-03-27
Fixes #1118 - Reuse build artifacts

### LINKED ISSUE #1118: Re-use build artifacts
**Scenario**
If a user attempts to re-publish without changing the contents of the `requirements.txt` file, avoid triggering a full dependency build (pip install). This feature would especially benefit the native-dependency scenario that requires running docker each time to build the required binaries. 

**Experience**

The functionality needs to be integrated into the following commands: 

1. func publish
2. func publish --build-native-dependencies
3. func pack 
4. func pack --build-native-dependencies

ankitkumarr:
@ahmelsayed do you think I should also check and add the file `.python_packages/requirements.txt.md5` I am creating to `.funcignore`?

@asavaritayal, ~~I called the flag to force update `--force-sync`. Let me know if you have a better name.~~ On Asavari's suggestion, I have removed the flag to force update.

This needs a doc update. And this should be the last item for changing the Python publishing workflow for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ image: Visual Studio 2017
 clone_folder: c:\azure-functions-cli
 
 build_script:
+- "SET PATH=C:\\Python36;C:\\Python36\\Scripts;%PATH%"
 - ps: |
       .\build.ps1
       if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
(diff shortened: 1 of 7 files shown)


ankitkumarr:
@ahmelsayed, added a basic e2e test for functionality. Let me know if that looks good, and I will merge this change.

ankitkumarr:
Apologies for multiple force push, I was experimenting with appveyor. Finally got it to run Python 3. Tests should be good now.

ahmelsayed:
Looks great!

PR #1158, opened 2019-03-27
Enable managed dependencies for PowerShell function apps.

Francisco-Gamino:
*  Adding support in func init to enable managed dependencies for PowerShell.
* Unit test for 'init --worker-runtime powershell --managed-dependencies'
* Adding PowerShell (preview) to drop down list of available runtime workers.

**Installation flow**
At function app creation, the user has the option to opt into Managed Dependencies for PowerShell function apps using the --managed-dependencies switch. Managed dependencies enables modules to be automatically managed by the Functions service. Currently, only the Azure Az module is supported. 

If the user opts in, we add an entry to host.json and create a requirements.psd1.

Scenario 1: User opts into managed dependencies for PowerShell
```
init --worker-runtime powershell --managed-dependencies
Writing profile.ps1
Writing requirements.psd1
Writing .gitignore
Writing host.json
Writing local.settings.json
Writing E:\MyFunctions\MyFuncApp\.vscode\extensions.json
```
Running this command adds a managedDependency entry in host.json and generates a requirements.psd1 file.

```
# Contents of host.json
{
  "version": "2.0",
  "managedDependency": {
    "enabled": true
  }
}

# Contents of requirements.psd1

# Only the Azure Az module is supported in preview.
#
@{
    'Az' = '1.*'
}
```

Scenario 2: User creates a PowerShell function app. If the –managed-dependencies switch is not specified, the user will get prompted. 
```
init --worker-runtime powershell
Would you like to install the Azure modules and have these automatically managed in Azure?
Yes
No
```
If the customer selects ‘yes’, we do the same as in Scenario 1.

Scenario 3: User creates a PowerShell function app and opts out of managed dependencies. This means that no requirements.psd1 file is generated and no managedDependency entry is added to host.json.
```
init --worker-runtime powershell
Would you like to install the Azure modules and have these automatically managed in Azure?
Yes
No
# Select no, or
init --worker-runtime powershell --managed-dependencies=false
```


Francisco-Gamino:
cc: @anirudhgarg @eamonoreilly

Francisco-Gamino:
This PR is being replaced by https://github.com/Azure/azure-functions-core-tools/pull/1159.

PR #1159, opened 2019-03-27
Enable managed dependencies for PowerShell function apps.

Francisco-Gamino:
*  Adding support in func init to enable managed dependencies for PowerShell.
* Unit test for 'init --worker-runtime powershell --managed-dependencies'
* Adding PowerShell (preview) to drop down list of available runtime workers.

**Installation flow**
At function app creation, the user has the option to opt into Managed Dependencies for PowerShell function apps using the --managed-dependencies switch. Managed dependencies enables modules to be automatically managed by the Functions service. Currently, only the Azure Az module is supported. 

If the user opts in, we add an entry to host.json and create a requirements.psd1.

Scenario 1: User opts into managed dependencies for PowerShell
```
init --worker-runtime powershell --managed-dependencies
Writing profile.ps1
Writing requirements.psd1
Writing .gitignore
Writing host.json
Writing local.settings.json
Writing E:\MyFunctions\MyFuncApp\.vscode\extensions.json
```
Running this command adds a managedDependency entry in host.json and generates a requirements.psd1 file.

```
# Contents of host.json
{
  "version": "2.0",
  "managedDependency": {
    "enabled": true
  }
}

# Contents of requirements.psd1

# Only the Azure Az module is supported in preview.
#
@{
    'Az' = '1.*'
}
```

Scenario 2: User creates a PowerShell function app. If the –managed-dependencies switch is not specified, the user will get prompted. 
```
init --worker-runtime powershell
Would you like to install the Azure modules and have these automatically managed in Azure?
Yes
No
```
If the customer selects ‘yes’, we do the same as in Scenario 1.

Scenario 3: User creates a PowerShell function app and opts out of managed dependencies. This means that no requirements.psd1 file is generated and no managedDependency entry is added to host.json.
```
init --worker-runtime powershell
Would you like to install the Azure modules and have these automatically managed in Azure?
Yes
No
# Select no, or
init --worker-runtime powershell --managed-dependencies=false
```

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -11,7 +11,6 @@ using Colors.Net;
 using Fclp;
 using Microsoft.Azure.WebJobs.Script;
 using static Azure.Functions.Cli.Common.OutputTheme;
-using static Colors.Net.StringStaticMethods;
 
 namespace Azure.Functions.Cli.Actions.LocalActions
 {
@@ -36,10 +35,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
         public string Language { get; set; }
 
+        public bool? ManagedDependencies { get; set; }
+
         internal static readonly Dictionary<Lazy<string>, Task<string>> fileToContentMap = new Dictionary<Lazy<string>, Task<string>>
         {
-            { new Lazy<string>(() => ".gitignore"), StaticResources.GitIgnore },
-            { new Lazy<string>(() => ScriptConstants.HostMetadataFileName), StaticResources.HostJson },
+            { new Lazy<string>(() => ".gitignore"), StaticResources.GitIgnore }
         };
 
         private readonly ITemplatesManager _templatesManager;
@@ -95,6 +95,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 .SetDefault(null)
                 .WithDescription("Initialize a language specific project. Currently supported when --worker-runtime set to node. Options are - \"typescript\" and \"javascript\"")
                 .Callback(l => Language = l);
+
+            Parser
+                .Setup<bool>("managed-dependencies")
+                .WithDescription("Installs managed dependencies. Currently, only the PowerShell worker runtime supports this functionality.")
+                .Callback(f => ManagedDependencies = f);
 
             if (args.Any() && !args.First().StartsWith("-"))
             {
@@ -157,8 +162,10 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             }
             else
             {
-                await InitLanguageSpecificArtifacts(workerRuntime, language);
+                bool managedDependenciesOption = ResolveManagedDependencies(workerRuntime, ManagedDependencies);
+                await InitLanguageSpecificArtifacts(workerRuntime, language, managedDependenciesOption);
                 await WriteFiles();
+                await WriteHostJson(workerRuntime, managedDependenciesOption);
                 await WriteLocalSettingsJson(workerRuntime);
             }
 
@@ -212,7 +219,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             return string.Empty;
         }
 
-        private static async Task InitLanguageSpecificArtifacts(WorkerRuntime workerRuntime, string language)
+        private static async Task InitLanguageSpecificArtifacts(WorkerRuntime workerRuntime, string language, bool managedDependenciesOption)
         {
             if (workerRuntime == Helpers.WorkerRuntime.python)
             {
@@ -221,6 +228,16 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             else if (workerRuntime == Helpers.WorkerRuntime.powershell)
             {
                 await WriteFiles("profile.ps1", await StaticResources.PowerShellProfilePs1);
+
+                if (managedDependenciesOption)
+                {
+                    string majorVersion = await PowerShellHelper.GetLatestAzModuleMajorVersion();
+
+                    var requirementsContent = await StaticResources.PowerShellRequirementsPsd1;
+                    requirementsContent = requirementsContent.Replace("MAJOR_VERSION", majorVersion);
+
+                    await WriteFiles("requirements.psd1", requirementsContent);
+                }
             }
 
             if (language == Constants.Languages.TypeScript)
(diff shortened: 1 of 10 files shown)


Francisco-Gamino:
cc: @anirudhgarg @eamonoreilly

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -157,8 +162,10 @@ private async Task InitFunctionAppProject()
             }
             else
             {
-                await InitLanguageSpecificArtifacts(workerRuntime, language);
+                bool managedDependenciesOption = ResolveManagedDependencies(workerRuntime, ManagedDependencies);
+                await InitLanguageSpecificArtifacts(workerRuntime, language, managedDependenciesOption);
```
ankitkumarr: Should we move the call to `ResolveManagedDependencies` in  `InitLanguageSpecificArtifacts` because it should be only done and used for `powershell` artifacts? `ResolveManagedDependencies` can update `ManagedDependencies` flag if needed.

This way `WriteHostJson` can also use `ManagedDependencies` directly and won't need to expect a flag.

Not a big deal. If you prefer it this way, that's fine with me too.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -317,5 +334,37 @@ private static async Task WriteFiles(string fileName, string fileContent)
...
+                }
+
+                return managedDependenciesOption.Value;
+            }
+
+            bool result = false;
+            ColoredConsole.Write("Would you like to install the Azure modules and have these automatically managed in Azure? ");
+            var answer = SelectionMenuHelper.DisplaySelectionWizard(WorkerRuntimeLanguageHelper.ManagedDependenciesInstallationOptions());
+            ColoredConsole.WriteLine(TitleColor(answer.ToString()));
+            if (answer.ToString().Equals(InstallManagedDependencies.Yes.ToString(), StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
Francisco-Gamino: I agree. Fixed.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -317,5 +334,37 @@ private static async Task WriteFiles(string fileName, string fileContent)
...
+            if (managedDependenciesOption.HasValue)
+            {
+                if (workerRuntime != Helpers.WorkerRuntime.powershell)
+                {
+                    throw new CliException("Managed dependencies is only supported for PowerShell.");
+                }
+
+                return managedDependenciesOption.Value;
+            }
+
```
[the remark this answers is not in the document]
Francisco-Gamino: I fixed the ResolveManagedDependencies method. Now the code looks like this:

```
if (workerRuntime != Helpers.WorkerRuntime.powershell)
{
	if (managedDependenciesOption.HasValue)
	{
		throw new CliException("Managed dependencies is only supported for PowerShell.");
	}

	return false;
}

ColoredConsole.Write("Would you like to install the Azure modules and have these automatically managed in Azure? ");
var answer = SelectionMenuHelper.DisplaySelectionWizard(WorkerRuntimeLanguageHelper.ManagedDependenciesInstallationOptions()).ToString();
ColoredConsole.WriteLine(TitleColor(answer));
return (answer.Equals(InstallManagedDependencies.Yes.ToString(), StringComparison.OrdinalIgnoreCase));
```

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+                    numberOfRetries--;
+                } while (!successfulRequest && numberOfRetries <= 0);
+            }
+
+            if (response == null)
+            {
+                throw new CliException(errorMessage);
+            }
+
+            var stream = response.Content.ReadAsStreamAsync().Result;
```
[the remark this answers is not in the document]
Francisco-Gamino: Fixed.

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                bool successfulRequest = false;
+                do
+                {
+                    response = await client.GetAsync(address);
+                    successfulRequest = response.IsSuccessStatusCode;
+                    numberOfRetries--;
+                } while (!successfulRequest && numberOfRetries <= 0);
+            }
+
```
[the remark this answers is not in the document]
Francisco-Gamino: Yes, we are going to throw a CliException. The error message will include the exception message as well, e.g.,

```
catch (Exception ex)
{
	if (noOfRetries == maxNumberOfTries)
	{
		var errorMsg = $@"Fail to get module version for {AzModuleName}. Errors: {ex.Message}";
		throw new CliException(errorMsg);
	}
	noOfRetries++;
}
```

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
```diff
@@ -0,0 +1,91 @@
...
+                // Find the version information
+                XmlNode root = doc.DocumentElement;
+                var props = root.SelectNodes("//m:properties/d:Version", nsmgr);
+                var latestVersion = new Version("0.0");
+
+                if (props != null && props.Count > 0)
+                {
+                    for (int i = 0; i < props.Count; i++)
+                    {
```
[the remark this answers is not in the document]
Francisco-Gamino: Thanks for this suggestion. This makes the code a bit more readable.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
Francisco-Gamino: If managed dependencies is enable, we need to do two things: create a requirements.psd1 file and add an entry to host.json. This the reason why we resolved it before calling into InitLanguageSpecificArtifacts and WriteHostJson. Is there a better way to do this?

Francisco-Gamino:
Thanks @ankitkumarr for the code review!

@ahmelsayed : Could you please take a look whenever you have a moment? Thanks!

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -317,5 +334,37 @@ private static async Task WriteFiles(string fileName, string fileContent)
...
+                return false;
+            }
+
+            if (managedDependenciesOption.HasValue)
+            {
+                return managedDependenciesOption.Value;
+            }
+
+            ColoredConsole.Write("Would you like to install the Azure modules and have these automatically managed in Azure? ");
+            var answer = SelectionMenuHelper.DisplaySelectionWizard(WorkerRuntimeLanguageHelper.ManagedDependenciesInstallationOptions()).ToString();
```
ahmelsayed: I think the `DisplaySelectionWizard` along with the `InstallManagedDependencies` Enum is a bit of an overkill for a `[y/n]` scenario. I think a generic helper like `bool SelectionHelper.Confirm(string question)` would be simpler?

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            Uri address = new Uri($"{PowerShellGalleryFindPackagesByIdUri}'{AzModuleName}'");
+            int maxNumberOfTries = 3;
+            int tries = 1;
+
+            string latestMajorVersion = null;
+            Stream stream = null;
+            using (HttpClient client = new HttpClient())
+            {
+                while (tries <= maxNumberOfTries)
```
ahmelsayed: is it possible to use `Helpers.RetryHelper` here?

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// </summary>
+        public static async Task<string> GetLatestAzModuleMajorVersion()
+        {
+            Uri address = new Uri($"{PowerShellGalleryFindPackagesByIdUri}'{AzModuleName}'");
+            int maxNumberOfTries = 3;
+            int tries = 1;
+
+            string latestMajorVersion = null;
+            Stream stream = null;
+            using (HttpClient client = new HttpClient())
```
ahmelsayed: add cli user agent to the client.

ahmelsayed:
just a couple of small comments from me. Let me know if you have any questions.

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
Francisco-Gamino: Thanks for pointing this out. Yes, we can. I have fixed the code accordingly.

src/Azure.Functions.Cli/Helpers/PowerShellHelper.cs
Francisco-Gamino: Fixed.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
Francisco-Gamino: Great suggestion. I have added a helper Confirm in SelectionMenuHelper. The code looks like this:

```
public static bool Confirm(string question)
{
	if (string.IsNullOrEmpty(question))
	{
		throw new CliException("Question cannot be null or empty.");
	}

	List<string> options = new List<string>() { "Yes", "No" };
	ColoredConsole.Write(question);
	var answer = DisplaySelectionWizard(options);
	ColoredConsole.WriteLine(TitleColor(answer));

	return answer.Equals("Yes", StringComparison.OrdinalIgnoreCase);
}
```

PR #1167, opened 2019-04-01
Update docker version

ankitkumarr:
Technically, this change wouldn't be needed unless with update python-worker version. But probably a good idea to keep things updated.


ahmelsayed:
also the new image is 700 MB smaller than the old one :)

PR #1180, opened 2019-04-05
fixes #1164 - verify version before start action

### LINKED ISSUE #1164: `func host start` should either spawn the right `python` process or error out
Right now `func host start` will spawn a `python` process for Python functions. For some users, `python` refers to Python 2 while `python3` refers to Python 3. 

Given that it is the host that spawns the `python` process, `func` should either supply the right information to the host (using `worker.config.json`) or check the Python version using `python` and fail with the right information to let the users know that they need to alias `python` to Python 3.

\cc @asavaritayal @mebibou

ankitkumarr:


ahmelsayed:
I think the title of the PR (and commit) is wrong, but other than that, it might be also nice to print the actual version of python being used.

ankitkumarr:
@ahmelsayed, oops, sorry, fixing the heading and the commit name. Regarding the version, we do print it if it's not the right version - https://github.com/Azure/azure-functions-core-tools/blob/4adac7830008a1d6118428edbfe3923488b59f60/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs#L88

PR #1189, opened 2019-04-09
updating Functions host nugets

brettsam:


src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -90,7 +90,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.JavaWorker" Version="1.3.1-SNAPSHOT" />
     <PackageReference Include="Microsoft.Azure.Functions.NodeJsWorker" Version="1.0.2" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker" Version="0.1.90-preview" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Script.WebHost" Version="2.0.12382" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.WebHost" Version="2.0.12405" />
```
soninaren: Also update the min version
https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj#L9

brettsam: Has that happened with every release? Surely we've done more than 6 releases of the CLI by now...

soninaren: it has not been consistently done, process is not heavily enforced. The app veyor build increases patch version automatically generally it does cause an issue in case it is missed. Wanted to update minor version because the host version update includes bundling changes.

soninaren: typo in the comment above, fixed :)

brettsam: Makes sense, thanks for the explanation.

PR #1193, opened 2019-04-11
Updating, "func extensions install", "func new", "func init ." and "func host start" actions

### LINKED ISSUE #1105: Load extensions from extension bundle
related to https://github.com/Azure/azure-functions-host/issues/3562

### LINKED ISSUE #1106: Fetch extension bundle configuration for host.json from CDN

### LINKED ISSUE #1108: Load templates from extension bundles for "func new" command

### LINKED ISSUE #1109: Make "func extensions install" a no-op if extension bundle is configured

soninaren:
`func host start` will load extensions from extension bundle if it is configured in host.json
closes #1105 

`func init . --extension-bundle` will add default extension bundling config to host.json
closes #1106 

`func extensions install` will be a no-op with a warning if extension bundle is configured in host.json
closes #1109 

`func new` will load the templates from extension bundle if it is configured in host.json
closes #1108

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -10,6 +10,8 @@ using Azure.Functions.Cli.Interfaces;
 using Colors.Net;
 using Fclp;
 using Microsoft.Azure.WebJobs.Script;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Actions.LocalActions
@@ -32,6 +34,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         public bool Force { get; set; }
 
         public bool Csx { get; set; }
+
+        public bool ExtensionBundle { get; set; }
 
         public string Language { get; set; }
 
@@ -101,6 +105,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 .WithDescription("Installs managed dependencies. Currently, only the PowerShell worker runtime supports this functionality.")
                 .Callback(f => ManagedDependencies = f);
 
+            Parser
+                .Setup<bool>("extension-bundle")
+                //.WithDescription("use default extension bundle configuration in host.json")
+                .Callback(e => ExtensionBundle = e);
+
             if (args.Any() && !args.First().StartsWith("-"))
             {
                 FolderName = args.First();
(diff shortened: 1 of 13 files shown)


src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public void Configure(IConfigurationBuilder builder)
+        {
+            var bundleId = ExtensionBundleHelper.GetExtensionBundleOptions(_hostOptions).Id;
+            if (!string.IsNullOrEmpty(bundleId))
+            {
+                builder.AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", Path.Combine(Path.GetTempPath(), "Functions", ScriptConstants.ExtensionBundleDirectory, bundleId)}
```
ahmelsayed: will this work on linux?

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -101,6 +105,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .WithDescription("Installs managed dependencies. Currently, only the PowerShell worker runtime supports this functionality.")
                 .Callback(f => ManagedDependencies = f);
 
+            Parser
+                .Setup<bool>("extension-bundle")
+                //.WithDescription("use default extension bundle configuration in host.json")
+                .Callback(e => ExtensionBundle = e);
```
ahmelsayed: shouldn't this just be the default? Maybe even the only behavior?

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -70,6 +72,14 @@ public override ICommandLineParserResult ParseArgs(string[] args)
 
         public async override Task RunAsync()
         {
+            var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
+            if (extensionBundleManager.IsExtensionBundleConfigured())
+            {
+                var hostFilePath = Path.Combine(Environment.CurrentDirectory, ScriptConstants.HostMetadataFileName);
+                ColoredConsole.WriteLine(WarningColor($"No action performed. Extension bundle is configured in {hostFilePath}"));
```
ahmelsayed: I think few actions already call into the InstallExtensionsAction, like the start, and publish actions. Ideally it wouldn't display a warning for those commands.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
soninaren: keep it non-default for internal testing, we need to look at the cold start numbers for this. This is why i am hiding the description.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
soninaren: That is a good call, We need to prevent `func extension install` via those actions as well. May need to pass a flag to not show warnings if `func extension install` is run via that command

src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
soninaren: as long as the tempPath can resolve it should work.

PR #1213, opened 2019-04-23
Add null propagation for the process

fraxedas:
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=netframework-4.8#System_Diagnostics_Process_Start_System_Diagnostics_ProcessStartInfo_ 

Process.Start(processStartInfo) will return a new Process that is associated with the process resource, or null if no process resource is started.


ahmelsayed:
Thanks for your contribution @fraxedas!

PR #1218, opened 2019-04-24
Fixes #1217 - Elastic Premium issues

### LINKED ISSUE #1217: Publishing to Elastic Premium shows an error message that it should not show
`Assuming --nozip (do not run from package) for publishing to Linux dedicated plan.`

It also does not sync triggers.

ankitkumarr:
Tried and tested.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -274,7 +274,8 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
 
             // For dedicated linux apps, we do not support run from package right now
-            if (functionApp.IsLinux && !functionApp.IsDynamic && RunFromPackageDeploy)
+            var isFunctionAppDedicated = !functionApp.IsDynamic && !functionApp.IsElasticPremium;
+            if (functionApp.IsLinux && isFunctionAppDedicated && RunFromPackageDeploy)
             {
                 ColoredConsole.WriteLine("Assuming --nozip (do not run from package) for publishing to Linux dedicated plan.");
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -321,14 +322,21 @@ private async Task PublishFunctionApp(Site functionApp, GitIgnoreParser ignorePa
                 await Task.Delay(TimeSpan.FromSeconds(5));
                 await SyncTriggers(functionApp);
             }
+
+            // Linux Elastic Premium functions take longer to deploy. In order to view the functions properly,
+            // we need to wait for 15 extra seconds
```
ahmelsayed: how did you land at 15 seconds? From what I've seen, this can be significantly longer.

ankitkumarr: Through trial and error, 15 seconds seemed to work for all of my cases. However, my functions were very small, and if that's not a constant time, I think we should not show functions at all for Elastic Premium. What's your thought?

ahmelsayed: I think we shouldn't show it if there is a chance (which there is) they will frequently be outdated. 

We need a reliable way to know that the runtime restarted and that we're getting the list from the new runtime, which as far as I know there isn't one for App Service Linux.

ankitkumarr: Fair! I updated the PR.

ahmelsayed: I'll be also good to communicate that to the premium PMs

ankitkumarr: I told @dariagrigoriu about this! Will let others know

PR #1223, opened 2019-04-25
Update Python versions

maiqbal11:
Waiting on https://github.com/Azure/azure-functions-docker/pull/98 to get merged and a new Docker image to get pushed out before updating the Docker image which this core tools is pinned to.


maiqbal11:
@ahmelsayed, the Python version updates are merged to Docker. You mentioned a change to the Docker image build process; how can I retrieve the version of the Docker image that contains the new Python versions for the release?

ankitkumarr:
@maiqbal11, going forward you shouldn't need to update docker image everytime, because we stopped using bundling process. We don't have a dependency on the worker like that anymore. I will update the release notes. This PR can be merged as is.

PR #1254, opened 2019-04-26
Fixing the error in the func new action with extension bundles is configured

soninaren:


soninaren:
@ahmelsayed need to add it to the current release. Some docs are blocked on it.

ahmelsayed:
Can you set it in `Main` instead?

ahmelsayed:
Only set it if it's not already set.

ahmelsayed:
also move it to a method call maybe `SetCoreToolsEnvironmentVariables` or something like that.

Issue #1259, opened 2019-04-29
func new should work with non initialized folders


ahmedelnably:
if the user run func new in a non initialized folder, we should prompt the user to provide the language to initialize the folder first.

ahmelsayed:
related to https://github.com/Azure/azure-functions-core-tools/issues/1210

PR #1267, opened 2019-04-30
Fix func new creating host.json

ankitkumarr:
On an empty directory, `_templatesManager` seems to create a `host.json` when there isn't one. This causes `WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage` to believe that there is a function app in the current directory. This messes up our logic for function app creation.

Running the template manager after trying to get the worker runtime fixes this scenario, as now `GetCurrentWorkerRuntimeLanguage` can throw an error when `host.json` is not present.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
@@ -73,8 +73,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 }
             }
 
-            var templates = await _templatesManager.Templates;
             var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
+            var templates = await _templatesManager.Templates;
 
             if (workerRuntime != WorkerRuntime.None && !string.IsNullOrWhiteSpace(Language))
             {


ankitkumarr:
Need to add a test

soninaren:
> On an empty directory, `_templatesManager` seems to create a `host.json` when there isn't one. This causes `WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage` to believe that there is a function in the current directory. This messes up our logic for function creation.
> 
> Running the template manager after trying to get the worker runtime fixes this scenario, as now `GetCurrentWorkerRuntimeLanguage` can throw an error when `host.json` is not present.

Why would `GetCurrentWorkerRuntimeLanguage` think there is a function present in the current directory because of host.json?

ankitkumarr:
@soninaren, that's the logic we use to check if function app is present https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Helpers/ScriptHostHelpers.cs#L73. We look at the presence of `host.json` or `local.settings.json`, because these two files are the common files that will be created anytime you create a function app.

ahmelsayed:
I'd like to re-think the whole `init` and `new` flows now with all the new options. We should better define what we want the behavior to be with these new improvments in mind. https://github.com/Azure/azure-functions-core-tools/issues/1259 https://github.com/Azure/azure-functions-core-tools/issues/1235 and --<language>

ankitkumarr:
@ahmelsayed, I agree with you. I made this quick PR to mitigate the issue for the next release.

PR #1273, opened 2019-05-01
Fixes #1229 - Python version resolutions

### LINKED ISSUE #1229: Auto-resolve Python versions when running locally
**Scenario**

1. If a user has multiple versions of Python in their local environment, each version is referenced by a different path. e.g. `/usr/local/bin/python` resolves to version 2 and `/usr/local/bin/python3` resolves to version 3. 

2. If a single version is present, the default path refers to that version. e.g. `/usr/local/bin/python` will resolve to version 3.

Core Tools should identify which path version 3.6 is present on and set it up accordingly to be used with the runtime. 

**Scratch**

Per the wiki here - https://github.com/Azure/azure-functions-host/wiki/Authoring-&-Testing-Language-Extensions, we should be able to use the `languageWorkers:python:defaultExecutablePath` environment variable to declare the correct alias.

On `func init`, and `func host start` the following workflow must apply: 

1. If `python3 --version == python3.6.*`
        use python3
2. Elif `python --version == python3.6.*`
        use python
3. Else throw an error

ankitkumarr:
fixes #1229

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -20,7 +20,7 @@ namespace Azure.Functions.Cli.Helpers
 
         public static async Task SetupPythonProject()
         {
-            await VerifyVersion();
+            await VerifyPythonVersions();
             CreateRequirements();
             await EnsureVirtualEnvrionmentIgnored();
         }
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -89,12 +130,7 @@ public static async Task VerifyVersion()
                     trials++;
                     await Task.Delay(TimeSpan.FromMilliseconds(200));
                 }
-
-                var output = sb.ToString().Trim();
-                if (output.IndexOf("3.6") == -1)
-                {
-                    throw new CliException($"Python 3.6 is required. Current python version is '{output}'");
-                }
+                return sb.ToString().Trim(); ;
```
ahmelsayed: remove extra semicolon

PR #1282, opened 2019-05-03
Add all powershell runtimes for windows

ahmelsayed:
It looks like Powershell launches `dotnet` process which then will load the right `runtime`. The bitness of `dotnet` will depend on what the user has installed, not the bitness of `func`. So we need to include both bitness runtimes in the both cli packages for Windows.

### DIFF
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -48,8 +48,8 @@ namespace Build
 
         public static readonly Dictionary<string, string[]> ToolsRuntimeToPowershellRuntimes = new Dictionary<string, string[]>
         {
-            { "win-x86", new [] { "win-x86", "win", "win10-x86", "win8-x86", "win81-x86", "win7-x86" } },
-            { "win-x64", new [] { "win-x64", "win", "win10-x64", "win8-x64", "win81-x64", "win7-x64" } },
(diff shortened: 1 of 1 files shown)


ahmelsayed:
@Francisco-Gamino would you mind giving the builds here a final sanity check https://ci.appveyor.com/project/appsvc/azure-functions-cli/builds/24295981/artifacts

build/Settings.cs
```diff
@@ -48,8 +48,8 @@ private static string config(string @default = null, [CallerMemberName] string k
 
         public static readonly Dictionary<string, string[]> ToolsRuntimeToPowershellRuntimes = new Dictionary<string, string[]>
         {
-            { "win-x86", new [] { "win-x86", "win", "win10-x86", "win8-x86", "win81-x86", "win7-x86" } },
-            { "win-x64", new [] { "win-x64", "win", "win10-x64", "win8-x64", "win81-x64", "win7-x64" } },
+            { "win-x86", new [] { "win-x86", "win", "win10-x86", "win8-x86", "win81-x86", "win7-x86", "win-x64", "win", "win10-x64", "win8-x64", "win81-x64", "win7-x64" } },
```
Francisco-Gamino: There is duplicate entry for "win". Also, since both x86 and x64 will have the same values. We can create one list and reference it "win-x64" and "win-x86".

Francisco-Gamino:
@ahmelsayed : everything works as expected after you code change. I validated this on x86 and x64.

PR #1290, opened 2019-05-07
HotFix: Downgrading PowerShellWorker 0.1.95

soninaren:


soninaren:
re-applying this hotfix commit 
https://github.com/Azure/azure-functions-core-tools/pull/1260

PR #1323, opened 2019-05-20
Fixes #931 - Add support for other clouds

### LINKED ISSUE #931: Running Azure CLI command func azure functionapp in Azure Government tries to connect to public cloud instead of Government cloud, resulting in 401
Running func azure functionapp publish YourFunctionApp (where YourFunctionApp is the name of a created Function App in the Government subscription) returns 401/Unauthorized.

Running this command in debug mode shows that it is trying to make an ARM call to the public cloud endpoint (management.azure.com ) instead of to the Government cloud endpoint (eg management.usgovcloudapi.net) and is returning 401 because the endpoint is expecting a public cloud token audience instead of the Government token audience that is provided. See sample request below for further details.
This occurs even with current Function Core Tools (2.3.148) and CLI (2.0.52)

Running non-func CLI commands from the same Government account connects to the Government cloud endpoint and works as expected (az functionapp create, az storage account create, and  az group create the respective resources ok).

Request/Response when using func azure functionapp publish:
Request:
Method: GET, RequestUri: 'https://management.azure.com/subscriptions?api-version=2018-09-01', Version: 2.0, Content: <null>, Headers:
{
  Authorization: Bearer  {snipped}
User-Agent: functions-cli/2.0
  Accept: application/json
  x-ms-request-id: 8a334eb8-c971-41fd-be82-2292dfee4945
}

Response:
StatusCode: 401, ReasonPhrase: 'Unauthorized', Version: 1.1, Content: System.Net.Http.HttpConnection+HttpConnectionResponseContent, Headers:
{
  Cache-Control: no-cache
  Pragma: no-cache
  WWW-Authenticate: Bearer authorization_uri='https://login.windows.net/', error='invalid_token', error_description='The access token is from wrong audience or resource.'
  x-ms-failure-cause: gateway
  x-ms-request-id: df87e089-fb76-461f-9c73-f2225f5f2426
  x-ms-correlation-request-id: df87e089-fb76-461f-9c73-f2225f5f2426
  x-ms-routing-request-id: EASTUS:20181211T203228Z:df87e089-fb76-461f-9c73-f2225f5f2426
  Strict-Transport-Security: max-age=31536000; includeSubDomains
  X-Content-Type-Options: nosniff
  Date: Tue, 11 Dec 2018 20:32:28 GMT
  Connection: close
  Content-Type: application/json; charset=utf-8
  Expires: -1
  Content-Length: 378
}
{'error':{'code':'InvalidAuthenticationTokenAudience','message':'The access token has been obtained for wrong audience or resource 'https://management.core.usgovcloudapi.net/'. It should exactly match with one of the allowed audiences 'https://management.core.windows.net/','https://management.core.windows.net','https://management.azure.com/','https://management.azure.com'.'}}

ankitkumarr:
I am yet to test this thoroughly, to make sure all functionality is still fine. Still wanted to make the PR in case you had feedback, while I try and get testing environments.

I did not add getting management URL from powershell yet. If this looks good, let me know, I will have to figure out how to retrieve this same info from powershell.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/AddStorageAccountSettingAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/AddStorageAccountSettingAction.cs
@@ -42,7 +42,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
         public override async Task RunAsync()
         {
-            var storageAccount = await AzureHelper.GetStorageAccount(StorageAccountName, AccessToken);
+            var storageAccount = await AzureHelper.GetStorageAccount(StorageAccountName, AccessToken, ManagementURL);
 
             var name = $"{storageAccount.StorageAccountName}_STORAGE";
             _secretsManager.SetSecret(name, storageAccount.ConnectionString);
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -11,6 +11,7 @@ using Colors.Net;
 using Fclp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using static Colors.Net.StringStaticMethods;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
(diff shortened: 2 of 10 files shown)


ankitkumarr:
The `ListFunctions` test kept failing. It stopped failing as soon as I made the two `az` calls synchronous. I think the problem might have been spawning two concurrent Python processes (`az`) in Appveyor environment. I suspect if Appveyor had issues with this, maybe other environments will too.

I have made it synchronous for now. Although, each `az` call takes a few seconds and will not slow down all `func azure` commands by a small amount. @ahmelsayed, would love to also hear your thoughts.

ahmelsayed:
I think it's fine to call them one after the other.

PR #1324, opened 2019-05-20
Fixes #688 - Extensions Install should only update if needed

### LINKED ISSUE #688: Don't add or update extensions.csproj for http, timer only projects
Right now extensions.csproj would be created for a project regardless when `func extensiosn install` is run. This is because metadata generator is added to all projects by default. No binding and metadata generator is required if project only contains http and timer functions only.

ankitkumarr:
Need to verify `ExtensionsMetadataGeneratorPackage` with @soninaren, and need to ensure i didn't break extensions commands' expectations.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
@@ -142,19 +142,22 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         {
             string warningMessage = "No action performed because no functions in your app require extensions.";
 
-            var anyExtensionsToBeInstalled = ExtensionsHelper.GetExtensionPackages().Count() > 0;
-
-            // CASE 1: If there are any bindings that need to install extensions
-            if (anyExtensionsToBeInstalled)
+            // CASE 1: If users need a package to be installed
+            if (!string.IsNullOrEmpty(Package) || !string.IsNullOrEmpty(Version))
             {
                 return true;
             }
 
+            // CASE 2: If there are any bindings that need to install extensions
+            if (ExtensionsHelper.GetExtensionPackages().Count() > 0)
+            {
+                return true;
+            }
 
             var extensionsProjDir = string.IsNullOrEmpty(ConfigPath) ? Environment.CurrentDirectory : ConfigPath;
             var extensionsProjFile = Path.Combine(extensionsProjDir, Constants.ExtenstionsCsProjFile);
 
-            // CASE 2: No extensions.csproj
+            // CASE 3: No extensions.csproj
             if (!FileSystemHelpers.FileExists(extensionsProjFile))
             {
                 if (_showNoActionWarning)
@@ -164,7 +167,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 return false;
             }
 
-            // CASE 3: extensions.csproj present with only ExtensionsMetaDataGenerator in it
+            // CASE 4: extensions.csproj present with only ExtensionsMetaDataGenerator in it
             // We look for this special case because we had added ExtensionsMetaDataGenerator to all function apps.
             // These apps do not need to do a restore, so if only ExtensionsMetaDataGenerator is present, we don't need to continue
             var extensionsProject = ProjectHelpers.GetProject(extensionsProjFile);
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.ExtensionBundle;
@@ -16,7 +17,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
     internal class InstallExtensionAction : BaseAction
     {
         private readonly ISecretsManager _secretsManager;
-        private readonly bool _showExtensionBundleWarning;
+        private readonly bool _showNoActionWarning;
 
         public string Package { get; set; } = string.Empty;
         public string Version { get; set; } = string.Empty;
@@ -26,10 +27,10 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         public bool Csx { get; set; }
         public bool Force { get; set; } = false;
 
-        public InstallExtensionAction(ISecretsManager secretsManager, bool showExtensionBundleWarning = true)
+        public InstallExtensionAction(ISecretsManager secretsManager, bool showNoActionWarning = true)
         {
             _secretsManager = secretsManager;
-            _showExtensionBundleWarning = showExtensionBundleWarning;
+            _showNoActionWarning = showNoActionWarning;
         }
 
         public override ICommandLineParserResult ParseArgs(string[] args)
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -85,39 +85,42 @@ public async override Task RunAsync()
                 return;
             }
 
-            if (CommandChecker.CommandExists("dotnet"))
```
soninaren: I think we still need another if statement that checks for dotnet command and throws an exception with appropriate error message.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -85,39 +85,42 @@ public async override Task RunAsync()
...
             {
-                if (!string.IsNullOrEmpty(ConfigPath) && !FileSystemHelpers.DirectoryExists(ConfigPath))
-                {
-                    throw new CliArgumentsException("Invalid config path, please verify directory exists");
-                }
+                throw new CliArgumentsException("Invalid config path, please verify directory exists");
+            }
 
-                var extensionsProj = await ExtensionsHelper.EnsureExtensionsProjectExistsAsync(_secretsManager, Csx, ConfigPath);
+            var extensionsProj = await ExtensionsHelper.EnsureExtensionsProjectExistsAsync(_secretsManager, Csx, ConfigPath);
```
soninaren: This will add an extension csproj to the project. I think the correct fix is to not add this at all if there are no extensions to be installed.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -126,22 +129,25 @@ public async override Task RunAsync()
...
-            {
-                ColoredConsole.Error.WriteLine(ErrorColor(Constants.Errors.ExtensionsNeedDotnet));
-            }
         }
 
         private async Task AddPackage(string extensionsProj, string pacakgeName, string version)
         {
-            var args = $"add \"{extensionsProj}\" package {pacakgeName} --version {version}";
-            if (!string.IsNullOrEmpty(Source))
+            if (CommandChecker.CommandExists("dotnet"))
```
soninaren: This should be done with all the other validations. No need to check for dotnet before adding every single package.

src/Azure.Functions.Cli/Helpers/ExtensionsHelper.cs
```diff
@@ -63,7 +63,12 @@ public static IEnumerable<ExtensionPackage> GetExtensionPackages()
                 }
             }
 
-            packages.Add("ExtensionsMetadataGeneratorPackage", Constants.ExtensionsMetadataGeneratorPackage);
+            // We only need extensionsMetadatGeneratorPackage, if there is a trigger type that requires an extension
+            // So, if we didn't find any extension packages, we don't need to add this.
+            if (packages.Count != 0)
```
soninaren: This would break the logic here. since the package count will be different. need to update that as well.

https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/ExtensionBundle/JsonFileConfigurationSource.cs#L71-L73

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
ankitkumarr: I moved the check to AddPackage command before we actually run the dotnet command. This way we don't fail if the users don't have dotnet, and we don't end up having to run dotnet commands.

I agree with your comment there that we shouldn't check before every package add. Is there a way to add all the packages with one dotnet command? Running dotnet commands several times probably adds the overhead too.

src/Azure.Functions.Cli/Helpers/ExtensionsHelper.cs
ankitkumarr: That checks for less than equals to 1, and removes extensionsmeta if it is present. I don't think the logic there should break. But I can update that to 0 I suppose?

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
ankitkumarr: I agree. This probably adds an overhead. Will update.

soninaren:
Based on our discussion we need to do the following
- Make `func extension install` no-op only when the following conditions are met
    1. There is no `extensions.csproj` present.
    2. There are no known extensions that we can install.
- If `extension.csproj` is present.
    1. Run add new extensions if they are not present in the existing `extension.csproj`
    2. Perform `sync action`
- Publish action can simply call `func extension install`

What does `no-op` includes the following behavior change.
1. don't check for `dotnet` command being present.
2. don't create `extensions.csproj`
3. don't perform `sync action`

ankitkumarr:
@soninaren, I have updated the PR, as discussed offline. Please have a look.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -128,13 +124,68 @@ public async override Task RunAsync()
...
+            var extensionsInProject = extensionsProject.Items
+                    .Where(item => item.ItemType.Equals(Constants.PackageReferenceElementName, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
+            if (extensionsInProject.Count == 1 &&
+                extensionsInProject.FirstOrDefault(item =>
+                item.Include.Equals(Constants.ExtensionsMetadataGeneratorPackage.Name, StringComparison.OrdinalIgnoreCase)) != null)
+            {
+                if (_showNoActionWarning)
+                {
+                    ColoredConsole.WriteLine(WarningColor($"No action performed because only {Constants.ExtensionsMetadataGeneratorPackage.Name} reference was found." +
```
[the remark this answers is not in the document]
ankitkumarr: Do you think simplying would confuse the users? I was worried that if the ExtensionsMetadataGeneratorPackage was indeed present, users would get confused why it isn't doing an install like it used to.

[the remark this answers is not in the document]
ankitkumarr: Thinking again, maybe you are right in that this is an implementation detail and we can get away with leaving ambiguity.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -74,24 +75,19 @@ public override ICommandLineParserResult ParseArgs(string[] args)
 
         public async override Task RunAsync()
         {
-            var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
-            if (extensionBundleManager.IsExtensionBundleConfigured())
+
+            if (!string.IsNullOrEmpty(ConfigPath) && !FileSystemHelpers.DirectoryExists(ConfigPath))
```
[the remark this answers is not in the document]
ankitkumarr: I see what you are saying. But, I think whether or not extension bundles are configured, the first thing we should do is validate the arguments? Even if the users do not need that argument, if they explicitly specified it, I think it's not a bad idea to validate before executing the rest of the code. Ideally argument validation should even run when we parse the arguments.

src/Azure.Functions.Cli/Helpers/ExtensionsHelper.cs
```diff
@@ -63,7 +63,12 @@ public static IEnumerable<ExtensionPackage> GetExtensionPackages()
                 }
             }
 
-            packages.Add("ExtensionsMetadataGeneratorPackage", Constants.ExtensionsMetadataGeneratorPackage);
+            // We only need extensionsMetadatGeneratorPackage, if there is a trigger type that requires an extension
```
soninaren: Nit: not just trigger type, there are extensions that have only input or output bindings.

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -128,13 +134,62 @@ public async override Task RunAsync()
...
+            // CASE 2: extensions.csproj present with only ExtensionsMetaDataGenerator in it
+            // We look for this special case because we had added ExtensionsMetaDataGenerator to all function apps.
+            // These apps do not need to do a restore, so if only ExtensionsMetaDataGenerator is present, we don't need to continue
+            var extensionsProject = ProjectHelpers.GetProject(extensionsProjFile);
+            var extensionsInProject = extensionsProject.Items
+                    .Where(item => item.ItemType.Equals(Constants.PackageReferenceElementName, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
+            if (extensionsInProject.Count == 1 &&
+                extensionsInProject.FirstOrDefault(item =>
+                item.Include.Equals(Constants.ExtensionsMetadataGeneratorPackage.Name, StringComparison.OrdinalIgnoreCase)) != null)
```
soninaren: You need to check for `!anyExtensionsToBeInstalled` here as well.

ankitkumarr: Ah, good catch. I think I am just going to return true if `anyExtensionsToBeInstalled`

src/Azure.Functions.Cli/Actions/LocalActions/InstallExtensionAction.cs
```diff
@@ -78,20 +79,25 @@ public async override Task RunAsync()
...
+            if (!string.IsNullOrEmpty(ConfigPath) && !FileSystemHelpers.DirectoryExists(ConfigPath))
             {
-                if (!string.IsNullOrEmpty(ConfigPath) && !FileSystemHelpers.DirectoryExists(ConfigPath))
-                {
-                    throw new CliArgumentsException("Invalid config path, please verify directory exists");
-                }
+                throw new CliArgumentsException("Invalid config path, please verify directory exists");
+            }
+
+            if (!NeedsExtensionsInstall())
```
soninaren: `!string.IsNullOrEmpty(Package) && !string.IsNullOrEmpty(Version)` is being ignored here. We need to not skip in extension installation incase user has specified individual extension id and version.

Issue #1069, opened 2019-02-20
Use run from package by default on Linux App Service Plan


asavaritayal:

ankitkumarr:
@asavaritayal is this supported for certain? @ahmelsayed I thought synctriggers would not work for Linux App Service and so Run From Package wouldn't work?

ahmelsayed:
They are not really related. With ANT80, dedicated Linux should support run from package. We probably still shouldn't call synctriggers. Though I'm a bit unclear if premium needs sync triggers

ankitkumarr:
Oh, I see. I tried to create an App Service plan function app and published using `WEBSITE_RUN_FROM_PACKAGE`. The functions didn't seem to show up.

asavaritayal:
@ahmedelnably flagging this as P0 as it takes a really long time today to publish an app to Linux Dedicated. The scenario also impacts Functions on Linux Premium.

ankitkumarr:
I investigated why this does not currently work properly. 

I found two issues here:

1. If the `WEBSITE_RUN_FROM_PACKAGE` is off, `core-tools` turns it on and start a kudulite build. However, the setting update takes a couple of minutes. So, this makes kudu not pick up the setting yet. Kudu ends up building as if `WEBSITE_RUN_FROM_PACKAGE` is off, while platform expects the filesystem as if `WEBSITE_RUN_FROM_PACKAGE` is on. This causes things to break.

2. This needs a fix in kudulite's `/api/zipdeploy` endpoint. Kudulite currently sets `siteversion.txt` under `/home/data/SitePackages` to point to a zip file, when using `WEBSITE_RUN_FROM_PACKAGE`. 
However, this seems to not be picked up, and portal does not show function apps. When I manually rename that to `packagename.txt` as suggested here https://github.com/Azure/app-service-announcements/issues/84, functions did seem to appear. I think kudulite needs to set `packagename.txt` for function apps.

For 2., I will reproduce and confirm this. If that's the case, I will make a PR to kudulite..
For 1, I talked to @ahmelsayed, and there's not much we can do other than wait for kudulite to get updated (a couple of mins). This will only happen the first time users switch to `WEBSITE_RUN_FROM_PACKAGE`.

ankitkumarr:
For 2, kudulite PR- https://github.com/Azure-App-Service/KuduLite/pull/37

ankitkumarr:
[...]

PR #1329, opened 2019-05-20
Add poll logic for kudu env settings

ankitkumarr:
This is core-tools work item to support https://github.com/Azure/azure-functions-core-tools/issues/1069.

Unfortunately, I can't enable Run From Package for Linux dedicated yet. Will do that after ANT83, when the kudulite fix is deployed.

Already making this PR as it's useful for Windows cases as well to avoid race conditions, which I suspect some users might be hitting. And after this, Linux RunFromPackage support for core-tools would only need a minor change.


ankitkumarr:
hmmm, the ListFunctions test keeps failing. It's definitely not related to my PR, so I will merge it. @ahmelsayed, might be something with the App itself.

PR #1349, opened 2019-05-30
Minor code simplification for SyncTriggers

mathewc:


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -316,13 +316,8 @@ private async Task PublishFunctionApp(Site functionApp, GitIgnoreParser ignorePa
                 await PublishAppSettings(functionApp, new Dictionary<string, string>(), additionalAppSettings);
             }
 
-            // Syncing triggers is not required when using zipdeploy api
```
mathewc: Why is "Syncing triggers is not required when using zipdeploy api"? I wonder if we can't just always sync here to avoid confusion/complicated logic.

ankitkumarr: I believe it's because kudu takes care of calling synctriggers or settriggers.

mathewc: Can you link to the code that demonstrates that?

ankitkumarr: https://github.com/projectkudu/kudu/blob/b22dbdf3a520c6aec3b323d4b0f2799acd300787/Kudu.Core/Helpers/PostDeploymentHelper.cs#L113

Sorry, I should have been clear with the comment there.

mathewc: Ok, thanks. Updated the PR then to retain the current logic - I just made a code simplification to reduce the duplicated checks

PR #1351, opened 2019-05-31
Detect all Python versions, and do not fail unless needed

### LINKED ISSUE #1327: Handle Python versioning
**Scenario**
Currently, users with a Python version different than 3.6.x are blocked from creating and publishing Function Apps. We want to tweak the command line experience to unlock this scenario and surface a discoverable warning in the console during `func init`, `func start` and `func publish`.

ankitkumarr:
Will test out on different python binaries once the build is complete.
Closes #1327

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -459,7 +459,6 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
         public async Task PublishZipDeploy(Site functionApp, Func<Task<Stream>> zipFileFactory)
         {
-            (var content, var length) = CreateStreamContentZip(await zipFileFactory());
             await RetryHelper.Retry(async () =>
             {
                 using (var handler = new ProgressMessageHandler(new HttpClientHandler()))
@@ -470,6 +469,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
                     ColoredConsole.WriteLine("Creating archive for current directory...");
 
+                    (var content, var length) = CreateStreamContentZip(await zipFileFactory());
                     request.Content = content;
 
                     HttpResponseMessage response = null;
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -459,6 +459,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
         public async Task PublishZipDeploy(Site functionApp, Func<Task<Stream>> zipFileFactory)
         {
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -73,47 +74,83 @@ private static void CreateRequirements()
...
+
+            const string pythonStr = "python";
+            const string python3Str = "python3";
+            const string python36Str = "python3.6";
+            const string py36Str = "3.6";
+            const string py3Str = "3.";
+            const string warningMessage = "Python 3.6.x is recommended, and used in Azure Functions. You are using Python version {0}.";
+            const string errorMessageOldPy = "Python 3.x (recommended version 3.6.x) is required. Found python versions ({0}).";
+            const string errorMessageNoPy = "Python 3.x (recommended 3.6.x) is required. No Python versions were found.";
+
```
ankitkumarr: @asavaritayal to comment on the messages above we show the user.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -438,6 +438,7 @@ private async Task SetRunFromPackageAppSetting(Site functionApp, string runFromP
 
         public async Task PublishZipDeploy(Site functionApp, Func<Task<Stream>> zipFileFactory)
         {
+            (var content, var length) = CreateStreamContentZip(await zipFileFactory());
```
ahmelsayed: `content` will throw a `ObjectDisposedException` on retry below. It needs to be in the retry body

ankitkumarr: Oh right right! I will just move `await zipFileFactory()` out of retry then, and call `CreateStreamContentZip` inside the retry.

I did this because we were retrying if creating a zip file failed due to build problems. I don't think we need to do that.

ahmelsayed: this whole action needs to be rewritten.

ankitkumarr: I agree that it has become messy. Ok, I can leave this change out of this PR as it doesn't concern the issue directly. 
Feel free to assign me an issue for the rewrite if you'd like with any specific suggestions you had.

ankitkumarr:
In addition to supporting `python3.6`, this PR also allows users to publish if they have other Python versions. Although `init` and publishing would work for users, `func host start` may not work for users using Python versions other than 3.6, due to https://github.com/Azure/azure-functions-python-worker/issues/450

ankitkumarr:
On further discussion, we concluded that we want to let users Publish and Init if they are using Python 3.x. However, if users try to `func host start`, they need to be on a Python 3.6 environment (global or virtual).

PR #1353, opened 2019-06-03
Update azure-pipelines.yml for Azure Pipelines

soninaren:


azure-pipelines.yml
```diff
@@ -1,13 +1,15 @@
-# Starter pipeline
-# Start with a minimal pipeline that you can customize to build and deploy your code.
-# Add steps that build, run tests, deploy, and more:
-# https://aka.ms/yaml
+name: $(Build.SourceBranchName)_$(Build.Reason)_$(devops_buildNumber)
 
 pool:
-  vmImage: 'Ubuntu 16.04'
+  vmImage: 'vs2017-win2016'
```
brettsam: is there a vs2019 image?

azure-pipelines.yml
```diff
@@ -1,13 +1,15 @@
...
+
+variables:
+  devops_buildNumber: $[counter(format(''), 1300)]
+  APPVEYOR_REPO_BRANCH: $(Build.SourceBranchName)
+  APPVEYOR_REPO_COMMIT: $(Build.SourceVersion)
```
brettsam: Eventually we can remove these mentions of appveyor, i'm assuming?

build/Settings.cs
```diff
@@ -75,7 +75,7 @@ private static string config(string @default = null, [CallerMemberName] string k
 
         public static readonly string ProjectTemplates = $"https://www.nuget.org/api/v2/package/Microsoft.Azure.WebJobs.ProjectTemplates/{ProjectTemplatesVersion}";
 
-        public static string BuildNumber => config(null, "Build.BuildId") ?? config("9999", "APPVEYOR_BUILD_NUMBER");
+        public static string BuildNumber => config(null, "devops_buildNumber") ?? config("9999", "APPVEYOR_BUILD_NUMBER");
```
brettsam: Where did Build.BuildId come from? Can we just make devops_buildNumber == Build.BuildId so it stays the same? That way we can kick off from AppVeyor if we need? Or is this something else?

azure-pipelines.yml
soninaren: yes, i did this so that app veyor keeps working in case this doesn't. Just a fallback mechanism

build/Settings.cs
soninaren: The issue with build.BuildId is that we cannot start from a particular number and it also resets it 0 in certain conditions (not fully sure about this) since we already have a count built up in app veyor we need to start from there. That is why i set up devops_buildnumber

azure-pipelines.yml
soninaren: not sure, this i what we had of kafka so i used the same

PR #1357, opened 2019-06-03
white listing building only v1.x branch, bumping seed version for devops

soninaren:


brettsam:
Do we know why this build failed?

soninaren:
> Do we know why this build failed?

I thought it was because of variables not being available for pr.

brettsam:
They're available now... so in theory this should be working.

ahmelsayed:
Thanks @brettsam and @soninaren for getting this to work! Can you also get the artifacts to be available after a run? I'm not sure what the syntax for that is in azure-pipelines.yml. I think that's the last thing we need before merging this.

ahmelsayed:
actually I just noticed there are 2 PRs with the same name. Did you mean to close the other one?

brettsam:
We paused this while we work out some kinks over in naren's fork. We'll get a real PR here and clean this up shortly... I think I just figured out the last issue we were having.

ahmelsayed:
Sounds good. Thanks @brettsam

brettsam:
Okay I think we got it. That last build has artifacts.

PR #1362, opened 2019-06-05
publish updates

ahmelsayed:


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -291,10 +291,22 @@ private async Task PublishFunctionApp(Site functionApp, GitIgnoreParser ignorePa
...
+                if (workerRuntimeEnum == WorkerRuntime.python && !NoBuild && BuildNativeDeps)
+                {
+                    await PublishRunFromPackage(functionApp, await PythonHelpers.ZipToSquashfsStream(await zipStreamFactory()), $"{fileNameNoExtension}.squashfs");
+                }
+                else
+                {
+                    await PublishRunFromPackage(functionApp, await zipStreamFactory(), $"{fileNameNoExtension}.zip");
+                }
+            }
             // If Consumption Linux or Elastic Premium Linux
```
ankitkumarr: I suggest updating the comment to avoid confusion

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -18,16 +18,39 @@ public static class AzureHelper
     {
         private static string _storageApiVersion = "2018-02-01";
 
-        public static async Task<Site> GetFunctionApp(string name, string accessToken, string managementURL)
+        public static async Task<Site> GetFunctionApp(string name, string accessToken, string managementURL, string defaultSubscription = null)
         {
+            if (!string.IsNullOrEmpty(defaultSubscription))
+            {
```
ankitkumarr: Is the idea now that we will only look at one subscription (default or specified) for all Azure actions?

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -291,10 +291,22 @@ private async Task PublishFunctionApp(Site functionApp, GitIgnoreParser ignorePa
 
             Func<Task<Stream>> zipStreamFactory = () => ZipHelper.GetAppZipFile(workerRuntimeEnum, functionAppRoot, BuildNativeDeps, NoBuild, ignoreParser, AdditionalPackages, ignoreDotNetCheck: true);
 
+            var fileNameNoExtension = string.Format("{0}-{1}", DateTimeOffset.UtcNow.ToString("yyyyMMddHHmmss"), Guid.NewGuid());
+            if (functionApp.IsLinux && functionApp.IsDynamic)
+            {
+                if (workerRuntimeEnum == WorkerRuntime.python && !NoBuild && BuildNativeDeps)
```
ankitkumarr: Just wondering, why do squashfs only for `--build-native-deps`? Is it because of the common dependency on docker?

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -428,6 +440,11 @@ private async Task SetRunFromPackageAppSetting(Site functionApp, string runFromP
                 }
             }
 
+            if (functionApp.IsDynamic && functionApp.IsLinux && !functionApp.AzureAppSettings.ContainsKey("WEBSITE_MOUNT_ENABLED"))
+            {
+                functionApp.AzureAppSettings["WEBSITE_MOUNT_ENABLED"] = "1";
```
ankitkumarr: Please correct me if I am wrong, but this will be set even when users are not in `--build-native-deps`, aka when we don't do SquashFS. Just making sure that's what you intended.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
ahmelsayed: No, it's just to check that subscription first. I have about 50 subs in my account :). Everytime I do a publish it searches all 50 subs for the function app (in order) and my function app happens to be in sub number 40something. This will just put the default sub on the top of the list to check. If there is no default sub set, it'll just ignore it

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
ahmelsayed: yes. At least for now since this is the scenario that needs it the most. however the only dependency we need is `./mksquashfs` which is available for linux, mac and Windows. So we can check if it's installed and use it to package instead of requiring docker. 

I'll need to refactor all of this action and all the helper classes to get that wired up correctly though.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
ahmelsayed: Correct. Any new publish from the core-tools to consumption linux (zip or squashfs) will set this to `"1"`. Causing zip files to get mounted using `fuse-zip` and squashfs ones using `squashfuse`.

ahmelsayed: but it'll leave it alone if the user flips this to `"0"` if they wanna disable the mounting behavior.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
ahmelsayed: done

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
ankitkumarr: Oh right! I misread and didn't realize that the search actually goes through the others too. This is a great enhancement actually. It takes forever for me to find my function app as well. Thanks!

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
ankitkumarr: That makes sense. It's fair to leave it for `--build-native-deps` for now then. We can revisit later in needed.

PR #1368, opened 2019-06-06
Add ability to open appinsight live metrics in a browser

### LINKED ISSUE #1334: Log streaming for Python - Open Browser
On `func azure functionapp logstream <app_name>`, launch the **Live Metrics Stream** view in the browser.

ankitkumarr:
Fixes #1334 

Works well in Windows. Will test Mac and Linux once build is done.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
@@ -37,6 +37,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         public override async Task RunAsync()
         {
             var subscriptions = await AzureHelper.GetSubscriptions(AccessToken, ManagementURL);
+            ColoredConsole.WriteLine("Retrieving Function App...");
             var functionApp = await AzureHelper.GetFunctionApp(FunctionAppName, AccessToken, ManagementURL, allSubs: subscriptions);
             if (UseBrowser)
             {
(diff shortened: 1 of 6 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
```diff
@@ -1,27 +1,53 @@
...
+using Azure.Functions.Cli.Helpers;
+using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
     [Action(Name = "logstream", Context = Context.Azure, SubContext = Context.FunctionApp, HelpText = "Show interactive streaming logs for an Azure-hosted Function App")]
     class LogStreamAction : BaseFunctionAppAction
     {
+        private const string ApplicationInsightsIKeySetting = "APPINSIGHTS_INSTRUMENTATIONKEY";
+        private const string LiveMetricsUriTemplate = "https://ms.portal.azure.com/#blade/AppInsightsExtension/QuickPulseBladeV2/ComponentId/{0}/ResourceId/{1}";
```
ahmelsayed:
```suggestion
        private const string LiveMetricsUriTemplate = "https://portal.azure.com/#blade/AppInsightsExtension/QuickPulseBladeV2/ComponentId/{0}/ResourceId/{1}";
```

src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
```diff
@@ -42,5 +68,33 @@ public override async Task RunAsync()
...
+                    $"Please make sure you have Application Insights configured with you function app.");
+            }
+
+            var iKey = functionApp.AzureAppSettings[ApplicationInsightsIKeySetting];
+            if (string.IsNullOrEmpty(iKey))
+            {
+                throw new CliException("Invalid Instrumentation Key found. Please make sure that the Application Insights is configured correctly.");
+            }
+
+            var appId = await AzureHelper.GetApplicationInsightIDFromIKey(iKey, AccessToken, ManagementURL, allSubs: allSubscriptions);
```
ahmelsayed: what happens if the application insights resource doesn't exist?

src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
ankitkumarr: Sounds good. What's the difference btw? In my browser it always opened that. Is that internal?

src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
ankitkumarr: The function `GetApplicationInsightIDFromIKey` takes care of it and throws the right error - complaining about the instrumentation key.

If key doesn't exist, that is taken care of above.

src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
ahmelsayed: yes, if you log in using an @microsoft account, it redirects you to `ms.portal.azure.com` which is usually running a pre-release version of the UX.

ankitkumarr: Ah, that makes sense. Thank you!

src/Azure.Functions.Cli/Actions/AzureActions/LogStreamAction.cs
ahmelsayed: sounds good.

PR #1370, opened 2019-06-07
re-adding DependencyValidator

brettsam:
This will write out errors if someone is overriding (or adding additional) services that we require. For now, it's just an error message in the CLI but doesn't break anything.


brettsam:
Verified locally with build that this looks right. If you do something like register an additional `IHostedService`, you'll get a message locally with:

```
[6/7/2019 2:38:12 PM] Invalid host services detected.
[6/7/2019 2:38:12 PM] Microsoft.Azure.WebJobs.Script.WebHost: The following service registrations did not match the expected services:
[6/7/2019 2:38:12 PM]   [Invalid] ServiceType: Microsoft.Extensions.Hosting.IHostedService, Lifetime: Singleton, ImplementationType: MyFunctionApp55.MyHostedService
```

soninaren:
I am guessing we added the powershell dependency to host, because of which we had to remove dependency validator.

brettsam:
Yeah there was a conditional `IHostedService` that I didn't catch originally. So all powershell CLI apps were seeing the message. I added it last release -- but forgot we needed to remove the `NullDependencyValidator` here too.

PR #1390, opened 2019-06-18
Add support for server side build

Hazhzeng:
### Dependency
https://github.com/Azure/azure-functions-host/pull/4585

### Background
Allow user to use --server-side-build flag in core tools and enable 
The expected behaviors:
1. User use `func azure functionapp publish <function site name> --server-side-build` to trigger a new build.
2. The core tools will initiate a web request to `https://management.azure.com/subscriptions/<subscription>/resourceGroups/<resource group>/providers/Microsoft.Web/sites/<function site name>/hostruntime/**admin/getsitetoken**` to get an **x-ms-site-restricted-token**. This **x-ms-site-restricted-token** will be used as authentication token in scmsite endpoints.
3. Core tools packs the files in user's local source directory into a zip file.
4. Core tools upload the zip file to `scmsite/api/zipdeploy?isAsync=true`.
4. Core tools will pull the build status from `scmsite/deployments/<id>`.
5. Core tools will emit the server side build log from `scmsite/deployments/<id>/log`.
6. Core tools will finish the execution once it receives Deployment Success or Deployment Failure status.

### Notice
1. The --server-side-build flag is only available to Linux Consumption Python.
2. Every 3 seconds, the core tool checks the build status.
3. Every 3 seconds, the core tool pulls down the full server side build log, and concatenate the logs to console. The size of the log can go up to a few hundreds KBs. Since we have a 30MB response content limitation. I don't think this is an issue.

### CR Request
@ahmelsayed @ankitkumarr

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -240,7 +240,7 @@ namespace Azure.Functions.Cli.Helpers
             return packagesMd5 == requirementsTxtMd5;
         }
 
-        internal static async Task<Stream> GetPythonDeploymentPackage(IEnumerable<string> files, string functionAppRoot, bool buildNativeDeps, string additionalPackages)
+        internal static async Task<Stream> GetPythonDeploymentPackage(IEnumerable<string> files, string functionAppRoot, bool buildNativeDeps, BuildOption buildOption, string additionalPackages)
         {
             var reqTxtFile = Path.Combine(functionAppRoot, Constants.RequirementsTxt);
             if (!FileSystemHelpers.FileExists(reqTxtFile))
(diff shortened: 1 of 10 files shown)


ahmedelnably:
@ColbyTresness what is the official name of this feature?

ColbyTresness:
Per discussion offline, "server side build" is what we should be using for CLI options and such.

ColbyTresness:
Per discussion offline, we'll go with "--build" and various additional parameters (remote, local, container, none)

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -274,6 +280,10 @@ internal static async Task<Stream> GetPythonDeploymentPackage(IEnumerable<string
                     throw new CliException("Docker is required to build native dependencies for python function apps");
                 }
             }
+            else if (buildOption == BuildOption.Remote)
```
ahmelsayed: The md5 check below should be if not remote also.

ahmelsayed:
Just a minor comment. @ankitkumarr I'm assuming this will impact your tests?

@Hazhzeng I'm assuming we need ant83 before we can merge this, right?

ankitkumarr:
@ahmelsayed, yes, it shouldn't break anything ideally. I should add proper tests for server side build and remove tests for PyInstaller.

I also need to make an update in this flow for Linux dedicated. I am waiting for Ant83 too. And then, I will update the tests with the final publishing options.

Hazhzeng:
@ahmelsayed, yes, this change depends on ant83.

PR #1392, opened 2019-06-19
Initial --<lang> support

ahmelsayed:


src/Azure.Functions.Cli/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -19,15 +19,40 @@ public static WorkerRuntime CurrentWorkerRuntime
...
+            }
+        }
+
+        public static string CurrentLanguageOrNull { get; private set; } = null;
+
         public static void Init(ISecretsManager secretsManager, string[] args)
         {
             try
             {
-                if (args.Contains("--csharp") || args.Contains("--dotnet"))
```
Ma3yTa: @ahmelsayed Did you missed F#?

ahmelsayed: I don't think we have F# templates for the core-tools.

PR #1393, opened 2019-06-19
Don't suggest to install v1

forki:
I think it's not a good idea to suggest installing v1


ahmelsayed:
Makes sense. Thanks @forki

PR #1411, opened 2019-06-26
Add Azure Event Hub trigger type

t-shama:


t-shama:
@ahmelsayed - Made a small addition for Event Hub. If you could look at it whenever you get the chance, that would be great 😃

PR #1412, opened 2019-06-26
Check that dotnet is installed as a prerequisite for PowerShell function apps (in func start).

### LINKED ISSUE #1410: Add a prerequisite to validate that dotnet is installed for PowerShell function apps.
This is a customer reported issue. 

Repo steps:
User install the latest version of the Core Tools and creates a PowerShell function app and an HttpTrigger function.

``` 
Id CommandLine        
-- -----------                                                                                                                                                                                                      1 npm i -g azure-functions-core-tools --unsafe-perm true                                                                                                                                                           2 func
3 cd C:\temp\
4 func init --worker-runtime powershell
5 cd .\MyFunctionProj\
6 ls
7 func new -l powershell -t HttpTrigger -n DemoTrigger
8 func start
```

Func.exe fails with the following error:
```
A host error has occurred
Microsoft.Azure.WebJobs.Script: Failed to start Language Worker Channel for language :powershell. System.Diagnostics.Process: The system cannot find the file specified.
```
This issue happened because dotnet is not installed. I will update the Core Tools to validate that dotnet is installed for PowerShell function apps.

Francisco-Gamino:
Check that dotnet is installed as a prerequisite for PowerShell function apps (in func start). 
This fixes #1410

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -236,6 +236,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
         public override async Task RunAsync()
         {
             var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
+
             if (workerRuntime == WorkerRuntime.None)
             {
                 ColoredConsole.WriteLine(WarningColor("your worker runtime is not set. As of 2.0.1-beta.26 a worker runtime setting is required."))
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -296,6 +297,10 @@ private async Task PreRunConditions(WorkerRuntime workerRuntime)
                     ColoredConsole.WriteLine("Could not find a valid .csproj file. Skipping the build.");
                 }
             }
+            else if (workerRuntime == WorkerRuntime.powershell && !CommandChecker.CommandExists("dotnet"))
+            {
+                throw new CliException("dotnet sdk is required for PowerShell functions. Please install https://microsoft.com/net");
```
ankitkumarr: It would be good to specify .NET Core SDK is required (if so), because we have gotten questions on whether they need dotnet framework or core. And might be a good idea to use the same link as used below, just to keep things consistent :)
Or if you like this link better, feel free to update the constants below.
https://github.com/Azure/azure-functions-core-tools/blob/3815dbd2554f7658b0a1b8dd38c0498a1f346d2e/src/Azure.Functions.Cli/Common/Constants.cs#L56

Francisco-Gamino: I agree. We should keep consistency. Exception message and link updated.

PR #1430, opened 2019-07-07
Update path to bash

### LINKED ISSUE #1403: No such file or directory on macOS
Tools version: 2.7.1373
OS: macOS

I keep getting an error that states "No such file or directory" when I try to run `func init` or `func start`. With `init` I'm asked to select a worker runtime and then immediately see the error message after I do that. With `start` the message shows up immediately. This is not a problem when I run it in Ubuntu.

I have traced the issue to line 16 in `src/Azure.Functions.Cli/Common/CommandChecker.cs`. If I change `bash` to `/bin/bash` the error goes away. 

I can create a PR for this change, but since tests don't work with Core yet according to `Contributing.md`, I'm not confident this change will not break something else.

miguelzakharia:
- Fix path to bash so the tools can be used on macOS without throwing "No such file or directory" error.
- Fixes #1403


ahmelsayed:
Thank you @miguelzakharia for the contribution!

PR #1432, opened 2019-07-09
Perform sync trigger when server side build succeeds

Hazhzeng:
Previously, we run into an issue where the list function has the old content before the server side build.
To address this issue, after each server side build succeeds, we should perform a sync trigger before listing the functions information.


Hazhzeng:
> Why are we changing/removing app settings for server side build?

WEBSITE_RUN_FROM_PACKAGE app setting blocks server side build and it takes the precedence over SCM_RUN_FROM_PACKAGE when loading function app. We need to remove it to enable the server side build and allow the function host to select the correct package.

Hazhzeng:
@ahmedelnably  Since the `--build` command help is not public, will make a separate PR.

PR #1437, opened 2019-07-12
Remove WEBSITE_CONTENTSHARE for Linux Consumption server side build

Hazhzeng:
As per https://github.com/Azure/azure-cli/issues/9027

We should remove WEBSITE_CONTENTAZUREFILECONNECTIONSTRING and WEBSITE_CONTENTSHARE in functionapp appsettings.

These two settings are set by default when user creating a function app with **Azure CLI < 2.0.65**
To allow these old function apps to use our server side build, we should remove these settings beforehand.

Otherwise, the SCM worker in Linux Consumption is not going to start.


ahmelsayed:
Why does the scm worker care about these settings?

Hazhzeng:
@ahmelsayed If these two app settings are set for Linux Consumption, the datarole will throw **AzureFiles mounting blocked for site ...** and both host and scm worker will return 503.

ahmelsayed:
I see, we should track removing that.

Issue #1423, opened 2019-07-02
Add slots support for publish command


ahmedelnably:

ankitkumarr:
Closed by https://github.com/Azure/azure-functions-core-tools/pull/1454

PR #1454, opened 2019-07-23
Add slot support, and use ARG for function app

### LINKED ISSUE #1381: Try gathering Function App information from ARG instead of ARM
It looks like ARG will reduce the number of calls we need to make to retrieve a Function App information when performing `func azure functionapp` actions. This may speed up our commands for interacting with Azure.

Issue to track this test and implementation.

ankitkumarr:
Closes #1381, #1423 

- Using ARG is a lot faster as we don't need to go through subscription(s) page by page to find a Function App. It also saves an `az` call.
- Slot support for all azure commands.


ahmelsayed:
This is awesome! I just tried ARG the other day and the difference in perf while searching is very significant.

PR #1466, opened 2019-08-02
Add worker references directly to csproj temporarily

ankitkumarr:
Reverting this in a way -- https://github.com/Azure/azure-functions-core-tools/pull/1464

Also making Node.js worker to match the host version.

As discussed offline with @fabiocav, while we should try and remove the explicit dependencies from the csproj to make releases and version management easier, we are experiencing build issues, and can't delay the release. We may need to visit Nuget packaging logic of the language workers to see if we are able to get them as transient dependency later.

 @pragnagopa


ankitkumarr:
Fabio beat me to it

PR #1476, opened 2019-08-05
Add help description for remote build

Hazhzeng:
We add a description for --build flag

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -84,6 +84,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             Parser
                 .Setup<BuildOption>('b', "build")
                 .SetDefault(BuildOption.Default)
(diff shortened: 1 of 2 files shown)


ColbyTresness:
What about adding this to the list when the user runs func --help? This should show up as an option

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -84,6 +84,7 @@ public override ICommandLineParserResult ParseArgs(string[] args)
             Parser
                 .Setup<BuildOption>('b', "build")
                 .SetDefault(BuildOption.Default)
+                .WithDescription("Perform build action when deploying Linux function apps. (accepts: remote)")
```
ahmedelnably: `when deploying to a Linux Function app....`

Hazhzeng:
> What about adding this to the list when the user runs func --help? This should show up as an option

The CLI library (fluent-command-line-parser)'s --help context complies with **<command> <contexts> <actions>** hierarchy. Since the **--build** flag is under **azure functionapp publish** context, the help will only show up when user runs **func azure functionapp --help**.

Hazhzeng:
Close-Open pull request to retrigger build

PR #1497, opened 2019-08-15
Fix Zip issues when packaging in unix

ankitkumarr:
`System.io.Compression` causes issues with `fuse-zip` if the zip is created in a Unix environment. Even after adding explicit Unix permissions, the zip failed to be mounted properly.

Switching to `DotNetZip` (suggested by @ahmelsayed) seems to fix such mount issues.


src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -48,17 +48,22 @@ public static Stream CreateZip(IEnumerable<string> files, string rootPath)
...
                     zip.AddFile(fileName, fileName, rootPath);
                 }
+                zip.Save(fileStream);
             }
-
             fileStream.Seek(0, SeekOrigin.Begin);
             return fileStream;
         }
+
+        public static string FixFileNameForZip(this string value, string zipRoot)
```
ahmelsayed: no longer needed, right?

ankitkumarr: Oops! Yes, I will remove it.

src/Azure.Functions.Cli/Extensions/ZipFileExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public static void AddFile(this ZipFile file, string fileName, string zipRoot, OpenDelegate openDelegate, CloseDelegate closeDelegate)
+        {
+            var entry = file.AddEntry(fileName.FixFileNameForZip(zipRoot), openDelegate, closeDelegate);
+
+            // If not Windows, we can assume it's a Unix environment
+            // For Unix environments, we need to explicity add file attributes in the Zipfile
+            if (!PlatformHelper.IsWindows)
+            {
+                if (TryGetUnixFileAttributes(fileName, out int fileAttributes))
```
ankitkumarr: May need to get rid of this, as DotNetZip's ExternalAttributes are read differently. 
After several trail and error, I realized even with ExternalAttributes set properly such that permissions are propogated when unzipping, fuse-mount does not read it the same way as unzip would. 

Basically this is a bit tricky to make sure ExternalAttributes are read properly. I will try to find some other way.

ahmelsayed: Do you know if executable bit is set correctly when using `zip(1)` on linux or mac?

ahmelsayed: fuse has a bunch of options http://man7.org/linux/man-pages/man8/mount.fuse.8.html#OPTIONS that we can see if they make a difference. 

I'm trying to understand if this is a limitation of fuse-zip or of DotNetZip.

ankitkumarr: `zip(1)` works fine, with fuse-zip understanding it correctly. I think I need to look a bit deeper into how fuse-zip retrieves file permissions. It's either a limitation of DotNetZip, or I need to figure out a way to use ExternalAttributes in a way that is read same by `unzip` as well as `fuse-zip`. I will look at how `zip(1)` does it.

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
ankitkumarr: Actually, reverted the new commit. So it is needed now.

Issue #1194, opened 2019-04-11
[Python] requirements.txt empty on new projects


EricJizbaMSFT:
_originally filed by @jeffhollan in https://github.com/Microsoft/vscode-azurefunctions/issues/1178_

Realize this may be on another repo or the template feed - tried to poke around and couldn't figure out where it's pulled from, but today the project template for python results in a `requirements.txt` file that is empty.  This seems a bit strange to me as the project has `import azure.functions as func` which my VS Code instance can't give me nice intellisense for unless I have `azure-functions` as one of the pip packages to install.

Seems we should have the default project template include `azure-functions` as a single package to import by default.

ankitkumarr:
We recently made a change to tie `azure-functions` (pip package) with `azure-functions-core-tools`. So, it is expected that the users will get a blank `requirements.txt`. 

Having `azure-functions` in `requirements.txt` may result in users installing that package when they don't need to. @jeffhollan, would you suggest that we have users install `azure-functions` just for intellisense? 

This may be overkilling, but @EricJizbaMSFT, is it possible to create a project level `PYTHONPATH` for each Python app, where we add to that variable the location of the `azure-functions` inside the installed `azure-functions-core-tools`? I think that should enable intellisense.

jeffhollan:
I’m not sure. For me personally it helps as I don’t know all the API surface for the function library. But fine if we wait and get feedback before we include it always. It would be nice if there was a way to get the autocomplete still.

ankitkumarr:
I am with you that we definitely want features like auto-completion. I think we overlooked this part when we decided to bind the worker. Thank you for bringing this up. I will take a look at what we can do.

asavaritayal:
/cc @maiqbal11 

Want to echo the above that we absolutely want to have auto-complete and intellisense enabled for the user from the very beginning. We've received enough feedback to prove that this is valuable. 

> This may be overkilling, but @EricJizbaMSFT, is it possible to create a project level PYTHONPATH for each Python app, where we add to that variable the location of the azure-functions inside the installed azure-functions-core-tools? I think that should enable intellisense.

@ankitkumarr why do we need a project level PYTHON_PATH for this? Trying to understand if we can just use the current PYTHON_PATH whether it points to global python or a venv.

maiqbal11:
[...]

PR #1498, opened 2019-08-15
Add azure-functions to requirements

maiqbal11:
Fixes Azure/azure-functions-python-worker#511
Fixes https://github.com/Azure/azure-functions-core-tools/issues/1194

An important extension of the work already done in https://github.com/Azure/azure-functions-python-worker/pull/506 and https://github.com/Azure/azure-functions-python-worker/pull/409. With the separation of the worker and the library, the users should now have control over the `azure-functions` library in their `requirements.txt`. This should be included in the startup experience so users publishing to Azure have the knowledge that they are in control of this dependency.

### DIFF
--- a/src/Azure.Functions.Cli/Common/Constants.cs
+++ b/src/Azure.Functions.Cli/Common/Constants.cs
@@ -30,6 +30,7 @@ namespace Azure.Functions.Cli.Common
         public const string PythonDockerImageVersionSetting = "FUNCTIONS_PYTHON_DOCKER_IMAGE";
         public const string PythonDockerImageSkipPull = "FUNCTIONS_PYTHON_DOCKER_SKIP_PULL";
         public const string PythonDockerRunCommand = "FUNCTIONS_PYTHON_DOCKER_RUN_COMMAND";
+        public const string PythonFunctionsLibrary = "azure-functions";
         public const string FunctionsCoreToolsEnvironment = "FUNCTIONS_CORETOOLS_ENVIRONMENT";
 
         public static string CliVersion => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.ToString(3);
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -66,7 +66,7 @@ namespace Azure.Functions.Cli.Helpers
         {
             if (!FileSystemHelpers.FileExists(Constants.RequirementsTxt))
             {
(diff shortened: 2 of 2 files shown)


Hazhzeng:
By merging this PR, resolves the issue in https://github.com/Azure/azure-functions-python-worker/issues/511

maiqbal11:
> By merging this PR, resolves the issue in [Azure/azure-functions-python-worker#511](https://github.com/Azure/azure-functions-python-worker/issues/511)

Thanks for pointing this out! Updated the description to reflect this.

maiqbal11:
> This is great! Assuming you have tested this E2E

I was able to verify that this works in the following scenarios for Consumption and Dedicated:

1. Stand alone without any other dependencies (default startup experience). 
2. Other `azure` dependency (specifically `azure-cosmos`). 

The user included `azure-functions` in used in both of these cases. Inclusion of other, non-`azure`, dependencies does not have any bearing on this behavior. Main point of concern has always been whether this conflicts with `azure` packages and it definitely doesn't.

ahmelsayed:
didn't we explicitly remove this because we wanted the runtime version to dictate the worker version, not the user package?

maiqbal11:
> didn't we explicitly remove this because we wanted the runtime version to dictate the worker version, not the user package?

This is the `azure-functions` library which contains the types that the user references in their code. The `azure-functions-worker` package is the worker and is still not exposed to the user.

ahmelsayed:
so there are no compatibility requirements between these 2 packages?

maiqbal11:
> so there are no compatibility requirements between these 2 packages?

There are, but including the `azure-functions` for the init experience ensures that users will get the latest version. Versions older than `1.0.0b5` will have compatibility issues.

ahmelsayed:
I meant more moving forward. Packages published today with whatever version of `azure-functions` is there today will need to always work regardless of the future versions of `azure-functions-worker`. These 2 packages were tightly tied together, and updating one without the other would cause issues. As long that's not the case anymore, it's fine. Just wanted to make sure

maiqbal11:
The coupling is much looser now than it was in the past. We can version 99% of the logic in the worker without having to worry about compatibility. There's one particular entry point in `azure-functions` that would always needs to be present, however. Otherwise the worker would not be able to support rich bindings in their current form. Certainly possible to refactor that away as well, but I believe the current state should suit us well. With the investment that is upcoming for live bindings, and the past investment that we made in generic bindings, it's very unlikely that we'll be expanding the scope of, or introducing breaks, in `azure-functions`.

PR #1499, opened 2019-08-15
Add remote build warning message when SCM site is not available

Hazhzeng:
If a function app is created before ANT83, the datarole does not contain SCM site.
Thus, core tools cannot find the scm/api/zipdeploy endpoints.
We should warn users from it.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -635,6 +635,12 @@ namespace Azure.Functions.Cli.Actions.AzureActions
 
         public async Task<DeployStatus> PerformServerSideBuild(Site functionApp, Func<Task<Stream>> zipFileFactory, Func<HttpClient, Task<DeployStatus>> deploymentStatusPollTask, bool attachRestrictedToken = true)
         {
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -635,8 +636,15 @@ public async Task PublishZipDeploy(Site functionApp, Func<Task<Stream>> zipFileF
 
         public async Task<DeployStatus> PerformServerSideBuild(Site functionApp, Func<Task<Stream>> zipFileFactory, Func<HttpClient, Task<DeployStatus>> deploymentStatusPollTask, bool attachRestrictedToken = true)
         {
+            Uri scmUri = null;
+            if (!Uri.TryCreate($"https://{functionApp.ScmUri}", UriKind.Absolute, out scmUri))
+            {
+                throw new CliException($"Your function app {functionApp.SiteName} does not support remote build. " + 
+                    "To enable remote build, please update your function app to the latest verison by recreating it.");
+            }
```
ankitkumarr: Is the only way to recreate it? Or can they use AppSettings update or something?

Hazhzeng: We checked that, the app settings cannot change function app metadata in datarole. Sadly.

PR #1504, opened 2019-08-16
Update remote build unavailable warning text change

Hazhzeng:
We add a precise date for function apps that doesn't support remote build.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -637,8 +637,8 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         {
             if (string.IsNullOrEmpty(functionApp.ScmUri))
             {
-                throw new CliException($"Your function app {functionApp.SiteName} does not support remote build. " + 
-                    "To enable remote build, please update your function app to the latest verison by recreating it.");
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -637,8 +637,8 @@ public async Task<DeployStatus> PerformServerSideBuild(Site functionApp, Func<Ta
         {
             if (string.IsNullOrEmpty(functionApp.ScmUri))
             {
-                throw new CliException($"Your function app {functionApp.SiteName} does not support remote build. " + 
-                    "To enable remote build, please update your function app to the latest verison by recreating it.");
+                throw new CliException($"Your function app {functionApp.SiteName} does not support remote build due to the app was created before 8/1/2019. " +
```
ahmedelnably: How about this?

"Your function app {functionApp.SiteName} does not support remote build, the cause could be that the app was created before 8/1/2019. "

ankitkumarr: I would rather be direct as we know for sure that's the reason.
"Remote build is a new feature added to function apps. Your function app {functionApp.SiteName} does not support remote build as it was created before Aug 1, 2019. "

ahmedelnably: Even better :)

PR #1511, opened 2019-08-20
Fix dedicated linux build tiny bug

### LINKED ISSUE #1510: Handle invalid response for /api/isdeploying in Linux Dedicated builds

ankitkumarr:
Fixes #1510 

with ANT 84, kudulite's `/api/isdeploying` returns a JSON response with multiple elements, some of which are not boolean. We should only extract and parse to boolean what we know for to be boolean.


ahmelsayed:
what is the payload like? You're still throwing an error if it's not boolean, right?

ankitkumarr:
Yes, the payload should be
{
"value": <boolean>,
"msg" : <string> (recently added)
}

Initially we were assuming it was a dictionary of string to boolean. But now it's not. 

I could add a retry if `value` is not a boolean. But it should always be. At least at the moment

Issue #1389, opened 2019-06-18
DockerFile PowerShell


bgelens:
Currently, when using `--docker` when PowerShell is chosen, only a dockerignore file is created. The dockerfile is not.

Would be great if the dockerfile for PowerShell would be created as well.

cc @TylerLeonhardt

ahmedelnably:
@eamonoreilly  and @Francisco-Gamino 
Is this a supported scenario for PowerShell?

guojian83:
Will the following Dockerfile do the work?

FROM mcr.microsoft.com/azure-functions/powershell:2.0

ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
    AzureFunctionsJobHost__Logging__Console__IsEnabled=true

COPY . /home/site/wwwroot
It seems to be working when i tested it.

I can submit a PR if that's all about it.

Regards,
Jian

bgelens:
Works fine! Just add `\` between the two environment variables

```docker
FROM mcr.microsoft.com/azure-functions/powershell:2.0

ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
  AzureFunctionsJobHost__Logging__Console__IsEnabled=true

COPY . /home/site/wwwroot
```

guojian83:
Yup. Sorry for the problem caused by online editor.
Updated in PR #1508.

Regards,
Jian

PR #1513, opened 2019-08-21
Fix: add Dockerfile for powershell worker runtime

guojian83:
The current cli is missing the Dockerfile when the worker runtime is powershell.

Please see: DockerFile PowerShell #1389

Regards,
Jian


Francisco-Gamino:
Hello @guojian83, thanks again for sending this!

Hello @ankitkumarr, the code change LGTM; however, could you please take a quick look as you are more familiar with this code base? Thank you!

PR #1532, opened 2019-09-04
Initial work to add telemetry

ankitkumarr:
Most of the telemetry client code comes from https://github.com/dotnet/cli/tree/8c678f74f037aba3600ad490d0fb6772d2811c2d/src/dotnet/Telemetry

Thanks to the dotnet team!

Closes https://github.com/Azure/azure-functions-core-tools/issues/1535

### DIFF
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -13,6 +13,7 @@ using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Threading.Tasks;
+using System.Text.RegularExpressions;
 
 namespace Build
 {
@@ -40,6 +41,22 @@ namespace Build
             .Aggregate(string.Empty, (a, b) => $"{a} --source {b}");
 
             Shell.Run("dotnet", $"restore {Settings.ProjectFile} {feeds}");
+        }
+
+        public static void ReplaceTelemetryInstrumentationKey()
+        {
+            var instrumentationKey = Settings.TelemetryInstrumentationKey;
+            if (!string.IsNullOrEmpty(instrumentationKey))
+            {
+                // Given the small size of the file, it should be ok to load it in the memory
+                var constantsFileText = File.ReadAllText(Settings.ConstantsFile);
+                if (Regex.Matches(constantsFileText, Settings.TelemetryKeyToReplace).Count != 1)
+                {
+                    throw new Exception($"Could not find exactly one {Settings.TelemetryKeyToReplace} in {Settings.ConstantsFile} to replace.");
+                }
+                constantsFileText = constantsFileText.Replace(Settings.TelemetryKeyToReplace, instrumentationKey);
+                File.WriteAllText(Settings.ConstantsFile, constantsFileText);
+            }
         }
 
         private static string GetRuntimeId(string runtime)
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -23,6 +23,8 @@ namespace Build
         public const string ItemTemplatesVersion = "2.0.10369";
         public const string ProjectTemplatesVersion = "2.0.10369";
 
+        public static readonly string ConstantsFile = Path.Combine(SrcProjectPath, "Common", "Constants.cs");
+
         public static readonly string SrcProjectPath = Path.GetFullPath("../src/Azure.Functions.Cli/");
 
         public static readonly string TestProjectPath = Path.GetFullPath("../test/Azure.Functions.Cli.Tests/");
@@ -75,9 +77,13 @@ namespace Build
 
         public static readonly string ProjectTemplates = $"https://www.nuget.org/api/v2/package/Microsoft.Azure.WebJobs.ProjectTemplates/{ProjectTemplatesVersion}";
 
+        public static readonly string TelemetryKeyToReplace = "00000000-0000-0000-0000-000000000000";
+
         public static string BuildNumber => config(null, "devops_buildNumber") ?? config("9999", "APPVEYOR_BUILD_NUMBER");
 
         public static string CommitId => config(null, "Build.SourceVersion") ?? config("N/A", "APPVEYOR_REPO_COMMIT");
+
+        public static string TelemetryInstrumentationKey => config(null, "TELEMETRY_INSTRUMENTATION_KEY");
 
         public static string BuildArtifactsStorage => config(null);
 
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -22,7 +22,7 @@ namespace Build
 
         public const string ItemTemplatesVersion = "2.0.10369";
         public const string ProjectTemplatesVersion = "2.0.10369";
-        
+
         public static readonly string SrcProjectPath = Path.GetFullPath("../src/Azure.Functions.Cli/");
 
         public static readonly string ConstantsFile = Path.Combine(SrcProjectPath, "Common", "Constants.cs");
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -22,10 +22,10 @@ namespace Build
 
         public const string ItemTemplatesVersion = "2.0.10369";
         public const string ProjectTemplatesVersion = "2.0.10369";
+        
+        public static readonly string SrcProjectPath = Path.GetFullPath("../src/Azure.Functions.Cli/");
 
         public static readonly string ConstantsFile = Path.Combine(SrcProjectPath, "Common", "Constants.cs");
-
-        public static readonly string SrcProjectPath = Path.GetFullPath("../src/Azure.Functions.Cli/");
 
         public static readonly string TestProjectPath = Path.GetFullPath("../test/Azure.Functions.Cli.Tests/");
 
--- a/src/Azure.Functions.Cli/Actions/BaseAction.cs
+++ b/src/Azure.Functions.Cli/Actions/BaseAction.cs
@@ -7,6 +7,8 @@ using Fclp.Internals;
 using System.Linq;
 using Azure.Functions.Cli.Telemetry;
 using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Common;
+using Colors.Net;
 
 namespace Azure.Functions.Cli.Actions
 {
@@ -16,9 +18,12 @@ namespace Azure.Functions.Cli.Actions
 
         public IEnumerable<ICommandLineOption> MatchedOptions { get; private set; }
 
+        public IDictionary<string, string> TelemetryCommandEvents { get; private set; }
+
         public BaseAction()
         {
             Parser = new FluentCommandLineParser();
+            TelemetryCommandEvents = new Dictionary<string, string>();
         }
 
         public virtual ICommandLineParserResult ParseArgs(string[] args)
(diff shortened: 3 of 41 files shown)


ankitkumarr:
@ahmelsayed / @fabiocav / @ahmedelnably, this should be ready to merged in. As telemetry will be disabled until we replace the instrumentation key as part of our build.

We can do that later once we have other components such as confirmation from partners, merge opt-out variable in homebrew, apt and chocolaty.

ankitkumarr:
Looks like github checks froze. Re-opening the same PR to see if that fixes it.

build/BuildSteps.cs
```diff
@@ -42,6 +43,22 @@ public static void RestorePackages()
             Shell.Run("dotnet", $"restore {Settings.ProjectFile} {feeds}");
         }
 
+        public static void ReplaceTelemetryInstrumentationKey()
+        {
+            var instrumentationKey = Settings.TelemetryInstrumentationKey;
+            if (!string.IsNullOrEmpty(instrumentationKey))
+            {
+                // Given the small size of the file, it should be ok to load it in the memory
```
ahmelsayed: also it's just a build script :)

src/Azure.Functions.Cli/Common/StaticSettings.cs
```diff
@@ -1,9 +1,13 @@
...
+﻿using Azure.Functions.Cli.Helpers;
+using System;
 
 namespace Azure.Functions.Cli.Common
 {
     public static class StaticSettings
     {
         public static bool IsDebug => Environment.GetEnvironmentVariable(Constants.CliDebug) == "1";
+
+        public static bool IsTelemetryEnabled => !(EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.TelemetryOptOutVariable)
```
ahmelsayed: shouldn't this just check if that variable is defined? do we need to take its content into account?

ankitkumarr: I figured we want to opt out if it's set to `1` or to `true`. That's what the .NET core CLI does as well (and I have been following their pattern :) ). I can change it to just look for the variable. I think that makes sense to be rather cautious and just assume that if the variable is set, user really wants to opt out.

src/Azure.Functions.Cli/Helpers/TelemetryHelpers.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                LogEventIfAllowed(telemetry, telemetryEvent);
+            }
+            catch
+            { 
+                // oh well!
+            }
+        }
+
+        public static void LogEventIfAllowed(ITelemetry telemetry, TelemetryEvent telemetryEvent)
```
fabiocav: Wondering if this return a  boolean to indicate whether the event was logged.

src/Azure.Functions.Cli/Telemetry/DockerContainerDetectorForTelemetry.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        private static bool ReadProcToDetectDockerInLinux()
+        {
+            return File
+                .ReadAllText("/proc/1/cgroup")
+                .Contains("/docker/");
+        }
+    }
+
+    internal enum IsDockerContainer
```
fabiocav: nit: drop the `Is` from the enum name

src/Azure.Functions.Cli/Telemetry/MacAddressGetter.cs
```diff
@@ -0,0 +1,151 @@
...
+                    return GetMacAddressByNetworkInterface();
+                }
+                else
+                {
+                    throw;
+                }
+            }
+        }
+
+        private static string ParseMACAddress(string shelloutput)
```
fabiocav: nit: Inconsistent MAC/Mac in this class.

src/Azure.Functions.Cli/Telemetry/MacAddressGetter.cs
```diff
@@ -0,0 +1,151 @@
...
+    internal static class MacAddressGetter
+    {
+        private const string MacRegex = @"(?:[a-z0-9]{2}[:\-]){5}[a-z0-9]{2}";
+        private const string ZeroRegex = @"(?:00[:\-]){5}00";
+        private const int ErrorFileNotFound = 0x2;
+        public static string GetMacAddress()
+        {
+            try
+            {
+                var shelloutput = GetShellOutMacAddressOutput().Result;
```
fabiocav: Use `GetAwaiter().GetResult()` if you must go with this

src/Azure.Functions.Cli/Telemetry/PersistenceChannel/PersistenceChannelDebugLog.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Azure.Functions.Cli.Telemetry.PersistenceChannel
+{
+    internal static class PersistenceChannelDebugLog
+    {
+        private static readonly bool _isEnabled = IsEnabledByEnvironment();
+
+        private static bool IsEnabledByEnvironment()
+        {
+            return EnvironmentHelper.GetEnvironmentVariableAsBool("FUNCTIONS_CORE_TOOLS_ENABLE_PERSISTENCE_CHANNEL_DEBUG_OUTPUT");
```
fabiocav: Should this be moved to a constant?

src/Azure.Functions.Cli/Actions/BaseAction.cs
```diff
@@ -27,6 +42,23 @@ public void SetFlag<T>(string longOption, string description, Action<T> callback
             }
         }
 
+        public virtual void UpdateTelemetryEvent(TelemetryEvent telemetryEvent)
```
fabiocav: This should likely be moved to `TelemetryHelpers` and take the events as a parameter.

src/Azure.Functions.Cli/ConsoleApp.cs
```diff
@@ -71,6 +91,16 @@ public static async Task RunAsync<T>(string[] args, IContainer container)
                     Console.ReadKey(true);
                 }
 
+                stopWatch.Stop();
+
+                // Log the event if we did recognize an event
+                if (!string.IsNullOrEmpty(app._telemetryEvent?.CommandName))
```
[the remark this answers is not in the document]
ankitkumarr: The reason I didn't is we have `Environment.Exit` in this block which would not execute the `finally` block.

I can restructure to move the exit to finally block as well. I took the lazy approach here.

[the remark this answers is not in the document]
fabiocav: I see... yeah, I think we can change this a bit to consolidate this code. 👍

src/Azure.Functions.Cli/Actions/BaseAction.cs
ankitkumarr: I think I created this virtual method with the idea that any Actions can override and manipulate the `telemetryEvent`. But then I moved to having a local `TelemetryCommandEvents` for commands to write to. I still left this method here, as -- 

a) it does have a dependency on the local `TelemetryCommandEvents` (which I think I should have made private), and 
b) still allows commands to override this and manipulate the telemetryevent, as this is latery called by the ConsoleApp block to update the TelemetryEvent.


If we think that's a better design, I can move it to `TelemetryHelpers`, and call that from ConsoleApp, would just need to leave `TelemetryCommandEvents` with a public getter.

build/Settings.cs
```diff
@@ -22,9 +22,11 @@ private static string config(string @default = null, [CallerMemberName] string k
 
         public const string ItemTemplatesVersion = "2.0.10369";
         public const string ProjectTemplatesVersion = "2.0.10369";
-
+
```
ankitkumarr: Note to self: remove this

src/Azure.Functions.Cli/Actions/BaseAction.cs
fabiocav: Unless there's a strong reason you can see to keep this here, I would recommend moving this to the helper as it is out of place in this type (this logic doesn't really belong here). Passing the events as an argument to this would be a better option.

src/Azure.Functions.Cli/npm/lib/install.js
```diff
@@ -47,6 +47,14 @@ const proxy = process.env.npm_config_https_proxy ||
             process.env.HTTP_PROXY ||
             process.env.http_proxy;
 
+const telemetryInfo = os.EOL
+    + 'Telemetry' + os.EOL
+    + '---------' + os.EOL
+    + 'The Azure Functions Core tools collect usage data in order to help us improve your experience. ' + os.EOL
+    + 'The data is anonymous and doesn\'t include any user specific or personal information. The data is collected by Microsoft.' + os.EOL
+    + os.EOL
+    + 'You can opt-out of telemetry by setting the FUNCTIONS_CORE_TOOLS_TELEMETRY_OPTOUT environment variable to \'1\' or \'true\' using your favorite shell.' + os.EOL
```
ahmelsayed: btw, we will need this for homebrew, chocolatey and apt. Homebrew is probably the easiest to update, but we should look into others too before we release

ankitkumarr: Yes, i think I still have the PRs from our last attempt :). Will just update them

src/Azure.Functions.Cli/Common/StaticSettings.cs
```diff
@@ -1,9 +1,13 @@
...
+                || Constants.TelemetryInstrumentationKey == "00000000-0000-0000-0000-000000000000");
```
ankitkumarr: Also this should be "!=" :sigh:

src/Azure.Functions.Cli/Actions/BaseAction.cs
ankitkumarr: Yeah, thinking about it, you are right. It doesn't belong here. Nothing specific to the action is happening in the function. I will update this.

src/Azure.Functions.Cli/Helpers/TelemetryHelpers.cs
ankitkumarr: I can make that. But right now, I don't see the need to use the boolean value. We could have a debug log for whether or not we logged it I suppose. But I figured, if Telemetry is off, I would make it completely off (no logs about it etc).
But I also don't see a strong reason to not log that "telemetry logging was not allowed".

fabiocav: Not super important... just something I noticed and wanted to bring up.

ankitkumarr:
One thing I wanted to explicitly call out here is -- currently we flush in the beginning and right at the end. Flush creates a new thread but does not block on it. So any time a command is executed, a telemetry item gets created, and right when flush is called, the process usually completes. So a command telemetry usually won't be flushed until the next command.

I thought of adding 1 second delay after flush, voted against it's probably ok to have telemetry for until N - 1 command execution.

src/Azure.Functions.Cli/npm/lib/install.js
ankitkumarr: @ahmelsayed / @fabiocav , I turned telemetry [off for now](https://github.com/Azure/azure-functions-core-tools/pull/1532/files#diff-cb827ae1b3f8f061cc7b979172f1223bR12-R13). 
I moved this out to a separate PR. Once we are ready to start the telemetry, we can merge these PRs below (will add another one later to turn it on) and turn it on. I would prefer to not block this PR.
- https://github.com/Azure/azure-functions-core-tools/pull/1555
- https://github.com/Azure/homebrew-functions/pull/15
- https://github.com/ahmelsayed/publish-script/pull/1

PR #1544, opened 2019-09-06
Added host and trigger name mapping for rabbitmq

katiecai:
Maps the string "rabbitmqtrigger" to "rabbitmq" for Kubernetes. Also populates "host" in metadata with the RabbitMQ connection string.

Details: https://github.com/Azure/azure-functions-rabbitmq-extension/issues/55

@pragnagopa @ahmelsayed @jeffhollan

### DIFF
--- a/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
+++ b/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
@@ -442,7 +442,7 @@ namespace Azure.Functions.Cli.Kubernetes
                                     .Where(i => i.Value.Type == JTokenType.String)
                                     .ToDictionary(k => k.Key, v => v.Value.ToString());
 
-            if (t["type"].ToString().ToLower() == "rabbitmqtrigger")
+            if (t["type"].ToString() == "rabbitMQTrigger")
             {
                 metadata["host"] = metadata["connectionStringSetting"];
                 metadata.Remove("connectionStringSetting");
--- a/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
+++ b/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
@@ -442,7 +442,7 @@ namespace Azure.Functions.Cli.Kubernetes
                                     .Where(i => i.Value.Type == JTokenType.String)
                                     .ToDictionary(k => k.Key, v => v.Value.ToString());
 
-            if (t["type"].ToString() == "rabbitMQTrigger")
+            if (t["type"].ToString().Equals("rabbitMQTrigger", StringComparison.InvariantCultureIgnoreCase))
             {
                 metadata["host"] = metadata["connectionStringSetting"];
                 metadata.Remove("connectionStringSetting");
--- a/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
+++ b/src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
@@ -436,7 +436,7 @@ namespace Azure.Functions.Cli.Kubernetes
             };
         }
 
-        private static IDictionary<string, string> PopulateMetadataDictionary(JToken t)
+        internal static IDictionary<string, string> PopulateMetadataDictionary(JToken t)
         {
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -430,14 +430,26 @@ private static ScaledObjectV1Alpha1 GetScaledObject(string name, string @namespa
...
             };
         }
 
+        private static IDictionary<string, string> PopulateMetadataDictionary(JToken t, IDictionary<string, string> secrets)
+        {
+            IDictionary<string, string> metadata = t.ToObject<Dictionary<string, JToken>>()
+                                    .Where(i => i.Value.Type == JTokenType.String)
+                                    .ToDictionary(k => k.Key, v => v.Value.ToString());
+
+            if (t["type"].ToString().ToLower() == "rabbitmqtrigger")
```
pragnagopa: Use  StringComparison.InvariantCultureIgnoreCase instead of toLower()

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -430,14 +430,26 @@ private static ScaledObjectV1Alpha1 GetScaledObject(string name, string @namespa
...
-                            Metadata = t.ToObject<Dictionary<string, JToken>>()
-                                    .Where(i => i.Value.Type == JTokenType.String)
-                                    .ToDictionary(k => k.Key, v => v.Value.ToString())
+                            Metadata = PopulateMetadataDictionary(t, secrets)
                         })
                 }
             };
         }
 
+        private static IDictionary<string, string> PopulateMetadataDictionary(JToken t, IDictionary<string, string> secrets)
```
pragnagopa: make this method internal and add unit test

jeffhollan:
This looks good. Just want to confirm one aspect.  The way we generate function.json is actually accurate, and a "bug" with KEDA I can try to fix this weekend.  The eventual metadata it should generate is something like:

```yaml
triggerType: rabbitMq
metadata:
   host: RabbitMqConnection
   queueName: myQueue
```

Today it looks like this in KEDA
```yaml
triggerType: rabbitMq
metadata:
   host: amqp://user:password@host:port
   queueName: myQueue
```

I wanted to confirm this PR isn't going to "resolve" the app setting and put in the actual value they have from secrets, but instead just put in the value of ConnectionStringSetting (some string that references the setting) but does rename it to host.

( [relevant work item in KEDA to fix this](https://github.com/kedacore/keda/issues/89) )

Hopefully that makes sense

jeffhollan:
Let me know which one this PR does - we may need to huddle on this on Monday too as even if I make the change to the rabbitMQ KEDA piece I'll have to figure out when / how we ship it

// @ahmelsayed @anirudhgarg

pragnagopa:
This PR is extracting the secrets for host property. @katiecai  - please update the PR to use setting name for host property. Thanks!

jeffhollan:
Thanks @pragnagopa - and thinking about how we are going to release this fix in KEDA just flagging it may also be adding a new "non-breaking" property called `connection` that will be the value.  We can chat monday but that would mean in this PR it looks good but:

- Change the mapping from `connectionStringSetting` -> `host` to `connectionStringSettings` -> `connection`
- Change so it doesn't extract the secret but just puts the environment variable name (e.g. RabbitMqConnection)

Apologies for last minute changes but just trying to make sure all these changes align for a full working e2e with KEDA, App Services, Core Tools

jeffhollan:
And I'd need @ahmelsayed to agree to my proposal above, so don't take it as doctrine :smile:

katiecai:
@jeffhollan @ahmelsayed told me about the bug in KEDA and told me to use the connection string value for now. Let me know what I should do when we decide how we want to deal with this. Thanks!

jeffhollan:
Will do.  Let me chat with @ahmelsayed today and will look to give you an answer in a few horus

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
pragnagopa: @katiecai -please address CR feedback and update PR

pragnagopa:
Please rebase

PR #1545, opened 2019-09-06
add gozip

ahmelsayed:
```
Usage of ./gozip:
  -base-dir string
        Base dir for the zip archive
  -dir string
        dir to zip
  -input-file string
        a file containing a list of files to archive
  -output string
        Output file name
  -quiet
        no output
```

### DIFF
--- a/tools/go/gozip/main.go
+++ b/tools/go/gozip/main.go
@@ -20,6 +20,8 @@ var (
 	fquiet     bool
 )
 
+const pathSeparator = string(os.PathSeparator)
+
 func init() {
 	flag.StringVar(&fbaseDir, "base-dir", "", "Base dir for the zip archive")
 	flag.StringVar(&fdir, "dir", "", "dir to zip")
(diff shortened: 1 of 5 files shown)


tools/go/gozip/main.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+
+	header, err := zip.FileInfoHeader(info)
+	if err != nil {
+		return err
+	}
+
+	baseLength := len(baseInZip)
+	if !strings.HasSuffix(baseInZip, "/") {
```
ankitkumarr: Would `go` always use `/` separator regardless of the OS? I am guessing yes, because of the use of `path`. But just wanted to confirm. :)

ahmelsayed: good point. I'll fix it

ahmelsayed: good point. I'll fix it

ahmelsayed:
let me know if you are able to try it. We will need to update installers to chmod +x the executable.

ankitkumarr:
@ahmelsayed, I will give it a shot today. Thanks

ankitkumarr:
Tried on a mac, noticed two minor things --
- If `-dir` is set to `.`, it emits the first two characters of every file name while zipping.
- Because we create the zipfile, before indexing the `-dir`, if the `-output` points to the same directory as `-dir`, the zip file will include a corrupted reference to itself.

These are not super big deals. But other than that, all looks good so far.

ankitkumarr:
@ahmelsayed, if you want to merge this PR. I can then update the installers, and migrate our packaging to use go-zip. Guessing that shouldn't block this PR.

PR #1550, opened 2019-09-16
Add DotNetZip.dll in the list to be signed

ankitkumarr:
I am still not 100% sure why `sigcheck.exe` test did not report this to us. When I ran it locally, the sign build failed as expected. Not sure why it went through in our pipeline.

I will look more into this later. But did not want to block on unsigned dlls.


ankitkumarr:
@soninaren changed the base branch to master. Would you mind reviewing again?

PR #1568, opened 2019-09-23
initial .NET Core 3 checkin

brettsam:


azure-pipelines.yml
```diff
@@ -23,6 +25,11 @@ variables:
 steps:
 - pwsh: |
     Write-Host "Target branch: '$(APPVEYOR_REPO_BRANCH)'"
+- task: UseDotNet@2
+  inputs:
+    packageType: 'sdk'
+    version: '3.0.100-rc1-014190'
```
timheuer: Shouldn't this be the release bits now and not the rc1?

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
ahmelsayed: Let's also set 

```xml
<PublishReadyToRun>true</PublishReadyToRun>
<PublishSingleFile>true</PublishSingleFile>
```

ahmelsayed:
looks like there might be few issues with these settings. `PublishReadyToRun` requires the build platform to match the target platform (we only build on Windows. Will need linux and mac CI machines), `PublishSingleFile` creates 1 ~600MB file => 240 zipped (at least for `linux-x64` runtime). I don't think it's trimming things correctly because I think our own trimming logic results in smaller (370MB => 120MB zipped) output. 


For `PublishReadyToRun` is nice to have, but we will need to update the CI flow to use 3 different build machines.

ahmelsayed:
Thank you @brettsam for sorting the readytorun stuff out. There are definitely some rough edges there :)

brettsam:
Agree we need to investigate whether PublishSingleFile is worth it. It'll need some extra work to incorporate some of the trimming/copying logic that currently happens in the C# build script into MSBuild.

I just went through this build of the CLI and did a smoke test with each language and it looks like it's all working.

PR #1586, opened 2019-10-08
Remove preview flag from PowerShell

Francisco-Gamino:
Remove preview flag from PowerShell.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
@@ -78,9 +78,6 @@ namespace Azure.Functions.Cli.Helpers
             {
                 switch (wr)
                 {
-                    case WorkerRuntime.powershell:
-                        workerToDisplayStrings[wr] = "powershell (preview)";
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -78,9 +78,6 @@ public static class WorkerRuntimeLanguageHelper
             {
                 switch (wr)
                 {
-                    case WorkerRuntime.powershell:
-                        workerToDisplayStrings[wr] = "powershell (preview)";
-                        break;
```
ahmelsayed: I think we can get rid of the method, right?

Francisco-Gamino: Hi @ahmelsayed, my original thought was to keep it as is, so next time we add a new preview language, we can just add a few lines and we are ready to go. However, you are right, we can remove this method and simplify the logic in ```InitAction.cs```.

Francisco-Gamino:
Thanks for the review and merging the code @ahmelsayed.

PR #1588, opened 2019-10-09
Leverage Docker build cache

arijitdas:
If we are adding any additional dependencies in `requirements.txt` for `pip install`, it takes a long time to install them again and again for any minor code change. Moving the steps to copy `requirements.txt` then `pip install` before copying the code takes advantage of the docker cache and drastically reduces build time.

Same would be applicable for other languages as well, e.g. node.


ahmelsayed:
Sorry I forgot to mention it. Thank you very much @arijitdas for your contribution!

arijitdas:
Happy to be of any help! 

Just a note, similar changes in Dockerfile.node file for npm install step will help the node build time as well

ahmelsayed:
will update the node file too. Thanks for the heads up

PR #1597, opened 2019-10-15
Refactor worker version extractor

Hazhzeng:
Update **PythonHelper** **ValidatePythonVersion** to extract python version info from a local devbox in a more organized way.

This also benefits my upcoming PR to generate different Dockerfile depends on different python versions.

Feedbacks are welcome.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -73,65 +73,64 @@ namespace Azure.Functions.Cli.Helpers
             }
         }
 
-        public static async Task<string> ValidatePythonVersion(bool setWorkerExecutable = false, bool errorIfNoExactMatch = false, bool errorOutIfOld = true)
+        public static async Task<WorkerLanguageVersionInfo> ValidatePythonVersion(
+            bool setWorkerExecutable = false,
+            bool errorIfNoExactMatch = false,
+            bool errorOutIfOld = true)
         {
             // If users are overriding this value, we don't have to worry about verification
-            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(_pythonDefaultExecutableVar)))
+            string pythonDefaultExecutablePath = Environment.GetEnvironmentVariable(_pythonDefaultExecutableVar);
+            if (!string.IsNullOrEmpty(pythonDefaultExecutablePath))
             {
-                return Environment.GetEnvironmentVariable(_pythonDefaultExecutableVar);
+                return await GetVersion(pythonDefaultExecutablePath);
             }
 
-            const string pythonStr = "python";
-            const string python3Str = "python3";
-            const string python36Str = "python3.6";
-            const string py36Str = "3.6";
-            const string py3Str = "3.";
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -73,65 +73,66 @@ private static void CreateRequirements()
...
-            var exeToVersion = new Dictionary<string, string>()
+            var workers = new List<WorkerLanguageVersionInfo>
             {
-                { pythonStr, pythonExeVersion },
-                { python3Str, python3ExeVersion },
-                { python36Str, python36ExeVersion }
+                await GetVersion("python"),
+                await GetVersion("python3"),
+                await GetVersion("python3.6"),
+                await GetVersion("python3.7")
```
ankitkumarr: You can make this parallel like how it was before.

Also, I am not sure if we want to continue with `python3.x` binary version checking.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -156,9 +157,16 @@ private static void SetWorkerPathIfNeeded(bool setWorker, string pyExe)
...
+        {
+            var pythonExeVersionTask = VerifyVersion(pythonExe);
+            var pythonExeVersion = await Utilities.SafeExecution(async () => await pythonExeVersionTask) ?? string.Empty;
+            return new WorkerLanguageVersionInfo(WorkerRuntime.python, pythonExeVersion, pythonExe);
+        }
+
         public static async Task<string> VerifyVersion(string pythonExe = "python")
         {
-            var exe = new Executable(pythonExe, "--version");
+            var exe = new Executable(pythonExe, "-c \"import sys; print(str(sys.version_info[0]) + \'.\' + str(sys.version_info[1]) + \'.\' + str(sys.version_info[2]))\"");
```
ankitkumarr: I would rather stay away from Python code to detect version. Also because syntax has changed over time. This may fail for older versions of Python.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -73,65 +73,64 @@ private static void CreateRequirements()
...
-            var pythonExeVersionTask = VerifyVersion(pythonStr);
-            var python3ExeVersionTask = VerifyVersion(python3Str);
-            var python36ExeVersionTask = VerifyVersion(python36Str);
-
-            var pythonExeVersion = await Utilities.SafeExecution(async () => await pythonExeVersionTask) ?? string.Empty;
-            var python3ExeVersion = await Utilities.SafeExecution(async () => await python3ExeVersionTask) ?? string.Empty;
-            var python36ExeVersion = await Utilities.SafeExecution(async () => await python36ExeVersionTask) ?? string.Empty;
-
-            var exeToVersion = new Dictionary<string, string>()
+            const string infoVersionSelectMessage = "We select Python interpreter '{0}' with version {1} for your project.";
```
ankitkumarr:
```suggestion
            const string infoVersionSelectMessage = "Found Python {1} ({0})";
```
You might also want to change the ordering (1, 0) in the above suggestion.

PR #1602, opened 2019-10-16
Generate python runtime dockerfile depends on local interpreter…

Hazhzeng:
Depends on the version the python user is using, we will decide to generate python3.6 or python3.7 docker file for their custom image.


ahmelsayed:
not exactly this PR, but related. We need to also conditionally select the image used with --build-native-deps

Hazhzeng:
@ahmelsayed add another PR for buildenv image change. https://github.com/Azure/azure-functions-core-tools/pulls
Please check when available.

PR #1606, opened 2019-10-17
Generate python buildenv dockerfile depends on local interpreter version

Hazhzeng:
We also need to change --build-native-deps image if python 3.7 interpreter is detected.


ahmelsayed:
It would be nice if we can get an e2e test for this, but it needs to check the python version on the ci machine

PR #1609, opened 2019-10-19
PowerShell function apps. Ignore prerelease versions of the Az module.

Francisco-Gamino:
The PowerShell gallery contains both prerelease and regular versions of the Az module. 
Updating logic in the ```PowerShellHelper.cs``` to skip prerelease versions.


ankitkumarr:
Looks like Durable tests are failing as a `DurableTask.AzureStorage.dll` is missing. Will take a look at what's going on. I don't think we changed any durable packages recently.

cc: @ahmelsayed

ankitkumarr:
I made a PR to see if we can fix that test quick https://github.com/Azure/azure-functions-core-tools/pull/1610

If it doesn't fix it, I think we'd have to ignore / temporarily remove those tests until we get them fixed.

PR #1611, opened 2019-10-21
Fix --build-native-deps validation

Hazhzeng:
Previously, due to the change on remote build default, a validation of not consistent build options will always block the --build-native-deps.

This issue is addressed in this PR. Now --no-build and --build-native-deps will have higher precedence over --build flag

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/PublishHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/PublishHelper.cs
@@ -24,8 +24,19 @@ namespace Azure.Functions.Cli.Helpers
             return null;
         }
 
-        public static BuildOption UpdateBuildOption(BuildOption currentBuildOption, WorkerRuntime runtime, Site site)
+        public static BuildOption ResolveBuildOption(BuildOption currentBuildOption, WorkerRuntime runtime, Site site, bool buildNativeDeps, bool noBuild)
         {
+            // --no-build and --build-native-deps will take precedence over --build local and --build remote
+            if (noBuild)
+            {
+                return BuildOption.None;
+            }
+
+            if (buildNativeDeps)
+            {
+                return BuildOption.Container;
+            }
+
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Helpers/PublishHelper.cs
```diff
@@ -24,8 +24,19 @@ public static GitIgnoreParser GetIgnoreParser(string workingDir)
             return null;
         }
 
-        public static BuildOption UpdateBuildOption(BuildOption currentBuildOption, WorkerRuntime runtime, Site site)
+        public static BuildOption UpdateBuildOption(BuildOption currentBuildOption, WorkerRuntime runtime, Site site, bool buildNativeDeps, bool noBuild)
```
ahmelsayed: I think `ResolveBuildOption` would be a more descriptive name for this method.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -276,11 +276,6 @@ internal static async Task<Stream> GetPythonDeploymentPackage(IEnumerable<string
             FileSystemHelpers.EnsureDirectory(packagesLocation);
 
             // Only one of the remote build or build-native-deps flag can be chosen
-            if (buildNativeDeps && buildOption == BuildOption.Remote)
-            {
-                throw new CliException("Cannot perform '--build-native-deps' along with '--build remote'");
```
ankitkumarr: I am curious, why remove this?

Hazhzeng: Good catch.

Hazhzeng: Actually, this condition is not reachable since the `buildNativeDeps` flag will switch `buildOption` to `BuildOption.Container`. Basically, this is a no ops.

PR #1614, opened 2019-10-22
Allow Python 3.7 to run locally

### LINKED ISSUE #1534: Allow Python 3.7 to perform func host start
Currently we block users with Python 3.7 to run `func host start`. This is because the Python worker binaries that comes with core-tools is compatible with Python 3.6 only.

But, we have a work item to add Python 3.7 worker binaries with core-tools. As soon as that's in, we should not block users to perform `func host start` with Python 3.7.

ankitkumarr:
Will make another PR to check and warn user when publishing environment does not match local environment (in terms of language version). 

Need to test it before merge

Closes #1534

### DIFF
--- a/src/Azure.Functions.Cli/Common/Constants.cs
+++ b/src/Azure.Functions.Cli/Common/Constants.cs
@@ -17,6 +17,7 @@ namespace Azure.Functions.Cli.Common
         public const string DotNetClrProcessId = @"${command:pickProcess}";
         public const string FuncIgnoreFile = ".funcignore";
         public const string FunctionsWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
+        public const string FunctionsWorkerRuntimeVersion = "FUNCTIONS_WORKER_RUNTIME_VERSION";
         public const string RequirementsTxt = "requirements.txt";
         public const string FunctionJsonFileName = "function.json";
         public const string ExtenstionsCsProjFile = "extensions.csproj";
@@ -46,7 +47,7 @@ namespace Azure.Functions.Cli.Common
         {
             { WorkerRuntime.dotnet, new [] { "mcr.microsoft.com/azure-functions/dotnet", "microsoft/azure-functions-dotnet-core2.0", "mcr.microsoft.com/azure-functions/base", "microsoft/azure-functions-base" } },
             { WorkerRuntime.node, new [] { "mcr.microsoft.com/azure-functions/node", "microsoft/azure-functions-node8" } },
-            { WorkerRuntime.python, new [] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6"  } },
+            { WorkerRuntime.python, new [] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6" } },
             { WorkerRuntime.powershell, new [] { "mcr.microsoft.com/azure-functions/powershell", "microsoft/azure-functions-powershell" } }
         };
 
--- a/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
@@ -17,10 +17,13 @@ namespace Azure.Functions.Cli.Helpers
         private static readonly string _pythonDefaultExecutableVar = "languageWorkers:python:defaultExecutablePath";
         private static bool InVirtualEnvironment => !string.IsNullOrEmpty(VirtualEnvironmentPath);
         public static string VirtualEnvironmentPath => Environment.GetEnvironmentVariable("VIRTUAL_ENV");
+        private static WorkerLanguageVersionInfo _pythonVersionCache = null;
 
         public static async Task SetupPythonProject()
         {
-            await ValidatePythonVersion(errorOutIfOld: false);
+            var pyVersion = await GetEnvironmentPythonVersion();
+            AssertPythonVersion(pyVersion, errorIfNoVersion: false);
+
             CreateRequirements();
             await EnsureVirtualEnvrionmentIgnored();
         }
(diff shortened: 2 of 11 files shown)


ankitkumarr:
@soninaren, I had to update some references for `HostJsonFileConfigurationSource` as looks like `IMetricsLogger` was added as a dependency.

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -46,7 +47,7 @@ internal static class Constants
         {
             { WorkerRuntime.dotnet, new [] { "mcr.microsoft.com/azure-functions/dotnet", "microsoft/azure-functions-dotnet-core2.0", "mcr.microsoft.com/azure-functions/base", "microsoft/azure-functions-base" } },
             { WorkerRuntime.node, new [] { "mcr.microsoft.com/azure-functions/node", "microsoft/azure-functions-node8" } },
-            { WorkerRuntime.python, new [] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6"  } },
+            { WorkerRuntime.python, new [] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6", "microsoft/azure-functions-python3.7" } },
```
ahmelsayed: Is that an actual image?

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -167,7 +167,18 @@ private static async Task<WorkerLanguageVersionInfo> GetVersion(string pythonExe
             return null;
         }
 
-        public static async Task<string> VerifyVersion(string pythonExe = "python")
+        public static void SetWorkerRuntimeVersionPython(WorkerLanguageVersionInfo version)
+        {
+            if (version == null)
+            {
+                return;
```
Hazhzeng: Do we want to throw an exception when worker version cannot be resolved? This is an issue if the customer is not configuring their PATH properly.

src/Azure.Functions.Cli/Common/Constants.cs
ankitkumarr: Ahh, I confused this with `mcr.microsoft.com/azure-functions/python:2.0-python3.7-appservice`.
Will remove.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
ankitkumarr: Technically this will never happen, as `ValidatePythonVersion` will throw an exception before it gets here. But I will add ArgumentNullException as a safeguard. Thanks

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -110,7 +110,7 @@ await python37GetVersionTask
             if (python36_37worker != null)
             {
                 SetWorkerPathIfNeeded(setWorkerExecutable, python36_37worker.ExecutablePath);
-                ColoredConsole.WriteLine(AdditionalInfoColor(string.Format(infoVersionSelectMessage, python36_37worker.ExecutablePath, python36_37worker.Version)));
+                ColoredConsole.WriteLine(AdditionalInfoColor(string.Format(infoVersionSelectMessage, python36_37worker.Version, python36_37worker.ExecutablePath)));
```
Hazhzeng: Sorry, I should've test the docker init scenario in the last version.

This line is printing out twice when doing docker init or build-native-deps.
I think it we should remove this line and would be sufficient to only print the python version on **func host start** [PythonHelpers.cs:177](https://github.com/Azure/azure-functions-core-tools/pull/1614/files#diff-88f494e2db490871070e912928cacfbcR177)

```
Found Python version 3.6.8 (python).
Writing .gitignore
Writing host.json
Writing local.settings.json
Writing C:\Users\hazeng\Projects\Functions\Python36Docker\.vscode\extensions.json
Found Python version 3.6.8 (python).
Writing Dockerfile
Writing .dockerignore
```

ankitkumarr: This function also has other print statements that may be printed twice during above.

I think we should avoid calling this function twice. Will take a look

ahmelsayed: It seems to me that this method is taking on too many roles. 
It's resolving the correct `WorkerLanguageVersionInfo` for the current environment, setting an environment variable that's needed for the worker, communicate the selection to the user, and throwing if the state is invalid based on few different options.

That could be broken down into multiple different methods that each take one role, and we can leave the orchasteration of these methods to the upstream caller. 

Something like `ResolveWorkerLanguageVersionInfo()` that just returns the right `WorkerLanguageVersionInfo`. An `AssertPythonVersion()` that will throw if the state is invalid. Maybe a `SetWorkerEnvironmentVariable` or `EnsureWorkerEnvironment()`, that initialized the environment, and the confirmation of the selection could be explicit by code paths that need it.

ankitkumarr: Thanks @ahmelsayed, that does sounds like a cleaner approach. I will update the PR.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -73,85 +76,100 @@ private static void CreateRequirements()
...
             {
                 Environment.SetEnvironmentVariable(_pythonDefaultExecutableVar, pyExe, EnvironmentVariableTarget.Process);
-                if (StaticSettings.IsDebug)
-                {
-                    ColoredConsole.WriteLine(VerboseColor($"{_pythonDefaultExecutableVar} set to {pyExe}"));
-                }
+            }
+            if (StaticSettings.IsDebug)
+            {
+                ColoredConsole.WriteLine(VerboseColor($"{_pythonDefaultExecutableVar} set to {pyExe}"));
```
Hazhzeng: I think when overwrite is set to False, and _pythonDefaultExecutableVar has value, the debug log will not reflects the actual _pythonDefaultExecutableVar value since pyExe is not set to the worker path.

ankitkumarr: Right, I will change the wording to be `$"{_pythonDefaultExecutableVar} is set to {pyExe}"`
Still good to show in debug what it's actually set to.

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -73,85 +76,100 @@ private static void CreateRequirements()
...
-            // If users are overriding this value, we don't have to worry about verification
+            if (pythonVersion?.Version == null)
+            {
+                var message = "Could not find a Python version. Python 3.6.x or 3.7.x is recommended, and used in Azure Functions.";
+                if (errorIfNoVersion) throw new CliException(message);
+                ColoredConsole.WriteLine(WarningColor(message));
+                return;
+            }
+
+            ColoredConsole.WriteLine(AdditionalInfoColor($"Found Python version {pythonVersion.Version} ({pythonVersion.ExecutablePath})."));
```
Hazhzeng: I think the proper place to report the python version is in **GetEnvironmentPythonVersion**.
Since we had a cache there, we only need to report the python version when we doing the first detection.
Any thoughts?

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
Hazhzeng: Then I think the second value needs to change to **pyExe** -> **Environment.GetEnvironmentVariable(_pythonDefaultExecutableVar)**.

ankitkumarr: Right right. duh!

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
ankitkumarr: That's fair. But I think when asserting is really when it matters what version users have. So, say if you don't have 3.7, the output would be --

```
Found Python version {pythonVersion.Version} ({pythonVersion.ExecutablePath}).
Python 3.6.x or 3.7.x is required for this operation.
Please install Python 3.6 or 3.7, and use a virtual environment to switch to Python 3.6 or 3.7.
```

Hazhzeng: Hmm, you made a good point. Yea, let's keep this line here.

PR #1626, opened 2019-10-25
Updating WebHost to 2.0.12850

yojagad:


ankitkumarr:
@Francisco-Gamino  @AnatoliB, the Powershell worker is failing to initialize in the CI test with the new Host. Did anything change that you might be aware of?

AnatoliB:
@ankitkumarr 
> It could be CI missing dotnet but seems unlikely as only the builds with the new Host are failing.

One important thing that changed in the host is how it spawns the dotnet process for the PowerShell worker. Previously, it would just start `dotnet` without specifying any path, so the OS would look it up according to its rules (e.g. searching the PATH env variable). After the change, the host specifies the full path to the dotnet executable explicitly: %ProgramFiles%\dotnet\dotnet.exe (on Windows). If the CI environment does not have dotnet installed at this location, this explains the failure.

AnatoliB:
This should resolve the issue: https://github.com/Azure/azure-functions-host/pull/5157

PR #1632, opened 2019-10-28
Error out if remote build runs on a broken SCM site

### LINKED ISSUE #1622: Remote build should check for scm hostname before removing Run from pkg app setting
Remote build on old linux consumption apps can leave the app broken. we remove the appsetting and later if the app doesn't support remote build we don't add it back.

Move the check to start of the command.

func azure functionapp publish <APP> --force

**Removing WEBSITE_RUN_FROM_PACKAGE app setting** 

Remote build is a new feature added to function apps.

Your function app mediareport does not support remote build as it was created before August 1st, 2019.

Please use '--build local', '--build-native-deps' or manually enable remote build.

For more information, please visit https://aka.ms/remotebuild

Hazhzeng:
### Background
Previously, we allow customer to update their function app with Azure CLI to create a new SCM site. Unfortunately, the SCM site lacks the SCM_RUN_FROM_PACKAGE setting, and thus, the remote build fails to update the artifact to a valid blob url.

This issue affects Linux Consumption function app created before 8/1/2019

### Mitigation
@balag0 suggested a mitigation for current broken SCM sites.
Remove and reset the app setting for AzureWebJobsStorage. However, this method will take the function app offline and it is not recoverable if the core tool fails in the middle of a remote build deployment.

### Proper Solution
Bala may introduce a solution to refresh the SCM_RUN_FROM_PACKAGE setting when the user updates any of the app setting. Before that, we should discourage our user from manually enabling the SCM site.

### Miscellaneous
resolves #1622

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -452,14 +452,11 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         private async Task<bool> HandleLinuxConsumptionPublish(Site functionApp, Func<Task<Stream>> zipFileFactory)
         {
             string fileNameNoExtension = string.Format("{0}-{1}", DateTimeOffset.UtcNow.ToString("yyyyMMddHHmmss"), Guid.NewGuid());
-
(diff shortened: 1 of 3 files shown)


ankitkumarr:
@Hazhzeng, do you need this change out in this week's release (before Ignite)? The release is about to start. Lmk if you do asap.

Issue #1574, opened 2019-10-01
Telemetry - do not collect if running as part of VS


ankitkumarr:
For VS extension, we package a specific version of core-tools `win.min`. In our build pipeline, we could drop a sentinel in that package such that we don't collect telemetry if that sentinel exists. Unless, VS extension specifically asks us to opt-in and collect usage.

ankitkumarr:
This will likely be moved to the next Sprint as other items with higher priority are rolling in.

ankitkumarr:
PRs:
- https://github.com/ahmelsayed/publish-script/pull/1
- https://github.com/Azure/azure-functions-core-tools/pull/1636
- https://github.com/Azure/homebrew-functions/pull/15
- https://github.com/Azure/azure-functions-core-tools/pull/1555

Once all above are merged, this can be closed.

fabiocav:
Closing as resolved.

PR #1636, opened 2019-10-30
Add telemetry to be enabled via file

ankitkumarr:
Part of work for https://github.com/Azure/azure-functions-core-tools/issues/1574

Pending other parts once this is in.

### DIFF
--- a/src/Azure.Functions.Cli/Common/StaticSettings.cs
+++ b/src/Azure.Functions.Cli/Common/StaticSettings.cs
@@ -1,15 +1,20 @@
-﻿using Azure.Functions.Cli.Helpers;
-using System;
+﻿using System;
+using Azure.Functions.Cli.Helpers;
 
 namespace Azure.Functions.Cli.Common
 {
     public static class StaticSettings
     {
+        private static Lazy<bool> _isTelemetryEnabledCache = new Lazy<bool>(() => TelemetryHelpers.CheckIfTelemetryEnabled());
+
         public static bool IsDebug => Environment.GetEnvironmentVariable(Constants.CliDebug) == "1";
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Common/StaticSettings.cs
```diff
@@ -1,15 +1,24 @@
...
-        // TODO: Enable this once it's ready to be released.
-        public static bool IsTelemetryEnabled => false && (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.TelemetryOptOutVariable))
-                && Constants.TelemetryInstrumentationKey != "00000000-0000-0000-0000-000000000000");
+        public static bool IsTelemetryEnabled
+        {
+            get
+            {
+                bool telemetryVal = _isTelemetryEnabledCache ?? TelemetryHelpers.CheckIfTelemetryEnabled();
+                _isTelemetryEnabledCache = telemetryVal;
+                return telemetryVal;
```
ahmelsayed: This pattern in .NET is encapsulated in [`Lazy<T>`](https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization). This is not relevant for the cli since it's mostly single threaded, but this isn't thread safe either. You'd need to use a lock, but `Lazy<T>` gives you for free.

ankitkumarr: Ooooh! Didn't know Lazy was thread safe. I guess that makes sense. Cool! I will update it. Figured if we are doing this check multiple times in one execution, we probably don't need to do this file check everytime.
I will update. Thanks!

PR #1648, opened 2019-11-05
Update ARM Api

ahmelsayed:


src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -147,12 +141,12 @@ internal static async Task<string> GetFunctionKey(string functionName, string ap
...
             }
-            var url = new Uri($"{managementURL}{appId}/hostruntime/admin/functions/{functionName}/keys?api-version={ArmUriTemplates.WebsitesApiVersion}");
+            var url = new Uri($"{managementURL}{appId}/functions/{functionName}/listKeys?api-version={ArmUriTemplates.WebsitesApiVersion}");
             var key = string.Empty;
             try
             {
-                var keysJson = await ArmHttpAsync<JToken>(HttpMethod.Get, url, accessToken);
-                key = (string)(keysJson["keys"] as JArray).First()["value"];
+                var keysJson = await ArmHttpAsync<JToken>(HttpMethod.Post, url, accessToken);
+                key = keysJson["default"].ToString();
```
Hazhzeng: Just to remind the **default** key can be removed from a function.
I think it would be better if we can cast keysJson to JArray and select the FirstOrDefault() key.

ahmelsayed: Good catch. Thanks Roger.

PR #1651, opened 2019-11-05
[E2E Test] Add features to deploy Azure resources for testing

Hazhzeng:
### Background
The deployment helper for Azure Functions Core Tools does not contain tests so far.
This due to a lack of tool-set to change Azure Resource via ARM 

### azure-pipelines.yml
1. Create a test resource group in an Azure Testing Subscription.
2. Generate a service principle to the test resource group.
3. In Azure DevOps pipelines, use **AzureCLI@2** and `az account get-access-token --query "accessToken"` to generate an access token in every build.
4. Pass the access token to **AZCLI_ACCESS_TOKEN**
5. Parse the AZCLI_ACCESS_TOKEN from core tool test helper
6. Use xunit test framework to perform E2E test

### AzureResourceManagers
- **Common**: contains data models and enums for ARM requests and responses
- **BaseAzureResourceManager**: used for providing Bearer AccessToken, Subscription ID, Resource Group name
- **MultiOsResourcesManager**: used for distinguishing Windows / Linux resources in different resource groups
- **FunctionAppManager / ServerFarmManager / StorageAccountManager**: used for creating/querying/deleting function app
- **DeploymentTests**: Basic remote build for Python consumption function app. (will add more tests in separate PRs)

### Pipeline Variables
- **ENABLE_DEPLOYMENT_TEST**: setting this to **true** to enable deployment tests. Set to **false** to opt out deployment tests.
- **TEST_SUBSCRIPTION_ID**: the subscription for creating resources
- **TEST_RESOURCE_GROUP_NAME_WINDOWS**: the name of resource group for creating windows resources
- **TEST_RESOURCE_GROUP_NAME_LINUX**: the name of resource group for creating linux resources

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,12 +23,22 @@ variables:
 steps:
 - pwsh: |
     Write-Host "Target branch: '$(APPVEYOR_REPO_BRANCH)'"
-- task: NodeTool@0 
+- task: NodeTool@0
   inputs:
     versionSpec: '10.x'
 - task: NuGetToolInstaller@1
   inputs:
-    versionSpec: 
+    versionSpec:
+- task: AzureCLI@2
+  displayName: Login via Azure CLI to acquire access token
+  inputs:
+    azureSubscription: $(E2ETestServiceConnectionName)
+    scriptType: ps
+    scriptLocation: inlineScript
+    inlineScript: |
+      # acquire access token from Azure CLI and export it to AZURE_MANAGEMENT_ACCESS_TOKEN
+      $accessToken = (az account get-access-token --query "accessToken" | % { $_.Trim('"') })
+      echo "##vso[task.setvariable variable=azure_management_access_token]$accessToken"
 - pwsh: |
     .\build.ps1
   env:
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -12,6 +12,7 @@ using Fclp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Colors.Net.StringStaticMethods;
+using Azure.Functions.Cli.Helpers;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
@@ -29,7 +30,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         // Windows PowerShell is PowerShell version 5.1 and lower that only works on Windows
         private const string _windowsPowerShellExecutable = "powershell";
 
-        private const string _defaultManagementURL = "https://management.azure.com/";
+        private const string _defaultManagementURL = Constants.DefaultManagementURL;
 
         public string AccessToken { get; set; }
         public bool ReadStdin { get; set; }
@@ -131,7 +132,18 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             (bool powershellSucceeded, string psToken) = await TryGetAzPowerShellToken();
             if (powershellSucceeded) return psToken;
 
+            if (TryGetTokenFromTestEnvironment(out string envToken))
+            {
+                return envToken;
+            }
+
             throw new CliException($"Unable to connect to Azure. Make sure you have the `az` CLI or `{_azProfileModuleName}` PowerShell module installed and logged in and try again");
+        }
+
+        private bool TryGetTokenFromTestEnvironment(out string token)
+        {
+            token = Environment.GetEnvironmentVariable(Constants.AzureManagementAccessToken);
+            return !string.IsNullOrEmpty(token);
         }
 
         private async Task<(bool succeeded, string token)> TryGetAzCliToken()
--- a/src/Azure.Functions.Cli/Common/Constants.cs
+++ b/src/Azure.Functions.Cli/Common/Constants.cs
@@ -41,6 +41,8 @@ namespace Azure.Functions.Cli.Common
         public const string WebsiteContentAzureFileConnectionString = "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING";
         public const string WebsiteContentShared = "WEBSITE_CONTENTSHARE";
         public const string TelemetrySentinelFile = "telemetryDefaultOn.sentinel";
+        public const string DefaultManagementURL = "https://management.azure.com/";
+        public const string AzureManagementAccessToken = "AZURE_MANAGEMENT_ACCESS_TOKEN";
 
         public static string CliVersion => typeof(Constants).GetTypeInfo().Assembly.GetName().Version.ToString(3);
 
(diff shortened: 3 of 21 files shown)


Hazhzeng:
There's a security concern if we allow the DevOps pipeline access the test resource group.
Malicious users may be able to create their own resources in our subscription.
We can follow the azure-functions-host approach to label the PR for full-ci.

azure-pipelines.yml
```diff
@@ -25,7 +25,16 @@ steps:
...
-    versionSpec: 
+    versionSpec:
+- task: AzureCLI@2
+  displayName: Login via Azure CLI to acquire access token
+  inputs:
+    azureSubscription: $(AzureSubscriptionName)
+    scriptType: ps
+    scriptLocation: inlineScript
+    inlineScript: |
+      $accessToken = (az account get-access-token --query "accessToken" | % { $_.Trim('"') })
```
ahmelsayed: Do you not need to login first?

ahmelsayed:
@Hazhzeng we shouldn't be running tests on external contributions automatically. I think this is the case already. But you can verify with Naren.

azure-pipelines.yml
Hazhzeng: Good catch, but I think since the AzureCLI task depends on the service connection in Azure DevOps, and the example does not have `az login` command, I think it will login automatically.

Let me check with my private DevOps pipeline tomorrow.

Hazhzeng: @ahmelsayed since the **subscriptionName** parameter accepts a specific service principal. I can confrim that the `az login` is automatically generated. https://dev.azure.com/hazeng0405/guessme/_build/results?buildId=7

ahmelsayed: Got it. Shouldn't you be able to reuse the code here then https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
to get a token?

Hazhzeng: It is a bit unfortunate that the `az` executable will no longer be available outside AzCLI task. So we may still need to export the AccessToken https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=2785&view=logs&jobId=12f1170f-54f2-53f3-20dd-22fc7dff55f9&taskId=036ed663-514a-59f3-0c88-23d2cf182147&lineStart=120&lineEnd=120&colStart=81&colEnd=96

ahmelsayed: I see. that makes sense then. Maybe add that as a comment so we remember that. Thanks Roger!

azure-pipelines.yml
```diff
@@ -22,10 +22,20 @@ variables:
 
 steps:
 - pwsh: |
-    Write-Host "Target branch: '$(APPVEYOR_REPO_BRANCH)'"
+    Write-Host Target branch: '$(APPVEYOR_REPO_BRANCH)'
```
Hazhzeng: I remove these double quotes to restore the syntax highlighting in vscode

Hazhzeng:
Hi @ahmelsayed, I think the e2e test facility is ready for review. Please take a look and leave some feedbacks when available. Much appreciated.

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -134,6 +141,16 @@ private async Task<string> GetAccessToken()
             throw new CliException($"Unable to connect to Azure. Make sure you have the `az` CLI or `{_azProfileModuleName}` PowerShell module installed and logged in and try again");
         }
 
+        private (bool succeeded, string token) TryGetTokenFromTestEnvironment()
```
ahmelsayed: pretty minor nit: but unless your `Try...` method is async, you can use out parameters instead of a tuple of `(success, result)`. It's a bit more idiomatic in C#. 

```suggestion
        private bool TryGetTokenFromTestEnvironment(out string token)
```

then the usage becomes a bit clearer

```csharp
if (TryGetTokenFromTestEnvironment(out string token))
{
    return token;
}
```

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -99,6 +100,16 @@ public static class KuduLiteDeploymentConstants
             };
         }
 
+        public static class E2ETestConstants
```
ahmelsayed: I'd move this to the test project. I know it's used in a couple of places in this project, but I think we can change that usage to avoid having this here.

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -125,6 +126,12 @@ private async Task<(bool, string)> TryGetAzCLIManagementURL()
 
         private async Task<string> GetAccessToken()
         {
+            if (EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.E2ETestConstants.EnableDeploymentTests))
```
ahmelsayed: Can this move below the `TryGetAzCliToken` and `TryGetAzPowerShellToken`, so you're sure you're not altering existing behavior, then come up with some descriptive env var name, like `AZURE_MANAGEMENT_ACCESS_TOKEN` and we can just use that as another source of the access token.

test/Azure.Functions.Cli.Tests/E2E/DeploymentTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        private readonly static string linuxStorageAccountName = $"lcte2estorage{id}";
+        private readonly static string linuxConsumptionServerFarm = $"lcte2ecserverfarm{id}";
+        private readonly static string linuxConsumptionPythonFunctionApp = $"lcte2ecpython{id}";
+
+        public DeploymentTests(ITestOutputHelper output, StorageAccountManager saManager, ServerFarmManager sfManager, FunctionAppManager faManager) : base(output)
+        {
+            _storageAccountManager = saManager;
+            _serverFarmManager = sfManager;
+            _functionAppManager = faManager;
+            Environment.SetEnvironmentVariable("FUNC_PATH", @"C:\Users\hazeng\AppData\Roaming\npm\node_modules\azure-functions-core-tools\bin\func.exe");
```
ahmelsayed: don't forget to remove this :)

test/Azure.Functions.Cli.Tests/E2E/AzureResourceManagers/BaseAzureResourceManager.cs
```diff
@@ -0,0 +1,96 @@
+﻿using Azure.Functions.Cli.Common;
+using System;
+
+namespace Azure.Functions.Cli.Tests.E2E.AzureResourceManagers
+{
+    public class BaseAzureResourceManager: IDisposable
```
ankitkumarr: You can probably mark this and `MultiOsResourcesManager` class as `abstract`. Or remove the `throw new NotImplementedException();` in `Cleanup` here. I say this as creating an instance of this and disposing would throw an exception.
And it seems like you'd not want to instantiate this directly, so maybe make this abstract?

PR #1652, opened 2019-11-05
Update sync triggers api

ahmelsayed:


ahmelsayed:
yes it is, and yes we should :)

PR #1668, opened 2019-11-19
Add extension bundle config by default

soninaren:


azure-pipelines.yml
```diff
@@ -31,11 +31,18 @@ steps:
     versionSpec: 
 - pwsh: |
     .\build.ps1
+  errorActionPreference: 'continue'
+  ignoreLASTEXITCODE: true
```
ankitkumarr: I don't think we should continue if the build failed. This will give incorrect impression for CI results.

soninaren: have removed this and have updated the next task to run even if this one failed

PR #1671, opened 2019-11-20
Remove logic to skip adding extension bundles,  download latest version of extensionbundle if available

### LINKED ISSUE #1664: Remove the logic to conditionally add extension bundle configuration
Currently we only apply extension bundle configuration if the functions you are using have reference extensions other than http or timer. We need to remove this since the runtime should now selectively load extensions based on what you are using.

### LINKED ISSUE #1665: Force fetch the latest version of extension bundle.
Check and fetch the latest version of extension bundle, if you don't have it already.

soninaren:
closes #1665
closes #1664


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -4,9 +4,10 @@
 using System.IO;
 using System.Linq;
 using System.Net;
-using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
+using static Azure.Functions.Cli.Common.OutputTheme;
+using static Colors.Net.StringStaticMethods;
```
ankitkumarr: Curious, did you accidentally move this up?

soninaren: it was no in the right order. so fixed it.

soninaren: Well VS moved it up. :) when i selected sort

PR #1676, opened 2019-11-21
Add extension bundle config by default, publish test results

soninaren:
Add extension bundle config by default, publish test results


test/Azure.Functions.Cli.Tests/ExtensionsTests/InstallExtensionsTests.cs
```diff
@@ -17,7 +20,7 @@ public Task try_install_with_no_valid_trigger()
             return CliTester.Run(new RunConfiguration
             {
                 Commands = new[] {
-                    "init . --worker-runtime node",
+                    "init . --worker-runtime node --no-bundle true",
```
ankitkumarr: You should be able to do just `--no-bundle`, right? Instead of `--no-bundle true`?

soninaren: yes you are right.

PR #1686, opened 2019-12-02
-Update the zulu maven image and use v3 version

amamounelsayed:


ahmelsayed:
I think you want to change the base branch to 3.x. If we're not gonna have v2 runtime with Java on linux, then we can remove the one in the dev branch.

amamounelsayed:
Fair point, and just java.sh should be for v3.

PR #1558, opened 2019-09-19
Adding a cli flag to start with full exception stack traces

aaronpowell:
This addresses #1557 by adding a new CLI flag so you can do `func start --verbose-errors` and get full stack traces in the logging when an internal component fails.

Issue #1557, opened 2019-09-19
Exceptions within dependency components lack details


aaronpowell:
I'm trying to work through a problem I have with a TypeScript Functions app which results in an exception being thrown somewhere within the Functions runtime.

In the console all I get is the following error message:

```
[19/09/2019 5:32:02 AM] Executed 'Functions.newUser' (Failed, Id=4696838d-988a-4f7e-b3d7-135ec7405b82)
[19/09/2019 5:32:02 AM] System.Private.CoreLib: Exception while executing function: Functions.newUser. Microsoft.WindowsAzure.Storage: Not Found.
```

_Note: This bug isn't about the **cause** of that exception, it's about the information in the logs about that exception._

Since the exception details are truncated it's really difficult to work out what the failing component is, and in turn fix said failing component.

Investigation led me to find that [this line](https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs#L55) (and thus the call out to the WebJobs SDK) results in the summarised exception message.

It'd be preferable if there was an opt-in way to show the full stack trace for debugging so that you can get enough information to rectify the problem.

aaronpowell:
I've created PR #1558 which allows me to now get log messages like so:

```
[19/09/2019 5:44:27 AM] Executed 'Functions.newUser' (Failed, Id=938020f2-1785-45ce-9410-97c5929b8f52)                                                                                                                                                    
[19/09/2019 5:44:27 AM] Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.newUser ---> Microsoft.WindowsAzure.Storage.StorageException: Not Found                                                   
[19/09/2019 5:44:27 AM]    at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteAsyncInternal[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext, CancellationToken token)                                           
[...]

PR #1701, opened 2019-12-10
Fixing #1557 by checking for debug mode and unpacking the exception if it is debug mode

brettsam:
Rebasing #1558 for @aaronpowell to get this merged.


brettsam:
@ankitkumarr -- this is a re-do of the PR at #1558, which we already approved. Just rebased it so we can get it in today.

PR #1703, opened 2019-12-11
Use go zip to preseve file permissions

ankitkumarr:
Need to test it in all environments and make sure that permissions are actually preserved.
Also need to chmod +x gozip in all installers


build/Settings.cs
```diff
@@ -114,6 +114,7 @@ public class SignInfo
                 "DurableTask.Core.Internal.dll",
                 "func.dll",
                 "func.exe",
+                "gozip.exe",
```
ankitkumarr: Updated PR to add the binary to sign.

PR #1708, opened 2019-12-12
Fix signing build sigcheck test

ankitkumarr:


ankitkumarr:
Also need to merge this to `dev`

build/BuildSteps.cs
```diff
@@ -339,6 +339,12 @@ public static void TestSignedArtifacts()
                 client.DownloadFile(Settings.SignInfo.SigcheckDownloadURL, sigcheckPath);
             }
 
+            // https://peter.hahndorf.eu/blog/post/2010/03/07/WorkAroundSysinternalsLicensePopups
+            // Can't use sigcheck without signing the License Agreement
+            Console.WriteLine("Signing EULA");
+            Console.WriteLine(Shell.GetOutput("reg.exe", "ADD HKCU\\Software\\Sysinternals /v EulaAccepted /t REG_DWORD /d 1 /f"));
+            Console.WriteLine(Shell.GetOutput("reg.exe", "ADD HKU\\.DEFAULT\\Software\\Sysinternals /v EulaAccepted /t REG_DWORD /d 1 /f"));
```
ankitkumarr: I could just do `reg.exe ADD "HKCU\Software\Sysinternals\Sigcheck" /v EulaAccepted /t REG_DWORD /d 1 /f`

But probably fine to accept for all Sysinternals tools.

soninaren: Would prefer to have an alternate way of accepting EULA then reg.exe. We can use this for now to unblock ourselves.

ankitkumarr:
\azp run

PR #1715, opened 2019-12-13
Initial PR for the changes to manages function keys in Kubernetes

SatishRanjan:
The changes include creation of below three flags in core tools to manages function keys in Kubernetes

- funcappkeys-secret-name: Kubernetes secret name where default function keys (host, system and function keys) will be created and can be used from the function run time container using secretRef
- funcappkeys-config-map-name: The Kubernetes configMap name where default function keys (host, system and function keys) will be created and can be used from the function run time container using configMapRef
- mount-funcappkeys-as-containervolume: A flag indicating if function keys needs to be mounted in the function run-time container as volume mount at the path "/run/secrets/functions-keys" from either from Kubernetes secret or configMap

Also, service account, role, and role-binding is created and Pod identity of service account is assigned with the correct permissions to add/modify the Kubernetes secrets and config maps

### DIFF
--- a/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
+++ b/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
@@ -7,6 +7,7 @@ using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Helpers;
 using Azure.Functions.Cli.Interfaces;
 using Azure.Functions.Cli.Kubernetes;
+using Azure.Functions.Cli.Kubernetes.FuncKeys;
 using Azure.Functions.Cli.Kubernetes.Models;
 using Colors.Net;
 using Fclp;
@@ -25,11 +26,14 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
         public string Namespace { get; set; } = "default";
         public string PullSecret { get; set; } = string.Empty;
         public bool NoDocker { get; set; }
-        public bool UseConfigMap { get; set; }
+        public bool UseConfigMapForAppSettings { get; set; }
         public bool DryRun { get; private set; }
         public string ImageName { get; private set; }
-        public string ConfigMapName { get; private set; }
-        public string SecretsCollectionName { get; private set; }
+        public string AppSettingsConfigMapName { get; private set; }
+        public string AppSettingsSecretsCollectionName { get; private set; }
+        public string FunAppKeysConfigMapName { get; private set; }
+        public string FunAppKeysSecretsCollectionName { get; private set; }
+        public bool MountFuncAppKeysAsContainerVolume { get; private set; }
         public int? PollingInterval { get; private set; }
         public int? CooldownPeriod { get; private set; }
         public string ServiceType { get; set; } = "LoadBalancer";
@@ -58,8 +62,11 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
             SetFlag<int>("cooldown-period", "The cooldown period for the deployment before scaling back to 0 after all triggers are no longer active. Default: 300 (seconds)", p => CooldownPeriod = p);
             SetFlag<int>("min-replicas", "Minimum replica count", m => MinReplicaCount = m);
             SetFlag<int>("max-replicas", "Maximum replica count to scale to by HPA", m => MaxReplicaCount = m);
-            SetFlag<string>("secret-name", "The name of a kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => SecretsCollectionName = sn);
-            SetFlag<string>("config-map-name", "The name of a config map to use in the deployment", cm => ConfigMapName = cm);
+            SetFlag<string>("funcappkeys-secret-name", "The name of a kubernetes secret collection to use for the function app keys (host keys, function keys etc.)", ksn => FunAppKeysSecretsCollectionName = ksn);
+            SetFlag<string>("funcappkeys-config-map-name", "The name of a kubernetes config map to use for the function app keys (host keys, function keys etc.)", kcm => FunAppKeysConfigMapName = kcm);
+            SetFlag<bool>("mount-funcappkeys-as-containervolume", "The flag indicating to mount the func app keys as container volume", kmv => MountFuncAppKeysAsContainerVolume = kmv);
+            SetFlag<string>("appsettings-secret-name", "The name of a kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => AppSettingsSecretsCollectionName = sn);
+            SetFlag<string>("appsettings-config-map-name", "The name of a config map to use in the deployment", cm => AppSettingsConfigMapName = cm);
             SetFlag<string>("service-type", "Kubernetes Service Type. Default LoadBalancer  Valid options: " + string.Join(",", ServiceTypes), s =>
             {
                 if (!string.IsNullOrEmpty(s) && !ServiceTypes.Contains(s))
@@ -69,7 +76,7 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
                 ServiceType = s;
             });
             SetFlag<bool>("no-docker", "With --image-name, the core-tools will inspect the functions inside the image. This will require mounting the image filesystem. Passing --no-docker uses current directory for functions.", nd => NoDocker = nd);
-            SetFlag<bool>("use-config-map", "Use a ConfigMap/V1 instead of a Secret/V1 object.", c => UseConfigMap = c);
+            SetFlag<bool>("use-config-map-for-appsettings", "Use a ConfigMap/V1 instead of a Secret/V1 object for function app settings configurations", c => UseConfigMapForAppSettings = c);
             SetFlag<bool>("dry-run", "Show the deployment template", f => DryRun = f);
             SetFlag<bool>("ignore-errors", "Proceed with the deployment if a resource returns an error. Default: false", f => IgnoreErrors = f);
             return base.ParseArgs(args);
@@ -102,21 +109,27 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
                 triggers = await DockerHelpers.GetTriggersFromDockerImage(resolvedImageName);
             }
 
-            var resources = KubernetesHelper.GetFunctionsDeploymentResources(
+            var functionNames = triggers.FunctionsJson.Select(f => f.Key);
+            var funcAppKeys = FuncAppKeysHelper.CreateKeys(functionNames);
+            var resources = await KubernetesHelper.GetFunctionsDeploymentResources(
                 Name,
                 resolvedImageName,
                 Namespace,
                 triggers,
                 _secretsManager.GetSecrets(),
+                funcAppKeys,
                 PullSecret,
-                SecretsCollectionName,
-                ConfigMapName,
-                UseConfigMap,
+                AppSettingsSecretsCollectionName,
+                AppSettingsConfigMapName,
+                UseConfigMapForAppSettings,
                 PollingInterval,
                 CooldownPeriod,
                 ServiceType,
                 MinReplicaCount,
-                MaxReplicaCount);
+                MaxReplicaCount,
+                FunAppKeysSecretsCollectionName,
+                FunAppKeysConfigMapName,
+                MountFuncAppKeysAsContainerVolume);
 
             if (DryRun)
             {
--- a/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
+++ b/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
@@ -65,8 +65,8 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
             SetFlag<string>("funcappkeys-secret-name", "The name of a kubernetes secret collection to use for the function app keys (host keys, function keys etc.)", ksn => FunAppKeysSecretsCollectionName = ksn);
             SetFlag<string>("funcappkeys-config-map-name", "The name of a kubernetes config map to use for the function app keys (host keys, function keys etc.)", kcm => FunAppKeysConfigMapName = kcm);
             SetFlag<bool>("mount-funcappkeys-as-containervolume", "The flag indicating to mount the func app keys as container volume", kmv => MountFuncAppKeysAsContainerVolume = kmv);
-            SetFlag<string>("appsettings-secret-name", "The name of a kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => AppSettingsSecretsCollectionName = sn);
-            SetFlag<string>("appsettings-config-map-name", "The name of a config map to use in the deployment", cm => AppSettingsConfigMapName = cm);
+            SetFlag<string>("appsettings-secret-name", "The name of a existing func app settings kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => AppSettingsSecretsCollectionName = sn);
+            SetFlag<string>("appsettings-config-map-name", "The name of an existing config map with func app settings to use in the deployment", cm => AppSettingsConfigMapName = cm);
             SetFlag<string>("service-type", "Kubernetes Service Type. Default LoadBalancer  Valid options: " + string.Join(",", ServiceTypes), s =>
             {
                 if (!string.IsNullOrEmpty(s) && !ServiceTypes.Contains(s))
--- a/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
+++ b/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Helpers;
@@ -9,8 +10,10 @@ using Azure.Functions.Cli.Interfaces;
 using Azure.Functions.Cli.Kubernetes;
 using Azure.Functions.Cli.Kubernetes.FuncKeys;
 using Azure.Functions.Cli.Kubernetes.Models;
+using Azure.Functions.Cli.Kubernetes.Models.Kubernetes;
 using Colors.Net;
 using Fclp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -61,7 +64,7 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
             SetFlag<int>("cooldown-period", "The cooldown period for the deployment before scaling back to 0 after all triggers are no longer active. Default: 300 (seconds)", p => CooldownPeriod = p);
             SetFlag<int>("min-replicas", "Minimum replica count", m => MinReplicaCount = m);
             SetFlag<int>("max-replicas", "Maximum replica count to scale to by HPA", m => MaxReplicaCount = m);
-            SetFlag<string>("keys-secret-name", "The name of a kubernetes secret collection to use for the function app keys (host keys, function keys etc.)", ksn => KeysSecretCollectionName = ksn);    
+            SetFlag<string>("keys-secret-name", "The name of a kubernetes secret collection to use for the function app keys (host keys, function keys etc.)", ksn => KeysSecretCollectionName = ksn);
             SetFlag<bool>("mount-funckeys-as-containervolume", "The flag indicating to mount the func app keys as container volume", kmv => MountFuncKeysAsContainerVolume = kmv);
             SetFlag<string>("secret-name", "The name of an existing kubernetes secret collection, containing func app settings, to use in the deployment instead of creating new a new one based upon local.settings.json", sn => SecretsCollectionName = sn);
             SetFlag<string>("config-map-name", "The name of an existing config map with func app settings to use in the deployment", cm => ConfigMapName = cm);
@@ -106,8 +109,8 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
                 }
                 triggers = await DockerHelpers.GetTriggersFromDockerImage(resolvedImageName);
             }
-           
-            var resources = await KubernetesHelper.GetFunctionsDeploymentResources(
+
+            (var resources, var existingKeysSecret, var newKeysSecret) = await KubernetesHelper.GetFunctionsDeploymentResources(
                 Name,
                 resolvedImageName,
                 Namespace,
@@ -145,6 +148,9 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
                 {
                     await KubectlHelper.KubectlApply(resource, showOutput: true, ignoreError: IgnoreErrors, @namespace: Namespace);
                 }
+
+                //Print the function keys message to the console
+                FuncAppKeysHelper.FunKeysMessage(existingKeysSecret, newKeysSecret);
             }
         }
 
--- a/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
+++ b/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
@@ -26,14 +26,13 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
         public string Namespace { get; set; } = "default";
         public string PullSecret { get; set; } = string.Empty;
         public bool NoDocker { get; set; }
-        public bool UseConfigMapForAppSettings { get; set; }
+        public bool UseConfigMap { get; set; }
         public bool DryRun { get; private set; }
         public string ImageName { get; private set; }
-        public string AppSettingsConfigMapName { get; private set; }
-        public string AppSettingsSecretsCollectionName { get; private set; }
-        public string FunAppKeysConfigMapName { get; private set; }
-        public string FunAppKeysSecretsCollectionName { get; private set; }
-        public bool MountFuncAppKeysAsContainerVolume { get; private set; }
+        public string ConfigMapName { get; private set; }
+        public string SecretsCollectionName { get; private set; }
+        public string KeysSecretCollectionName { get; private set; }
+        public bool MountFuncKeysAsContainerVolume { get; private set; }
         public int? PollingInterval { get; private set; }
         public int? CooldownPeriod { get; private set; }
         public string ServiceType { get; set; } = "LoadBalancer";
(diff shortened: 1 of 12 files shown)


src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -102,21 +109,27 @@ public override async Task RunAsync()
                 triggers = await DockerHelpers.GetTriggersFromDockerImage(resolvedImageName);
             }
 
-            var resources = KubernetesHelper.GetFunctionsDeploymentResources(
+            var functionNames = triggers.FunctionsJson.Select(f => f.Key);
+            var funcAppKeys = FuncAppKeysHelper.CreateKeys(functionNames);
```
ahmelsayed: I think this can be encapsulated in the `GetFunctionsDeploymentResources` call to keep the action itself simpler.

src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -58,8 +62,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
             SetFlag<int>("cooldown-period", "The cooldown period for the deployment before scaling back to 0 after all triggers are no longer active. Default: 300 (seconds)", p => CooldownPeriod = p);
             SetFlag<int>("min-replicas", "Minimum replica count", m => MinReplicaCount = m);
             SetFlag<int>("max-replicas", "Maximum replica count to scale to by HPA", m => MaxReplicaCount = m);
-            SetFlag<string>("secret-name", "The name of a kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => SecretsCollectionName = sn);
-            SetFlag<string>("config-map-name", "The name of a config map to use in the deployment", cm => ConfigMapName = cm);
+            SetFlag<string>("funcappkeys-secret-name", "The name of a kubernetes secret collection to use for the function app keys (host keys, function keys etc.)", ksn => FunAppKeysSecretsCollectionName = ksn);
+            SetFlag<string>("funcappkeys-config-map-name", "The name of a kubernetes config map to use for the function app keys (host keys, function keys etc.)", kcm => FunAppKeysConfigMapName = kcm);
```
ahmelsayed: let's only do secrets for now.

src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -58,8 +62,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
...
+            SetFlag<bool>("mount-funcappkeys-as-containervolume", "The flag indicating to mount the func app keys as container volume", kmv => MountFuncAppKeysAsContainerVolume = kmv);
+            SetFlag<string>("appsettings-secret-name", "The name of a existing func app settings kubernetes secret collection to use in the deployment instead of generating one based on local.settings.json", sn => AppSettingsSecretsCollectionName = sn);
+            SetFlag<string>("appsettings-config-map-name", "The name of an existing config map with func app settings to use in the deployment", cm => AppSettingsConfigMapName = cm);
```
ahmelsayed: I'd rather not rename existing flags if possible. I think we can get rid of `funcappkeys-config-map-name`, and rename `funcappkeys-secret-name` to `keys-secret-name`, then leave `secret-name` and `config-map-name` as is

src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -69,7 +76,7 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 ServiceType = s;
             });
             SetFlag<bool>("no-docker", "With --image-name, the core-tools will inspect the functions inside the image. This will require mounting the image filesystem. Passing --no-docker uses current directory for functions.", nd => NoDocker = nd);
-            SetFlag<bool>("use-config-map", "Use a ConfigMap/V1 instead of a Secret/V1 object.", c => UseConfigMap = c);
+            SetFlag<bool>("use-config-map-for-appsettings", "Use a ConfigMap/V1 instead of a Secret/V1 object for function app settings configurations", c => UseConfigMapForAppSettings = c);
```
ahmelsayed: revert to `use-config-map` since keys are always secrets

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+
+                if (!await ResourceExists("secret", funcAppKeysSecretsCollectionName, @namespace))
+                {
+                    result.Insert(resourceIndex, appKeysSecret);
+                    resourceIndex++;
+                }
+
+                foreach (var deployment in deployments)
+                {
+                    deployment.Spec.Template.Spec.Containers.First().EnvFrom.Append(
```
ahmelsayed: This will have no effect on an IEnumerable

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+
+                if (!await ResourceExists("configMap", funcAppKeysConfigMapName, @namespace))
+                {
+                    result.Insert(resourceIndex, appKeysConfigMap);
+                    resourceIndex++;
+                }
+
+                foreach (var deployment in deployments)
+                {
+                    deployment.Spec.Template.Spec.Containers.First().EnvFrom.Append(
```
ahmelsayed: This will have no effect on an IEnumerable

src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
ahmelsayed: also keys are only needed for http functions. In `GetFunctionsDeploymentResources`, functions are split into 2 collections, and we only need to create the keys for http ones

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -170,21 +181,23 @@ internal static string GetKedaResources(string @namespace)
...
             // Set worker runtime if needed.
-            if (!secrets.ContainsKey(Constants.FunctionsWorkerRuntime))
+            if (!appSettingsSecrets.ContainsKey(Constants.FunctionsWorkerRuntime))
             {
-                secrets[Constants.FunctionsWorkerRuntime] = GlobalCoreToolsSettings.CurrentWorkerRuntime.ToString();
+                appSettingsSecrets[Constants.FunctionsWorkerRuntime] = GlobalCoreToolsSettings.CurrentWorkerRuntime.ToString();
             }
 
-            if (useConfigMap)
+            int resourceIndex = 0;
```
ahmelsayed: I don't think tracking the index is necessary. I just wanted the output to have all dependent resources before the `Deployment` resource, but I don't think it's actually worth introducing a tracking variable. just inserting at 0 should be fine

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+            }
+
+            //if function keys Secrets/configMaps needs to be mounted as volume in the function runtime container
+            if (mountFuncKeysAsContainerVolume)
+            {
+                FuncAppKeysHelper.CreateFuncAppKeysVolumeMountDeploymentResource(deployments, funcAppKeysSecretsCollectionName, funcAppKeysConfigMapName);
+            }
+            else
+            {
+                var svcActName = "function-identity-svc-act";
```
ahmelsayed: service account name should drive from the deployment name to be tied to this particular deployment. Otherwise all function apps in a given namespace will be sharing the same identity.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+                var svcAct = GetServiceAccount(svcActName, @namespace);
+                if (!await ResourceExists("ServiceAccount", svcActName, @namespace))
```
ahmelsayed: This shouldn't be needed for service accounts as the deployment should be idempotent.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+                {
+                    result.Insert(resourceIndex, svcAct);
+                    resourceIndex++;
+                }
+
+                var secretManagerRoleName = "secrets-manager-role";
```
ahmelsayed: This could be shared, but maybe rename to `functions-keys-manager-role`

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+                var secretManagerRole = GetRole(secretManagerRoleName, @namespace);
+                if (!await ResourceExists("Role", secretManagerRoleName, @namespace))
```
ahmelsayed: should also be idempotent.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+                {
+                    result.Insert(resourceIndex, svcAct);
+                    resourceIndex++;
+                }
+
+                var roleBindingName = "function-identity-svcact-to-secret-manager-rolebinding";
```
ahmelsayed: the role binding name should drive from the service account name.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,88 @@ internal static string GetKedaResources(string @namespace)
...
+                var secretRoleBinding = GetRoleBinding(roleBindingName, @namespace, secretManagerRoleName, svcActName);
+                if (!await ResourceExists("RoleBinding", secretManagerRoleName, @namespace))
```
ahmelsayed: should also be idempotent.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -528,5 +627,71 @@ private static ConfigMapV1 GetConfigMap(string name, string @namespace, IDiction
...
+                Kind = "ServiceAccount",
+                Metadata = new ObjectMetadataV1
+                {
+                    Name = name,
+                    Namespace = @namespace
+                }
+            };
+        }
+
+        public static RoleV1 GetRole(string name, string @namespace)
```
ahmelsayed: rename to reflect role permissions specified below since this is not a generic "Role" creator.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
SatishRanjan: This index is used to insert the Kubernetes resources in the right order, e.g. to create ServiceAccount, Role and RoleBinding resource in the order

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
SatishRanjan: wouldn't it be required to check for existence if created as a shared resource?

ahmelsayed: It should always be idempotent though, right?

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -152,6 +161,8 @@ internal static string GetKedaResources(string @namespace)
             {
                 int position = 0;
                 var enabledFunctions = httpFunctions.ToDictionary(k => $"AzureFunctionsJobHost__functions__{position++}", v => v.Key);
+                //Add environment variables for the func app keys
+                FuncAppKeysHelper.AddAppKeysEnvironVariableNames(enabledFunctions, keysSecretCollectionName, mountKeysAsContainerVolume);
```
ahmelsayed: It's hard to reason what this method actually doing just looking at the name, passed parameters, and their usage below.

I'd rather have this method return the dictionary of additional env var, then concat or merge the dictionaries when/before calling `GetDeployment`

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,6 +264,54 @@ internal static string GetKedaResources(string @namespace)
...
+                else
+                {
+                    var svcActName = $"{name}-function-keys-identity-svc-act";
+                    var svcActDeploymentResource = GetServiceAccount(svcActName, @namespace);
+                    result.Insert(resourceIndex, svcActDeploymentResource);
+                    resourceIndex++;
+
+                    var funcKeysManagerRoleName = "functions-keys-manager-role";
+                    var secretManagerRole = GetSecretManagerRole(funcKeysManagerRoleName, @namespace);
+                    if (!await ResourceExists("Role", funcKeysManagerRoleName, @namespace))
```
ahmelsayed: shouldn't be needed for the role since it's idempotent, right?

src/Azure.Functions.Cli/Kubernetes/FuncKeys/FuncAppKeysHelper.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            if (!envVariables.ContainsKey(AzureWebJobsKubernetesSecretNameEnvVariableName)
+                && !mountKeysAsContainerVolume)
+            {
+                if (!string.IsNullOrWhiteSpace(keysSecretCollectionName))
+                {
+                    envVariables.Add(AzureWebJobsKubernetesSecretNameEnvVariableName, $"secrets/{keysSecretCollectionName}");
+                }
+            }
```
ahmelsayed: I think the logic of this method can be significantly simplified if you just default the `keysSecretCollectionName` in `GetFunctionsDeploymentResources()` to a value if it's null. That way you just need to check `mountKeysAsContainerVolume` in one if statement, right?

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -140,7 +153,9 @@ internal static string GetKedaResources(string @namespace)
             int? cooldownPeriod = null,
             string serviceType = "LoadBalancer",
             int? minReplicas = null,
-            int? maxReplicas = null)
+            int? maxReplicas = null,
+            string keysSecretCollectionName = "func-keys-secret",
```
ahmelsayed: The default name for the secret collection should drive from the deployment name to account for multiple deployments.

ahmelsayed: also if you don't pass `--keys-secret-name` this will be null, which will cause the deployment to fail trying to create a resource with an empty name

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -250,11 +275,68 @@ internal static string GetKedaResources(string @namespace)
...
+                            keysSecret = existingFuncKeysSecret;
+                        }
+                    }
+                }
+
+                if (funcKeys.Any())
+                {
+                    newKeysSecret = GetSecret(keysSecretCollectionName, @namespace, funcKeys);
+                    keysSecret = newKeysSecret;
+                }
```
ahmelsayed: This logic is incorrect. Try running the example in this comment https://github.com/Azure/azure-functions-core-tools/pull/1715#pullrequestreview-338871909 

And verify the end result in the cluster.

src/Azure.Functions.Cli/Kubernetes/FuncKeys/FuncAppKeysHelper.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        private static void PrintKeyOutputMessage(IEnumerable<KeyValuePair<string, string>> functionKeys, string keyMsg)
+        {
+            var keyOutputMsgTemplate = "http://[ip]/api/{0}?code={1}";
+            foreach (var funcKey in functionKeys)
+            {
+                var functionName = funcKey.Key.Split('.')[1];
+                var functionKey = Encoding.UTF8.GetString(Convert.FromBase64String(funcKey.Value));
+                ColoredConsole.WriteLine(string.Concat("\t", string.Format(keyOutputMsgTemplate, functionName, functionKey), keyMsg));
+            }
```
ahmelsayed: This is also not really correct. You can't build the URL just like that because we allow the users to customize the route per function and per host. You must query the runtime to get the actual list of routes. See the [`AzureHelper.PrintFunctionsInfo()`](https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Helpers/AzureHelper.cs#L350) for more details 

You should be able to resolve the actual IP by polling the service created in the command above. If the service was of type `LoadBalancer`, you should be able to just hit it directly using the master key. We can leave `ClusterIP` for now because you'll need to `port-forward` to it.

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +159,28 @@ internal static string GetKedaResources(string @namespace)
...
+            string keysSecretCollectionName = null,
+            bool mountKeysAsContainerVolume = false)
         {
             ScaledObjectV1Alpha1 scaledobject = null;
             var result = new List<IKubernetesResource>();
             var deployments = new List<DeploymentV1Apps>();
             var httpFunctions = triggers.FunctionsJson
                 .Where(b => b.Value["bindings"]?.Any(e => e?["type"].ToString().IndexOf("httpTrigger", StringComparison.OrdinalIgnoreCase) != -1) == true);
             var nonHttpFunctions = triggers.FunctionsJson.Where(f => httpFunctions.All(h => h.Key != f.Key));
+            keysSecretCollectionName = string.IsNullOrEmpty(keysSecretCollectionName) ? string.Concat($"func-keys-kube-secret-{name}") : keysSecretCollectionName;
```
ahmelsayed:
```suggestion
            keysSecretCollectionName = string.IsNullOrEmpty(keysSecretCollectionName)
            	? $"func-keys-kube-secret-{name}"
            	: keysSecretCollectionName;
```

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +159,28 @@ internal static string GetKedaResources(string @namespace)
...
             if (httpFunctions.Any())
             {
                 int position = 0;
                 var enabledFunctions = httpFunctions.ToDictionary(k => $"AzureFunctionsJobHost__functions__{position++}", v => v.Key);
+                //Environment variables for the func app keys
+                var funcKeyEnvironmentVariables = FuncAppKeysHelper.FuncKeysKubernetesEnvironVariables(keysSecretCollectionName, mountKeysAsContainerVolume);
+                foreach (var environmentVar in funcKeyEnvironmentVariables)
+                {
+                    enabledFunctions.TryAdd(environmentVar.Key, environmentVar.Value);
```
ahmelsayed: variables should describe their content. Create a new dictionary that merges the 2 collections with a proper name

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -256,11 +282,97 @@ internal static string GetKedaResources(string @namespace)
...
+                else if (unchangedFuncKeys?.Any() == true)
+                {
+                    kubernetesKeysSecretResource = allExistingFuncKeys;
+                }
+                //All keys are the new keys
+                else
+                {
+                    kubernetesKeysSecretResource = GetSecret(keysSecretCollectionName, @namespace, currentImageFuncKeys);
+                    newFuncKeys = kubernetesKeysSecretResource.Data;
+                }
```
ahmelsayed: these conditions are not mutually exclusive, right?

ahmelsayed: also I'd encapsulate this logic in another method to make it easier to read

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -256,11 +282,97 @@ internal static string GetKedaResources(string @namespace)
...
+                    //add service account identity to the pod
+                    foreach (var deployment in deployments)
+                    {
+                        deployment.Spec.Template.Spec.ServiceAccountName = svcActName;
+                    }
+                }
+            }
+
+            result = result.Concat(deployments).ToList();
+            return (scaledobject != null ? result.Append(scaledobject) : result, unchangedFuncKeys, newFuncKeys);
```
ahmelsayed: merge `unchangedFuncKeys` and `newFuncKeys` to simplify `GetFunctionsDeploymentResources()` contract

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +165,25 @@ internal static string GetKedaResources(string @namespace)
...
+            bool mountKeysAsContainerVolume = false)
         {
             ScaledObjectV1Alpha1 scaledobject = null;
             var result = new List<IKubernetesResource>();
             var deployments = new List<DeploymentV1Apps>();
             var httpFunctions = triggers.FunctionsJson
                 .Where(b => b.Value["bindings"]?.Any(e => e?["type"].ToString().IndexOf("httpTrigger", StringComparison.OrdinalIgnoreCase) != -1) == true);
             var nonHttpFunctions = triggers.FunctionsJson.Where(f => httpFunctions.All(h => h.Key != f.Key));
+            keysSecretCollectionName = string.IsNullOrEmpty(keysSecretCollectionName)
+                ? string.Concat($"func-keys-kube-secret-{name}")
```
ahmelsayed:
```suggestion
                ? $"func-keys-kube-secret-{name}"
```

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +165,25 @@ internal static string GetKedaResources(string @namespace)
...
+                : keysSecretCollectionName;
             if (httpFunctions.Any())
             {
                 int position = 0;
                 var enabledFunctions = httpFunctions.ToDictionary(k => $"AzureFunctionsJobHost__functions__{position++}", v => v.Key);
+                //Environment variables for the func app keys kubernetes secret              
+                enabledFunctions = enabledFunctions.Concat(FuncAppKeysHelper.FuncKeysKubernetesEnvironVariables(keysSecretCollectionName, mountKeysAsContainerVolume)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
ahmelsayed: create a new variable name

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +165,25 @@ internal static string GetKedaResources(string @namespace)
...
                 .Where(b => b.Value["bindings"]?.Any(e => e?["type"].ToString().IndexOf("httpTrigger", StringComparison.OrdinalIgnoreCase) != -1) == true);
             var nonHttpFunctions = triggers.FunctionsJson.Where(f => httpFunctions.All(h => h.Key != f.Key));
+            keysSecretCollectionName = string.IsNullOrEmpty(keysSecretCollectionName)
+                ? string.Concat($"func-keys-kube-secret-{name}")
+                : keysSecretCollectionName;
             if (httpFunctions.Any())
             {
                 int position = 0;
                 var enabledFunctions = httpFunctions.ToDictionary(k => $"AzureFunctionsJobHost__functions__{position++}", v => v.Key);
+                //Environment variables for the func app keys kubernetes secret
```
ahmelsayed:
```suggestion
                //Environment variables for the func app keys kubernetes secret
```

test/Azure.Functions.Cli.Tests/KubernetesHelperUnitTests.cs
```diff
@@ -57,5 +59,52 @@ public void PopulateMetadataDictionary_CorrectlyPopulatesRabbitMQMetadata()
...
+            var @namespace = "funcappkeys-test-ns0";
+            var resource = KubernetesHelper.GetRoleBinding(roleBindingName, @namespace, roleName, svcActName);
+            var payload = JsonConvert.SerializeObject(resource, Formatting.Indented,
+               new JsonSerializerSettings
+               {
+                   NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore
+               });
+
+            Assert.True(JToken.Parse(payload) is JToken);
+        }
```
ahmelsayed: These tests are not validating anything. I'd remove them instead

SatishRanjan:
> I think the flow can be changed a little bit to be more idempotent when creating/updating a deployment.
> 
> The main scenario we're trying to enable is
> 
> ```shell
> $ func init --docker --python
> $ func new --name http1 --template http
> $ func kubernetes deploy --name test1
> 
> {... deployment messages ...}
> Http Functions:
>     http://{ip}/api/http1?key=111111111111111
> 
> # add a new function
> $ func new --name http2 --template http
> $ func kubernetes deploy --name test1
> 
> {... deployment messages ...}
> Http Functions:
>     http://{ip}/api/http1?key=111111111111111 # this didn't change
>     http://{ip}/api/http2?key=222222222222222 # this was added
> ```
> 
> I think the code will read easier if you do a "Get" before you create the resource, then assume that every `kubectl apply -f` is an idempotent update to all the resources associated with this deployment.

Resolved

SatishRanjan:
closed by mistake

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -146,18 +165,26 @@ internal static string GetKedaResources(string @namespace)
...
+            keysSecretCollectionName = string.IsNullOrEmpty(keysSecretCollectionName)
+                ? $"func-keys-kube-secret-{name}"
+                : keysSecretCollectionName;
             if (httpFunctions.Any())
             {
                 int position = 0;
                 var enabledFunctions = httpFunctions.ToDictionary(k => $"AzureFunctionsJobHost__functions__{position++}", v => v.Key);
+                //Environment variables for the func app keys kubernetes secret
+                var kubernetesSecretEnvironmentVariable = FuncAppKeysHelper.FuncKeysKubernetesEnvironVariables(keysSecretCollectionName, mountKeysAsContainerVolume);
+                enabledFunctions = enabledFunctions.Concat(kubernetesSecretEnvironmentVariable).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
ahmelsayed:
```suggestion
                var additionalEnvVars = enabledFunctions.Concat(kubernetesSecretEnvironmentVariable).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```

and replace `enabledFunctions` in the `GetDeployment()` call

ahmelsayed:
@SatishRanjan can you rebase your branch?

saedabdu:
Appreciate a working example on how to use the new flags for a dotnet function app.
Thanks in advance!

PR #1720, opened 2019-12-18
Update version and signing as well

yojagad:


ankitkumarr:
I am assuming you ran `redirectCompare` and `packageConfigParser`? Also if you wouldn't mind squashing the commits. Looks good to me otherwise.

build.fsx
```diff
@@ -187,6 +189,7 @@ Target "GenerateZipToSign" (fun _ ->
         "Ignite.SharpNetSH.dll"
         "NCrontab.dll"
         "Newtonsoft.Json.dll"
+        "protobuf-net.dll"
```
ahmelsayed: just curious why protobuf dll is in v1? was that added to the v2 webjobs project?

yojagad: Actually, good point. I'm not sure why it was ever installed in v1 to begin with. Let me see if I can find out

ankitkumarr: Looks like it comes from https://github.com/Azure/azure-functions-host/blob/v1.x/src/WebJobs.Script/WebJobs.Script.csproj#L231-L233?

ahmelsayed: oh, so it has been there for a while and we just missed it. thanks @ankitkumarr  and @yojagad, I was just curious where it was coming from.

yojagad:
Ran packageConfigParser, redirectCompare and squashed the commits in the latest push.

PR #1730, opened 2019-12-19
Fixed .funcignore parser to ignore files/directories based on an exact match rather than on contains

### LINKED ISSUE #1507: core-tools ignores everything that starts with .funcignore entry
1. func init
2. func new (select HttpTrigger)
3. echo "Http" > `.funcignore`
4. func pack

The zip file is missing `HttpTrigger` files.

madelinegordon:
fixes #1507

### DIFF
--- a/test/Azure.Functions.Cli.Tests/GitIgnoreParserTests.cs
+++ b/test/Azure.Functions.Cli.Tests/GitIgnoreParserTests.cs
@@ -38,6 +38,10 @@ namespace Azure.Functions.Cli.Tests
 
 /foo/*.wat
 
+/test1
+
+test2
+
 # Ignore some deep sub folders
 /othernonexistent/**/what
 
@@ -56,7 +60,16 @@ namespace Azure.Functions.Cli.Tests
         {
             _gitignore.Accepts("test/index.js").Should().BeTrue();
             _gitignore.Accepts("wat/test/index.js").Should().BeTrue();
+            _gitignore.Accepts("/othernonexistent/blah/adsasd/whatads").Should().BeTrue();
+            _gitignore.Accepts("/othernonexistent/blah/adsasd/what/foo").Should().BeTrue();
             _gitignoreNoNegatives.Accepts("test/index.js").Should().BeTrue();
+        }
+
+        [Fact]
+        public void AcceptShouldAcceptFileNameThatContainsNameIgnored()
+        {
+            _gitignore.Accepts("test1").Should().BeFalse();
+            _gitignore.Accepts("test1File.wat").Should().BeTrue();
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Azure.Functions.Cli.Tests/GitIgnoreParserTests.cs
```diff
@@ -59,6 +61,13 @@ public void AcceptShouldAcceptTheGivenFilenames()
             _gitignoreNoNegatives.Accepts("test/index.js").Should().BeTrue();
         }
 
+        [Fact]
+        public void AcceptShouldAcceptFileNameThatContainsNameIgnored()
+        {
```
fabiocav: @ankitkumarr @ahmelsayed could you also take a quick look at the test cases we have in this file (outside of the context of this changes) and see if we have the coverage we want? If there's anything else we should add, we should do that as part of this changes.

ahmelsayed: Honestly, I'm not too sure about this implementation of gitignore. I don't really understand it, and always meant to revisit it, but never did. 

here are some tests that I think should have been there. 

``` bash
file.log # should ignore
dir/dir/file.log # should ignore
baz # should ignore
dir/file.wat # should ignore

/othernonexistent/blah/what # should ignore
/othernonexistent/blah/adsasd/whatads # should accept
/othernonexistent/blah/adsasd/what/foo # should accept
```

Though I just tried them, and this simple one `dir/dir/file.log` doesn't pass for some reason. a couple above needed @madelinegordon's fix to pass. I honestly wouldn't have been able to figure that out. I couldn't find a dotnet parser at the time, and didn't wanna look at git's code so just picked that random lib

ankitkumarr: Another one I can think of is --
adding `test2` in the gitignore, and

```
test2 # should ignore
/test2 # should ignore
/directory/subdirectory/test2 # should ignore
/directory/subdirectory/test2/supersubdirectory # should ignore
/test22 # accept
/directory/subdirectory/test21 # accept
```

madelinegordon: By adding these tests I’ve discovered that there are more issues which were already present in this implementation: 
•	When the .gitignore contains “test2”, the regex should match “/test2”, and “abc/test2”, and with and without my changes it does not match. 
•	Not unignoring files correctly (in current test, “/nonexistent/foo” should be unignored but is a match). 
The change I’ve made thus far fixed the issue of matching “test1abc” when only “test1” was supposed to be ignored. 

The regex generated by the previous implementation is: `^((([^\/]+)\.log)|(([^\/]+)\.swp)|(baz)|(foo\/([^\/]+)\.wat)|(node_modules)|(nonexistent)|(othernonexistent\/(.+)\/what)|(test1)|(test2))` , and by the new implementation: `^((([^\/]+)\.log)|(([^\/]+)\.swp)|(baz)|(foo\/([^\/]+)\.wat)|(node_modules)|(nonexistent)|(othernonexistent\/(.+)\/what)|(test1)|(test2))\b`. You can test these regex at https://regex101.com/. 

The .gitignore rules (which I assume this parser is modeled after) can be found at https://git-scm.com/docs/gitignore.  

Adding these tests presented more holes in this parser implementation, and as far as I can tell, my changes did not introduce any issues that weren’t already there. @fabiocav  should I continue to work on figuring out how to fix those other issues for this same PR?

Issue #1724, opened 2019-12-18
If too many subscriptions are available, core-tools may fail to find the app


ankitkumarr:
This is because `azure-functions-core-tool` uses "https://management.azure.com/subscriptions" to get  all the subscriptions. This API, similar to other ARM API, may return partial result with a `nextLink` to an API for next set of results.
Today, we completely ignore that. We need to handle this case.

We started doing this in the past to query function apps. But since we moved to "ARG", it wasn't needed anymore.

Additionally, today we only use `--subscription` for `func azure functionapp publish`. We need to start using it for all `func azure functionapp <sub-action>` commands.

This can be broken into two issues. But adding `--subscription` is a super small change.

Two workitems --
1. If `nextLink` present in ARM response, honor it.
2. Use `--subscription` when set for all `func azure functionapp <sub-action>` commands.

PR #1746, opened 2020-01-06
Fix bug when too many subscriptions are available

ankitkumarr:
Fixes https://github.com/Azure/azure-functions-core-tools/issues/1724


ahmelsayed:
Maybe I misunderstood ARG, but I thought this was replaced by an ARG query, right?

ankitkumarr:
For ARG, we need the list of the subscriptions to run the query on. So we still call the ARM  endpoint to get all the subscriptions. The response may return partial result if user has too many subscriptions.

ahmelsayed:
Got it. Thanks for explaining that.

PR #1795, opened 2020-02-04
Extensions install will fetch latest version if not specified.

### LINKED ISSUE #1773: Update "Func extensions install" command to fetch the latest version from nuget.org

gzuber:
Fixes #1773


soninaren:
We also need to update the templates to latest to make sure they don't break

soninaren:
@gzuber can you cherry pick this to v3.x as well?

PR #1799, opened 2020-02-04
Fetching templates from cdn at build time instead of including it as static resource

### LINKED ISSUE #1740: Fetch templates from tooling feed instead of hard coding
We need to look up the latest release for the corresponding runtime and download the VS and Scripting templates

https://github.com/Azure/azure-functions-tooling-feed/blob/master/cli-feed-v3.json

**Example**
Get latest release version for v3 (`tags -> v3 -> release -> 3.2.0`)
- https://github.com/Azure/azure-functions-tooling-feed/blob/d833d6c709a1654508bd53a5ef1f47e2caeab63b/cli-feed-v3.json#L60

Get the VS templates (C# precompiled) for v3 (`releases -> 3.2.0`)
- https://github.com/Azure/azure-functions-tooling-feed/blob/d833d6c709a1654508bd53a5ef1f47e2caeab63b/cli-feed-v3.json#L2990-L2991

Get the non dotnet templates for v3 (`releases -> 3.2.0`)
- https://github.com/Azure/azure-functions-tooling-feed/blob/d833d6c709a1654508bd53a5ef1f47e2caeab63b/cli-feed-v3.json#L2992

soninaren:
closes #1740


build/BuildSteps.cs
```diff
@@ -14,6 +14,7 @@
 using System.Security.Cryptography;
 using System.Threading.Tasks;
 using System.Text.RegularExpressions;
+using Newtonsoft.Json.Linq;
```
ankitkumarr: leftover, please remove :)

build/FileHelpers.cs
```diff
@@ -6,6 +6,20 @@ namespace Build
...
+            if (!Directory.Exists(dirPath))
+            {
+                Directory.CreateDirectory(dirPath);
+            }
+        }
+
+        public static bool FileExists(string filePath)
+        {
+            return File.Exists(filePath);
+        }
```
ankitkumarr: no biggie, but out of curiosity - does this add any benefit to just calling `File.Exists`?

soninaren: No benefit here, i had a bunch of code here to do some additional validation, that i removed but left the method

soninaren: have removed this

build/BuildSteps.cs
soninaren: done

PR #1807, opened 2020-02-05
disabling sampling of requests in default host.json

soninaren:
closes https://github.com/Azure/azure-functions-templates/issues/906

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/host.json
+++ b/src/Azure.Functions.Cli/StaticResources/host.json
@@ -1,3 +1,11 @@
 {
-    "version":  "2.0"
+  "version": "2.0",
+  "logging": {
+    "applicationInsights": {
+      "samplingExcludedTypes": "Request",
+      "samplingSettings": {
+        "isEnabled": true
+      }
+    }
+  }
 }
--- a/test/Azure.Functions.Cli.Tests/E2E/InitTests.cs
+++ b/test/Azure.Functions.Cli.Tests/E2E/InitTests.cs
@@ -421,7 +421,7 @@ namespace Azure.Functions.Cli.Tests.E2E
         {
             return CliTester.Run(new RunConfiguration
             {
-                Commands = new[] { "init ." },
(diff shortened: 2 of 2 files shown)


src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -434,5 +436,15 @@ private static async Task<string> AddBundleConfig(string hostJsonContent)
             hostJsonObj.Add("extensionBundle", bundleConfig);
             return JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
         }
+
+
+        private static async Task<string> AddLoggingConfig(string hostJsonContent)
```
ahmelsayed: any reason for this approach as opposed to just have it in the default host.json?

soninaren: I followed the sample of setting bundles as default in host.json. I think the only advantage we get by following this approach, is that we can introduce a flag to skip this logic in the init command `no-loggingConfig`. While that made sense for bundles, I don't think it does for Application Insight configuration. I will revert this change to simple add this configuration to default host.json

ahmelsayed: sounds good

brettsam:
While you are doing this -- @jeffhollan (or @paulbatum... I can't remember who brought it up), weren't we discussing also adding this to the default?

```
"samplingSettings": {
    "isEnabled": true 
}
```

So it'd be much more "in your face" that sampling is enabled by default?

soninaren:
> While you are doing this -- @jeffhollan (or @paulbatum... I can't remember who brought it up), weren't we discussing also adding this to the default?
> 
> ```
> "samplingSettings": {
>     "isEnabled": true 
> }
> ```
> 
> So it'd be much more "in your face" that sampling is enabled by default?

Jeff added that to the issue after the PR was out. Will add that to the config as well

Issue #1804, opened 2020-02-05
Any `func` command does not work


seed-of-apricot:
I installed the package globally, created a new project through VSCode extension and tried to run the sample without any changes. The error I get is:

```node
> func start

events.js:298
      throw er; // Unhandled 'error' event
      ^

Error: spawn C:\Users\apricot\AppData\Roaming\nvm\v13.7.0\node_modules\azure-functions-core-tools\bin/func ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:467:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:273:12)
    at onErrorNT (internal/child_process.js:467:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn C:\\Users\\apricot\\AppData\\Roaming\\nvm\\v13.7.0\\node_modules\\azure-functions-core-tools\\bin/func',
  path: 'C:\\Users\\apricot\\AppData\\Roaming\\nvm\\v13.7.0\\node_modules\\azure-functions-core-tools\\bin/func',
  spawnargs: [ 'start' ]
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! hogehoge@1.0.0 start: `func start`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the hogehoge@1.0.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.        

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\apricot\AppData\Roaming\npm-cache\_logs\2020-02-05T09_13_13_764Z-debug.log
```

So, how about `func` itself?

```node
C:\Users\apricot\src\user\directory\hogehoge>func init fugafuga
events.js:298
      throw er; // Unhandled 'error' event
      ^

Error: spawn C:\Users\apricot\AppData\Roaming\nvm\v13.7.0\node_modules\azure-functions-core-tools\bin/func ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:467:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
[...]

PR #1812, opened 2020-02-07
Update unzipper npm package to resolve unzip issue during installation

ronbrogan:
I encountered an issue installing tools on a fresh install where the install script would fail silently before the download of "https://functionscdn.azureedge.net/public/2.4.290/Azure.Functions.Cli.win-x64.2.4.290.zip" would even start. Looking at the unzipper library that is used, I saw that a fix for the issue was merged here: 
https://github.com/ZJONSSON/node-unzipper/pull/169

This issue has similar symptoms to what I experienced, so this may be a fix for that as well:
https://github.com/Azure/azure-functions-core-tools/issues/1804


ahmelsayed:
Thanks @ronbrogan

ahmelsayed:
@ankitkumarr can we pull this change into the v3.x branch as well?

Issue #693, opened 2018-09-04
Choco should support 64 bit install of cli


EricJizbaMSFT:
As of today, choco only seems to include the x86 version of the functions cli: https://chocolatey.org/packages/azure-functions-core-tools
![screen shot 2018-09-04 at 9 48 10 am](https://user-images.githubusercontent.com/11282622/45045401-ae57c180-b027-11e8-9149-0e4d70c4ade1.png)

However, VS Code only supports debugging 64 bit processes (see https://github.com/OmniSharp/omnisharp-vscode/issues/1716) so it would be very helpful if you also added 64 bit :)

andreasblueher:
What is the current status on this issue? I can't debug my azure functions and this is suboptimal at best.

Arnavion:
As of today, the only workaround is to install using npm / yarn instead.

cgillum:
For those who prefer to stay away from npm/yarn, here is a workaround that worked for me (copied from [this StackOverflow answer](https://stackoverflow.com/a/55120037/2069)):

1. Run `choco uninstall azure-functions-core-tools`
2. Download nupkg file from [here](https://chocolatey.org/packages/azure-functions-core-tools) (see the _Download_ link)
3. Open the nupkg in [Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) and edit the `tools\chocolateyinstall.ps1` script (change `x86` to `x64` in the package URL)
4. Run `choco install azure-functions-core-tools -source . --ignore-checksums` in the folder where edited nupkg file is.

This installed the x64 version of the Azure Functions Core Tools. After I did this, debugging .NET code in VS Code started working for me.

gep13:
@EricJizbaMSFT I work for Chocolatey, and will happily help to get this package working again.  Do you know who is behind the account here on chocolatey.org: https://chocolatey.org/profiles/nugetazurefunctions  It would be good to work with them to get this package updated.

ahmedelnably:
@EricJizbaMSFT as part of the v3 release of the core tools to choco, we will default to x64 of the tools (all other installers are x64 only). v2 will stay on x86, to minimize any chance of breaking users.
@gep13 thanks for offering help, @fabiocav and @ahmelsayed own the package

gep13:
@ahmedelnably thanks for getting back to me, I appreciate it, and am happy to help, so feel free to reach out if you have any questions.  

[...]

PR #1813, opened 2020-02-08
Add x64 notice installing tools for debugging

DenisBalan:
For debugging under vscode, x64 installer is required.
After issue #693 x64 install option is available.
Comment in https://github.com/Azure/azure-functions-core-tools/issues/693#issuecomment-573294750 says that clarification in README file is needed for installing x64 bitness from choco.

This PR adds a notice in README.md file on how specify x64 bitness when installing from choco package manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ choco install azure-functions-core-tools
 ```
 #### Notice: To debug functions under vscode, x64 bitness is required
 ```bash
-choco install azure-functions-core-tools --params "'/x64:true'"
+choco install azure-functions-core-tools --params "'/x64'"
 ```
 ### Mac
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,7 +39,10 @@ To install with chocolatey:
 ```bash
 choco install azure-functions-core-tools
 ```
-
+#### Notice: To debug functions under vscode, x64 bitness is required
+```bash
+choco install azure-functions-core-tools --params "'/x64:true'"
```
ankitkumarr: `"'/x64'" should also work.
We can maybe just put that instead of `"'/x64:true'"`

ankitkumarr:
Sorry this slipped and took me a while to get. Looks good to me, if you could just address one tiny comment and update with dev, we can merge this.

README.md
DenisBalan: Done

DenisBalan:
@ankitkumarr done, also merged dev into patch branch

PR #1846, opened 2020-02-25
Added MSI creation assets and CI/CD integration scripts.

kashimiz:
Note for @fabiocav: the versioning issue we were running into (getting a 'Another version of this product is already installed.' error on trying to upgrade) was due to having the same ProductId each time, thus it's set to autogenerate. The [How-To guide](https://wixtoolset.org/documentation/manual/v3/howtos/updates/major_upgrade.html) also seems to allude to this as the correct method, though it's a bit counterintuitive.


build/BuildSteps.cs
```diff
@@ -425,15 +425,6 @@ public static void Zip()
                 ColoredConsole.WriteLine($"Creating {shaPath}");
                 File.WriteAllText(shaPath, ComputeSha256(zipPath));
 
-                try
-                {
-                    Directory.Delete(path, recursive: true);
```
soninaren: Why did we remove this?

build.ps1
```diff
@@ -34,3 +48,31 @@ else {
...
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
 }
+
+# Get runtime version
+$artifactsPath = "$baseDir\artifacts"
+$cli = Get-ChildItem -Path $artifactsPath -Include func.dll -Recurse | Select-Object -First 1
+$cliVersion = [System.Diagnostics.FileVersionInfo]::GetVersionInfo($cli).FileVersion
+
+# Generate MSI installers for Windows
+@('x64', 'x86') | ForEach-Object {
```
soninaren: May want to add msi files to the CLI release pipeline @yojagad has created, so the MSI files are also automatically uploaded to each release / tag we create on github.

azure-pipelines.yml
```diff
@@ -53,10 +53,44 @@ steps:
     testResultsFiles: '**/*.trx'
     failTaskOnFailedTests: true
   condition: succeededOrFailed()
+- task: EsrpCodeSigning@1
```
soninaren: Should this be part of the release pipeline instead, or conditionally sign builds only from master / v3.x branch or build text in subject [Build MSI]? That would avoid generating and signing MSI for every commit.

build/BuildSteps.cs
ankitkumarr: I imagine it's removed to be able to use the artifacts for MSI creating after. Maybe at the end of the powershell script, we should delete this directory?

ahmelsayed:
The overall change looks fine to me in general. I'd move all the msi artifacts (icons, license, etc) into the build folder rather than top level. Also the powershell script is Windows only, right? Do we need a cross-plat way of building this?

build.ps1
kashimiz: Added.

build/BuildSteps.cs
kashimiz: Leaving this as a future enhancement for now.

azure-pipelines.yml
kashimiz: Done; moved to release pipeline.

kashimiz:
@ahmelsayed Cross-plat build.ps1 is being tracked on #1004. Spoke with Naren offline; will be doing that as a future enhancement. Other changes requested have been made.

build.ps1
```diff
@@ -34,3 +35,54 @@ else {
     Invoke-Expression -Command  "dotnet run"
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
 }
+
+if ($MsiGenBranches.Contains($env:APPVEYOR_REPO_BRANCH)) {
```
soninaren: So if someone names their branch `v3.x_hotfix` this logic would still trigger.

soninaren: Its okay if this is intentional, You can use `-split ","` and compare exact values

kashimiz: Fixed so it's a string[].contains(), not string.contains().

build/BuildSteps.cs
kashimiz: This has been done.

Issue #1694, opened 2019-12-05
Surface remote build logs


kulkarnisonia16:

Hazhzeng:
@ankitkumarr can refer to https://github.com/Azure/azure-functions-core-tools/blob/dab8daca606d1a6f398b01b6f29ce8f081c34ec9/src/Azure.Functions.Cli/Helpers/KuduLiteDeploymentHelpers.cs#L101 for Linux Consumption remote build log streaming

Hazhzeng:
This is already done on March 3, 2020. Closing this issue.

PR #1857, opened 2020-02-29
Improve dedicated plan remote build experience v3

Hazhzeng:
### Background
In Python 3.8 bug bash, we decided to improve the experience for remote build.

### Fixes
1. Suppress **custom image** warning when deploying to dedicated plan (this is due to a recent change in app service plan supporting **PYTHON|** images as official images.
2. Enable log streaming for app service plan & elastic premium plan
3. Prioritize **py** Windows launcher for Python version detection
    - expected sequence is: [py, python3, python, python36, python37, python38]

### Rationale
py.exe is aware of virtual environment, we should add it into version detection
https://www.python.org/dev/peps/pep-0486/#rationale


Hazhzeng:
Resolve https://github.com/Azure/azure-functions-core-tools/issues/1694

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -125,21 +125,21 @@ public static async Task<WorkerLanguageVersionInfo> GetEnvironmentPythonVersion(
                 return await GetVersion(pythonDefaultExecutablePath);
             }
 
-            var pythonGetVersionTask = GetVersion("python");
+            var pyGetVersionTask = GetVersion("py"); // Windows Python Launcher
```
ankitkumarr: Maybe it's fine, but this will almost always be a useless operation in a non-Windows environment. I wonder if we should do
```csharp
var pyGetVersionTask = IsWindows ? GetVersion("py") : Task.FromResult<WorkerLanguageVersionInfo>(null);
```

or something like that.

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -55,10 +55,10 @@ internal static class Constants
 
         public static readonly Dictionary<WorkerRuntime, IEnumerable<string>> WorkerRuntimeImages = new Dictionary<WorkerRuntime, IEnumerable<string>>
         {
-            { WorkerRuntime.dotnet, new [] { "mcr.microsoft.com/azure-functions/dotnet", "microsoft/azure-functions-dotnet-core2.0", "mcr.microsoft.com/azure-functions/base", "microsoft/azure-functions-base" } },
-            { WorkerRuntime.node, new [] { "mcr.microsoft.com/azure-functions/node", "microsoft/azure-functions-node8" } },
-            { WorkerRuntime.python, new [] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6" } },
-            { WorkerRuntime.powershell, new [] { "mcr.microsoft.com/azure-functions/powershell", "microsoft/azure-functions-powershell" } }
+            { WorkerRuntime.dotnet, new [] { "mcr.microsoft.com/azure-functions/dotnet", "microsoft/azure-functions-dotnet-core2.0", "mcr.microsoft.com/azure-functions/base", "microsoft/azure-functions-base", "DOTNET" } },
```
ankitkumarr: I think you'd want to change the logic here a little bit -- https://github.com/Azure/azure-functions-core-tools/blob/dab8daca606d1a6f398b01b6f29ce8f081c34ec9/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L243

If we just add `DOTNET` etc, and I have a custom image, say `foo:bar-dotnet`, it is going to assume that's a known image which is not true.

Moreover, if I have an image with a weird name `myrepository:not-node` with `FUNCTIONS_WORKER_RUNTIME` as `dotnet`. our logic above may fail with this exception - https://github.com/Azure/azure-functions-core-tools/blob/dab8daca606d1a6f398b01b6f29ce8f081c34ec9/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L269

I would suggest changing that logic to be more like "LinuxFxVersion" startswith `"DOCKER|{knownimage}:"` or equals `"DOCKER|{knownimage}"`, or whatever that can take care of such weird corner cases.

PR #1868, opened 2020-03-03
Fixed PowerShell build script to properly parse MsiGenBranches arg.

kashimiz:


build.ps1
```diff
@@ -38,7 +38,7 @@ else {
     if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
 }
 
-if ($MsiGenBranches.Contains($env:APPVEYOR_REPO_BRANCH)) {
+if ($MsiGenBranches -Contains $env:APPVEYOR_REPO_BRANCH) {
```
Francisco-Gamino: I would probably add a `.Trim()` to `$env:APPVEYOR_REPO_BRANCH` before doing the `-Contains`, e.g., 
```powershell
PS E:\> @("master","v3.x") -contains "master ".Trim()
True
```
Other than that, LGTM.

PR #1882, opened 2020-03-06
Check if bundle config exists, hostJson exists

### LINKED ISSUE #1828: Throw an error if the customers with pre compiled function app have bundles configured
Extension bundles should not be configured for functions using C# precompiled functions. It lead to a variety of issues and has a potential of users running into issues when ever we deploy a new version of extension bundle. Core tools should throw and error when it sees a host.json file with extension bundle configured in it.

soninaren:
resolves #1828

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -34,6 +34,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Microsoft.Azure.WebJobs.Script.Description;
 
 namespace Azure.Functions.Cli.Actions.HostActions
 {
@@ -248,6 +249,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
             await PreRunConditions();
             Utilities.PrintLogo();
             Utilities.PrintVersion();
+            ValidateHostJsonConfiguration();
 
             var settings = SelfHostWebHostSettingsFactory.Create(Environment.CurrentDirectory);
 
@@ -266,6 +268,20 @@ namespace Azure.Functions.Cli.Actions.HostActions
             DisplayDisabledFunctions(scriptHost);
 
             await runTask;
+        }
+
+        private void ValidateHostJsonConfiguration()
+        {
+            bool IsPreCompiledApp = IsPreCompiledFunctionApp();
+            if (IsPreCompiledApp && !HostJsonExist())
+            {
+                throw new CliException($"Host.json file does not exist. Please make sure host.json file is preset at {Environment.CurrentDirectory}");
+            }
+
+            if (IsPreCompiledApp && BundleConfigurationExists())
+            {
+                throw new CliException($"Extension bundle is configured for the function app with pre-compiled functions. Please remove extension bundle configuration from host.json: {Path.Combine(Environment.CurrentDirectory, "host.json")}");
+            }
         }
 
         private async Task PreRunConditions()
@@ -299,6 +315,39 @@ namespace Azure.Functions.Cli.Actions.HostActions
             {
                 throw new CliException($"Port {Port} is unavailable. Close the process using that port, or specify another port using --port [-p].");
             }
+        }
+
+        private bool HostJsonExist()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
+            return File.Exists(hostJsonPath);
+        }
+
+        private bool BundleConfigurationExists()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
+            var hostJson = FileSystemHelpers.ReadAllTextFromFile(hostJsonPath);
+            return hostJson.Contains("extensionBundle", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private bool IsPreCompiledFunctionApp()
+        {
+            bool isPrecompiled = false;
+            foreach (var directory in FileSystemHelpers.GetDirectories(Environment.CurrentDirectory))
+            {
+                var functionMetadataFile = Path.Combine(directory, "function.json");
+                if (File.Exists(functionMetadataFile))
+                {
+                    var functionMetadataFileContent = FileSystemHelpers.ReadAllTextFromFile(functionMetadataFile);
+                    var functionMetadata = JsonConvert.DeserializeObject<FunctionMetadata>(functionMetadataFileContent);
+                    isPrecompiled = isPrecompiled || (functionMetadata?.ScriptFile?.EndsWith(".dll") != null && functionMetadata.ScriptFile.EndsWith(".dll"));
+                }
+                if (isPrecompiled)
+                {
+                    break;
+                }
+            }
+            return isPrecompiled;
         }
 
         private void DisplayDisabledFunctions(IScriptJobHost scriptHost)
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
...
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
+            var hostJson = FileSystemHelpers.ReadAllTextFromFile(hostJsonPath);
+            return hostJson.Contains("extensionBundle", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private bool IsPreCompiledFunctionApp()
+        {
+            bool isPrecompiled = false;
+            foreach (var directory in FileSystemHelpers.GetDirectories(Environment.CurrentDirectory))
```
pragnagopa: should this ScriptRootPath instead of current directory?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
...
+            {
+                var functionMetadataFile = Path.Combine(directory, "function.json");
```
pragnagopa: use `Constants.FunctionJsonFileName`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
...
 
+        private bool HostJsonExist()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
+            return File.Exists(hostJsonPath);
+        }
+
+        private bool BundleConfigurationExists()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
```
pragnagopa: Add host.json file name to Constants

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
             }
         }
 
+        private bool HostJsonExist()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
```
pragnagopa: hostJsonPath can be constant. This will avoid calculating path twice

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
...
+            return File.Exists(hostJsonPath);
+        }
+
+        private bool BundleConfigurationExists()
+        {
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, "host.json");
+            var hostJson = FileSystemHelpers.ReadAllTextFromFile(hostJsonPath);
+            return hostJson.Contains("extensionBundle", StringComparison.OrdinalIgnoreCase);
```
pragnagopa: Add `extensionBundle`  to constants as well

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -301,6 +317,39 @@ private async Task PreRunConditions()
...
+        {
+            bool isPrecompiled = false;
+            foreach (var directory in FileSystemHelpers.GetDirectories(Environment.CurrentDirectory))
+            {
+                var functionMetadataFile = Path.Combine(directory, "function.json");
+                if (File.Exists(functionMetadataFile))
+                {
+                    var functionMetadataFileContent = FileSystemHelpers.ReadAllTextFromFile(functionMetadataFile);
+                    var functionMetadata = JsonConvert.DeserializeObject<FunctionMetadata>(functionMetadataFileContent);
+                    isPrecompiled = isPrecompiled || (functionMetadata?.ScriptFile?.EndsWith(".dll") != null && functionMetadata.ScriptFile.EndsWith(".dll"));
```
pragnagopa: use `GetExtension` similar to https://github.com/Azure/azure-functions-host/blob/aa3d0576d5bc2f73a86bdb4c32af3c97be9b9f14/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L174

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -145,6 +145,86 @@ public async Task start_displays_error_on_invalid_function_json()
...
+                        // give the host time to load functions and print any errors
+                        await Task.Delay(TimeSpan.FromSeconds(10));
+                        p.Kill();
+                    }
+                }
+            }, _output);
+        }
+
+        [Fact]
+        public async Task start_displays_error_on_missing_host_json()
```
pragnagopa: Move tests to use theory similar to : https://github.com/Azure/azure-functions-host/blob/aa3d0576d5bc2f73a86bdb4c32af3c97be9b9f14/test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs#L76

Also, add test for non dotnet languages and bundles should be allowed.

soninaren: The tests for non dotnet languages are already present. Will double check to confirm

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
soninaren: The current directory is actually managed by core tools and changes based on the type of app. For example for dotnet it changes to `/bin/output/bin`. Changing to the scriptroot would not work or would require larger refactoring.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
soninaren: Can't move to theory since there is only a single data point to test

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -268,6 +270,21 @@ public override async Task RunAsync()
             await runTask;
         }
 
+        private void ValidateHostJsonConfiguration()
+        {
+            bool IsPreCompiledApp = IsPreCompiledFunctionApp();
+            var hostJsonPath = Path.Combine(Environment.CurrentDirectory, Constants.HostJsonFileName);
```
pragnagopa: hostJsonPath is being calculated at multiple places. Move this to constant?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -268,6 +270,21 @@ public override async Task RunAsync()
...
+            if (IsPreCompiledApp && !File.Exists(hostJsonPath))
+            {
+                throw new CliException($"Host.json file in missing. Please make sure host.json file is preset at {Environment.CurrentDirectory}");
+            }
+
+            if (IsPreCompiledApp && BundleConfigurationExists(hostJsonPath))
+            {
+                throw new CliException($"Extension bundle configuration should not be present for the function app with pre-compiled functions. Please remove extension bundle configuration from host.json: {Path.Combine(Environment.CurrentDirectory, "host.json")}");
```
pragnagopa: move hostJson path to constant and reuse

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
soninaren: I updated the logic to calculate the `hostJsonPath` only once. I cannot make it a constant because the value of the `Environment.CurrentDirectory` changes based on the context. If i make hostJson path constant the `Environment.CurrentDirectory` it would may end up using incorrect path.

PR #1938, opened 2020-04-19
Added useDefaultCert parameter to use HTTPS with the default certificate.

aaltotsky:
The parameter useDefaultCert allows configuring Kestrel to use HTTPS with the default certificate. This will simplify certificate management and allow users to run the command:
host start --csharp --useHttps --pause-on-error --useDefaultCert --port 44301
Users now may use "dotnet dev-certs" to create certificates and with useDefaultCert bind Kestrel to a specific port with a certificate.

The workaround for v1 is here: https://altotsky.wordpress.com/2019/03/02/run-azure-function-with-https-endpoint-under-localhost-from-visual-studio/


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -107,6 +109,12 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .SetDefault(false)
                 .Callback(s => UseHttps = s);
 
+            Parser
+                .Setup<bool>("useDefaultCert")
```
ahmelsayed, line 113: actually would you mind updating this to `use-default-cert` instead? I know this command isn't very consistent since it has `useHttps`, but that was one of the first flags added here and I think everything else started using `kebab-case`

aaltotsky:
@ahmelsayed: Could you review the changes?

soninaren:
Closing this the PR as it against a locked branch.
@aaltotsky Please open a new PR against v4.x in case the change is still relevant.

PR #1940, opened 2020-04-20
Using ESRP signing service instead of funkins, sign on for master and release/3.0

### LINKED ISSUE #1916: Remove binary signing dependency on funkins

### LINKED ISSUE #1917: Sign core tools binaries in the release pipeline

soninaren:
- Have added the logic to sign the binaries only on master and release/3.0
- Using ESRP service that signs the binaries significantly faster
- Added signing verification on all branches
- Added pre and post signing test
- Dynamically remove files that don't need to be signed

Fixes #1916 
Fixes #1917

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -126,7 +126,7 @@ steps:
   condition: and(succeeded(),or (eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Build.SourceBranch'], 'refs/heads/release/3.0')))
 - pwsh: |
     .\repackageBinaries.ps1
-  displayName: Sign and repackage binaries
+  displayName: Repackage signed binaries
   env:
     AzureBlobSigningConnectionString: $(AzureBlobSigningConnectionString)
     BuildArtifactsStorage: $(BuildArtifactsStorage)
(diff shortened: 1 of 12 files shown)


azure-pipelines.yml
```diff
@@ -46,13 +48,130 @@ steps:
...
+      $artifactsPath = Resolve-Path "..\artifacts\"
+      $zipFilesSearchPath = Join-Path $artifactsPath "*.zip"
+      $zipFiles  = Get-ChildItem -File $zipFilesSearchPath
+      
+      foreach($zipFile in $zipFiles)
+      {
+          $zipFullPath = $zipFile.FullName
+          $fileName = $zipFile.Name
+          GenerateSha $zipFullPath $artifactsPath $fileName
+      }
```
brettsam, line 173: I'd much rather have these be in their own scripts that get called inline. It makes the yaml a lot easier to read.

build/Settings.cs
```diff
@@ -181,7 +187,13 @@ public class SignInfo
...
                 "Microsoft.Build.Framework.dll",
                 "Microsoft.Build.dll",
-                "Microsoft.CodeAnalysis.*.dll",
+                "Microsoft.CodeAnalysis.CSharp.Scripting.dll",
+                "Microsoft.CodeAnalysis.CSharp.Workspaces.dll",
+                "Microsoft.CodeAnalysis.Razor.dll",
+                "Microsoft.CodeAnalysis.Scripting.dll",
+                "Microsoft.CodeAnalysis.VisualBasic.dll",
+                "Microsoft.CodeAnalysis.VisualBasic.Workspaces.dll",
+                "Microsoft.CodeAnalysis.Workspaces.dll",
```
brettsam, line 196: Why'd you have to expand these out? I see wildcards elsewhere in the list still...

azure-pipelines.yml
```diff
@@ -46,13 +48,130 @@ steps:
...
+  condition: and(succeeded(),or (eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Build.SourceBranch'], 'refs/heads/release/3.0')))
+- task: DotNetCoreCLI@2
+  inputs:
+    command: 'run'
+    workingDirectory: '.\build'
+    arguments: 'TestSignedArtifacts --signTest'
+  displayName: 'Verify signed binaries'
+  condition: and(succeeded(),or (eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Build.SourceBranch'], 'refs/heads/release/3.0')))
+- pwsh: |
+    .\generateMsiFiles.ps1 -MsiGenBranches @("v3.x") -CommitMessage '$(Build.SourceVersionMessage)'
```
brettsam, line 144: Should this pass `release/3.0` and not `v3.x`?

azure-pipelines.yml
```diff
@@ -46,13 +48,130 @@ steps:
...
+            "ToolVersion": "1.0"
+        }
+      ]
+    SessionTimeout: '60'
+    MaxConcurrency: '50'
+    MaxRetryAttempts: '5'
+  condition: and(succeeded(),or (eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Build.SourceBranch'], 'refs/heads/release/3.0')))
+- pwsh: |
+    .\repackageBinaries.ps1
+  displayName: Sign and repackage binaries
```
brettsam, line 129: This isn't actually signing is it? Maybe name should be "Repackage signed binaries"?

build/Settings.cs
soninaren: Wild card was covering a lot more files than actually needed. So expanded them. After doing further testing i found that more than 400 files being signed where as we only needed to sign 220 files. A lot of files were being re-signed. I added code to dynamically remove any files that do not need to be signed. But left this one as is.

PR #1951, opened 2020-04-29
Fixed issue where AzureWebJobsStorage wasn't being properly set in local.settings.json

### LINKED ISSUE #1927: Default AzureWebJobsStorage to  instead of {AzureWebJobsStorage}
`func new` creates a local.settings.json file with a `AzureWebJobsStorage` value of `AzureWebJobsStorage`. Can we default it to an empty string

gzuber:
Resolves #1927 

This looks like this is a bug that was introduced during a revert 2 years ago. Although I might be missing something since it's been around for so long unless it's pretty low impact. The code has been trying to set the value of AzureWebJobsStorage in local.settings.json, but it's been using the value it's trying to set as the value it's trying to replace (Setting y to a value of x by searching/replacing x).


gzuber:
cc @ahmelsayed because I was told you have context

PR #1981, opened 2020-05-12
Updated webhost

yojagad:


build/Settings.cs
```diff
@@ -223,6 +223,7 @@ public class SignInfo
                 "Microsoft.OData.Core.dll",
                 "Microsoft.OData.Edm.dll",
                 "Microsoft.Spatial.dll",
+				"Mono.Posix.NETStandard.dll",
```
soninaren, line 226: Nit: Fix the white space

PR #1958, opened 2020-04-30
Update PowerShellWorker to v3.0.279 and v3.0.283

AnatoliB:
PowerShell worker release notes:
- https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v3.0.279
- https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v3.0.283

PR #2007, opened 2020-05-26
Update Functions Runtime to 3.0.13614

ankitkumarr:


ankitkumarr:
@AnatoliB, fyi I updated the Powershell worker and added the new feed to core tools build pipeline.

AnatoliB:
> @AnatoliB, fyi I updated the Powershell worker and added the new feed to core tools build pipeline.

@ankitkumarr I see, thank you. I've been trying to do this in https://github.com/Azure/azure-functions-core-tools/pull/1958, but it keeps failing with a strange signing-related error (@soninaren is trying to figure out the reason). I would anticipate your PR to hit the same issue.

ankitkumarr:
Looks like this PR's CI failed a couple of times with `"System.IO.IOException: There is not enough space on the disk"`.

Azure DevOps agents have 10 GB available for us, and I won't be surprised if we are hitting that during the build when we have extracted and compressed versions of our artifacts. We may need to be careful with disk space.

@AnatoliB, I remember hearing that this Powershell worker binaries is also bigger compared to the past ones, is that correct? That would make sense why we start seeing this now.

cc: @ahmelsayed @soninaren

AnatoliB:
> @AnatoliB, I remember hearing that this Powershell worker binaries is also bigger compared to the past ones, is that correct? That would make sense why we start seeing this now.

The total size of these two nupkg increased by ~100 MB unzipped (from 354 MB to 452 MB). For the Functions Host package, we remove the unused runtimes, which brings the total down to ~150 MB. I don't know if we already do this for the core tools, but we should consider that if the size is the actual problem here.

ankitkumarr:
@AnatoliB, we had added code to manually do that during the build step here -- https://github.com/Azure/azure-functions-core-tools/blob/dev/build/BuildSteps.cs#L96-L131

Although, I am not sure how powershell 6 and powershell 7 are laid out, so we may need to update that code to do it for both of these versions?

ankitkumarr:
Ok, so I did a small hacky workaround that seemed to work for this PR. I reordered
https://github.com/Azure/azure-functions-core-tools/pull/2007/files#diff-5d321faaedf15437386ffed322280d2cR39 

to push `win` runtimes in the end.

I noticed that we [don't delete](https://github.com/Azure/azure-functions-core-tools/blob/v3.x/build/BuildSteps.cs#L497) binaries for `win` runtimes after creating zip packages as they are used by the MSI creation. So, by moving linux and osx to be packaged first, we clear off some space by deleting the artifacts, rather than having it accumulated if we start with `win`.

Anyway, this is definitely not a long term approach, and we should figure out a better way to manage space. But, for now, we should be unblocked.

Issue #1973, opened 2020-05-05
azure-functions-core-tools folder is created in current directory


anthonychu:
Core Tools uses [`LocalApplicationData`](https://github.com/Azure/azure-functions-core-tools/blob/16b36f985ecffdd40ab4fb974f5f576593980284/src/Azure.Functions.Cli/Common/Utilities.cs#L163) to create a folder for telemetry. This uses the `XDG_DATA_HOME` environment variable on Linux, and it's not always set.

If `XDG_DATA_HOME` isn't set, the folder will be created where `func` is first run.

/cc @johnpapa

johnpapa:
some notes ...

when I run `func start` in a functions app, it creates a `az func core tools/.telemetry` folder

when I stop the functions from running, it adds a file to that folder.

if I run `func start` in a non azure functions folder, it still adds the telemetry folder.

bpossolo:
How do we configure the location where this folder gets created?
I'm on mac os 10.15 catalina using the java runtime + mvn `azure-functions-maven-plugin`.
I tried setting the XDG_DATA_HOME before running `mvn azure-functions:run` but the folder keeps getting generated in my project root folder

PR #2012, opened 2020-05-27
Handle when directory is not available

ankitkumarr:
Fixes https://github.com/Azure/azure-functions-core-tools/issues/1973


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -160,7 +160,14 @@ internal static async Task<T> SafeExecution<T>(Func<Task<T>> action)
 
         internal static string EnsureCoreToolsLocalData()
         {
-            var localPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "azure-functions-core-tools");
+            var appDataDir = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+
+            if (string.IsNullOrEmpty(appDataDir))
+            {
+                throw new Exception("Cannot find the Local Application Data.");
```
ankitkumarr, line 167: throwing an exception here makes sense as this is an error condition. In case of telemetry, this will be swallowed by the caller.

ankitkumarr:
@anthonychu, were you able to repro this? I tried and for some reason I wasn't able to. If you had repro steps, I can quickly test this PR.

anthonychu:
Here's how to repro:

```bash
docker run --rm -it mcr.microsoft.com/azure-functions/dotnet:3.0-dotnet3-core-tools bash

# inside the container
mkdir test
cd test
func

# there should be an azure-functions-core-tools folder created
```

I couldn't figure out how to test the updated version though. Tried to checkout your branch in the container and build it. When I ran it with the same steps, indeed no folder was created. But if I run func that came with the container, it didn't create a folder either, so my test might be flawed.

ankitkumarr:
Thanks @anthonychu, your steps helped, and I was able to repro, and verify a fix.
I used an `ubuntu` image, installed core tools with apt, and managed to see the bug.

I then used the artifacts from my PR build and made sure the bug didn't repro. Then I set `XDG_DATA_HOME` and made sure the directory was created where I set it to. So, yeah we should be good with this PR.

Oh, and one thing, we only create that directory when core tools is installed via package managers. So if you just build my changes, telemetry will be off and you won't see it regardless. I had to manually turn it on to verify.

But in short, we should be good. Once approved, I will merge this and make another PR for v3.x

PR #2028, opened 2020-06-10
func new will now run init if the proper files are not present

### LINKED ISSUE #1259: func new should work with non initialized folders

gzuber:
Resolves #1259

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
@@ -10,6 +10,7 @@ using Azure.Functions.Cli.Interfaces;
 using Azure.Functions.Cli.Telemetry;
 using Colors.Net;
 using Fclp;
+using Microsoft.OData.UriParser;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.OutputTheme;
@@ -24,6 +25,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         private readonly ITemplatesManager _templatesManager;
         private readonly ISecretsManager _secretsManager;
 
+        private readonly InitAction _initAction;
+
         public string Language { get; set; }
         public string TemplateName { get; set; }
         public string FunctionName { get; set; }
@@ -33,6 +36,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         {
             _templatesManager = templatesManager;
             _secretsManager = secretsManager;
+            _initAction = new InitAction(_templatesManager, _secretsManager);
         }
 
         public override ICommandLineParserResult ParseArgs(string[] args)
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -75,6 +76,13 @@ public async override Task RunAsync()
             }
 
             var workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone;
+            if (!FileSystemHelpers.FileExists(Path.Combine(Environment.CurrentDirectory, "local.settings.json")))
+            {
+                // we're assuming "func init" has not been run
+                await InitApp(workerRuntime);
+                workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
```
gzuber, line 83: I'm concerned that this is the correct way to update `workerRuntime`. My worry is that `GlobalCoreToolsSettings` will continue to return `None` for runtime unless we run `Init` again, but as far as I can see we no longer have access to the `args` variable. Do we need to run `Init` with blank args or is that a concern?

src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -75,6 +76,17 @@ public async override Task RunAsync()
...
 
             var workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone;
+            if (!FileSystemHelpers.FileExists(Path.Combine(Environment.CurrentDirectory, "local.settings.json")))
+            {
+                // we're assuming "func init" has not been run
+                await InitApp(workerRuntime);
+
+                if (workerRuntime == WorkerRuntime.None)
+                {
+                    workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
```
gzuber, line 86: I'm concerned that this is the correct way to update workerRuntime. My worry is that GlobalCoreToolsSettings will continue to return None for runtime unless we run Init again, but as far as I can see we no longer have access to the args variable. Do we need to run Init with blank args or is that a concern?

ahmelsayed: so I think ideally what you want is if you detect that you need init, you can just call the init action.

How about creating the `InitAction` object in `ParseArgs()` calling parse args on the init object there, then only calling RunAsync here if needed? You might need a way to communicate back from init action to this action what the workerRuntime is. Does this make sense?

gzuber: Yeah I think that makes sense, and I made some changes that I think match up with what you're saying. I guess I was trying to make some guesses on the runtime just in case they passed in a language to skip the runtime selection screen if possible, but I think this is cleaner and more consistent (and I'm sure more error-resistant)

gzuber:
Unfortunately these changes mean that in order to create a function without having ran init first, the user _has_ to interact with the console other than if they pass in the `--csx` flag. Which doesn't seem like a big deal, but only one of my tests will work. Removing the others.

PR #2029, opened 2020-06-11
Update Functions Host version

ankitkumarr:


IanKemp:
Will this update the Tools to use Host version 3.0.13901 (https://github.com/Azure/azure-functions-host/releases/tag/v3.0.13901)?

Assuming this PR is accepted, when can we expect a new Tools build?

PR #1860, opened 2020-03-02
Update README with v3 apt

ankitkumarr:

PR #2030, opened 2020-06-12
Update README.md with apt installation command for v2

janpro0706:
It's has a typo or legacy instructions for apt installation in Ubuntu.
I thought that apt will implicitly install v2 without version number, but installed latest(v3).


janpro0706:
I found that 'azure-functions-core-tools-2' package was recently added(#1860), but maybe there's no updates in 'README.md'

PR #2092, opened 2020-07-15
Add MSI instructions

### LINKED ISSUE #2076: Add MSI instructions for installing Core Tools
Add instructions for installing Core Tools on Windows using MSI.

@ankitkumarr @fabiocav Are we ready to add this as the recommended method to acquire Core Tools on Windows?

anthonychu:
Also list v3 instructions ahead of v2

Resolves #2076 

/cc @ankitkumarr

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,14 +48,14 @@ npm i -g azure-functions-core-tools@2 --unsafe-perm true
 choco install azure-functions-core-tools-3
 ```
 
-##### v2
-```bash
-choco install azure-functions-core-tools-2
-```
-
 *Notice: To debug functions under vscode, the 64-bit version is required*
 ```bash
 choco install azure-functions-core-tools-3 --params "'/x64'"
+```
+
+##### v2
+```bash
+choco install azure-functions-core-tools-2
 ```
 
 ### Mac
--- a/README.md
+++ b/README.md
@@ -22,14 +22,14 @@ The Azure Functions Core Tools provide a local development experience for creati
 
 ### Windows
 
-To download and install with MSI:
+#### To download and install with MSI:
 
 **v3**
 
-- [Windows 64-bit](https://go.microsoft.com/fwlink/?linkid=2135274) - Required for VS Code debugging
+- [Windows 64-bit](https://go.microsoft.com/fwlink/?linkid=2135274) (Required for VS Code debugging)
 - [Windows 32-bit](https://go.microsoft.com/fwlink/?linkid=2135275)
 
-To install with npm:
+#### To install with npm:
 
 **v3**
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,53 +16,62 @@ The Azure Functions Core Tools provide a local development experience for creati
...
-To install runtime with npm:
+To download and install with MSI:
 
-**v2**
-```bash
-npm i -g azure-functions-core-tools@2 --unsafe-perm true
-```
+**v3**
+
+- [Windows 64-bit](https://go.microsoft.com/fwlink/?linkid=2135274) - Required for VS Code debugging
```
ankitkumarr, line 29: I think we should move `64 bit required for VS Code debugging` to a note at the top, as we also have the same note next to choco. And looking at it here, it may give the impression you need to install via MSI 64 bit for VS Code debugging (over say `npm` or `apt` or `choco`).

anthonychu: Do we have anyone who would need x86? Wonder if our instructions should just be x64.

ankitkumarr: hmm. I think in Azure we run 32 bit Host binaries, so to avoid disparity I think it makes sense to have x86 instructions.

fabiocav: Yes, we should have both here.

README.md
```diff
@@ -16,53 +16,62 @@ The Azure Functions Core Tools provide a local development experience for creati
...
 
-**v3**: Self-contained cross-platform package
+**v3**: (v3.x branch): Self-contained cross-platform package **(recommended)**
 
 ## Installing
 
 ### Windows
 
-To install runtime with npm:
+To download and install with MSI:
```
fabiocav, line 25: I realize this takes the change beyond the initial scope you were planning, but can we make these headings? The document options are a bit difficult to read.

README.md
```diff
@@ -16,53 +16,62 @@ The Azure Functions Core Tools provide a local development experience for creati
...
-"**v3**"
+
+**v2**
 ```bash
-choco install azure-functions-core-tools-3
+choco install azure-functions-core-tools-2
 ```
 
-#### Notice: To debug functions under vscode, x64 bitness is required
+*Notice: To debug functions under vscode, x64 bitness is required*
```
fabiocav, line 56: nit: Should the verbiage be:
`Notice: To debug functions under vscode, the 64-bit version is required`

README.md
anthonychu: Updated. Good suggestion

README.md
anthonychu: Updated the message to clarify that VS Code requires 64-bit version. Should be clearer now. If we say upfront that 64-bit is needed for VS Code debugging, people might miss it, or they see it and wonder how to get 64-bit from npm.

anthonychu:
@fabiocav @ankitkumarr Good to merge?

fabiocav:
👍

PR #2093, opened 2020-07-15
func init and new for custom handlers

### LINKED ISSUE #2052: func init for Custom Handlers
Parent: #1972 

Add `custom` as a worker runtime stack option.

```
func init --worker-runtime custom
```

If `custom` is selected, add empty `customHandler` section to host.json:

```json
{
  "version": "2.0",
  "logging": {
    "applicationInsights": {
      "samplingSettings": {
        "isEnabled": true,
        "excludedTypes": "Request"
      }
    }
  },
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[1.*, 2.0.0)"
  },
  "customHandler": {
    "description": {
      "defaultExecutablePath": "",
      "workingDirectory": "",
      "arguments": []
    },
    "enableForwardingHttpRequest": false
  }
}

```

@pragnagopa Is the schema correct? (forgot if we decided to remove `type`). Thanks.

yojagad:
Fixes #2052 
Corresponding template PR: https://github.com/Azure/azure-functions-templates/pull/964. Currently blocked until template is released which is tracked by this work item https://github.com/Azure/azure-functions-templates/issues/965.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -6,12 +6,11 @@ using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Extensions;
 using Azure.Functions.Cli.Helpers;
 using Azure.Functions.Cli.Interfaces;
-using Azure.Functions.Cli.Telemetry;
 using Colors.Net;
 using Fclp;
-using Microsoft.Azure.WebJobs.Script;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.OutputTheme;
@@ -424,33 +423,20 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
             if (workerRuntime == Helpers.WorkerRuntime.powershell && managedDependenciesOption)
             {
-                hostJsonContent = await AddManagedDependencyConfig(hostJsonContent);
+                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.ManagedDependenciesConfig, Constants.ManagedDependencyConfigPropertyName);
             }
 
             if (extensionBundle)
             {
-                hostJsonContent = await AddBundleConfig(hostJsonContent);
+                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.BundleConfig, Constants.ExtensionBundleConfigPropertyName);
+            }
+
+            if(workerRuntime == Helpers.WorkerRuntime.custom)
+            {
+                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.CustomHandlerConfig, Constants.CustomHandlerPropertyName);
             }
 
             await WriteFiles("host.json", hostJsonContent);
-        }
-
-        private static async Task<string> AddBundleConfig(string hostJsonContent)
-        {
-            var hostJsonObj = JsonConvert.DeserializeObject<JObject>(hostJsonContent);
-            var bundleConfigContent = await StaticResources.BundleConfig;
-            var bundleConfig = JsonConvert.DeserializeObject<JToken>(bundleConfigContent);
-            hostJsonObj.Add(Constants.ExtensionBundleConfigPropertyName, bundleConfig);
-            return JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
-        }
-
-        private static async Task<string> AddManagedDependencyConfig(string hostJsonContent)
-        {
-            var hostJsonObj = JsonConvert.DeserializeObject<JObject>(hostJsonContent);
-            var managedDependenciesConfigContent = await StaticResources.ManagedDependenciesConfig;
-            var managedDependenciesConfig = JsonConvert.DeserializeObject<JToken>(managedDependenciesConfigContent);
-            hostJsonObj.Add(Constants.ManagedDependencyConfigPropertyName, managedDependenciesConfig);
-            return JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
         }
 
         private static string AddWorkerVersion(string localSettingsContent, string workerVersion)
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -323,6 +323,10 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 await WriteFiles("Dockerfile", await StaticResources.DockerfilePowershell);
             }
+            else if(workerRuntime == Helpers.WorkerRuntime.custom)
+            {
+                await WriteFiles("Dockerfile", await StaticResources.DockerfileCustom);
+            }
             else if (workerRuntime == Helpers.WorkerRuntime.None)
             {
                 throw new CliException("Can't find WorkerRuntime None");
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -423,20 +423,20 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
             if (workerRuntime == Helpers.WorkerRuntime.powershell && managedDependenciesOption)
             {
-                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.ManagedDependenciesConfig, Constants.ManagedDependencyConfigPropertyName);
+                hostJsonContent = await hostJsonContent.AppendContent(Constants.ManagedDependencyConfigPropertyName, StaticResources.ManagedDependenciesConfig);
             }
 
             if (extensionBundle)
             {
-                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.BundleConfig, Constants.ExtensionBundleConfigPropertyName);
+                hostJsonContent = await hostJsonContent.AppendContent(Constants.ExtensionBundleConfigPropertyName, StaticResources.BundleConfig);
             }
 
             if(workerRuntime == Helpers.WorkerRuntime.custom)
             {
-                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.CustomHandlerConfig, Constants.CustomHandlerPropertyName);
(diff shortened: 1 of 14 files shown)


src/Azure.Functions.Cli/Common/TemplatesManager.cs
```diff
@@ -35,16 +34,18 @@ private static async Task<IEnumerable<Template>> GetTemplates()
...
                 var contentProvider = ExtensionBundleHelper.GetExtensionBundleContentProvider();
                 templatesJson = await contentProvider.GetTemplates();
             }
-            else
+
+            if(string.IsNullOrEmpty(templatesJson))
             {
                 templatesJson = GetTemplatesJson();
             }
+
```
yojagad, line 48: @soninaren Is this an acceptable workaround for bundles vs non-bundles templates ?

src/Azure.Functions.Cli/StaticResources/customHandler.json
```diff
@@ -0,0 +1,8 @@
+﻿{
+  "description": {
+    "defaultExecutablePath": "",
+    "workingDirectory": "",
+    "arguments": [ ]
+  },
+  "enableForwardingHttpRequest": true
```
anthonychu, line 7: Wonder if we should default `enableForwardingHttpRequest` to `false`.

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -39,7 +41,8 @@ public static class WorkerRuntimeLanguageHelper
             { WorkerRuntime.dotnet, Constants.Languages.CSharp },
             { WorkerRuntime.node, Constants.Languages.JavaScript },
             { WorkerRuntime.python, Constants.Languages.Python },
-            { WorkerRuntime.powershell, Constants.Languages.Powershell }
+            { WorkerRuntime.powershell, Constants.Languages.Powershell },
+            { WorkerRuntime.custom, Constants.Languages.Custom },
```
anthonychu, line 45: Can you confirm that this will also set `FUNCTIONS_WORKER_RUNTIME` to `custom`?

src/Azure.Functions.Cli/StaticResources/customHandler.json
```diff
@@ -0,0 +1,8 @@
+﻿{
```
soninaren, line 1: When you do `func init` with custom handler the config here would be added to host.json. If someone runs `func start` after that without making any changes, the function app would run into errors since the config is incomplete.

Instead we should add a wizard similar to node, where once you select node you further prompter to select language ` JavaScript or TypeScript` we should prompt for the required fields for the customer to input on the init command.

src/Azure.Functions.Cli/Common/TemplatesManager.cs
soninaren: Not sure what changed here. Instead of else it would go in the if block the endresult here still remains the same.

Did you want to merge the custom handler templates to templates offered by bundles?  The way it is set up in the PR right now, the custom handler template would only show up for customer who does not have bundles configured.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -424,33 +423,20 @@ private static async Task WriteHostJson(WorkerRuntime workerRuntime, bool manage
...
+            }
 
-        private static async Task<string> AddManagedDependencyConfig(string hostJsonContent)
-        {
-            var hostJsonObj = JsonConvert.DeserializeObject<JObject>(hostJsonContent);
-            var managedDependenciesConfigContent = await StaticResources.ManagedDependenciesConfig;
-            var managedDependenciesConfig = JsonConvert.DeserializeObject<JToken>(managedDependenciesConfigContent);
-            hostJsonObj.Add(Constants.ManagedDependencyConfigPropertyName, managedDependenciesConfig);
-            return JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
+            await WriteFiles("host.json", hostJsonContent);
```
pragnagopa, line 439: Nit: Move `host.json` to constants

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
```diff
@@ -424,33 +423,20 @@ private static async Task WriteHostJson(WorkerRuntime workerRuntime, bool manage
...
-        {
-            var hostJsonObj = JsonConvert.DeserializeObject<JObject>(hostJsonContent);
-            var bundleConfigContent = await StaticResources.BundleConfig;
-            var bundleConfig = JsonConvert.DeserializeObject<JToken>(bundleConfigContent);
-            hostJsonObj.Add(Constants.ExtensionBundleConfigPropertyName, bundleConfig);
-            return JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
-        }
+            if(workerRuntime == Helpers.WorkerRuntime.custom)
+            {
+                hostJsonContent = await hostJsonContent.AppendContent(StaticResources.CustomHandlerConfig, Constants.CustomHandlerPropertyName);
```
pragnagopa, line 436: nit: `Helpers.WorkerRuntime.Custom` ? casing on custom

src/Azure.Functions.Cli/StaticResources/customHandler.json
pragnagopa: by default `enableForwardingHttpRequest` is `false`. Ok to omit this altogether?

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
```diff
@@ -71,6 +71,23 @@ public async Task create_template_function_using_alias()
...
+            await CliTester.Run(new RunConfiguration
+            {
+                Commands = new[]
+                {
+                    "init . --worker-runtime custom",
+                    "new --template \"Http Trigger\" --name testfunc"
+                },
+                OutputContains = new[]
+                {
+                    "The function \"testfunc\" was created successfully from the \"Http Trigger\" template."
```
pragnagopa, line 86: should this verify function.json content?

src/Azure.Functions.Cli/StaticResources/StaticResources.cs
```diff
@@ -51,6 +51,8 @@ private static async Task<string> GetValue(string name)
 
         public static Task<string> BundleConfig => GetValue("bundleConfig.json");
 
+        public static Task<string> CustomHandlerConfig => GetValue("customHandler.json");
```
pragnagopa, line 54: nit: Rename  `customHandler.json` to `customHandlerConfig.json`

src/Azure.Functions.Cli/StaticResources/customHandler.json
anthonychu: IMO we should keep it there and set to `false`. It makes it easy to change, and also makes the setting discoverable.

I was actually wondering if we should add the extensions.http.routePrefix by default as well so custom handlers customers can change it from `"api"` to `""` easily. I think that might be a bit much though.

pragnagopa: @yojagad - let's set `enableForwardingHttpRequest` explicilty to `false`

Lets skip introducing route prefix here as it is specific to httpTrigger and host.json for other languages do not have it either.

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
yojagad: You mean in local.settings.json, right ? Yeah it had `FUNCTIONS_WORKER_RUNTIME` set to `custom`

src/Azure.Functions.Cli/Common/TemplatesManager.cs
yojagad: Discussed this offline. This does a filtration based on language which is why this works right now. Will change the code to make this more explicit.

src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
yojagad: The rest of the values in the enum are fully lowercased so just stuck with that.

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
yojagad: I'm just following the pattern , I'll explore if there's a way to verify content once templates are released.

src/Azure.Functions.Cli/StaticResources/customHandler.json
yojagad: Doing a func start even after doing func new, they would still run into errors unless they configured the right parameters, so not sure if it's worth adding a wizard like experience. I'll let @pragnagopa or @anthonychu chime in.

src/Azure.Functions.Cli/StaticResources/customHandler.json
yojagad: Based on the earlier conversation we had around supportability, etc, I removed this setting altogether. Let me know if you still think I should add it back with it set to false

anthonychu: I’m okay with taking it out for now.

src/Azure.Functions.Cli/StaticResources/Dockerfile.custom
```diff
@@ -0,0 +1,6 @@
+# To enable ssh & remote debugging on app service change the base image to the one below
+FROM mcr.microsoft.com/azure-functions/dotnet:3.0-appservice 
+ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
+    AzureFunctionsJobHost__Logging__Console__IsEnabled=true
```
pragnagopa, line 4: @ankitkumarr / @divyagandhii  - do you know if it is requrired to set `AzureFunctionsJobHost__Logging__Console__IsEnabled=true` ? when users deploy to azure, this causes duplicated logs in Kusto

ankitkumarr: I would imagine this is required when running or locally or some non App Service environment such as Kubernetes.

Wondering if adding a comment requesting to remove when publishing in Azure would be reasonable.

@ahmelsayed might know better.

yojagad: A separate tracking issue will be opened for this one.

yojagad:
Removing @soninaren since he's oof for a few weeks. I believe I addressed all of his concerns, but if he does have any more, I'll address them as they come.

PR #2107, opened 2020-07-22
changing the default bundle download path for osx

### LINKED ISSUE #1999: Automatically recover when bundles folder is corrupt
Sometimes, the bundles folder can become corrupt and the solution is to delete the folder and restart the app. Let's investigate how we can automatically recover for this situation.

Some instances where this was a problem:
* #1924 
* #1995

soninaren:
Changing the default bundle download path for OSX. 

fixes #1999

### DIFF
--- a/src/Azure.Functions.Cli/Common/Constants.cs
+++ b/src/Azure.Functions.Cli/Common/Constants.cs
@@ -46,6 +46,8 @@ namespace Azure.Functions.Cli.Common
         public const string DefaultManagementURL = "https://management.azure.com/";
         public const string AzureManagementAccessToken = "AZURE_MANAGEMENT_ACCESS_TOKEN";
         public const string ExtensionBundleConfigPropertyName = "extensionBundle";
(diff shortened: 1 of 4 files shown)


src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -46,6 +46,8 @@ internal static class Constants
         public const string DefaultManagementURL = "https://management.azure.com/";
         public const string AzureManagementAccessToken = "AZURE_MANAGEMENT_ACCESS_TOKEN";
         public const string ExtensionBundleConfigPropertyName = "extensionBundle";
+        public const string OSXCoreToolsTempDirectoryName = ".azure-functions-core-tools";
+        public const string OSXRootPath = "~/";
```
ankitkumarr, line 50: Can probably make this `~`. But either way should work (I think?).

soninaren: Yeah i checked both notation as same. Keep this as is.

PR #2114, opened 2020-07-25
Adding AuthLevel to HttpTrigger

### LINKED ISSUE #375: Add parameter for authorization level to "func new"
I believe it would be handy to have an `--authlevel` to control `authLevel` as it's written to `function.json`.

For example:

```bash
$ func new --language JavaScript --template HttpTrigger \
           --name HttpTriggerJavaScript --authlevel anonymous
```

produces

`function.json`:
```
{
  "bindings": [
    {
      "authLevel": "anonymous",
 ...
```

An extra neat touch would be to mention  _function and admin level have no effect when running locally_ in the usage output (when running `func` with no parameter):

```
$ func
                  %%%%%%
                 %%%%%%
            @   %%%%%%    @
          @@   %%%%%%      @@
       @@@    %%%%%%%%%%%    @@@
     @@      %%%%%%%%%%        @@
       @@         %%%%       @@
         @@      %%%       @@
           @@    %%      @@
                %%
                %

Azure Functions Core Tools (2.0.1-beta.22)
Function Runtime Version: 2.0.11415.0
Usage: func [context] [context] <action> [-/--options]

...
Actions:
new     Create a new function from a template. Aliases: new, create
    --language [-l]    Template programming language, such as C#, F#, JavaScript, etc.
    --template [-t]    Template name
    --name [-n]        Function name
    --authlevel [-a]   Authorization level (anonymous, function, admin)
                       [!] Note that authorization level has no effect when
                       running the function locally.
```

soninaren:
fixes #375

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
@@ -10,6 +10,7 @@ using Azure.Functions.Cli.Interfaces;
 using Azure.Functions.Cli.Telemetry;
 using Colors.Net;
 using Fclp;
+using Microsoft.Azure.WebJobs.Extensions.Http;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.Constants;
@@ -29,7 +30,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         public string TemplateName { get; set; }
         public string FunctionName { get; set; }
         public bool Csx { get; set; }
-        public HttpAuthorizationLevel? AuthorizationLevel { get; set; }
+        public AuthorizationLevel? AuthorizationLevel { get; set; }
 
         Lazy<IEnumerable<Template>> _templates;
 
@@ -59,7 +60,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 .Callback(n => FunctionName = n);
 
             Parser
-                .Setup<HttpAuthorizationLevel?>('a', "authorizationLevel")
+                .Setup<AuthorizationLevel?>('a', "authlevel")
                 .WithDescription("Authorization level is applicable to templates that use Http trigger, Allowed values: [function, anonymous, admin]. Authorization level is not enforced when running functions from core tools")
                 .Callback(a => AuthorizationLevel = a);
 
--- a/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
@@ -12,6 +12,7 @@ using Colors.Net;
 using Fclp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using static Azure.Functions.Cli.Common.Constants;
 using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Actions.LocalActions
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -52,6 +58,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .WithDescription("Function name")
                 .Callback(n => FunctionName = n);
 
+            Parser
+                .Setup<HttpAuthorizationLevel?>('a', "authorizationLevel")
```
pragnagopa, line 62: `authlevel` ? as that matches function.json schema

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -233,5 +235,12 @@ public static class StaticResourcesNames
                         Name = "Microsoft.Azure.WebJobs.Extensions.Twilio" }
                     }
                 });
+
+        public enum HttpAuthorizationLevel
+        {
```
pragnagopa, line 240: Instead of defining another enum, is it possible to reuse : https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/AuthorizationLevel.cs --> this will ensure we keep the code in sync

src/Azure.Functions.Cli/Common/ExtensionPackage.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Azure.Functions.Cli.Common
 {
-    class ExtensionPackage
+    public class ExtensionPackage
```
pragnagopa, line 7: intentionally making this public?

src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -21,18 +22,23 @@ namespace Azure.Functions.Cli.Actions.LocalActions
...
     {
-        private readonly ITemplatesManager _templatesManager;
+        private ITemplatesManager _templatesManager;
         private readonly ISecretsManager _secretsManager;
 
         public string Language { get; set; }
         public string TemplateName { get; set; }
         public string FunctionName { get; set; }
         public bool Csx { get; set; }
+        public HttpAuthorizationLevel? AuthorizationLevel { get; set; }
```
yojagad, line 32: > public HttpAuthorizationLevel? AuthorizationLevel { get; set; } [](start = 7, length = 64)

Is there a scenario where this can/should be null ?

soninaren: The main reason for us to have a null here is because the `authlevel` does not apply to other triggers. The `func new` and its parameters until now were applicable to all the trigger. So if the user decides to use this for other templates that does not accept the authLevel parameter, I wanted to provide an error message. Without a null value i would have no way to differentiate between no value passed vs default value without making it nullable.

src/Azure.Functions.Cli/Common/ExtensionPackage.cs
soninaren: yeah was needed for the enum. may end up removing this after switching to enum from webjobs sdk.

src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
soninaren: sure i can do that.

src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs
yojagad: I was thinking along the lines of making default to be "None" and if for other triggers , if it's anything other than "None" , an error could be displayed. But then this means httpTrigger should have some code changes where default cannot be "None". 

I don't really have much of a preference, either ways.

soninaren: Ahh i thought you were saying use one of  the values as default instead. I used "unassigned" as a default value to signal null equivalent. But then I found out that i would need to add custom parsing logic and validation logic. Adding nullable allows me use default validation from the framework. Also now I am using the Enum from webjobs sdk, So i can't add a default anyway.

soninaren: appreciate the feedback :)

src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -52,6 +52,8 @@ internal static class Constants
         public const string OSXRootPath = "~/";
         public const string ManagedDependencyConfigPropertyName = "managedDependency";
         public const string PowerShellWorkerDefaultVersion = "~6";
+        public const string AuthLevelErrorMessage = "Unable to configure Authorization level, the selected template does not use Http Trigger";
```
pragnagopa, line 55: Super Nit: `Unable to configure Authorization level.` Full stop, instead of comma :)

soninaren: updated

PR #2116, opened 2020-07-27
Warn users to stop using azure-functions-worker in requirements.txt (v2)

Hazhzeng:
1. Add a message for warning users to not include azure-functions-worker in requirements.txt.
```
# Do not include azure-functions-worker as it may conflict with the Azure Functions platform

azure-functions
```
2. Add a parser of requirements.txt in RequirementsTxtParser.cs
3. Add a warning message when `func azure functionapp publish` with `azure-functions-worker` in requirements.txt

### DIFF
--- a/src/Azure.Functions.Cli/Common/RequirementsTxtParser.cs
+++ b/src/Azure.Functions.Cli/Common/RequirementsTxtParser.cs
@@ -37,17 +37,17 @@ namespace Azure.Functions.Cli.Common
                 {
                     GroupCollection groups = match.Groups;
 
-                    groups.TryGetValue("name", out Group packageName);
-                    groups.TryGetValue("spec", out Group packageSpec);
-                    groups.TryGetValue("envmarker", out Group packageEnvMarker);
-                    groups.TryGetValue("directref", out Group packageDirectRef);
+                    Group packageName = groups["name"];
+                    Group packageSpec = groups["spec"];
+                    Group packageEnvMarker = groups["envmarker"];
+                    Group packageDirectRef = groups["directref"];
 
                     packages.Add(new PythonPackage()
                     {
-                        Name = packageName.Value.ToLower().Replace('_', '-').Replace('.', '-').Trim(),
-                        Specification = packageSpec?.Value?.Trim() ?? string.Empty,
-                        EnvironmentMarkers = packageEnvMarker?.Value?.Trim() ?? string.Empty,
-                        DirectReference = packageDirectRef?.Value?.Trim() ?? string.Empty
+                        Name = packageName.Success ? packageName.Value.ToLower().Replace('_', '-').Replace('.', '-').Trim() : string.Empty,
+                        Specification = packageSpec.Success ? packageSpec.Value.Trim() : string.Empty,
+                        EnvironmentMarkers = packageEnvMarker.Success ? packageEnvMarker.Value.Trim() : string.Empty,
+                        DirectReference = packageDirectRef.Success ? packageDirectRef.Value.Trim() : string.Empty
                     });
                 }
             });
--- a/src/Azure.Functions.Cli/Common/RequirementsTxtParser.cs
+++ b/src/Azure.Functions.Cli/Common/RequirementsTxtParser.cs
@@ -1,0 +1,58 @@
+﻿using Azure.Functions.Cli.Common;
+using System.Linq;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Text.RegularExpressions;
+using System.Collections.Concurrent;
+
+namespace Azure.Functions.Cli.Common
+{
+    // Requirements.txt PEP https://www.python.org/dev/peps/pep-0440/
+    public static class RequirementsTxtParser
+    {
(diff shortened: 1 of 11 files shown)


ankitkumarr:
I agree with the sentiment here that we should discourage users from having `azure-functions-worker` in the `requirements.txt`. But I don't think users manually add that. I think that they end up with this if they are using an older version of the core tools (when we used to add it for them).

Changing this in `func init` step doesn't solve the issue for users who still have that. If users are creating a new project with new core tools, I don't think they will have this anyway. And personally, I would avoid adding strong message starting with `"DO NOT" ` when users are starting and creating the functions project and just starting out.

I suggest that during the publish step, we add a validation such that if user's `requirements.txt` has `azure-functions-worker`, we warn them clearly. This also takes care of users who already have that in their `requirements.txt`, and avoids even mentioning `azure-functions-worker` unless necessary.

What are your thoughts?

anthonychu:
I agree we should warn, or even throw error in func start and publish.

Hazhzeng:
Thanks for the suggestions.

Since the old core tools has `azure-functions-worker==1.0.0b3`, we do still see customer manually adding the latest `azure-functions-worker` library into requirements.txt.

I partially agree with this idea but now we need to add the mechanism in core tools to parse requirements.txt, although it makes it cleaner, but still have a potential of breaking existing function apps. My intention is, instead of forcing them to change the existing apps, we should prevent the users from adding azure-functions-worker into their new project.

Hi @vrdmr, any suggestion on this one?

test/Azure.Functions.Cli.Tests/E2E/InitTests.cs
```diff
@@ -508,5 +508,35 @@ public Task init_python_app_twice()
...
+                {
+                    new FileResult
+                    {
+                        Name = "requirements.txt",
+                        ContentContains = new []
+                        {
+                            "# DO NOT include azure-functions-worker in this file",
+                            "# The Python Worker is managed by Azure Functions platform",
+                            "# Manually managing azure-functions-worker may cause unexpected issues",
+                            "azure-functions"
```
vrdmr, line 535: Could you please change the text to -

```python
# Do NOT include azure-functions-worker in this file
# The Python Worker is managed by Azure Functions platform
# Manually managing azure-functions-worker may cause unexpected issues
#
# If this file already contains azure-functions-worker dependency, kindly remove it
```

@anthonychu @ankitkumarr

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
```diff
@@ -64,11 +64,13 @@ public static async Task EnsureVirtualEnvrionmentIgnored()
...
-        private static void CreateRequirements()
+        private async static Task CreateRequirements()
         {
             if (!FileSystemHelpers.FileExists(Constants.RequirementsTxt))
             {
-                FileSystemHelpers.WriteAllTextToFile(Constants.RequirementsTxt, Constants.PythonFunctionsLibrary);
+                ColoredConsole.WriteLine($"Writing {Constants.RequirementsTxt}");
+                string requirementsTxtContent = await StaticResources.PythonRequirementsTxt;
+                await FileSystemHelpers.WriteAllTextToFileAsync(Constants.RequirementsTxt, requirementsTxtContent);
             }
```
vrdmr, line 74: Could you also add functionality in `func host start` for python - which reads the `requirments.txt` and checks for not required dependencies - and throw a warning suggesting the user to remove it?

test/Azure.Functions.Cli.Tests/E2E/InitTests.cs
vrdmr: @anthonychu - Could you please help writing a friendlier text?

vrdmr:
FYI @anirudhgarg.

test/Azure.Functions.Cli.Tests/E2E/InitTests.cs
anthonychu: Perhaps shorten it to something like “Do not include azure-functions-python-worker as it may conflict with the Azure Functions platform.”

src/Azure.Functions.Cli/Helpers/PythonHelpers.cs
Hazhzeng: Instead of adding it in `func host start`, I think `func azure functionapp publish` is a better place to warn our users, since requirements.txt is resolved at the time of deployment.

src/Azure.Functions.Cli/Common/RequirementsTxtParser.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Azure.Functions.Cli.Common
+{
+    // Requirements.txt PEP https://www.python.org/dev/peps/pep-0440/
+    public static class RequirementsTxtParser
+    {
+        public static async Task<List<PythonPackage>> ParseRequirementsTxtFile(string functionAppRoot)
+        {
+            // Check if requirements.txt exist
+            string requirementsTxtPath = $"{functionAppRoot}/{Constants.RequirementsTxt}";
```
ankitkumarr, line 16: Maybe use `Path.Join` here.

PR #2135, opened 2020-08-05
Force log-white listed logs

### LINKED ISSUE #2110: Force Log when host/worker is ready for VS Code debugger to attach  when loglevel is set to None
After discussing with @pragnagopa, it sounds like the initial change in Azure/azure-functions-host#6230 is not quite sufficient for the purposes of VS Code attaching the debugger reliably.

There are 2 issues in particular:
- In-proc .NET apps will not print the new message
- The message is printed with log level Information. If default log levels are changed to Warning, Error, or None, the message will not print

We need this message to be printed regardless of stack or log level. Also suggest updating the message to something more generic and doesn't mention "worker process". (something like "Function app is ready"?)

Also needed for this: https://github.com/Azure/azure-functions-core-tools/issues/1131

@ericjizbaMSFT

pragnagopa:
Fixes #2110


pragnagopa:
@brettsam - please take a look. If the approach looks ok, I will add unit tests. Thanks!

pragnagopa:
cc @anthonychu fyi

PR #2139, opened 2020-08-05
Less verbose logs on func start

pragnagopa:
Fixes #2141 Fixes #2143  Fixes #2110

This PR:
- Added `--verbose` flag. Running with `--verbose` flag will give users existing logging behavior
- If running in CI environment - https://github.com/watson/ci-info/blob/master/index.js#L52-L59, `--verbose` flag is set.
- Displays all the functions  (incorporated code from @anthonychu)
- Strips out date from log messages.
- By default `func host start` will have less verbose system logs:
    - Default log level for system logs is `Warning`
    - Deafult log level for user logs is `Information`
- If `Default` log level is specified in `host.json` - both user logs and system logs level is set to the value specified in `host.json`
- If `host.json` has default loglevel set to `None` - Only whitelisted logs that **VS Code** uses will be logged
- If `host.json` lists any category then defaults do not apply

Sample output on func host start with default host.json

```
func.exe host start
Azure Functions Core Tools (3.0.1 Commit hash: N/A)
Function Runtime Version: 3.0.1.0
Hosting environment: Production
Content root path: C:\pgopa\vsCode\jsDebug
Now listening on: http://0.0.0.0:7071
Application started. Press Ctrl+C to shut down.

Functions:

        HttpTrigger: [GET,POST] http://localhost:7071/api/HttpTrigger

For detailed output, run func with --verbose flag.
Worker process started and initialized.
Host lock lease acquired by instance ID '00000000000000000000000022DF9FE8'.
JavaScript HTTP trigger function processed a request.
```

Sample output on func host start with following host.json

```
{
	"version": "2.0",
	"logging": {
		"LogLevel": {
			"Function.HttpTrigger": "Trace",
			"Microsoft.Azure.WebJobs.Script.Description.WorkerFunctionInvoker": "Trace",
			"Microsoft.Azure.WebJobs.Script.WebHost.Middleware.SystemTraceMiddleware": "Information"
		}
	}
}
```

**NOTE: To see middleware tracing user explicitly needs to add that category in host.json**
Output :

```
Azure Functions Core Tools (3.0.1 Commit hash: N/A)
Function Runtime Version: 3.0.1.0
Hosting environment: Production
Content root path: C:\pgopa\vsCode\jsDebug
Now listening on: http://0.0.0.0:7071
Application started. Press Ctrl+C to shut down.

Functions:

        HttpTrigger: [GET,POST] http://localhost:7071/api/HttpTrigger

For detailed output, run func with --verbose flag.
[...]

PR #2145, opened 2020-08-06
Force logs only when Default LogLevel is None

### LINKED ISSUE #2110: Force Log when host/worker is ready for VS Code debugger to attach  when loglevel is set to None

pragnagopa:
Fixes #2110 

host.json has default loglevel set to `None`

```
{
  "version": "2.0",
   "Logging": {
      "LogLevel": {
		"Default": "None"
        }
   }
}
```

then add filter in JobHost logging builder to ensure all logs flow through and inspect each log. If Log starts with whitelisted prefix then log console otherwise do not log.

This keeps the existing behavior but allows core tools to *Force logs* ignoring the default log level

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -6,27 +6,19 @@ using System.Linq;
 using System.Net;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
-using Azure.Functions.Cli.Actions.HostActions.WebHost.Security;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Diagnostics;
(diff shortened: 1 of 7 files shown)


pragnagopa:
Closing in favor of https://github.com/Azure/azure-functions-core-tools/pull/2139

PR #2160, opened 2020-08-18
changing the default bundle download path

### LINKED ISSUE #1999: Automatically recover when bundles folder is corrupt

soninaren:
closes #1999

### DIFF
--- a/src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
+++ b/src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
@@ -1,12 +1,6 @@
 ﻿using System.Collections.Generic;
-using System.IO;
-using Azure.Functions.Cli.Common;
 using Microsoft.Azure.WebJobs.Script;
-using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Logging;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Azure.Functions.Cli.ExtensionBundle
 {
@@ -26,7 +20,7 @@ namespace Azure.Functions.Cli.ExtensionBundle
             {
                 builder.AddInMemoryCollection(new Dictionary<string, string>
                 {
-                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", Path.Combine(Path.GetTempPath(), "Functions", ScriptConstants.ExtensionBundleDirectory, bundleId)},
+                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", ExtensionBundleHelper.GetBundleDownloadPath(bundleId) },
                     { "AzureFunctionsJobHost:extensionBundle:ensureLatest", "true"}
                 });
             }
--- a/src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleHelper.cs
+++ b/src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleHelper.cs
@@ -6,9 +6,8 @@ using Microsoft.Azure.WebJobs.Script.ExtensionBundle;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging.Abstractions;
 using System;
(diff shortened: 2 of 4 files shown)


soninaren:
@pragnagopa can you take a look?

src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -6,9 +6,8 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging.Abstractions;
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Text;
+using static System.Environment;
```
pragnagopa, line 10: This is used for SpecialFolder? If yes, please use Environment.SpecialFolder to be more explicit

src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -42,5 +41,10 @@ public static ExtensionBundleContentProvider GetExtensionBundleContentProvider()
         {
             return new ExtensionBundleContentProvider(GetExtensionBundleManager(), NullLogger<ExtensionBundleContentProvider>.Instance);
         }
+
+        public static string GetBundleDownloadPath(string bundleId)
+        {
```
pragnagopa, line 46: Add unit test

src/Azure.Functions.Cli/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
```diff
@@ -26,7 +20,7 @@ public void Configure(IConfigurationBuilder builder)
             {
                 builder.AddInMemoryCollection(new Dictionary<string, string>
                 {
-                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", Path.Combine(Path.GetTempPath(), "Functions", ScriptConstants.ExtensionBundleDirectory, bundleId)},
+                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", ExtensionBundleHelper.GetBundleDownloadPath(bundleId) },
```
pragnagopa, line 23: should this be double underscore instead of `:`

soninaren: no `:` is the delimiter used by dotnet runtime to specify hierarchy. Putting double underscore would break the bundle options logic as it will all be in the same level.

soninaren:
Have addressed the pr feedback

test/Azure.Functions.Cli.Tests/ExtensionBundleHelperTests.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Azure.Functions.Cli.Tests
+{
+    public class ExtensionBundleHelperTests
+    {
+        [Fact]
+        public void VerifyGetBundleDownloadPathReturnCorrectPath()
+        {
+            var downloadPath = ExtensionBundleHelper.GetBundleDownloadPath("BundleId");
+            var expectedPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".azure-functions-core-tools", "Functions", "ExtensionBundles", "BundleId");
+            Assert.Equal(expectedPath, downloadPath);
```
pragnagopa, line 17: Can you update the test to verify path for each platform:

```
OSPlatform.Windows
OSPlatform.OSX
OSPlatform.Linux
```
You can use `System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform()` to detemine the platform and verify with a hardcoded path.

soninaren: We can get the root path based on `System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform()` but UserProfile would also include the user account used to execute the test. The only way to look that up would be by `Environment.SpecialFolder.UserProfile`

PR #2172, opened 2020-08-26
Clean host.json contents before test

pragnagopa:


test/Azure.Functions.Cli.Tests/LoggingFilterHelperTests.cs
```diff
@@ -1,26 +1,58 @@
...
+        private string _hostJsonFilePath;
 
         public LoggingFilterHelperTests()
         {
-            _hostOptions = new ScriptApplicationHostOptions
+            try
             {
-                ScriptPath = Directory.GetCurrentDirectory()
-            };
+                _workerDir = Path.GetTempFileName();
```
soninaren, line 22: Nit: According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netcore-3.1). `GetTempFileName` would create a temp file return full path. Not sure if we are using it right.

pragnagopa: Thanks. Skipping these tests for now. Need more time to investigate. Will address this as part of https://github.com/Azure/azure-functions-core-tools/issues/2174

PR #2186, opened 2020-09-02
Fix null ref when debugging from VS

### LINKED ISSUE #2183: Null reference exception when debugging from VS
Hi Guys,

I have installed Microsoft Visual Studio Enterprise 16.8.0 Preview 2.1 yesterday and now can't run Azure Functions locally.
image
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=func
  StackTrace:
   at Azure.Functions.Cli.Utilities.LogLevelExists(IConfigurationRoot hostJsonConfig, String category) in D:\a\1\s\src\Azure.Functions.Cli\Common\Utilities.cs:line 208
```

Azure Functions Core Tools (3.0.2852 Commit hash: d1c2c8e7dc10a94a399246dda94049a7dbf0383d)
Function Runtime Version: 3.0.14287.0

pragnagopa:
Fixes #2183

### DIFF
--- a/src/Azure.Functions.Cli/Common/Utilities.cs
+++ b/src/Azure.Functions.Cli/Common/Utilities.cs
@@ -205,12 +205,14 @@ namespace Azure.Functions.Cli
             string categoryKey = ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.Logging, LogLevelSection, category);
             try
             {
-                if (Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey].ToString(), true, out object outLevel))
+                if (hostJsonConfig[categoryKey] != null && Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey].ToString(), true, out object outLevel))
                 {
                     return true;
                 }
             }
-            catch { }
+            catch 
+            { 
+            }
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -205,12 +205,14 @@ internal static bool LogLevelExists(IConfigurationRoot hostJsonConfig, string ca
             string categoryKey = ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.Logging, LogLevelSection, category);
             try
             {
-                if (Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey].ToString(), true, out object outLevel))
+                if (hostJsonConfig[categoryKey] != null && Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey].ToString(), true, out object outLevel))
```
brettsam, line 208: What if you just removed the `.ToString()`? I assume that's where the null-ref was? That should already be a string and I think `Enum.TryParse()` can handle that.

pragnagopa: thanks. you are right. I assumed `Enum.TryParse` would throw on if the input was `null`

PR #2191, opened 2020-09-02
Add default webjobs loglevel filter and treat logs with category ILogger<T> as user logs

### LINKED ISSUE #2184: Core Tools 3.0.2852 does not print logs when ILogger<T> is used
Azure Functions 3.0.14287.0
C# .Net Core 3.1
Azure Functions v34
Visual Studio 2019 16.7.2
Windows 10

Symptom: When we log using ILogger those log lines do not show up in func.exe console window anymore when debugging in Visual Studio.

Also opened here: https://github.com/Azure/azure-functions-host/issues/6582

### LINKED ISSUE #2189: Massively verbose console output since Core Tools 2.7.2855
Since Visual Studio updated Core Tools to 2.7.2855 I now have massively verbose console output.  Lots of dark cyan lines everywhere from just about every library I am using.  This was not the previous behaviour. Screenshot of example output below.

I get this verbose output with Information, Debug or Trace default log levels  I might have expected it with Trace although it didn't previously.  Would not expect it with Information.

App is also running much slower but this is presumably because of all the logging.

image

pragnagopa:
cc @anthonychu 

Fixes #2189 Fixes #2184 

- Added missing Webjobs logging filters defined here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs#L28

 For  now this is a copy of the code until issue https://github.com/Azure/azure-functions-host/issues/6588 is fixed.

- Fixed user category filter to ensure logs with category ILogger<T> are treated as user logs

### DIFF
--- a/src/Azure.Functions.Cli/Common/Utilities.cs
+++ b/src/Azure.Functions.Cli/Common/Utilities.cs
@@ -250,7 +250,7 @@ namespace Azure.Functions.Cli
             {
                 return actualLevel >= userLogMinLevel;
             }
-            if (AllowedCategoryPrefixes.Where(p => category.StartsWith(p)).Any())
+            if (IsSystemLogCategory(category))
             {
                 // System logs
                 return actualLevel >= systemLogMinLevel;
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -262,6 +273,11 @@ internal static bool UserLoggingFilter(LogLevel actualLevel)
             return actualLevel >= LogLevel.Trace;
         }
 
+        internal static bool SystemLoggingFilter(string category, LogLevel actualLevel, LogLevel minLevel)
```
brettsam, line 276: is this needed?

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -16,15 +16,20 @@
...
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Azure.Functions.Cli
 {
     internal static class Utilities
     {
         public const string LogLevelSection = "loglevel";
         public const string LogLevelDefaultSection = "Default";
+        internal static readonly string[] AllowedCategoryPrefixes = new[]
```
brettsam, line 26: rename to `SystemCategoryPrefixes`

brettsam (approved):

ankitkumarr (approved):

PR #2204, opened 2020-09-05
Update Readme

antempus:
v2 beta was removed after v2 functions GA, this line doesn't seem to be required any longer and the associated documentation is gone.


fabiocav (approved):

anthonychu (approved):

fabiocav:
@antempus could you please squash the commits on the PR so we can have it merged? Thanks!

antempus:
@fabiocav 
Struggle with that a little more than I would've liked to, but this should be good to go.

fabiocav:
Thank you!

PR #2223, opened 2020-09-11
Fix default logging filters - rely on LoggerFilterOptions

### LINKED ISSUE #2216: Unable to override loglevel in local.settings.json
Reading all the docs on this I should be able to override my default loglevel locally but it seems the setting isn't getting picked up for some reason.

host.json

```json
{
  "version": "2.0",
  "functions": [ "RunSnapshot" ],
  "logging": {
    "logLevel": {
      "Default": "Information",
    },
    "applicationInsights": {
      "samplingExcludedTypes": "Request",
      "samplingSettings": {
        "isEnabled": true
      }
    }
  }
}
```

local.settings.json

```json
{
  "IsEncrypted": false,
  "Values": {

    "AzureFunctionsJobHost__logging__logLevel__Default": "Debug",

  }
}
```

Yet despite me setting the level to debug in `local.settings.json` my app is still logging at information level. I can see if I alter the setting in `host.json` I get the expected output. Is this a bug or can someone point me at what I'm doing wrong here?

>Azure Functions Core Tools (3.0.2881 Commit hash: fa9641ad925588192888bc977ff51d8a63536267)
Function Runtime Version: 3.0.14287.0

### LINKED ISSUE #2217: Updates to console logging based on feedback
There's been some feedback after release of #2141 

* Add timestamps back to logs - We've gotten some good feedback that this is useful for delimiting the start of multiline logs. https://github.com/Azure/azure-functions-core-tools/issues/2184#issuecomment-686836558
* Add function start/stop messages back
    ```
    Executing 'Functions.HttpTrigger' (Reason='This function was programmatically called via the host APIs.', Id=b7b32e4f-4c9f-4a48-b160-5d445644f182)
    Executed 'Functions.HttpTrigger' (Succeeded, Id=b7b32e4f-4c9f-4a48-b160-5d445644f182)
    ```

### LINKED ISSUE #2220: Unexpected verbose/informational logging in latest core tools release
### Issue
There are two issues that I'm seeing related to custom log sources in the latest core tools release - i.e. logs written from `ILogger` instances created by `ILoggerFactory.CreateLogger(...)`.

1. When I configure log levels for my various log sources, no matter what value I specify, I see **Verbose** logs written to the console. My expectation is that the log level I specify in host.json would be honored (this is similar to #2216 but slightly different because I see verbose logs regardless of what I have configured for the default logging level).
2. If I don't configure _any_ log levels for my various log sources, then I see **Informational** logs written to the console. This seems like a regression from previous releases, where I wouldn't see any log output from my custom log sources unless explicitly configured (and I preferred that behavior).

### Core Tools Version
```
Azure Functions Core Tools (3.0.2881 Commit hash: fa9641ad925588192888bc977ff51d8a63536267)
Function Runtime Version: 3.0.14287.0
```

### Repro Steps
You can use the [Durable Functions precompiled sample app](https://github.com/Azure/azure-functions-durable-extension/tree/dev/samples/precompiled) to see this behavior. This is the default host.json logging configuration you should see in the sample app (I removed some of the irrelevant parts):

```json
{
  "version": "2.0",
  "logging": {
    "logLevel": {
      "DurableTask.AzureStorage": "Warning",
      "DurableTask.Core": "Warning"
    }
  }
}
```

1. Use F5 in Visual Studio to start the sample app.
2. [Repro for (1) above] Notice the mix of blue and green logs that get emitted. These are Informational and Verbose logs emitted by the `DurableTask.AzureStorage` and `DurableTask.Core` log sources in the Durable Functions extension.

image

3. Update the **host.json** file by removing the `logging` section entirely.

4. [Repro for (2) above] Notice the blue logs that get emitted. These are Informational logs emitted by the Durable Functions `DurableTask.AzureStorage` log source and I don't expect them to be visible by default.

image

Again, in previous versions I didn't see this output by default, and my hope is that we can restore the previous behavior in terms of default log verbosity for custom log providers.

pragnagopa:
**Resolves #2216** and **Resolves #2220**
- Use registered `LoggerFilterOptions` to filter logs
    - Default log levels apply only if `LoggingFilterRule` does not have a `logLevel` for a category.

**Resolves #2217**
- Bring back time stamps in the logs
- Print Function Executing and Executed messages.

### DIFF
--- a/src/Azure.Functions.Cli/Common/Utilities.cs
+++ b/src/Azure.Functions.Cli/Common/Utilities.cs
@@ -24,13 +24,6 @@ namespace Azure.Functions.Cli
     {
         public const string LogLevelSection = "loglevel";
         public const string LogLevelDefaultSection = "Default";
-        internal static readonly string[] SystemCategoryPrefixes = new[]
-        {
-            "Microsoft.Azure.WebJobs.",
-            "Function.",
-            "Worker.",
-            "Host."
-        };
 
         internal static void PrintLogo()
         {
@@ -189,29 +182,12 @@ namespace Azure.Functions.Cli
             return localPath;
         }
 
-        internal static LogLevel GetHostJsonDefaultLogLevel(IConfigurationRoot hostJsonConfig)
-        {
-            string defaultLogLevelKey = ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.Logging, LogLevelSection, LogLevelDefaultSection);
-            try
-            {
-                if (Enum.TryParse(typeof(LogLevel), hostJsonConfig[defaultLogLevelKey].ToString(), true, out object outLevel))
-                {
-                    return (LogLevel)outLevel;
-                }
-            }
-            catch
-            {
-            }
-            // Default log level
-            return LogLevel.Information;
-        }
-
-        internal static bool LogLevelExists(IConfigurationRoot hostJsonConfig, string category)
+        internal static bool LogLevelExists(IConfigurationRoot hostJsonConfig, string category, out LogLevel outLogLevel)
         {
             string categoryKey = ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.Logging, LogLevelSection, category);
             try
             {
-                if (Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey], true, out object outLevel))
+                if (Enum.TryParse(hostJsonConfig[categoryKey], true, out outLogLevel))
                 {
                     return true;
                 }
@@ -219,6 +195,7 @@ namespace Azure.Functions.Cli
             catch 
             { 
             }
+            outLogLevel = LogLevel.Information;
             return false;
         }
 
@@ -247,7 +224,9 @@ namespace Azure.Functions.Cli
         /// <returns></returns>
         internal static bool DefaultLoggingFilter(string category, LogLevel actualLevel, LogLevel userLogMinLevel, LogLevel systemLogMinLevel)
         {
-            if (LogCategories.IsFunctionUserCategory(category) || category.Equals(WorkerConstants.FunctionConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase))
+            if (LogCategories.IsFunctionUserCategory(category)
+                || LogCategories.IsFunctionCategory(category)
+                || category.Equals(WorkerConstants.FunctionConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase))
             {
                 return actualLevel >= userLogMinLevel;
             }
@@ -260,28 +239,9 @@ namespace Azure.Functions.Cli
             return actualLevel >= userLogMinLevel;
         }
 
-        /// <summary>
-        /// Returns true for user logs >= Trace level. Returns false, if log level is explicitly set to None.
-        /// </summary>
-        /// <param name="actualLevel"></param>
-        /// <returns></returns>
-        internal static bool UserLoggingFilter(LogLevel actualLevel)
-        {
-            if (actualLevel == LogLevel.None)
-            {
-                return false;
-            }
-            return actualLevel >= LogLevel.Trace;
-        }
-
-        internal static bool SystemLoggingFilter(string category, LogLevel actualLevel, LogLevel minLevel)
-        {
-            return actualLevel >= minLevel && IsSystemLogCategory(category);
-        }
-
         internal static bool IsSystemLogCategory(string category)
         {
-            return SystemCategoryPrefixes.Where(p => category.StartsWith(p)).Any();
+            return ScriptConstants.SystemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any();
         }
 
         internal static IConfigurationRoot BuildHostJsonConfigutation(ScriptApplicationHostOptions hostOptions)
--- a/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
@@ -55,7 +55,7 @@ namespace Azure.Functions.Cli.Diagnostics
             {
                 return Utilities.DefaultLoggingFilter(category, logLevel, filterRule.LogLevel.Value, filterRule.LogLevel.Value);
             }
-            return Utilities.DefaultLoggingFilter(category, logLevel, filterRule.LogLevel.Value, filterRule.LogLevel.Value);
+            return Utilities.DefaultLoggingFilter(category, logLevel, _loggingFilterHelper.UserLogDefaultLogLevel, _loggingFilterHelper.SystemLogDefaultLogLevel);
         }
 
         public bool IsEnabled(LogLevel logLevel)
--- a/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
@@ -90,7 +90,7 @@ namespace Azure.Functions.Cli.Diagnostics
         {
             foreach (var line in GetMessageString(logLevel, formattedMessage, exception))
             {
-                ColoredConsole.WriteLine($"[{DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss")}] {line}");
+                ColoredConsole.WriteLine($"[{DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fff")}] {line}");
             }
         }
 
--- a/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
@@ -14,18 +14,53 @@ namespace Azure.Functions.Cli.Diagnostics
         private readonly bool _verboseErrors;
         private readonly string _category;
         private readonly LoggingFilterHelper _loggingFilterHelper;
+        private readonly LoggerFilterOptions _loggerFilterOptions;
         private readonly string[] allowedLogsPrefixes = new string[] { "Worker process started and initialized.", "Host lock lease acquired by instance ID" };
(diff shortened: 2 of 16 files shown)


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -189,36 +182,23 @@ internal static string EnsureCoreToolsLocalData()
...
-            {
-            }
-            // Default log level
-            return LogLevel.Information;
-        }
-
-        internal static bool LogLevelExists(IConfigurationRoot hostJsonConfig, string category)
+        internal static bool LogLevelExists(IConfigurationRoot hostJsonConfig, string category, out LogLevel outLogLevel)
         {
+            // Host creates user logger for function with Category: Function.{FunctionName}.User Remove ".User" to match the  category name to the funciton name
```
brettsam, line 187: nit: misspelling of `function` in comment

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -189,36 +182,23 @@ internal static string EnsureCoreToolsLocalData()
...
+            category = category.Replace(".User", "", StringComparison.OrdinalIgnoreCase);
             string categoryKey = ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.Logging, LogLevelSection, category);
             try
             {
                 if (Enum.TryParse(typeof(LogLevel), hostJsonConfig[categoryKey], true, out object outLevel))
```
brettsam, line 192: This shouldn't be a direct match -- we need to do a `StartsWith` here -- and choose the longest match. Then you don't need to do the replacement of `.User`. The built-in `LoggerFilterOptions` does all of this already so you shouldn't really need this logic if you switched to using that.

For example -- if someone uses two filters of `Host.` -> `LogLevel.None` and `Host.Startup` -> `LogLevel.Information`. it means they want all host logs to be hidden unless it's `Host.Startup`. I suspect this doesn't work right here?

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
```diff
@@ -55,12 +55,7 @@ private void LogToConsole(LogLevel logLevel, Exception exception, string formatt
         {
             foreach (var line in GetMessageString(logLevel, formattedMessage, exception))
             {
-                var outputline = $"{line}";
-                if (_loggingFilterHelper.VerboseLogging)
-                {
-                    outputline = $"[{DateTime.UtcNow}] {outputline}";
-                }
-                ColoredConsole.WriteLine($"{outputline}");
+                ColoredConsole.WriteLine($"[{DateTime.UtcNow}] {line}");
```
brettsam, line 58: We typically format this as `DateTime.UtcNow.ToString("O")` -- or even shorten the milliseconds down a bit sometimes with `ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK")`.

src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
```diff
@@ -65,16 +62,29 @@ public LoggingFilterHelper(IConfigurationRoot hostJsonConfig, bool? verboseLoggi
 
         internal void AddConsoleLoggingProvider(ILoggingBuilder loggingBuilder)
         {
-            // Filter is needed to force all the logs at jobhost level
-            loggingBuilder.AddFilter<ColoredConsoleLoggerProvider>((category, level) => true).AddProvider(new ColoredConsoleLoggerProvider(this));
+            // Filter is needed to force all the System Logs coming from jobhost level
+            loggingBuilder.AddFilter<ColoredConsoleLoggerProvider>((category, level) => Filter(category, level)).AddProvider(new ColoredConsoleLoggerProvider(this));
```
brettsam, line 66: This is now filtering at this level, as well as in the logger itself? Can you let the logger do all of the filtering, so it's all in one place?

src/Azure.Functions.Cli/Common/Utilities.cs
fabiocav: Can we please have a test in place for those scenarios?

fabiocav, line 192: If this code is staying, please use the generic version that outputs a typed enum

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -265,29 +247,39 @@ internal static bool DefaultLoggingFilter(string category, LogLevel actualLevel,
...
         {
-            return SystemCategoryPrefixes.Where(p => category.StartsWith(p)).Any();
+            return ScriptConstants.SystemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any();
         }
 
         internal static IConfigurationRoot BuildHostJsonConfigutation(ScriptApplicationHostOptions hostOptions)
         {
             IConfigurationBuilder builder = new ConfigurationBuilder();
             builder.Add(new HostJsonFileConfigurationSource(hostOptions, SystemEnvironment.Instance, loggerFactory: NullLoggerFactory.Instance, metricsLogger: new MetricsLogger()));
+            builder.Add(new ScriptEnvironmentVariablesConfigurationSource());
```
fabiocav, line 282: This doesn't seem like the right place to add another configuration source

fabiocav (changes requested):

src/Azure.Functions.Cli/Common/Utilities.cs
pragnagopa: Updated to rely on `LoggerFilterOptions`. Added unit test that verified rule selection.

src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
pragnagopa: done!

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
pragnagopa: fixed!

src/Azure.Functions.Cli/Common/Utilities.cs
pragnagopa: not needed anymore. Thanks!

src/Azure.Functions.Cli/Common/Utilities.cs
pragnagopa: Fixed. Thanks!

src/Azure.Functions.Cli/Common/Utilities.cs
pragnagopa: Need to figure out what is the right place to add this configuration source. Without this overriding via local.setting.json is not working.

pragnagopa: Using `LoggerFilterOptions` at webhost level fixed this.

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
```diff
@@ -14,18 +14,53 @@ public class ColoredConsoleLogger : ILogger
...
+            {
+                bool enabled = filterRule.Filter(ProviderType.FullName, category, logLevel);
+                if (!enabled)
+                {
+                    return false;
+                }
+            }
+            if (filterRule.LogLevel != null)
+            {
+                return Utilities.DefaultLoggingFilter(category, logLevel, filterRule.LogLevel.Value, filterRule.LogLevel.Value);
```
pragnagopa, line 56: @brettsam  / @fabiocav  - this takes care of picking the right log level based on filter options,

pragnagopa:
@brettsam / @fabiocav - PR is ready for review. Addressed feedback, added more unit tests.

pragnagopa:
@Francisco-Gamino  - please wait for this PR to be merged before preparing CLI build for OGFs. Thanks!

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
```diff
@@ -14,18 +14,53 @@ public class ColoredConsoleLogger : ILogger
...
+                return false;
+            }
+            if (filterRule.Filter != null)
+            {
+                bool enabled = filterRule.Filter(ProviderType.FullName, category, logLevel);
+                if (!enabled)
+                {
+                    return false;
+                }
+            }
```
brettsam, line 53: For these two steps, if the filter comes back as true, we keep going? Shouldn't we return back if there's an explicit filter or log level set?

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
```diff
@@ -55,12 +90,7 @@ private void LogToConsole(LogLevel logLevel, Exception exception, string formatt
         {
             foreach (var line in GetMessageString(logLevel, formattedMessage, exception))
             {
-                var outputline = $"{line}";
-                if (_loggingFilterHelper.VerboseLogging)
-                {
-                    outputline = $"[{DateTime.UtcNow}] {outputline}";
-                }
-                ColoredConsole.WriteLine($"{outputline}");
+                ColoredConsole.WriteLine($"[{DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss")}] {line}");
```
brettsam, line 93: it may be worth putting a few millisecond places here too so folks can see differences in really quick executions?

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
pragnagopa: - If filter returns `false`  - it returns  - that means it failed check for `DefaultWebJobsFilter` added at webhost level.
- Filter comes back as `true` - for system logs from jobhost level - if loglevel is overwritten in host.json / env vars - `LogLevel` on filter is not null - this log level is used instead of defaults.

Do let me know if this is not clear.

src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs
pragnagopa: updated format!

brettsam (approved):

fabiocav (approved):

PR #2260, opened 2020-10-09
Fixed warning for missing settings when they're set in user secrets file

### LINKED ISSUE #2219: Remove warning for missing setting from local.setting.json file in case they are present in user secrets file.
Some information in this PR https://github.com/Azure/azure-functions-core-tools/pull/2136

Repro posted by @vijayrkn 
> Function:
> 
> ```
>     public static class Function
>     {
>         [FunctionName("Function")]
>         public static void Run([QueueTrigger("myqueue-items", Connection = "ConnectionStrings:MyQueueConn")]string myQueueItem, ILogger log)
>         {
>             log.LogInformation($"C# Queue trigger function processed: {myQueueItem}");
>         }
>     }
> ```
> 
> local.settings.json
> 
> ```
> {
>     "IsEncrypted": false,
>     "Values": {
>         "AzureWebJobsStorage": "UseDevelopmentStorage=true",
>         "FUNCTIONS_WORKER_RUNTIME": "dotnet"
>     }
> }
> ```
> 
> user secrets:
> 
> ```
> {
>   "ConnectionStrings:MyQueueConn": "UseDevelopmentStorage=true"
> }
> ```
> 
> csproj:
> 
> ```
> <Project Sdk="Microsoft.NET.Sdk">
>   <PropertyGroup>
>     <TargetFramework>netcoreapp3.1</TargetFramework>
>     <AzureFunctionsVersion>v3</AzureFunctionsVersion>
>     <UserSecretsId>4a9f61f7-0f22-4d2d-bc40-6ef82a58df03</UserSecretsId>
>   </PropertyGroup>
>   <ItemGroup>
>     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.10" />
>     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="3.1.0" />
>     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.7" />
>   </ItemGroup>
>   <ItemGroup>
>     <None Update="host.json">
>       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
>     </None>
>     <None Update="local.settings.json">
>       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
>       <CopyToPublishDirectory>Never</CopyToPublishDirectory>
>     </None>
>   </ItemGroup>
> </Project>
> ```
> 
> Let me know if you need anything more.

gzuber:
Resolves #2219 

A few design decisions I wanted to explain:

- Because secrets.json is only read after the host has started (during startup), we didn't have access to secrets.json to see if the needed settings were defined. We can't parse the secrets.json ourselves because we don't necessarily know where it is and the format might change. So we create a dummy configuration which parses the secrets.json, and then use that configuration to make sure all necessary settings are defined.
- I changed all the `if (dotnet)`s to `if (userSecretsEnabled)`s because we're doing more (checks, etc...) with user secrets so I thought it would pay off to be more specific
    - As a result, I changed the flow so that we're only traversing folders and parsing the csproj for secrets id once. If I had kept it how it was, we would be doing it 3 times (to see if secrets are enabled, to make sure settings are defined, to actually include them in the config) so I wanted to be more efficient. Especially since I didn't want to come up with a weird fix so it only logs the traversal once.
- I changed the warnings/errors for if settings aren't defined slightly... but in a weird way. If the user is using a secrets.json, I wanted to include that in our list of files we checked and the list of files they could add the setting to in order to fix the problem. However, the exception for if `AzureWebJobsStorage` is not defined is worded in a way that it makes sense to say "local.settings.json AND User Secrets". All the rest are set up to use OR. I know that there are tests that count on this warning to be worded in a specific way and I'm assuming it might cause issues if I changed it, so I decided to accommodate it. If it's not a big deal and I can change that error I think that's the best course of action. This is what the warnings will look like:

If everything is fine:
specified

If AzureWebJobsStorage is undefined:
storage_unspecified

If a different, important setting is undefined:
trigger_var_unspecified

Note the build warnings are something that needs to be taken care of in azure-functions-vs-build-sdk. [Issue here](https://github.com/Azure/azure-functions-vs-build-sdk/issues/477)

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -64,6 +64,8 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         public List<string> EnabledFunctions { get; private set; }
 
+        public string UserSecretsId { get; private set; }
+
         public StartHostAction(ISecretsManager secretsManager)
         {
             _secretsManager = secretsManager;
@@ -152,12 +154,15 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         private async Task<IWebHost> BuildWebHost(ScriptApplicationHostOptions hostOptions, Uri listenAddress, Uri baseAddress, X509Certificate2 certificate)
         {
+            LoggingFilterHelper loggingFilterHelper = new LoggingFilterHelper(_hostJsonConfig, VerboseLogging);
+            if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.dotnet)
+            {
+                UserSecretsId = ProjectHelpers.GetUserSecretsId(hostOptions.ScriptPath, loggingFilterHelper, new LoggerFilterOptions());
+            }
+
             IDictionary<string, string> settings = await GetConfigurationSettings(hostOptions.ScriptPath, baseAddress);
-           
             settings.AddRange(LanguageWorkerHelper.GetWorkerConfiguration(LanguageWorkerSetting));
             UpdateEnvironmentVariables(settings);
-
-            LoggingFilterHelper loggingFilterHelper = new LoggingFilterHelper(_hostJsonConfig, VerboseLogging);
 
             var defaultBuilder = Microsoft.AspNetCore.WebHost.CreateDefaultBuilder(Array.Empty<string>());
             
@@ -193,7 +198,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
                     // This is needed to filter system logs only for known categories
                     loggingBuilder.AddDefaultWebJobsFilters<ColoredConsoleLoggerProvider>(LogLevel.Trace);
                 })
-                .ConfigureServices((context, services) => services.AddSingleton<IStartup>(new Startup(context, hostOptions, CorsOrigins, CorsCredentials, EnableAuth, loggingFilterHelper)))
+                .ConfigureServices((context, services) => services.AddSingleton<IStartup>(new Startup(context, hostOptions, CorsOrigins, CorsCredentials, EnableAuth, UserSecretsId, loggingFilterHelper)))
                 .Build();
         }
 
@@ -211,7 +216,16 @@ namespace Azure.Functions.Cli.Actions.HostActions
                     .GetEnvironmentVariables()
                     .Cast<DictionaryEntry>()
                     .ToDictionary(k => k.Key.ToString(), v => v.Value.ToString());
-            await CheckNonOptionalSettings(settings.Union(environment), scriptPath);
+
+            // Build user secrets in order to read secrets.json file
+            IEnumerable<KeyValuePair<string, string>> userSecrets = new Dictionary<string, string>();
+            bool userSecretsEnabled = !string.IsNullOrEmpty(UserSecretsId);
+            if (userSecretsEnabled)
+            {
+                userSecrets = Utilities.BuildUserSecrets(UserSecretsId, _hostJsonConfig, VerboseLogging);
+            }
+
+            await CheckNonOptionalSettings(settings.Union(environment).Union(userSecrets), scriptPath, userSecretsEnabled);
 
             // when running locally in CLI we want the host to run in debug mode
             // which optimizes host responsiveness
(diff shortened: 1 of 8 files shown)


gzuber:
@pragnagopa for logging
@brettsam for configuration
@soninaren for user secrets

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -211,7 +216,20 @@ private async Task<IWebHost> BuildWebHost(ScriptApplicationHostOptions hostOptio
                     .GetEnvironmentVariables()
                     .Cast<DictionaryEntry>()
                     .ToDictionary(k => k.Key.ToString(), v => v.Value.ToString());
-            await CheckNonOptionalSettings(settings.Union(environment), scriptPath);
+
+            // Build user secrets in order to read secrets.json file
+            IEnumerable<KeyValuePair<string, string>> userSecrets = new Dictionary<string, string>();
+            bool userSecretsEnabled = !string.IsNullOrEmpty(UserSecretsId);
+            if (userSecretsEnabled)
+            {
```
pragnagopa, line 224: Nit: Move this to `Utilities`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -389,9 +414,9 @@ internal static async Task CheckNonOptionalSettings(IEnumerable<KeyValuePair<str
 
                 if (string.IsNullOrWhiteSpace(azureWebJobsStorage) && !allNonStorageTriggers)
                 {
-                    throw new CliException($"Missing value for AzureWebJobsStorage in {SecretsManager.AppSettingsFileName}. " +
+                    throw new CliException($"Missing value for AzureWebJobsStorage in {andConfigSources}. " +
```
pragnagopa, line 417: nit: move exception message string to constants. This will avoid defining variables orConfigSources / andConfigSources

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -410,8 +435,8 @@ internal static async Task CheckNonOptionalSettings(IEnumerable<KeyValuePair<str
                                 else if (!secrets.Any(v => v.Key.Equals(appSettingName, StringComparison.OrdinalIgnoreCase)))
                                 {
                                     ColoredConsole
-                                        .WriteLine(WarningColor($"Warning: Cannot find value named '{appSettingName}' in {SecretsManager.AppSettingsFileName} that matches '{token.Key}' property set on '{binding["type"]?.ToString()}' in '{filePath}'. " +
-                                            $"You can run 'func azure functionapp fetch-app-settings <functionAppName>' or specify a connection string in {SecretsManager.AppSettingsFileName}."));
+                                        .WriteLine(WarningColor($"Warning: Cannot find value named '{appSettingName}' in {orConfigSources} that matches '{token.Key}' property set on '{binding["type"]?.ToString()}' in '{filePath}'. " +
```
pragnagopa, line 438: Same here. nit: move exception message string to constants. This will avoid defining variables orConfigSources / andConfigSources

src/Azure.Functions.Cli/Helpers/ProjectHelpers.cs
```diff
@@ -11,6 +11,21 @@ namespace Azure.Functions.Cli.Helpers
...
     {
+        public static string GetUserSecretsId(string scriptPath, LoggingFilterHelper loggingFilterHelper, LoggerFilterOptions loggerFilterOptions)
+        {
+            if (string.IsNullOrEmpty(scriptPath))
+            {
+                return null;
+            }
+            string projectFilePath = ProjectHelpers.FindProjectFile(scriptPath, loggingFilterHelper, loggerFilterOptions);
+            if (projectFilePath == null) return null;
+
```
pragnagopa, line 22: nit: if block

pragnagopa (commented):
Added minor comments. Please add E2E tests `starttests` to verify log output - for scenarios you mentioned as part of PR description.

gzuber:
@pragnagopa -- not sure if this is what you had in mind for moving the messages to constants, but I gave it a shot 😄

pragnagopa (approved):
Please make sure PR build is green. Please ping here if you need approval again if you need to make test fixes.

pragnagopa:
@gzuber - Can you please "Resolve Conversation" for the code review you have already addressed? Please ping if you need approval again.

gzuber:
@yojagad Is this related to the build issues we've been seeing recently by chance? The license check isn't running, is there a way for me to run that manually?

@pragnagopa Sorry about that, all your comment have been addressed (and I marked them resolved) I've just been trying to figure out why some of the random tests were failing. The tests have run successfully a few times in a row now, so I think we should be good.

pragnagopa:
Merging this PR. Build is green. Some issue with CLA webhook. Bypassing the check.

Issue #2144, opened 2020-08-06
Support KEDA 2.0 for func kubernetes commands


jeffhollan:
KEDA 2.0 will go into beta soon, with a 2.0 release planned for later this year.  There are some minor changes to the spec so the function core tools will need to be able to generate the latest one.  

https://keda.sh/docs/2.0/migration/

Ideally there could still be a flag for backwards compat to KEDA 1.0?  Or maybe we make people opt-in to keda 2.0?  `func kubernetes deploy --keda-version 2` or something?

Will just need live before like end of October

// @apawast @anirudhgarg @TsuyoshiUshio

tomkerkhove:
Would be good to use 2.0 as default and provide capability to use 1.0 for backward compatibility maybe?

anthonychu:
I think it might be safer for `deploy` to check which version is installed in the cluster and be smart about what to do. A `--keda-version` flag might still be needed when the version is unknown (like when using `--dry-run`).

tomkerkhove:
Even better proposal!

jeffhollan:
FYI - 2.0 is now in beta, scheduled to be default in coming months.  Just flagging in that we likely want to try to book the work for this before end of October

tomkerkhove:
If I'm bored I can give it a try, but no promises!

tomkerkhove:
RC is out - https://github.com/kedacore/keda/releases/tag/v2.0.0-rc

TsuyoshiUshio:
It is a good idea! I agree with @anthonychu 's idea. I have a customer who wants to use v2 for the production as a backend of a service. However, My scalelogic prevent to use it, So I provide them a hotfix with branch and push keda image to dockerhub. Probably, if we operate it, these case happens, then we can easy to provide hot fix and customer can rollback old version if there is a problem for a new release.

jeffhollan:
Just pinging here, looking for end of October / early November release for KEDA 2.0 and will start being default. Would be good to track these changes.  All the changes between 1 and 2.0 are covered here https://keda.sh/docs/2.0/migration/ - added to PM board for planning / prioritization

tomkerkhove:
I'll pick this up in one of the coming days

pragnagopa:
cc @fabiocav 
Tagging myself (@pragnagopa), @ankitkumarr , @gzuber  for reviews

pragnagopa:
Closing as addressed in referenced PR.

PR #2266, opened 2020-10-19
🚧 - Provide support for KEDA 2.0

tomkerkhove:
Provide support for KEDA 2.0.

- It will automatically determine the current version of KEDA and take according actions, otherwise it will go with 2.0.
- All actions support passing KEDA version if need be
- Using Kubernetes resource to delete required resources rather than keeping a list of resource names

Relates to #2144


tomkerkhove:
I'll need some help on following topics:
- Is there a full list of trigger types that I can lookup so that I can implement `GetKedaTrigger` accordingly to map to the KEDA scalers?
- `PopulateMetadataDictionary` is to provide the metadata for scalers but I think it uses ENV VAR references by default, is that correct? Just to make sure I update the breaking changes
- Is there a way to run this locally? Based on the contributing file there are no tests but I'd like to give `KedaHelper.DetermineCurrentVersion` a go.

I don't know who will be able to help with my questions here, maybe @ahmelsayed @fabiocav or @jeffhollan?

src/Azure.Functions.Cli/Kubernetes/KEDA/KedaHelper.cs
```diff
@@ -41,10 +43,20 @@ internal static IKubernetesResource GetScaledObject(string name, string @namespa
...
         {
-            // TODO: discover version
-            return KedaVersion.v2;
+            // Get KEDA resource information
+            var resourceInfo = await KubernetesHelper.ResourceExists("Deployment", "keda-metrics-apiserver", @namespace, returnJsonOutput: true);
+            if (resourceInfo.ResourceExists == false)
+            {
+                // We use KEDA v2 as a default
+                return KedaVersion.v2;
+            }
```
prqgdbc, line 54: Perhaps checking for the existence of the apiservice here instead might be more compatible with different methods of installing helm?

```suggestion
            var resourceInfoV2 = await KubernetesHelper.ResourceExists("apiservice", "v1alpha1.keda.sh", @namespace, returnJsonOutput: true);
            
            if(resourceInfoV2.ResourceExists)
            {
                return KedaVersion.v2;
            }
```

tomkerkhove: This way you rely on the resource type which is different between v1 & v2 which is not the case.

The ways to install KEDA all provide deployment with that name so think we are good to go here.

tomkerkhove:
Any thoughts on this @ahmelsayed @jeffhollan?

jeffhollan:
I have this flagged as an item to have a dev assigned for this sprint - planning is on Wednesday. If you don’t hear back by then that may be when @fabiocav assigns someone to lock and merge. Thank you a ton for doing this

fabiocav:
Thank you, @tomkerkhove !

Assigning this to be looked at in the current sprint and we'll have folks assigned and following up with you.

tomkerkhove:
Thanks @fabiocav !

fabiocav:
@tomkerkhove I'm also assuming this is for 3.0, correct? If so, we need to target the v3.x branch. We'll actually be changing the default branch target this week

tomkerkhove:
It is backwards compatible but happy to switch to v3 if you prefer?

tomkerkhove:
I'll clean things up and switch to v3

tomkerkhove:
There are too many changes, I'll just create a new branch based off of v3 to make sure I don't mess things up

tomkerkhove:
Closing in favor of https://github.com/Azure/azure-functions-core-tools/pull/2289 which is targeting 3.X

PR #2276, opened 2020-10-24
Merging v3.x to release_3.0

azfuncgh:


sidkri:
This was done before the hotfix was merged and so is not what will be released.

PR #2311, opened 2020-11-24
Update v2 Python Worker Version to 1.1.10

Hazhzeng:
Python Worker Release note [1.1.10](https://github.com/Azure/azure-functions-python-worker/releases/tag/1.1.10)


pragnagopa:
cc @fabiocav 

@Hazhzeng - Same as host. We will not be releasing regular updates to V2 Core-Tools . V2 core-tools will be updated only when V2 runtime would be updated.

Closing as not needed.

PR #2316, opened 2020-12-01
Sequential JobHost restart

mhoeger:
Commit https://github.com/Azure/azure-functions-host/commit/12fff582f5880ee120e3ccd0ed1ea7753e705469 added support for non-overlapping worker recycle 

This PR consumes the host change 

Resolves https://github.com/Azure/azure-functions-core-tools/issues/2317

Note that to get the continuous debug experience, there is still more work required on the VS Code side, but this unblocks those changes.

### DIFF
--- a/src/Azure.Functions.Cli/Program.cs
+++ b/src/Azure.Functions.Cli/Program.cs
@@ -48,6 +48,11 @@ namespace Azure.Functions.Cli
             {
                 Environment.SetEnvironmentVariable(Constants.FunctionsCoreToolsEnvironment, "True");
             }
+
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.SequentialJobHostRestart)))
+            {
+                Environment.SetEnvironmentVariable(Constants.SequentialJobHostRestart, "True");
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Program.cs
```diff
@@ -48,6 +48,11 @@ private static void SetCoreToolsEnvironmentVaraibles()
             {
                 Environment.SetEnvironmentVariable(Constants.FunctionsCoreToolsEnvironment, "True");
             }
+
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.SequentialJobHostRestart)))
```
pragnagopa, line 52: Use `SetEnvironmentVariableAsBoolIfNotExists` instead

pragnagopa:
Please open a separate issue in Core-Tools repo to link this PR . Also, Resolve the issue in host with the commit referenced above

src/Azure.Functions.Cli/Program.cs
mhoeger: Updated! :)

pragnagopa (approved):

PR #2323, opened 2020-12-04
[v2] Point templates to 2.1.1642

yojagad:


pragnagopa:
Auto-upgrade from Function V2 to V3 has started. We will not be releasing functions V2 runtime / Core tools unless a bug fix is needed. Closing as not required. Thanks!

yojagad:
@pragnagopa I think it will still be beneficial to merge this into dev so that if at all release is being cut off for a bug fix, this can go along.

pragnagopa:
For new features - we recommend customers to move to V3 :) V2 releases will be scoped to critical bug fixes if any

yojagad:
V2 should be cut for critical fixes only, I agree yes. 

But when bundles does complete, v2 core tools will anyway have the ability to use RMQ template even without a core tools release.

This PR merely enables the option to use the said template without bundle, which I agree is not critical , but would be good to have in case for some rare reason they couldn't create with bundles in the picture.. Do you still think that the right thing to do in that particular scenario would be to ask customers to upgrade to V3?

pragnagopa:
>  Do you still think that the right thing to do in that particular scenario would be to ask customers to upgrade to V3?

That is correct - For any new features recommendation is to move to V3.

PR #2326, opened 2020-12-07
set proper .net isolated settings when publishing

ankitkumarr:


src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -72,10 +72,12 @@ public static class WorkerRuntimeLanguageHelper
         public static string AvailableWorkersRuntimeString =>
             string.Join(", ", availableWorkersRuntime.Keys
                 .Where(k => (k != WorkerRuntime.java))
+                .Where(k => (k != WorkerRuntime.dotnetIsolated))
                 .Select(s => s.ToString()));
 
         public static IEnumerable<WorkerRuntime> AvailableWorkersList => availableWorkersRuntime.Keys
-            .Where(k => k != WorkerRuntime.java);
+            .Where(k => k != WorkerRuntime.java)
+            .Where(k => k != WorkerRuntime.dotnetIsolated);
```
ankitkumarr, line 80: This way we can hide the `dotnetIsolated` runtime from `func init` as we don't have those templates yet.

brettsam (approved):

Issue #2307, opened 2020-11-23
Outdated func warning breaks parsing of --version command


ejizba:
If I run `func --version` on v2, I'm seeing this:

> You are using an outdated version of the Azure Functions Core Tools. For more information, please see: https://aka.ms/func-v2-upgrade
> 2.7.3023

Please print the warning to stderr (it's currently going to stdout) or just remove it for this command. VS Code fails to parse the version as-is, affecting a number of version-specific scenarios

anthonychu:
Sorry we didn't consider this scenario. Let's see if we can get this fixed relatively quickly.

fabiocav:
@soninaren is currently working on this. Change is in place, working on the build for release.

soninaren:
This is released to tooling feed. Pending npm release.

soninaren:
npm release is complete

PR #2327, opened 2020-12-08
[v2] Only show upgrade warning when an action is invoked

ankitkumarr:
Fixes https://github.com/Azure/azure-functions-core-tools/issues/2307

Doing it this way ensures that we don't show the v2 upgrade warning when `func --version` is invoked. This change makes it so we show the warning if any action was found. So any command that does something other than `func --version` should show the warning.


kashimiz (approved):
:shipit:

soninaren (approved):

PR #2332, opened 2020-12-08
WEBSITE_MOUNT_ENABLED =  1 for linux consumption

yojagad:
This reverts commit cd9bfca26f9c7fe06ce245f5bf69bc6486a685dd.

This allows persisting the permissions of binaries being deployed in linux consumption (Which is needed in case of custom handlers)


pragnagopa:
Please update the description with issue we are trying to address.

pragnagopa (approved):
nit: Update PR description.

PR #2333, opened 2020-12-09
Fixed user secrets failing tests

gzuber:
These tests were failing and after debugging, it looks like they weren't successfully editing the files they needed to be editing. Directly using the dotnet APIs instead of our `FileSystemHelpers` seemed to do the trick.


gzuber:
@pragnagopa

pragnagopa (approved):

PR #2336, opened 2020-12-09
Add architecture (32 or 64 bit) note to start up message

JohnCarmichael3000:
Adds architecture "32-bit" or "64-bit" note to the version information displayed on func startup. Makes it easy to confirm which one is running.


src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -49,7 +49,7 @@ internal static void PrintVersion()
         {
             ColoredConsole
                 .WriteLine($"\nAzure Functions Core Tools")
-                .WriteLine($"Core Tools Version:       {Constants.CliDetailedVersion}".DarkGray())
+                .WriteLine($"Core Tools Version:       {Constants.CliDetailedVersion + (Environment.Is64BitProcess ? " (64-bit)" : " (32-bit)")}".DarkGray())
```
soninaren, line 52: Tools like VS Code parse the version information. This can potentially break the integration.

cc:@ejizba

ejizba: Thanks for the ping, but I don't think this would affect us. We only care about `func --version` which [seems to use](https://github.com/Azure/azure-functions-core-tools/blob/3a59efba0cbf74dfaf1911de5f073df97259d9f3/src/Azure.Functions.Cli/ConsoleApp.cs#L217) `Constants.CliVersion` instead of this `PrintVersion` method

JohnCarmichael3000: Alternatively, could add a new line in PrintVersion method, which would be less risky of breaking something. 

+ .WriteLine($"Core Tools Architecture:       {(Environment.Is64BitProcess ? " 64-bit" : "32-bit")}".DarkGray())

Though, hopefully nobody is relying on parsing data from a large startup text block. I installed the core tools from a package installer and then it took much more effort that I thought it would to determine which version had been installed (32 or 64  bit), which led to my PR.

soninaren (approved):

src/Azure.Functions.Cli/Common/Utilities.cs
soninaren: @JohnCarmichael3000, Can you rebase your branch and sign the license/cla?

JohnCarmichael3000:
A bit new to this, not sure why it says 2 expected checks still...?

azure-functions-core-tools Expected — Waiting for status to be reported
license/cla Expected — Waiting for status to be reported

I merged with current code and signed the license.

https://cla.dotnetfoundation.org/Azure/azure-functions-core-tools?pullRequest=2336
reports "You have signed the CLA for multiple repositories or organizations".

soninaren:
> A bit new to this, not sure why it says 2 expected checks still...?
> 
> azure-functions-core-tools Expected — Waiting for status to be reported
> license/cla Expected — Waiting for status to be reported
> 
> I merged with current code and signed the license.
> 
> https://cla.dotnetfoundation.org/Azure/azure-functions-core-tools?pullRequest=2336
> reports "You have signed the CLA for multiple repositories or organizations".

The license/cla check seems to be stuck / broken. I don't have permissions to overwrite it. Will get someone to do that.

JohnCarmichael3000:
Not sure why this isn't moving ahead?

soninaren:
> Not sure why this isn't moving ahead?

Sorry about the delay here. I have scheduled a build. Will merge it as soon as it passes.

src/Azure.Functions.Cli/Common/Utilities.cs
JohnCarmichael3000: Great, thanks. I was thinking it was something in my end I hadn't completed. :)

soninaren:
@JohnCarmichael3000, I had to take your commit and merge it as part of seperate PR. Because of the movement on the v3.x branch your branch was getting out of sync between builds. The change in this PR, https://github.com/Azure/azure-functions-core-tools/pull/2439

It would still shows you as the author of the commit.

PR #2372, opened 2021-01-04
Add new files to sign

ankitkumarr:
Recent changes on us not including workers made it so that the workers like Powershell that had additional dependencies are not resolved anymore. This cause some dependencies that were common to Host and Powershell worker to be resolved differently. These dependencies now include some unsigned dlls which used to be signed previously.

More details:

- Up until recently, core tools always referenced workers for all versions (including minified ones). We used to manually delete the "workers" folder after build for min versions. However, we changed that to conditionally include workers references only for non-minified versions.
- Both Functions Host and Powershell worker have dependencies on libraries like "Grpc.Core.Api".
- Some libraries (such as Grpc.Core.Api) only sign binaries that are "AnyCpu"
- In core tools, we try to build ReadyToRun binaries for windows runtimes. When we do so, Functions Host (which also can provide R2R binaries) gets the targeted dlls for the architecture. So, when we build ready to run in core tools with "-r win-x64", it gets the unsigned dlls that target "x64" architecture instead of "AnyCpu". However, if we include Powershell worker, because that doesn't do R2R binaries, it brings the signed "AnyCpu" dlls that are common to both Host and worker.


Francisco-Gamino (approved):
LGTM

sidkri (approved):

PR #2378, opened 2021-01-08
revert to exisiting way of removing workers

ankitkumarr:
cc: @kashimiz

### DIFF
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -128,6 +128,7 @@ namespace Build
                                 (string.IsNullOrEmpty(Settings.IntegrationBuildNumber) ? string.Empty : $"/p:IntegrationBuildNumber=\"{Settings.IntegrationBuildNumber}\" ") +
                                 $"-o {outputPath} -c Release ");
         }
+
         public static void DotnetPublish()
         {
             foreach (var runtime in Settings.TargetRuntimes)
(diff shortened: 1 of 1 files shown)


build/BuildSteps.cs
```diff
@@ -583,5 +588,17 @@ private static List<string> GetV3PackageList()
 
             return packageList;
         }
+
+        private static void RemoveLanguageWorkers(string outputPath)
```
fabiocav, line 592: I realize this is a tactical fix, but if we're not adding tests here, or intend to change the approach, can we please have an issue filed and linked for follow up?

fabiocav (approved):

build/BuildSteps.cs
ankitkumarr: Yes absolutely. I will be opening an issue with proper details for a follow up.

Francisco-Gamino:
Hi @ankitkumarr -- should we also revert these?

1) https://github.com/Azure/azure-functions-core-tools/blob/v3.x/build/BuildSteps.cs#L126
2) https://github.com/Azure/azure-functions-core-tools/blob/v3.x/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj#L147

ankitkumarr:
Hi @Francisco-Gamino, I don't think so. That's for the nupkg that we started creating for core tools if I understand correctly. I think we can leave that as is for now. @kashimiz can confirm.

Francisco-Gamino:
> Hi @Francisco-Gamino, I don't think so. That's for the nupkg that we started creating for core tools if I understand correctly. I think we can leave that as is for now. @kashimiz can confirm.

Thanks @ankitkumarr  for clarifying.

build/BuildSteps.cs
ankitkumarr: https://github.com/Azure/azure-functions-core-tools/issues/2380

PR #2409, opened 2021-01-28
Add Durable triggers to TriggersWithoutStorage

cgillum:
This is to support running Durable Functions with the new SQL backend that's currently under development.


cgillum:
Hi @ankitkumarr 👋. I tagged you for PR review since you made a similar change for Kafka triggers. :)  Let me know if there is any other work I should consider. I wasn't sure if I only need to worry about this check in core-tools or if there are similar checks in other places, like the Functions host...

ankitkumarr (approved):

PR #2436, opened 2021-02-11
Parallelize the Kubernetes build tasks to improve latency.

brendandburns:
Signed-of-by: Brendan Burns

This improves the overall deployment latency by parallelizing the Docker push and the Kubernetes setup. It still waits for the push to complete before creating the final deployment.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
+++ b/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
@@ -137,6 +137,7 @@ namespace Azure.Functions.Cli.Actions.KubernetesActions
             }
             else
             {
+                List<Task> tasks = new List<Task>();
                 if (!await KubernetesHelper.NamespaceExists(Namespace))
                 {
                     await KubernetesHelper.CreateNamespace(Namespace);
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -137,21 +137,25 @@ public override async Task RunAsync()
...
                 }
 
                 if (shouldBuild)
                 {
-                    await DockerHelpers.DockerPush(resolvedImageName);
+                    // TODO: Join this with the build above so that it can run in parallel to the Kubernetes object create
+                    tasks.Add(DockerHelpers.DockerPush(resolvedImageName, false));
                 }
-
+                // TODO: Convert these to YAML so that we can send them as a single file
```
ahmelsayed, line 151:
```csharp
var yaml = KubernetesHelper.SerializeResources(resources, OutputSerializationOptions.Yaml);
``` 

should convert them to yaml, then the other overload of `KubectlApply()` should take care of it. I thought applying them one by one would make errors a bit easier to read, but I'm not sure how true that is.

src/Azure.Functions.Cli/Helpers/DockerHelpers.cs
```diff
@@ -62,6 +62,7 @@ internal static async Task<TriggersPayload> GetTriggersFromDockerImage(string im
             var containerId = string.Empty;
             try
             {
+                // TODO: This could be a single Docker run command
```
ahmelsayed, line 65: I used to have issues with sharing volumes from a Windows machine to a container that would work with all configs for Docker for Windows, so used `docker cp` instead.

ahmelsayed (approved):

ahmelsayed (approved):
LGTM

PR #2434, opened 2021-02-10
func kubernetes deploy should deploy to the current namespace, not the default

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Open work:
- [x] Check if look-ups specify namespace

### Example: Specifying namespace

```yaml
data:
  AzureWebJobsStorage: VXNlRGV2ZWxvcG1lbnRTdG9yYWdlPXRydWU=
  FUNCTIONS_WORKER_RUNTIME: ZG90bmV0
apiVersion: v1
kind: Secret
metadata:
  name: foo
  namespace: example
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
  namespace: example
  labels:
    app: foo
spec:
  selector:
    matchLabels:
      app: foo
  template:
    metadata:
      labels:
        app: foo
    spec:
      containers:
      - name: foo
        image: bar.io/foo
        env:
        - name: AzureFunctionsJobHost__functions__0
          value: test
        envFrom:
        - secretRef:
            name: foo
        readinessProbe:
          failureThreshold: 3
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 240
          httpGet:
            path: /
            port: 80
            scheme: HTTP
        startupProbe:
          failureThreshold: 3
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 240
          httpGet:
            path: /
            port: 80
            scheme: HTTP
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: foo
  namespace: example
  labels: {}
spec:
  scaleTargetRef:
    name: foo
  triggers:
  - type: azure-queue
    metadata:
      queueName: myqueue-items
      connectionFromEnv: ''
---


```

### Example: No namespace specified

```yaml
data:
  AzureWebJobsStorage: VXNlRGV2ZWxvcG1lbnRTdG9yYWdlPXRydWU=
  FUNCTIONS_WORKER_RUNTIME: ZG90bmV0
apiVersion: v1
kind: Secret
metadata:
  name: foo
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
  labels:
    app: foo
spec:
  selector:
    matchLabels:
      app: foo
  template:
    metadata:
      labels:
        app: foo
    spec:
      containers:
      - name: foo
        image: bar.io/foo
        env:
        - name: AzureFunctionsJobHost__functions__0
          value: test
        envFrom:
[...]

PR #2448, opened 2021-02-16
Fixes a bug in empty namespace handling

brendandburns:
Bug introduced by #2434

cc @tomkerkhove

### DIFF
--- a/src/Azure.Functions.Cli/Kubernetes/KubectlHelper.cs
+++ b/src/Azure.Functions.Cli/Kubernetes/KubectlHelper.cs
@@ -135,16 +135,25 @@ namespace Azure.Functions.Cli.Kubernetes
                 }
             }
 
+            static string MaybeNamespace(string @namespace)
+            {
+                if (string.IsNullOrWhiteSpace(@namespace) == false)
+                {
+                    return $" --namespace {@namespace}";
+                }
+                return String.Empty;
+            }
+
             string GetResourceFullName(IKubernetesResource r)
             {
                 switch (r)
                 {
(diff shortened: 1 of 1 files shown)


tomkerkhove (approved):
I've missed that one, sorry about that @brendandburns .

Another confirmation we need those unit/integration tests for the Kubernetes features

tomkerkhove:
@ahmelsayed Can you take a look at this one please?

src/Azure.Functions.Cli/Kubernetes/KubectlHelper.cs
```diff
@@ -135,16 +135,25 @@ async Task TimeoutFunc()
                 }
             }
 
+            static string MaybeNamespace(string @namespace)
+            {
+                if (string.IsNullOrWhiteSpace(@namespace) == false)
+                {
+                    return $" --namespace {@namespace}";
+                }
+                return " --namespace default";
```
ahmelsayed, line 144: I think this should return `string.Empty` to use the default namespace from kubeconfig rather than `default`, right?

tomkerkhove: Jup, should be ommited then indeed

brendandburns: done.

brendandburns:
Comment addressed, please take another look.

tomkerkhove (approved):

ahmelsayed (approved):

Issue #2236, opened 2020-09-21
Build native deps should check Azure Python version


Hazhzeng:
Assume on local devbox, user uses python 3.6 for development, when they deploy to a python 3.7 app, due to the python dependencies are resolved using the 3.6 images (e.g.  mcr.microsoft.com/azure-functions/python:2.0.12493-python3.6-buildenv), the app will not work.

We should at least add a warning message for this scenario.

https://github.com/Azure/azure-functions-core-tools/blob/55f206c9eb26cea0a801090284293d2334765841/src/Azure.Functions.Cli/Helpers/PythonHelpers.cs#L442-L453

cc: @ankitkumarr

Hazhzeng:
@stefanushinardi

ankitkumarr:
I wonder if even out of `--build-native-deps`, we should check if Python version locally doesn't match `FUNCTIONS_WORKER_RUNTIME_VERSION` and throw a warning when publishing to a Python function app. We should do this even for `--build remote` or other scenarios, to encourage consistency between local and Azure environment.

stefanushinardi:
Hey @fabiocav, if you could take a look at this during sprint planning and possibly prioritize it for this sprint that would be great. Being pinged by DevDiv as they are tracking this in ML e2e review.

PR #2453, opened 2021-02-17
Python version warning (when local != remote)

mhoeger:
Resolves https://github.com/Azure/azure-functions-core-tools/issues/2236

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -292,11 +292,10 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 // Check if remote LinuxFxVersion exists and is different from local version
                 if (!string.IsNullOrEmpty(functionApp.LinuxFxVersion))
                 {
-                    string localImageVersion = await PythonHelpers.ChoosePythonBuildEnvImage();
-                    if (!string.Equals(localImageVersion, functionApp.LinuxFxVersion))
+                    string localVersion = (await PythonHelpers.GetEnvironmentPythonVersion()).Version;
+                    if (!string.IsNullOrEmpty(localVersion) && !functionApp.LinuxFxVersion.Contains(localVersion))
                     {
-                        string localVersion = (await PythonHelpers.GetEnvironmentPythonVersion()).Version;
-                        ColoredConsole.WriteLine(WarningColor($"Local python version '{localVersion}' is different from the published version that will be run in Azure on image '{functionApp.LinuxFxVersion}'."));
+                        ColoredConsole.WriteLine(WarningColor($"Local python version '{localVersion}' is different from the version expected for your deployed Function App. This may result in dependencies being built incorrectly. The image running your deployed Function App is '{functionApp.LinuxFxVersion}'."));
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -276,10 +276,21 @@ public override async Task RunAsync()
...
+                // Check if azure-functions-worker exists in requirements.txt for Python function app
                 await PythonHelpers.WarnIfAzureFunctionsWorkerInRequirementsTxt();
+                // Check if remote LinuxFxVersion exists and is different from local version
+                if (!string.IsNullOrEmpty(functionApp.LinuxFxVersion))
+                {
+                    string localImageVersion = await PythonHelpers.ChoosePythonBuildEnvImage();
+                    if (!string.Equals(localImageVersion, functionApp.LinuxFxVersion))
+                    {
+                        string localVersion = (await GetEnvironmentPythonVersion()).Version;
+                        ColoredConsole.WriteLine(WarningColor($"Local python version '{localVersion}' is different from the published version that will be run in Azure on image '{functionApp.LinuxFxVersion}'."));
```
stefanushinardi, line 291: nit: might be good to call out that your app might not work in the cloud due to the difference ?  I remember you might get a weird error like "module not found" when deploying 3.6 to 3.7 because the location of the dep folder moved, which makes me wonder if it is best to just fail the operation @Hazhzeng wdyt?

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -276,10 +276,21 @@ public override async Task RunAsync()
...
             }
 
-            // Check if azure-functions-worker exists in requirements.txt for Python function app
+            // Special checks for python dependencies
             if (workerRuntime == WorkerRuntime.python)
             {
+                // Check if azure-functions-worker exists in requirements.txt for Python function app
                 await PythonHelpers.WarnIfAzureFunctionsWorkerInRequirementsTxt();
+                // Check if remote LinuxFxVersion exists and is different from local version
+                if (!string.IsNullOrEmpty(functionApp.LinuxFxVersion))
```
Hazhzeng, line 285: Also my need an `else` statement, since when `functionApp.LinuxFxVersion` is empty, that stands for `Python|3.6`.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -276,10 +276,21 @@ public override async Task RunAsync()
...
+                {
+                    string localImageVersion = await PythonHelpers.ChoosePythonBuildEnvImage();
+                    if (!string.Equals(localImageVersion, functionApp.LinuxFxVersion))
```
Hazhzeng: The `LinuxFxVersion` cannot be used to compare with the localImageVersion directly. Since localImageVersion and LinuxFxVersion returns different format.
```
functionApp.LinuxFxVersion = "PYTHON|<major>.<minor>"
localImageVersion  = "mcr.microsoft.com/azure-functions/python:2.0.14786-python3.6-buildenv"
```

I would suggest moving this piece of logic into PythonHelper.cs, parsing the <major> and <minor> version from LinuxFxVersion, and compare the <major> <minor> with `GetEnvironmentPythonVersion().Major` `GetEnvironmentPythonVersion().Minor`.

You can find `AssertPythonVersion` as a reference.

Hazhzeng (approved):
LGTM :shipit:

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
Hazhzeng: Since we allowed customer to publish onto different Python version in old days, and this operation occurs in their local machine, we may not want to completely block this scenario. I think adding one more line in the warning should be sufficient:
`$"This may result in 'ModuleNotFound' error in Azure Functions. Please create a Python {localVersion.Major}.{localVersion.Minor} virtual environment on your local machine for development."`

PR #2460, opened 2021-02-19
Add the ability to emit a config file instead of deploying.

brendandburns:


src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs
```diff
@@ -148,12 +152,20 @@ public override async Task RunAsync()
...
                     tasks.Add(DockerHelpers.DockerPush(resolvedImageName, false));
                 }
+                if (WriteConfigs)
+                {
+                    var yaml = KubernetesHelper.SerializeResources(resources, OutputSerializationOptions.Yaml);
+                    await File.WriteAllTextAsync(ConfigFile, yaml);
+                    Console.Write($"Configuration written to {ConfigFile}");
+                    return;
+                }
+
```
ahmelsayed, line 162: Should this be an `else if` on the outer if/else? 

```
if (DryRun)
{
    ....
}
else if (WriteConfigs)
{
    ....
}
else
{
    ...
}
```

Otherwise we're not awaiting the createNamespace or the docker build tasks which will cause them to start but finish randomly.

brendandburns: Cleaned up the logic here a bunch, please take another look.

brendandburns:
Comment addressed. Please take another look.

ahmelsayed (approved):

PR #2476, opened 2021-03-01
Add .NET Isolated support for all commands

ankitkumarr:


src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs
```diff
@@ -155,33 +174,85 @@ public static string GetCsprojOrFsproj()
...
+            {
+                return WebJobsTemplateOpetation(action);
+            }
+        }
+
+        private static async Task IsolatedTemplateOperation(Func<Task> action)
+        {
+            try
+            {
+                await UninstallWebJobsTemplates();
```
soninaren, line 195: Why is this needed? Is there an identity conflict?

soninaren (approved):

src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs
ankitkumarr: As discussed offline, we will follow-up with a separate PR as changing the ids require confirmation with tooling partners and we will look into that.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -306,6 +314,43 @@ public override async Task RunAsync()
...
+            if (functionApp.IsDynamic)
+            {
+                if (!string.IsNullOrEmpty(functionApp.LinuxFxVersion))
+                {
+                    updatedSettings = new Dictionary<string, string>
+                    {
+                        [Constants.LinuxFxVersion] = ""
+                    };
+                }
+            }
```
Hazhzeng, line 329: Hmm, even if this is Linux Consumption, we should still set it to "DOTNET-ISOLATED|5.0" right? This should be similar to "PYTHON|3.6", which will use the default mesh image.

ankitkumarr: It unfortunately doesn't work right now. Linux consumption blocks setting it to unknown values and it doesn't know this value. The fix is merged in the platform, but it will take some time for it to be rolled.

PR #2479, opened 2021-03-02
Updated WebJost to 3.0.15417, workers to:

kashimiz:
* Java: 1.8.2-SNAPSHOT
* Python: 3.1.1.12

### DIFF
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -211,6 +211,7 @@ namespace Build
                 "System.Diagnostics.PerformanceCounter.dll",
                 "System.IdentityModel.Tokens.Jwt.dll",
                 "System.Interactive.Async.dll",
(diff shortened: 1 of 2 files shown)


build/Settings.cs
```diff
@@ -211,6 +211,7 @@ public class SignInfo
                 "System.Diagnostics.PerformanceCounter.dll",
                 "System.IdentityModel.Tokens.Jwt.dll",
                 "System.Interactive.Async.dll",
+                "System.Memory.Data.dll",
```
pragnagopa, line 214: intentional?

kashimiz: Yes; we treat DLLs outside of the Functions and WebJobs namespaces as third-party.

pragnagopa (approved):

PR #2483, opened 2021-03-03
Improve githash calculation by adding an implementation that matches github actions.

brendandburns:
See https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

and

https://github.com/actions/runner/blob/master/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts


ahmelsayed (approved):
interesting. Thanks for the contribution!

PR #2503, opened 2021-03-17
K8s deployment for MSSQL-based Durable Functions

cgillum:
### Context for this PR
This PR adds support for deploying a Durable Functions app to Kubernetes using the `func kubernetes deploy` command. It's specifically designed to use the new [MSSQL backend for Durable Functions](https://microsoft.github.io/durabletask-mssql) and relies on the [MSSQL KEDA scaler](https://keda.sh/docs/2.2/scalers/mssql/) that is part of KEDA v2.2.  Support for other backends, like the default Azure Storage backend or the new [Netherite backend](https://microsoft.github.io/durabletask-netherite) will come at another time.

### The changes
I had to refactor some of the existing code because unlike all other trigger types, Durable Functions triggers are _not_ 1:1 with respect to scalers. Rather, all Durable triggers in a function app are powered by a single scaler.

### Testing
I tested this using a Durable Functions app named **durabletask-mssql-app** that I'm working on. Prior to this PR, the `func kubernetes deploy` command resulted in the following yaml for the ScaledObject:

```yaml
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: durabletask-mssql-app
  namespace: default
  labels: {}
spec:
  scaleTargetRef:
    name: durabletask-mssql-app
  triggers:
  - type: orchestrationtrigger
    metadata: {}
  - type: entitytrigger
    metadata: {}
  - type: orchestrationtrigger
    metadata: {}
  - type: orchestrationtrigger
    metadata: {}
  - type: activitytrigger
    metadata: {}
  - type: activitytrigger
    metadata: {}
  - type: activitytrigger
    metadata: {}
```

The tool doesn't know anything about Durable, so it creates a bogus set of KEDA triggers, one for each of my orchestration, activity, and entity triggers. However, after my changes, the same command now correctly generates a single scaler that covers all the Durable triggers:

```yaml
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: durabletask-mssql-app
  labels: {}
spec:
  scaleTargetRef:
    name: durabletask-mssql-app
  triggers:
  - type: mssql
    metadata:
      connectionStringFromEnv: SQLDB_Connection
```

### DIFF
--- a/src/Azure.Functions.Cli/Kubernetes/KEDA/V2/KedaV2Resource.cs
+++ b/src/Azure.Functions.Cli/Kubernetes/KEDA/V2/KedaV2Resource.cs
@@ -41,20 +41,56 @@ namespace Azure.Functions.Cli.Kubernetes.KEDA.V2
                         .Where(b => b?["type"] != null)
                         .Where(b => b["type"].ToString().IndexOf("Trigger", StringComparison.OrdinalIgnoreCase) != -1)
                         .Where(b => b["type"].ToString().IndexOf("httpTrigger", StringComparison.OrdinalIgnoreCase) == -1)
-                        .Select(t => new ScaledObjectTriggerV1Alpha1
-                        {
-                            Type = GetKedaTriggerType(t["type"]?.ToString()),
-                            Metadata = PopulateMetadataDictionary(t)
-                        })
+                        .GroupBy(b => IsDurable(b)) // Multiple durable triggers map to a single scaler
+                        .SelectMany(group => group.Key ? GetDurableScalar(triggers.HostJson) : group.Select(GetStandardScalar))
                 }
             };
         }
 
-        private const string ConnectionField = "connection";
-        private const string ConnectionFromEnvField = "connectionFromEnv";
+        private static bool IsDurable(JToken trigger) => 
+            trigger["type"].ToString().Equals("orchestrationTrigger", StringComparison.OrdinalIgnoreCase) ||
+            trigger["type"].ToString().Equals("activityTrigger", StringComparison.OrdinalIgnoreCase) ||
+            trigger["type"].ToString().Equals("entityTrigger", StringComparison.OrdinalIgnoreCase);
+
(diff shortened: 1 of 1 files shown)


cgillum:
Also adding @TsuyoshiUshio since I know you created an external scaler for Durable Functions with the Azure Storage backend. I'm not sure whether the `func kubernetes deploy` command is a good way to integrate with external scalers, but it would be good to get your opinion.

src/Azure.Functions.Cli/Kubernetes/KEDA/V2/KedaV2Resource.cs
```diff
@@ -41,20 +41,56 @@ public class KedaV2Resource : KedaResourceBase
...
+            // Custom storage types are supported starting in Durable Functions v2.4.2
+            if (string.Equals(storageType, "MicrosoftSQL", StringComparison.OrdinalIgnoreCase))
+            {
+                yield return new ScaledObjectTriggerV1Alpha1
+                {
+                    // MSSQL scaler reference: https://keda.sh/docs/2.2/scalers/mssql/
+                    Type = "mssql",
+                    Metadata = new Dictionary<string, string>
+                    {
+                        ["connectionStringFromEnv"] = storageProviderConfig?["connectionStringName"]?.ToString(),
```
ahmelsayed, line 70: do you need to add other metadata for the scaler, `query`, `targetValue`?

cgillum: Yes. I left it out because none of the other Functions triggers actually included this info.  Is that because they have default values if not specified? I can add this.

ahmelsayed: correct. I think the `query` here is analogous to `queueName` for example. We don't set "targetValue" by default since it's not a concept in Azure/Azure Functions, but Keda has always a default for that.

ahmelsayed (approved):

PR #2506, opened 2021-03-19
Kubernetes deploy fixes

ahmelsayed:
- don't append ":latest" to image name
  - replace \r\n line endings in shell script
  - double quote numerical strings in YamlDotnet


cgillum (approved):
I verified that my deploy scenario is working after taking these fixes. Thanks!

PR #2521, opened 2021-03-24
Add x permissions for customHandler executable when publishing

ahmelsayed:


src/Azure.Functions.Cli/Helpers/HostHelpers.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class HostHelpers
+    {
+        public static async Task<string> GetCustomHandlerExecutable()
+        {
+            if (!FileSystemHelpers.FileExists(Constants.HostJsonFileName))
+            {
+                throw new InvalidOperationException();
+            }
+
+            var hostJson = JsonConvert.DeserializeObject<JToken>(await FileSystemHelpers.ReadAllTextFromFileAsync(Constants.HostJsonFileName));
```
pragnagopa, line 19: This would set permissions just for one file. We do want to preserve permissions that users build with. 
Fix in the host for Linux Consumption was a work-around - but does not solve the issue completely as launching any other executable will fail.

ahmelsayed: I'm not sure how to do that correctly when the cli is packaging files on Windows. Windows files don't have an executable permission. When running on Linux, this zip code should read the correct permissions and set them on the file in the zip archive. But when running on Windows, no file is ever marked as executable. We can set them all as executable, but we should avoid that. 

With RUN_FROM_PACKAGE in linux, the file system is readonly, so I think the host setting the permission on the file probably doesn't work.

pragnagopa: Thanks for clarifying.  I missed this is specific to building on windows.

src/Azure.Functions.Cli/Helpers/HostHelpers.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Threading.Tasks;
+using Azure.Functions.Cli.Common;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Azure.Functions.Cli.Helpers
+{
+    public static class HostHelpers
+    {
+        public static async Task<string> GetCustomHandlerExecutable()
```
pragnagopa, line 12: Add unit tests

pragnagopa (approved):

src/Azure.Functions.Cli/Helpers/HostHelpers.cs
ahmelsayed: done

pragnagopa (approved):

PR #2539, opened 2021-04-06
Update app kind check for kubernetes

ahmelsayed:


pragnagopa (approved):
Unit tests?

TsuyoshiUshio (approved):
Change it's self looks good. However, I don't know the context of the change It change the kind from `kubeapp` to `kubernetes` and for `KubeApp` it enables to show the FunctionsInfo. 
Why this change is happening?

PR #2566, opened 2021-04-21
first move to v4

brettsam:
now ready to merge

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -25,10 +25,11 @@ namespace Azure.Functions.Cli.Actions.AzureActions
     [Action(Name = "publish", Context = Context.Azure, SubContext = Context.FunctionApp, HelpText = "Publish the current directory contents to an Azure Function App. Locally deleted files are not removed from destination.")]
     internal class PublishFunctionAppAction : BaseFunctionAppAction
     {
-        private const string _hostVersion = "3";
+        private const string _hostVersion = "4";
 
         private readonly ISettings _settings;
         private readonly ISecretsManager _secretsManager;
+        private const string _requiredNetFrameworkVersion = "6.0";
 
         public bool PublishLocalSettings { get; set; }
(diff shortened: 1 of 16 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -198,11 +199,17 @@ public override async Task RunAsync()
...
-                            "You can pass --force to force update the app to v3, or downgrade to v1 or v2 tooling for publishing.");
+                            throw new CliException($"You're trying to use v4 tooling to publish to a non-v4 function app ({Constants.FunctionsExtensionVersion} is set to {version}).\n" +
+                            "You can pass --force to force update the app to v4, or downgrade tooling for publishing.");
                         }
                     }
                 }
+
+                if (string.Compare(functionApp.NetFrameworkVersion, _requiredNetFrameworkVersion, StringComparison.OrdinalIgnoreCase) != 0)
+                {
+                    await DotnetIsolatedWindowsValidation(functionApp, _requiredNetFrameworkVersion,
```
ankitkumarr, line 210: Would you mind explaining this addition here?

Is this to make sure the `netFrameworkVersion` is `v6.0` for all apps? If so, the function name `DotnetIsolatedWindowsValidation` might be confusing. Also this code-path (`DotnetIsolatedWindowsValidation`) is called again for .NET isolated in Windows below, which could again overwrite the `netFrameworkVersion` (if trying to run .NET 5). Do we need to remove that call?
If this is temporary, we can add some TODO to re-visit the reorganization.

brettsam: For v4, everything needs to run on v6.0 (or v7.0 eventually if you want to target that with dotnet-isolated). You can't deploy a v4 app to .NET 5.0 as the host itself will require .NET 6.0.

I agree that method needs a rename and/or refactor. Just so I don't bork up the v3.x -> v4.x merges, I will follow up with an update to the v3.x branch and then merge that forward to v4.x.

ankitkumarr (approved):

PR #2568, opened 2021-04-22
Fixing publish action for kubenertes

TsuyoshiUshio:
I create a simple fix for `func azure publish` command 

## Known Issue

The command will success the deployment, however, after the deployment, when the core tools fetch the functions information that fails with 500. It is known issue. That will be fixed on the Geo master side, so that I keep the current code. 

CC @pragnagopa

```
https://management.azure.com/subscriptions/{SUBSCRIPTION_ID}/resourceGroups/LimaArcNcsEnv05/providers/Microsoft.Web/sites/tsushihttp6/functions?api-version=2015-08-01

-> Internal Server Error  (500)
```

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -367,7 +367,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             // For dedicated linux apps, we do not support run from package right now
             var isFunctionAppDedicatedLinux = functionApp.IsLinux && !functionApp.IsDynamic && !functionApp.IsElasticPremium &&!functionApp.IsKubeApp;
 
-            // For Lima app, we do not support run from package right now
+            // For Kubernetes Environment app, we do not support run from package right now
             var isKubeApp = functionApp.IsLinux && functionApp.IsKubeApp;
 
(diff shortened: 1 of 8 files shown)


pragnagopa (changes requested):
support for run from package will be added soon. For existing issues - they should be fixed  as well. We should not put in temporary work-arounds

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -506,6 +515,11 @@ private async Task HandleLinuxDedicatedPublish(Site functionApp, Func<Task<Strea
             await PerformAppServiceRemoteBuild(zipStreamFactory, functionApp);
         }
 
+        private async Task HandleLimaAppPublish(Site functionApp, Func<Task<Stream>> zipStreamFactory)
```
pragnagopa, line 518: rename to ` HandleKubeAppPublish`

pragnagopa: this method is not used anywhere?

pragnagopa (changes requested):
added couple of comments

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
TsuyoshiUshio: Sorry it was old method that is unused.

TsuyoshiUshio:
Latest v3 branch already has fix.

PR #2569, opened 2021-04-22
Upgrade pipeline task PublishBuildArtifacts to PublishPipelineArtifacts

Hazhzeng:
As per https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/build-artifacts?view=azure-devops&tabs=yaml.
We should use PublishPipelineArtifacts instead of PublishBuildArtifacts to yield more uploading speed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -78,7 +78,7 @@ jobs:
       Pattern: '*.dll, *.exe'
       signConfigType: 'inlineSignParams'
       inlineOperation: |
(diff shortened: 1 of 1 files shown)


TsuyoshiUshio (commented):
LGTM

TsuyoshiUshio:
We should change other ones if we use it. Thank you Roger!

TsuyoshiUshio (approved):
LGTM

PR #2570, opened 2021-04-27
Support PrintFunctionsInfo for Kubernetes Environment

TsuyoshiUshio:
PrintFunctionsInfo doesn't show anything if we deploy to the Kubernetes Environment. The reason is, just after the deployment, the restart of the worker process will started. So that we need to wait until the process is re-started again. 

We use the API `/admin/host/status` on Azure Functions Host that checks `processUptime` for watching the time. 
Currently, I set it as 60 sec according to the log and the timeout time of the `worker` process startup. 

I'm limited this behavior is happens only on `Kubernetes Environment`. 

# Sample output 

```bash
$ func.exe azure functionapp publish tsushihttp56
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
Uploading 1.45 KB [###############################################################################]
Upload completed successfully.
Waiting for the functions host up and running ....... done
Functions in tsushihttp56:
    HttpTrigger - [httpTrigger]
        Invoke url: https://tsushihttp56.<<mydomain>>/api/httptrigger
```

CC @pragnagopa

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -459,7 +459,7 @@ namespace Azure.Functions.Cli.Helpers
                 await WaitUntilFunctionAppReadyAsync(functionApp, accessToken, managementURL, showKeys);
             }
 
-            ArmArrayWrapper<FunctionInfo> functions = null;
+            ArmArrayWrapper<FunctionInfo> functions = await GetFunctions(functionApp, accessToken, managementURL);
 
             functions = await GetFunctions(functionApp, accessToken, managementURL);
 
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -421,6 +421,7 @@ namespace Azure.Functions.Cli.Helpers
                       };
 
                       var response = await functionAppReadyClient.SendAsync(request);
+                      ColoredConsole.Write(".");
 
                       if (response.IsSuccessStatusCode)
                       {
@@ -430,6 +431,7 @@ namespace Azure.Functions.Cli.Helpers
                           // Wait 60sec for the readiness of the worker process rebooting by deployment.
                           if (processUpTime >= Constants.DefaultWorkerProcessUptimeReadiness)  
                           {
+                              ColoredConsole.WriteLine(" done");
                               return;
                           }
                           else
@@ -448,8 +450,8 @@ namespace Azure.Functions.Cli.Helpers
 
 
     public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, string managementURL, bool showKeys)
-        {
-
+    {
+            ColoredConsole.Write("Waiting for the functions host up and running ");
             await WaitUntilFunctionAppReadyAsync(functionApp, accessToken, managementURL, showKeys);
 
             ArmArrayWrapper<FunctionInfo> functions = null;
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -410,19 +410,19 @@ namespace Azure.Functions.Cli.Helpers
                 functionAppReadyClient.DefaultRequestHeaders.Add("Accept", jsonContentType);
             }
 
-            functionAppReadyClient.DefaultRequestHeaders.Add("x-ms-request-id", Guid.NewGuid().ToString());
-            var uri = new Uri($"https://{functionApp.HostName}/admin/host/status?code={masterKey}");
-            var request = new HttpRequestMessage()
-            {
-                RequestUri = uri,
-                Method = HttpMethod.Get
-            };
-
             await RetryHelper.Retry(async () =>
                   {
+                      functionAppReadyClient.DefaultRequestHeaders.Add("x-ms-request-id", Guid.NewGuid().ToString());
+                      var uri = new Uri($"https://{functionApp.HostName}/admin/host/status?code={masterKey}");
+                      var request = new HttpRequestMessage()
+                      {
+                          RequestUri = uri,
+                          Method = HttpMethod.Get
+                      };
+
                       var response = await functionAppReadyClient.SendAsync(request);
 
-                      if (!response.IsSuccessStatusCode)
+                      if (response.IsSuccessStatusCode)
                       {
                           var json = await response.Content.ReadAsAsync<JToken>();
                           var processUpTime = json["processUptime"]?.ToObject<int>() ?? 0;
@@ -453,11 +453,8 @@ namespace Azure.Functions.Cli.Helpers
             await WaitUntilFunctionAppReadyAsync(functionApp, accessToken, managementURL, showKeys);
 
             ArmArrayWrapper<FunctionInfo> functions = null;
-            await RetryHelper.Retry(async () =>
-            {
-                functions = await GetFunctions(functionApp, accessToken, managementURL);
-                if (!functions.value.Any()) throw new Exception($"Cannot get functions information: {functionApp.HostName}");
-            }, 3, TimeSpan.FromSeconds(5));
+
+            functions = await GetFunctions(functionApp, accessToken, managementURL);
 
             ColoredConsole.WriteLine(TitleColor($"Functions in {functionApp.SiteName}:"));
             foreach (var function in functions.value.Select(v => v.properties))
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -9,6 +9,8 @@ using Azure.Functions.Cli.Arm.Models;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Extensions;
 using Colors.Net;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.Extensions.Http.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.OutputTheme;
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -359,7 +359,13 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
 
         public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, string managementURL, bool showKeys)
         {
-            var functions = await GetFunctions(functionApp, accessToken, managementURL);
+            ArmArrayWrapper<FunctionInfo> functions = null;
+            await RetryHelper.Retry(async () =>
+            {
+                functions = await GetFunctions(functionApp, accessToken, managementURL);
+                if (!functions.value.Any()) throw new Exception($"GetFunctions is not ready for {functionApp}");
```
pragnagopa, line 366: it is possible to publish a function app without any functions. We should only poll if function app is just published and if the content of the function app locally has any functions

TsuyoshiUshio:
Hi @ahmelsayed 
I change this PR as you suggested. Could you review this? 

The change introduced.   Before Feting GetFunction(), It wait until the `/admin/host/status` response `processUptime` reaches 60000ms (I've got the time from the log  and it is also the worker startup timeout).  Then call GetFunction(). 

CC: @pragnagopa

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: Hi @pragnagopa  
I made a change. try not to retry the GetFunctions. Sounds good?

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -155,6 +157,36 @@ internal static async Task<string> GetFunctionKey(string functionName, string ap
             return key;
         }
 
+        internal static async Task<string> GetMasterKeyAsync(string appId, string accessToken, string managementURL) {
```
ahmelsayed, line 160:
```suggestion
        internal static async Task<string> GetMasterKeyAsync(string appId, string accessToken, string managementURL)
        {
```

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +389,75 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
             }
         }
 
-        public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, string managementURL, bool showKeys)
+        private static HttpClient functionAppReadyClient = null;
```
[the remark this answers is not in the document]
TsuyoshiUshio: Hi @ahmelsayed Thank you for the review! I try not to instantiate HttpClient every time. Is there any way to keep the static variables inside of the method? 

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-3.1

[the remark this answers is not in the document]
ahmelsayed: In the CLI context, the process always exists after it's done with the command, so we don't have to worry about connections leaking over time. `WaitUntilFunctionAppReadyAsync()` is called once and the process terminates. But yeah, if this was a long running service then we would want to cache the client.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +390,75 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+
+        internal static async Task WaitUntilFunctionAppReadyAsync(Site functionApp, string accessToken, string managementURL,
+            bool showKeys, HttpMessageHandler messageHandler = null)
         {
-            var functions = await GetFunctions(functionApp, accessToken, managementURL);
+            var masterKey = await GetMasterKeyAsync(functionApp.SiteId, accessToken, managementURL);
```
pragnagopa, line 398: GetMasterKeyAsync can return null. Add check for null before using GetMasterKeyAsync

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +390,75 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+                  }, 6, TimeSpan.FromSeconds(10)
+            );
+        }
+
+
+
+    public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, string managementURL, bool showKeys)
+    {
+            ColoredConsole.Write("Waiting for the functions host up and running ");
+            await WaitUntilFunctionAppReadyAsync(functionApp, accessToken, managementURL, showKeys);
```
pragnagopa, line 456: Can we scope the this change to just kubernetes? Otherwise this is a change in behavior for existing apps

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +390,75 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+                              return;
+                          }
+                          else
+                          {
+                              throw new Exception($"GetFunctions is not ready. Hostname: {functionApp.HostName}");
+                          }
+                      }
+                      else
+                      {
+                          throw new Exception($"Status check returns unsuccessful status. Hostname: {functionApp.HostName}");
```
pragnagopa, line 445: do not throw exceptions as this is being retried. Only throw exception if last retry is unsuccessfull

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +390,75 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+                      var response = await functionAppReadyClient.SendAsync(request);
+                      ColoredConsole.Write(".");
+
+                      if (response.IsSuccessStatusCode)
+                      {
+                          var json = await response.Content.ReadAsAsync<JToken>();
+                          var processUpTime = json["processUptime"]?.ToObject<int>() ?? 0;
+
+                          // Wait 60sec for the readiness of the worker process rebooting by deployment.
+                          if (processUpTime >= Constants.DefaultWorkerProcessUptimeReadiness)
```
pragnagopa, line 433: waiting 60secs does seem like a long delay. As Core Tools has knowledge of count of functions deployed , make this check dynamic - instead of relying on processUpTime - if host status returns 200 - make call to get function apps - if response matches list of functions in core tools - no need to wait for 60secs

pragnagopa (changes requested):
Added few comments

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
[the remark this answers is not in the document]
TsuyoshiUshio: sure. Then, I'll include in the method.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: The `processUpTime` spec is decided when I talk with @ahmelsayed, so that I'd like to include him to this conversation.

Do you know where `count of functions deployed` ? How can I get to know the host is ready in this case? the increase of the functions deployed means host is ready?

pragnagopa: As long as this change is scoped to KubeApp, this is ok for now. We can revisit improvement later.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
pragnagopa: Will approve, once this is addressed!

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: Thanks. Very good point. Done.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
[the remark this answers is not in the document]
TsuyoshiUshio: Done the change! Thanks.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: Great! I narrow the scope!

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: Hi @pragnagopa 
Could you help me to clarify this?  I see the `Retry` functions looks retry by Exception. https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Helpers/RetryHelper.cs#L23
I thought that we have two opportunity to retry. 1.  status is not good. (probably functions host is not started yet) and 2. functions host is running but worker is not ready. 

In this case, How can I detect the case 1 without Exception?

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
TsuyoshiUshio: Done!

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
pragnagopa: Ahh thanks for clafication. Retryhelper expects an exception to be thrown and only displays if flag is set. This is OK!

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -357,9 +390,79 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+
+    public static async Task PrintFunctionsInfo(Site functionApp, string accessToken, string managementURL, bool showKeys)
+    {
+            if (functionApp.IsKubeApp)
+            {
+                ColoredConsole.Write("Waiting for the functions host up and running ");
+                await WaitUntilFunctionAppReadyAsync(functionApp, accessToken, managementURL, showKeys);
+            }
+
+            ArmArrayWrapper<FunctionInfo> functions = null;
```
pragnagopa, line 462:
```suggestion
            ArmArrayWrapper<FunctionInfo> functions = await GetFunctions(functionApp, accessToken, managementURL);
```

pragnagopa (approved):

PR #2588, opened 2021-05-12
adding support for dotnet-framework-version during publishing

### LINKED ISSUE #2573: Add "netFrameworkVersion" option to publish for dotnet-isolated
With .NET 6 coming, we need a way to also publish dotnet-isolated to a .NET 6 App Service. Today we assume it is .NET 5.

brettsam:
fixes #2573 

In dotnet-isolated scenarios, we need to allow the customer to specify what version of .NET they want running in their Function app. Today we assume .NET 5, but with .NET 6 coming, we need to allow this to be specified.

Ideally, we'll be able to auto-detect, but we'll still want this to allow for overrides.

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -19,13 +19,14 @@ using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
 using Newtonsoft.Json;
 using static Azure.Functions.Cli.Common.OutputTheme;
-using static Colors.Net.StringStaticMethods;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
     [Action(Name = "publish", Context = Context.Azure, SubContext = Context.FunctionApp, HelpText = "Publish the current directory contents to an Azure Function App. Locally deleted files are not removed from destination.")]
     internal class PublishFunctionAppAction : BaseFunctionAppAction
     {
+        private const string _hostVersion = "3";
+
         private readonly ISettings _settings;
         private readonly ISecretsManager _secretsManager;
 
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -243,14 +249,13 @@ public override async Task RunAsync()
...
-                // For dotnet isolated, as function create options are limited, we update an exisiting App to .NET 5 isolated app,
-                // if the user is trying to deploy that. This is why we need to special case this scenario and set the proper
-                // LinuxFxVersion properties when missing.
-                await DotnetIsolatedLinuxValidation(functionApp);
+                await UpdateDotNetIsolatedFrameworkVersion(functionApp, DotnetFrameworkVersion,
+                    (settings) => AzureHelper.UpdateWebSettings(functionApp, settings, AccessToken, ManagementURL));
             }
-            else if (functionApp.IsLinux && !functionApp.IsDynamic && !string.IsNullOrEmpty(functionApp.LinuxFxVersion))
+
+            if (workerRuntime != WorkerRuntime.dotnetIsolated && functionApp.IsLinux && !functionApp.IsDynamic && !string.IsNullOrEmpty(functionApp.LinuxFxVersion))
```
ankitkumarr, line 258: curious, why not `else if (functionApp.IsLinux &&  ...)`?

ankitkumarr (approved):

ankitkumarr:
@brettsam, do we want to throw if passing `--dotnet-framework-version 6.0` for Linux? It won't be supported right now.

brettsam:
> do we want to throw if passing --dotnet-framework-version 6.0 for Linux? It won't be supported right now.
We can come back and add restrictions if it's a problem. My goal was to open it up a bit for when 6.0 is supported.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
brettsam: good question... think it was an accident. I'll adjust.

PR #2592, opened 2021-05-14
fix: fetching old value of function information after deployment for Kubernetes Environment

TsuyoshiUshio:
When we use Kubernetes Environment, when we deploy the app, it returns immediately, and returns old information of the Azure Functions App. The problem is, the current version is just watch the `/host/admin/status` 's body `processUptime` is exceed `60 sec`. 

I observe when the new Function App information is available. It is on average 15 sec around. 
New Logic will be `processUptime` should be between 30 sec and 45 sec includes some slack by the experiment.  The polling was 5sec, I make it 3 sec for now.  Through the experiment, I experience some customer's pains, so that I improve that points as well.

This PR is the fix and provide additional feature for improving the customer experience of the deployment.  We need to assume several state for this sub command. 

## NOTICE

For the review. For AzureHelper, I just change small place, however, Visual Studio fix the wrong indentation and format document.   What I changed on the AzureHelper is `WaitUntilFunctionAppReadyAsync L447 - L480` and ` PrintFunctionsInfo L513 - 517` If you guys are not comfortable for this review, I'll newly create a branch and introduce a change without formatting. Feel free to let me know.  

CC: @pragnagopa @ahmelsayed 

## 1.  Just after the Function App deployment, publish your app. 
The deployment is not ready. We will see following error. 

```
$ func.exe azure functionapp publish tsushiqueue1906
Can't find app with name "tsushiqueue1906"
```
## 2.  Function App is ready however, the pod processUptime is less than 45 sec. 
The deployment is ready and ok, however, it is too fast, so that fetching Function App info wont' available. There is no way to identify if it is the first deployment or not.  I show the warning message for it.  Deployment itself is fine.

```
$ func.exe azure functionapp publish tsushiqueue1906
Getting site publishing info...
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
Uploading 2.3 KB [################################################################################]
Upload completed successfully.
Waiting for the functions host up and running . done
Functions in tsushiqueue1906:
The deployment happens soon after the Function App deployment. Wait for 40 sec to fetch the trigger info after the function app provisioning has been finished.
```

## 3. Function App lives more than 45 sec 

```
$ func.exe azure functionapp publish tsushiqueue1907
Getting site publishing info...
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
[...]

PR #2593, opened 2021-05-15
fix: fetching old value of function information after deployment for Kubernetes Environment

TsuyoshiUshio:
When we use Kubernetes Environment, when we deploy the app, it returns immediately, and returns old information of the Azure Functions App. The problem is, the current version is just watch the `/host/admin/status` 's body `processUptime` is exceed `60 sec`. 

I observe when the new Function App information is available. It is on average 15 sec around. 
New Logic will be `processUptime` should be between 30 sec and 45 sec includes some slack by the experiment.  The polling was 5sec, I make it 3 sec for now.  Through the experiment, I experience some customer's pains, so that I improve that points as well.

This PR is the fix and provide additional feature for improving the customer experience of the deployment.  We need to assume several state for this sub command. 
Move from https://github.com/Azure/azure-functions-core-tools/pull/2592

CC: @pragnagopa @ahmelsayed 

## 1.  Just after the Function App deployment, publish your app. 
The deployment is not ready. We will see following error. 

```
$ func.exe azure functionapp publish tsushiqueue1906
Can't find app with name "tsushiqueue1906"
```
## 2.  Function App is ready however, the pod processUptime is less than 45 sec. 
The deployment is ready and ok, however, it is too fast, so that fetching Function App info wont' available. There is no way to identify if it is the first deployment or not.  I show the warning message for it.  Deployment itself is fine.

```
$ func.exe azure functionapp publish tsushiqueue1906
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
Uploading 2.3 KB [################################################################################]
Upload completed successfully.
Waiting for the functions host up and running . done
Functions in tsushiqueue1906:
The deployment happens soon after the Function App deployment. Wait for 40 sec to fetch the trigger info after the function app provisioning has been finished.
```

## 3. Function App lives more than 45 sec 

```
$ func.exe azure functionapp publish tsushiqueue1907
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
Uploading 2.3 KB [################################################################################]
Upload completed successfully.
Waiting for the functions host up and running ........... done
Functions in tsushiqueue1907:
    QueueTrigger12 - [queueTrigger]
```

##  4. The case customers don't want to wait 30 sec
Customers need to wait until the new Function App available. For the new change, it will be 30 sec. 
So that customer might want to skip fetching information especially CI scenario. 

I introduce a new flag `--deploy-only` it skip to fetch the function info. Just Deploy the app. 

```
$ func.exe azure functionapp publish tsushiqueue1908 --deploy-only
Creating archive for current directory...
Could not find gozip for packaging. Using DotNetZip to package. This may cause problems preserving file permissions when using in a Linux based environment.
Uploading 2.3 KB [################################################################################]
Upload completed successfully.
```

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -44,7 +44,6 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         public bool NoBuild { get; set; }
         public string DotnetCliParameters { get; set; }
         public string DotnetFrameworkVersion { get; set; }
-        public bool DeployOnly { get; set; }
 
         public PublishFunctionAppAction(ISettings settings, ISecretsManager secretsManager)
         {
@@ -117,10 +116,6 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 .Setup<string>("dotnet-framework-version")
                 .WithDescription("Only applies to dotnet-isolated applications. Specifies the .NET Framework version for the function app. For example, set to '5.0' when publishing a .NET 5 app.")
                 .Callback(s => DotnetFrameworkVersion = s);
-            Parser
-                .Setup<bool>("deploy-only")
-                .WithDescription("Skip showing function trigger info after the deployment.")
-                .Callback(f => DeployOnly = f);
             return base.ParseArgs(args);
         }
 
@@ -492,8 +487,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             // So, we only show the info, if Function App is not Linux Elastic Premium
             // or a Linux Dedicated Function App with remote build
             if (!(functionApp.IsLinux && functionApp.IsElasticPremium)
-                && !(isFunctionAppDedicatedLinux && PublishBuildOption == BuildOption.Remote)
-                && !DeployOnly)
+                && !(isFunctionAppDedicatedLinux && PublishBuildOption == BuildOption.Remote))
             {
                 await AzureHelper.PrintFunctionsInfo(functionApp, AccessToken, ManagementURL, showKeys: true);
             }
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -44,6 +44,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         public bool NoBuild { get; set; }
         public string DotnetCliParameters { get; set; }
         public string DotnetFrameworkVersion { get; set; }
+        public bool DeployOnly { get; set; }
 
         public PublishFunctionAppAction(ISettings settings, ISecretsManager secretsManager)
         {
@@ -116,7 +117,10 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 .Setup<string>("dotnet-framework-version")
                 .WithDescription("Only applies to dotnet-isolated applications. Specifies the .NET Framework version for the function app. For example, set to '5.0' when publishing a .NET 5 app.")
                 .Callback(s => DotnetFrameworkVersion = s);
-
+            Parser
+                .Setup<bool>("deploy-only")
+                .WithDescription("Skip showing function trigger info after the deployment.")
+                .Callback(f => DeployOnly = f);
             return base.ParseArgs(args);
         }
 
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -116,7 +117,10 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .Setup<string>("dotnet-framework-version")
                 .WithDescription("Only applies to dotnet-isolated applications. Specifies the .NET Framework version for the function app. For example, set to '5.0' when publishing a .NET 5 app.")
                 .Callback(s => DotnetFrameworkVersion = s);
-
+            Parser
+                .Setup<bool>("deploy-only")
```
pragnagopa, line 121: Avoid adding `deploy-only` flag  - i do not see any other reasons why will users explicitly pass this flag . If this is intended experience for KubeApp - we should do that by default and let the users pass in verbose flag to see rest of the logs

TsuyoshiUshio: Let me clarify your idea and understand the context. I'd happy to remove the `deploy-only` option.  I'd like to explain the context when I tested it. 
func azure functionapp publish command  is deploy app and shows the list of the triggers.  The behavior is fine, however, we need to wait around 30 sec to see the updated value.  Usually it's useful, however, some people don't need to see the information is updated and don't want to wait to finish the command.

```
Functions in tsushiqueue1907:
    QueueTrigger12 - [queueTrigger]
```

So that I just want to give the customer to have an option whether if they need this info with 30 sec or not. (at least for Kubernetes Environment, I don't know when I do it for other platform) 
In this case, in you idea, in case, by default, in kubernetes environment, we don't need the trigger info, and `--verbose` option reside in the `StartHostAction` In this case, should we move the `VersboseLogging` from `StartHostAction` to `BaseAction` for references from `PublishFunctionAppAction`?  I also think `--verbose` option looks little different for showing trigger information.  The other idea is removing the showing info feature for Kubernetes Environment. 
Do you want to avoid to increase a new behavior for avoiding to introduce new bugs?

pragnagopa: Thanks for explanation. My recommendation is do not add a separate flag - longer delay to display Functions Info for KubeApps is ok - as long as the long polling is scoped to KubeApp.

Also, displaying logs while polling for functions host status is OK!

TsuyoshiUshio: Sure. Then I'll simply remove the `--deploy-only` flag. Thank you for clarifying it!

TsuyoshiUshio: Hi @pragnagopa I removed the debug only flag and test it.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -464,6 +465,11 @@ public static async Task PrintFunctionsInfo(Site functionApp, string accessToken
             functions = await GetFunctions(functionApp, accessToken, managementURL);
 
             ColoredConsole.WriteLine(TitleColor($"Functions in {functionApp.SiteName}:"));
+            if (functionApp.IsKubeApp && !functions.value.Any())
+            {
+                ColoredConsole.WriteLine(WarningColor(
+                    $"The deployment success. Wait for {Constants.DefaultRestartedWorkerProcessUptimeWithin / 1000} sec to fetch the trigger info after the function app provisioning has been finished."));
```
pragnagopa, line 471: change to 
`The deployment succeeded. Waiting for {Constants.DefaultRestartedWorkerProcessUptimeWithin / 1000} sec to fetch the list of functions deployed in the function app.")`

TsuyoshiUshio: Thank you, @pragnagopa 
I love the sentence. Updated.

pragnagopa (approved):

TsuyoshiUshio:
Hi @yojagad 
Do you know why the CI fails? (It is not the place I change. ) Looks Template testing. 

```
C:\Users\tsushi\Code\Lima\test\CoreTools\CIFail>C:\Users\tsushi\Code\CoreTools\azure-functions-core-tools\src\Azure.Functions.Cli\bin\Debug\netcoreapp3.1\func.exe new --template "Http Trigger" --name testfunc
One or more errors occurred. (Can't find templates location. Looked at 'C:\Users\tsushi\Code\CoreTools\azure-functions-core-tools\src\Azure.Functions.Cli\bin\Debug\netcoreapp3.1\templates\templates.json')
```

Code 
https://github.com/Azure/azure-functions-core-tools/blob/v3.x/test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs#L145-L158

Probably, related bundle template release?

TsuyoshiUshio:
Looks Transient!

PR #2617, opened 2021-06-08
refactoring publish action

brettsam:
Creating the `UpdateFrameworkVersions` method to make this easier to test in v4, where the logic has a few more steps.


ankitkumarr (approved):
Looks good to me

PR #2623, opened 2021-06-15
fix typo in RunAsync()

mburleigh:
Other console messages are in the form of "Press any key to <do something>"; I added the word "key" the console output.


michaelpeng36 (approved):
LGTM

PR #2659, opened 2021-07-26
Multi-stage docker builds for python

peterkh:
More closely match what happens when publishing a python function app.

Using the `-buildenv` image for doing pip installs. Without this packages like `pyodbc` fail to install as there is no gcc installed in function rumtime image.


kshyju:
@peterkh Is this change still valid/relevant?

@vrdmr @gavin-aguiar Could you please review?

soninaren:
@peterkh v3.x is not in active development. Please send a PR against v4.x branch if the change is still valid / relevant.

PR #2667, opened 2021-07-28
Adding ability to select the .NET language on create

aaronpowell:
When adding the F# templates for dotnet-isolated (see: https://github.com/Azure/azure-functions-templates/pull/1074) I dug around to figure out just how the tools worked to create new projects.

This led me to figure out how the dotnet setup was done and that adding a language selector into the `func init` pipeline wouldn't be that hard. So now we have the ability to select which .NET language we want to use for both in-proc and isolated models.

As a result, F# projects can be managed using the `func` CLI rather than having to go via the `dotnet` cli.


aaronpowell:
I've done some local testing of this, using the new templates, from the above mentioned pack, and it was able to create both C# and F# across dotnet and isolated.

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -41,7 +41,7 @@ public static class WorkerRuntimeLanguageHelper
         private static readonly IDictionary<WorkerRuntime, string> workerToDefaultLanguageMap = new Dictionary<WorkerRuntime, string>
         {
             { WorkerRuntime.dotnet, Constants.Languages.CSharp },
-            { WorkerRuntime.dotnetIsolated, Constants.Languages.CSharp },
+            { WorkerRuntime.dotnetIsolated, Constants.Languages.CSharpIsolated },
```
dsyme, line 44: Does this need entries for F#?

dsyme: I see it's the defaults

aaronpowell: Yeah, figured that the default of C# makes the most sense

anthonychu:
@fabiocav Adding it to the next sprint to see if we can have someone review this. It's only going into v4. I think @aaronpowell can help create a PR to back port it to v3 as well if this is approved.

fabiocav:
Apologies for the delay here. Reassigning this for review. 

@aaronpowell would it be possible to update this PR against the latest code? Thanks!

aaronpowell:
@fabiocav done

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -25,7 +25,7 @@ public static class WorkerRuntimeLanguageHelper
         private static readonly IDictionary<WorkerRuntime, IEnumerable<string>> availableWorkersRuntime = new Dictionary<WorkerRuntime, IEnumerable<string>>
         {
             { WorkerRuntime.dotnet, new [] { "c#", "csharp", "f#", "fsharp" } },
-            { WorkerRuntime.dotnetIsolated, new [] { "dotnet-isolated" } },
+            { WorkerRuntime.dotnetIsolated, new [] { "dotnet-isolated", "c#-isolated", "csharp-isolated", "f#-isolated", "fsharp-isolated" } },
```
michaelpeng36, line 28: Do we still need `dotnet-isolated` in the array containing `c#-isolated` and `csharp-isolated` here?

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -55,7 +55,7 @@ public static class WorkerRuntimeLanguageHelper
             { Constants.Languages.TypeScript, new [] { "ts" } },
             { Constants.Languages.Python, new [] { "py" } },
             { Constants.Languages.Powershell, new [] { "pwsh" } },
-            { Constants.Languages.CSharp, new [] { "csharp", "dotnet", "dotnet-isolated", "dotnetIsolated" } },
+            { Constants.Languages.CSharpIsolated, new [] { "csharp", "dotnet", "dotnet-isolated", "dotnetIsolated" } },
```
michaelpeng36, line 58: I believe there should be separate entries here: one with `{ Constants.Languages.CSharp, new [] { "csharp", "dotnet" } }` and another with `{ Constants.Languages.CSharpIsolated, new [] { "dotnet-isolated", "dotnetIsolated" } }`

michaelpeng36 (changes requested):
@aaronpowell his PR seems mostly correct, but there are a few minor issues that need to be addressed, which I have added in the comments.

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
aaronpowell: I'll have to have a second look at it, but I _think_ the reason I kept `dotnet-isolated` in there was for backwards compatibility, as prior to this change that's how things were identified.

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
aaronpowell: Yes that's probably true, otherwise it won't detect in-proc properly

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
aaronpowell: I've had a bit of a poke around and I don't think it can be removed with a lot of other cascading changes.

Removing `dotnet-isolated` will cause existing projects to fail to launch using `func start` as it's unable to detect the worker type. I also tested changing the worker type in `local.settings.json` to `c#-isolated` but that then errors within the WebJobs dependency as it couldn't find the functions language `c#-isolated`, so there'd be a downstream impact.

Additionally, all templates would need to be updated to either say `c#-isolated` or `f#-isolated` in them before the change would usable.

Ultimately, the additional worker types are probably not doing anything of value at present, they are more future proofing 😅

src/Azure.Functions.Cli/Helpers/WorkerRuntimeLanguageHelper.cs
aaronpowell: Done but I've had to move `csharp` out of the CSharpIsolated constant, otherwise it'll explode later on in `WorkerRuntimeStringToLanguage`, but I'm not quite sure how they all hang together to know if that's a problem (hasn't seem to be in my testing)

michaelpeng36 (approved):
LGTM

dsyme:
Yay!!!!

PR #2677, opened 2021-08-06
feat: Use KEDA v2.4

tomkerkhove:
Change CLI to use KEDA v2.4 CRDs to use the correct deployment for KEDA.

It would be ideal if somebody could pull this version and deploy an application to verify that it works.


tomkerkhove:
Any thoughts @ahmelsayed ?

tomkerkhove:
> @tomkerkhove - changes look straight forward and its minor version update. Can you please Update the description with information on what tests were to verify this change? I believe we do not have good e2e test coverage to verify these changes in the CI
> 
> Tagging @lpapudippu for context

Done!

michaelpeng36:
Closing as this issue is addressed in #2797 and #2799

tomkerkhove:
Too bad my effort was not merged faster but it is what it is 🤷

PR #2682, opened 2021-08-12
Use next available port when the default port is already used

PramodValavala-MSFT:
When working with multiple function apps locally, every function host apart from the first needs to have a different port manually set either via the command line argument `--port` or the `Host.LocalHttpPort` property in `local.settings.json`. Also, in some cases the default port might be blocked with another process (sometimes an older `func` itself that didn't terminate).

This PR adds support to pick up the next available port (7072, 7073, ...) when none is provided. If a port is defined in either place mentioned above, this would not do anything.

One known issue is that multiple function apps run at the exact same time might end up picking the same port, causing one of them to error out. This is present even today with the static default port, primarily because the validation is done before the host starts up.

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
@@ -1,4 +1,8 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Net;
+using System.Net.NetworkInformation;
 using System.Net.Sockets;
 
 namespace Azure.Functions.Cli.Helpers
@@ -32,6 +36,34 @@ namespace Azure.Functions.Cli.Helpers
             {
                 listener.Stop();
             }
+        }
+
+        public static int GetNextAvailablePort(int startPort)
+        {
+            var usedPorts = new List<int>();
+            var ipGlobalProperties = IPGlobalProperties.GetIPGlobalProperties();
+
+            usedPorts.AddRange(ipGlobalProperties
+                                .GetActiveTcpConnections()
+                                .Where(c => c.LocalEndPoint.Port >= startPort)
+                                .Select(c => c.LocalEndPoint.Port));
+            usedPorts.AddRange(ipGlobalProperties
+                                .GetActiveTcpListeners()
+                                .Where(l => l.Port >= startPort)
+                                .Select(l => l.Port));
+
+            usedPorts.Sort();
+
+            foreach (var usedPort in usedPorts)
+            {
+                if (startPort < usedPort)
+                {
+                    return startPort;
+                }
+                startPort++;
+            }
+
+            return GetAvailablePort();
         }
     }
 }
--- a/src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
+++ b/src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
@@ -40,29 +40,36 @@ namespace Azure.Functions.Cli.Helpers
 
         public static int GetNextAvailablePort(int startPort)
         {
-            var usedPorts = new List<int>();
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
```diff
@@ -33,5 +37,33 @@ public static int GetAvailablePort()
...
+            usedPorts.AddRange(ipGlobalProperties
+                                .GetActiveTcpConnections()
+                                .Where(c => c.LocalEndPoint.Port >= startPort)
+                                .Select(c => c.LocalEndPoint.Port));
+            usedPorts.AddRange(ipGlobalProperties
+                                .GetActiveTcpListeners()
+                                .Where(l => l.Port >= startPort)
+                                .Select(l => l.Port));
+
+            usedPorts.Sort();
```
swapnil-nagar, line 55: Why sorting is needed for all used ports, thats an additional complexity of O(NlogN). You are iterating over all the values in usedPorts.

You have two options,
1. Used HashSet and iterate over all the values and lookup will be O(N), then you can do startPort++ and keep looking.
2. Used TreeSet or SortedSet to store and look for value less than startPort.

src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
```diff
@@ -33,5 +37,33 @@ public static int GetAvailablePort()
...
+
+            foreach (var usedPort in usedPorts)
+            {
+                if (startPort < usedPort)
+                {
+                    return startPort;
+                }
+                startPort++;
+            }
+
```
swapnil-nagar, line 65: What if all ports are exhausted, then should we log the errors?

src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
```diff
@@ -33,5 +37,33 @@ public static int GetAvailablePort()
                 listener.Stop();
             }
         }
+
+        public static int GetNextAvailablePort(int startPort)
+        {
```
swapnil-nagar, line 42: ArgumentExceptin when startPort is either using for system process or upper limit of 65536

src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
PramodValavala-MSFT: Done! 👍

src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
PramodValavala-MSFT: Done! 👍

src/Azure.Functions.Cli/Helpers/NetworkHelpers.cs
PramodValavala-MSFT: After checking the ports after the start port, the `GetAvailablePort` method will try to get any available port at random.

jviau:
My concern with this is we are introducing a silent behavior change between debug sessions of the same application. In a complex environment this could lead to very subtle issues. If it is part of an automation environment, or a dev just has a bunch of scripts or something like postman set up, with the known port and then it changes on them, they may investigate for a while before figuring out the port changed.

I personally would prefer a hard failure at startup to tell me that something else is using the default port, and I either need to specify a port or deal with whatever is using that port.

jviau:
We have discussed this internally and we have decided to not take this approach due to the above concern (silent port changes). We have also made improvements to alert of port issues earlier in the startup process since this PR was opened, alleviating the severity of this issue.

PramodValavala-MSFT:
@jviau Thanks for sharing that insight! That does make sense, and I would likely make the same decision now myself. 😄

PR #2711, opened 2021-09-17
Use "release" configuration when building dotnet function apps by default

### LINKED ISSUE #2628: "azure functionapp publish" deploys Debug builds for net5
If we publish our net5 standalone function app via "azure functionapp publish" it gets build in Debug mode. This is a very very bad default for production deployments and also breaks all out `#if DEBUG` compiler directives

kshyju:
Fixes #2628 

I verified it by having [a dotnet function app](https://shkr-cli-publishconfig-test.azurewebsites.net/api/function1) with `#if DEBUG` conditional expressions as below and verifying the result of the endpoint before & after publishing the app using CLI with change from this PR.

    string responseMessage = string.Empty;

    #if DEBUG
        responseMessage = "Debug build - from local CLI-1228"; 
    #else
        responseMessage = "Release build - from local CLI-1228"; 
    #endif

    return new OkObjectResult(responseMessage);

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -44,7 +44,7 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         public bool NoBuild { get; set; }
 
         // For .net function apps, build using "release" configuration by default. User can override using "--dotnet-cli-params" as needed.
-        public string DotnetCliParameters { get; set; } = "-c release";
+        public string DotnetCliParameters { get; set; } = "--configuration release";
         public string DotnetFrameworkVersion { get; set; }
 
         public PublishFunctionAppAction(ISettings settings, ISecretsManager secretsManager)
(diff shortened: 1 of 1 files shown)


brettsam (commented):
Two things:
- We need to update this comment to make sense as well: `.WithDescription("When publishing dotnet functions, the core tools calls 'dotnet build --output bin/publish'. Any parameters passed to this will be appended to the command line.")`
- There's some bug where you need to use a leading space (like `" -c release"`)-- does that not apply here? Can we fix that at the same time?

kshyju:
> We need to update this comment to make sense as well: `.WithDescription("When publishing dotnet functions, the core tools calls 'dotnet build --output bin/publish'. Any parameters passed to this will be appended to the command line.")`

Thanks! Updated the description text to include the configuration option value as well. Feel free to suggest a more correct language if you have one.

> There's some bug where you need to use a leading space (like `" -c release"`)-- does that not apply here? Can we fix that at the same time?

Yes. Either a leading space or a explicit escaping of double quotes is needed when user is explicitly passing that value from command line as the value of `--dotnet-cli-params` option . This is because, the Console app(CLI) will read the arguments as a string array and pass it to the Fclp parser where it will call the [CommandLineOptionGrouper.GroupArgumentsByOption](https://github.com/fclp/fluent-command-line-parser/blob/develop/FluentCommandLineParser/Internals/Parsing/CommandLineOptionGrouper.cs#L57) method, which reads this single args array and convert it to a collection of options (keys & it's values). This method internally [determines an item in the array is an option key](https://github.com/fclp/fluent-command-line-parser/blob/develop/FluentCommandLineParser/Internals/Parsing/CommandLineOptionGrouper.cs#L188) by checking whether the value starts with one of the [option prefix special characters](https://github.com/fclp/fluent-command-line-parser/blob/develop/FluentCommandLineParser/Internals/SpecialCharacters.cs#L47) ( `-`, `--`, `/`)

There is an open issue on this FCLP parser repo for this. See https://github.com/fclp/fluent-command-line-parser/issues/99

User can either use a leading space char or explicitly escape the double quotes.

Example of explicit escaping of quotes:
` --dotnet-cli-params "\"--configuration debug"\"`
Example of using leading space to trick the parser so that it will not be picked up as an option key.
` --dotnet-cli-params " --configuration debug"`

Multiple options will also work in both cases.

We do not need the leading space when we set the default value in code. It is only needed when user provides it via command line.

fabiocav (approved):

PR #2724, opened 2021-09-24
Update v3 Python Worker Version to 1.2.6

vrdmr:
Python Worker Release note [1.2.6](https://github.com/Azure/azure-functions-python-worker/releases/tag/1.2.6)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,12 +1,17 @@
+# =======================================================
 # See https://help.github.com/articles/about-codeowners/
 # for more info about CODEOWNERS file
 #
 # It uses the same pattern rule for gitignore file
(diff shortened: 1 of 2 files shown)


michaelpeng36:
For the CODEOWNERS file, it seems that the last matching pattern takes the most precedence (see the example in https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) so regardless of if `src/Azure.Functions.Cli/Azure.Functions.Cli.csproj` is the only file modified, it seems Anatoli and I will be the only ones notified. Also, shall we add owners of other workers for `src/Azure.Functions.Cli/Azure.Functions.Cli.csproj`?

michaelpeng36 (approved):
LGTM

PR #2730, opened 2021-09-28
Adding dapr triggers to not require storage in core-tools

### LINKED ISSUE #2065: Add daprTriggers to those that don't require storage
I wonder if we're better at this point to only enforce `AzureWebJobsStorage` if it's a timer trigger or event hub trigger rather than today how we allow-list certain ones.  Regardless we are now launching 3 new triggers that don't require a storage account that the core tools block today:

`daprBindingTrigger`
`daprServiceInvocationTrigger`
`daprTopicTrigger`

// @anthonychu @ConnorMcMahon

apawast:
We launched 3 new triggers that don't require a storage account that are blocked in core tools today. This PR should unblock them. 

Fixes #2065


kshyju:
@paulyuk Could you review this change and confirm if it is still valid and relevant?

fabiocav:
Closing this as stale. Following up offline to understand if this is still a requirement.

PR #2734, opened 2021-09-30
Fix help message for base actions. Closes #2612

### LINKED ISSUE #2612: Argument --slot not being printed for func azure functionapp publish
As mentioned in https://github.com/Azure/azure-functions-core-tools/issues/2610 the `--slot` setting isn't being generated with the help output for `func azure functionapp publish`. Per @ahmelsayed, the [order of adding the arg and throwing the exception](https://github.com/Azure/azure-functions-core-tools/blob/v3.x/src/Azure.Functions.Cli/Actions/AzureActions/BaseFunctionAppAction.cs#L16-L32) is probably what’s causing it to not print out.

ahmelsayed:


michaelpeng36 (approved):
LGTM

PR #2738, opened 2021-09-30
Disabling start_nodejs test to allow other tests

michaelpeng36:
The suspected cause is updates to ASP.NET Core. The issue will require further investigation and will be listed as an issue on the repo, but it will be disabled for now to allow other tests to take place.

### DIFF
--- a/test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
+++ b/test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
@@ -23,7 +23,7 @@ namespace Azure.Functions.Cli.Tests.E2E
         private const string _serverNotReady = "Host was not ready after 10 seconds";
         public StartTests(ITestOutputHelper output) : base(output) { }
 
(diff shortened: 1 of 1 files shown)


test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -23,7 +23,7 @@ public class StartTests : BaseE2ETest
         private const string _serverNotReady = "Host was not ready after 10 seconds";
         public StartTests(ITestOutputHelper output) : base(output) { }
 
-        [Fact]
+        [Fact(Skip = "Test broken due to ASP.NET Core message suppression not taking effect.")]
```
Francisco-Gamino, line 26: Could you please open an issue for the follow-up investigation and reference it here? Thanks.

Francisco-Gamino (changes requested):

Francisco-Gamino (approved):

PR #2741, opened 2021-09-30
suppressing grpc server startup logs

brettsam:

PR #2742, opened 2021-09-30
Re-enable nodejs start test case

michaelpeng36:
Re-enabling the test case involving nodejs startup given the new logging hotfix for ASP.NET Core (see #2741)


Francisco-Gamino (approved):
LGTM

soninaren (approved):

PR #2768, opened 2021-10-15
Prevent warnings with identity-based connections

### LINKED ISSUE #2755: Errors and warnings when using identity-based connections in bindings
Identity-based connections are still not honored except when used for `AzureWebJobsStorage`.

## Repro Steps

1. Run `func init --worker-runtime node`
2. Modify `host.json` to use the preview extension bundle:

  ```json
  {
    "version": "2.0",
    "logging": {
      "applicationInsights": {
        "samplingSettings": {
          "isEnabled": true,
          "excludedTypes": "Request"
        }
      }
    },
    "extensionBundle": {
      "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
      "version": "[3.*, 4.0.0)"
    }
  }
  ```

3. Run `func new -t "Azure Blob Storage Trigger" -n BlobTrigger`
4. Update the default connection in `function.json` for the blob trigger:

  ```json
  {
    "bindings": [
      {
        "name": "myBlob",
        "type": "blobTrigger",
        "direction": "in",
        "path": "samples-workitems/{name}",
        "connection": "mystorage"
      }
    ]
  }
  ```
5. Add the identity-based connection to `local.settings.json`:

  ```json
  {
    "IsEncrypted": false,
    "Values": {
      "FUNCTIONS_WORKER_RUNTIME": "node",
      "mystorage__accountName": "MY_STORAGE_ACCOUNT_NAME"
    }
  }
  ```
  I had previously assigned myself the Storage Blob Data Owner role over the storage account represented by "MY_STORAGE_ACCOUNT_NAME". So in a local development context with the above configuration, the app should be able to read and write.

6. Run `func host start`

Here's what I get:
```
Azure Functions Core Tools
Core Tools Version: 3.0.3785 Commit hash: db6fe71b2f05d09757179d5618a07bba4b28826f (64-bit)
Function Runtime Version: 3.2.0.0

Warning: Cannot find value named 'mystorage' in local.settings.json that matches 'connection' property set on 'blobTrigger' in 'PATH_OBFUSCATED\BlobTrigger1\function.json'. You can run 'func azure functionapp fetch-app-settings <functionAppName>' or specify a connection string in local.settings.json.
[2021-10-08T17:35:48.650Z] There was an error performing a read operation on the Blob Storage Secret Repository. Please ensure the 'AzureWebJobsStorage' connection string is valid.
[2021-10-08T17:35:48.943Z] A host error has occurred during startup operation '784aa661-69e8-4a39-8181-854c467d9dba'.
[2021-10-08T17:35:48.950Z] Microsoft.Azure.WebJobs.Script.WebHost: Could not create BlobContainerClient.
[2021-10-08T17:35:49.031Z] There was an error performing a read operation on the Blob Storage Secret Repository. Please ensure the 'AzureWebJobsStorage' connection string is valid.
Value cannot be null. (Parameter 'provider')
```

## Discussion

That error seems to come from here: https://github.com/Azure/azure-functions-core-tools/blob/5e114c97ed391449fefcae1a93767f6e9fd8f49a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs#L428-L429

It's really odd to me that the secret repository error is triggered by that. Removal of the BlobTrigger is sufficient to make that go away, so it seems like some kind of an exception catching issue - I didn't look into that, though.

There is a [StorageConnectionExists()](https://github.com/Azure/azure-functions-core-tools/blob/5e114c97ed391449fefcae1a93767f6e9fd8f49a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs#L446) method that we use only to check on AzureWebJobsStorage earlier, but it is not applied to the binding connections. This was added in response to https://github.com/Azure/azure-functions-core-tools/issues/2552. I did confirm that I could set `AzureWebJobsStorage__accountName` to a storage account, run it locally, and that everything worked (including timer trigger locks being taken on the blob).

I do have some minor concerns about Storage-specific behavior being included here in general. This behavior also needs to hold for Service Bus, Event Hubs, Cosmos, etc. The method seems like it's sufficiently generic aside from the name. But we do need to call it from the binding loop.

CC @cachai2 (and thanks for help identify and assembling this info)

michaelpeng36:
Ensure that warnings are no longer emitted when customers use managed identity to connect to service. 

Fixes #2755

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/PackAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/PackAction.cs
@@ -14,7 +14,7 @@ using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Actions.LocalActions
 {
-    [Action(Name = "pack", HelpText = "Pack function app into a zip that's ready to run.", ShowInHelp = false)]
+    [Action(Name = "pack", HelpText = "Pack function app into a zip that's ready to run.", ShowInHelp = true)]
     internal class PackAction : BaseAction
     {
         private readonly ISecretsManager _secretsManager;
@@ -39,7 +39,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             Parser
                 .Setup<bool>("build-native-deps")
                 .SetDefault(false)
-                .WithDescription("Skips generating .wheels folder when publishing python function apps.")
+                .WithDescription("Build python packages in a container image")
                 .Callback(f => BuildNativeDeps = f);
             Parser
                 .Setup<bool>("no-bundler")


mattchenderson:
I know there was a question on cherry picking this into 4.x as well - just wanted to note that too.

PR #2785, opened 2021-10-26
Copy into src folder directly to match path of pdb

### LINKED ISSUE #2784: PDB symbol path don't add up when debugging (attaching) Azure Functions inside a container
It looks like the paths don't match when you attach to a running container hosting Azure Functions.

Steps to reproduce:
1. Follow steps 1 to 3 in this guide: https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-function-linux-custom-image?tabs=isolated-process%2Cbash%2Cazure-cli&pivots=programming-language-csharp
2. Put a break-point in VS Code in the Test Function
3. Attach to the container through VS Code (see luanch.json below)
 The following error occurs:

image

The problem seems to be in [Dockerfile.dotnet](https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/StaticResources/Dockerfile.dotnet) where it copies to source `COPY . /src/dotnet-function-app`. Changing it to  `COPY . /src` seems to fix the issue. (I don't know if it will have other implications)

_luanch.json_
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Docker .NET Core Attach",
            "type": "docker",
            "request": "attach",
            "platform": "netCore",
            "sourceFileMap": {
                "/src": "${workspaceFolder}/src/scanner-api"
            }
        }
    ]
}
```

CasperJ:
### PDB symbol path don't add up when debugging (attaching) Azure Functions inside a container
It looks like the paths don't match when you attach to a running container hosting Azure Functions.
resolves #2784

### Pull request checklist

* [X] My changes **should not** be added to the release notes for the next release

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/Dockerfile.dotnet
+++ b/src/Azure.Functions.Cli/StaticResources/Dockerfile.dotnet
@@ -1,7 +1,7 @@
 FROM microsoft/dotnet:2.2-sdk AS installer-env
 
-COPY . /src/dotnet-function-app
-RUN cd /src/dotnet-function-app && \
+COPY . /src
+RUN cd /src && \
     mkdir -p /home/site/wwwroot && \
     dotnet publish *.csproj --output /home/site/wwwroot
 


soninaren:
@CasperJ the PR is against an inactive development branch. Please send a PR against v4.x branch if the change is still relevant.

PR #2814, opened 2021-11-06
Filter both shm and app insights warnings

ejizba:
It appears only the last filter is used, so to filter both we would need to combine it into one filter.


michaelpeng36 (approved):
Would you be able to add this change to the `v4.x` branch? Otherwise, LGTM

PR #2852, opened 2021-12-02
Update Node.js package.json templates for `func init`

ejizba:
- Updated the version of `@azure/functions` to v3 (see [here](https://github.com/Azure/azure-functions-nodejs-worker/issues/428) for versioning info). This was the main motivation for the PR
- Simplified the scripts. For example, `func extensions install` is no longer necessary. Also I chose to get rid of `npm-run-all`. This [package](https://www.npmjs.com/package/npm-run-all) has not been updated in 3 years and has 9 of its own dependencies, both things I avoid when adding npm packages. IMO its better for the template to be clean & simple
- Made it so that it exactly matches the templates users get in VS Code

I will make a similar PR to the v3.x branch.

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/javascriptPackage.json
+++ b/src/Azure.Functions.Cli/StaticResources/javascriptPackage.json
@@ -1,9 +1,11 @@
 {
   "name": "",
-  "version": "",
+  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


michaelpeng36 (approved):
LGTM

PR #2853, opened 2021-12-02
Port "Update Node.js package.json templates for func init" to v3.x

ejizba:
Only difference is `@azure/functions` version should be 2 instead of 3 (see [here](https://github.com/Azure/azure-functions-nodejs-worker/issues/428) for versioning info)

Original PR: https://github.com/Azure/azure-functions-core-tools/pull/2852


michaelpeng36 (approved):
LGTM

PR #2859, opened 2021-12-04
Ignore Azurite Files in Function App (CLI v3)

lechnerc77:

When initializing an Azure Functions via the CLI and switching to VSCode for development the Azurite Extension in Visual Studio Code (<https://marketplace.visualstudio.com/items?itemName=Azurite.azurite>) often gets used for storage emulation in local development. The extension creates several files in the Azure Functions project directory that are only relevant for local development. In analogy to e.g. `local.settings.json` they should be ignored by git by default.
It would be very comfortable if this would also be the case for the CLI.

This pull request contains the required enhancement `.gitignore` file as a proposal to cover the scenario described above.

### Pull request checklist

* [X ] My changes **should not** be added to the release notes for the next release


lechnerc77:
This is a follow up to pull-request #2725  to match the right branch

michaelpeng36 (approved):
LGTM

Issue #2819, opened 2021-11-08
Kafka Dotnet function template option is not available while creating from core tools


shrohilla:
While creating a function from core tools, if we select the option of dotnet, in the triggering option the Kafka trigger option is absent.

While the template is available in function template 3.x branch -- https://github.com/Azure/azure-functions-templates/tree/v3.x/Functions.Templates/Templates

![dotnet-template](https://user-images.githubusercontent.com/80326489/140788926-01293129-8afb-4efe-9e44-8c3c30fee897.png)

PFA of the image 

cc @pragnagopa @TsuyoshiUshio @fabiocav @lpapudippu

pragnagopa:
@TsuyoshiUshio  - Please confirm if the dotnet templates already include Kafka template.  If yes, this issue can be scoped to updating templates version referred in Core Tools

cc @soninaren

TsuyoshiUshio:
Code Pointer for func new azure-functions-core-tools/CreateFunctionAction.cs at dev · Azure/azure-functions-core-tools (github.com)

In this case, current func command reference the old template. 

For dotnet, we can find the template version in here. 
[azure-functions-core-tools/Settings.cs at v3.x · Azure/azure-functions-core-tools (github.com)](https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Actions/LocalActions/CreateFunctionAction.cs)
It looks 3.1.1648. 

We can update it with new version listed in here. 
[Releases · Azure/azure-functions-templates (github.com)](https://github.com/Azure/azure-functions-core-tools/blob/v3.x/build/Settings.cs#L23-L27)

Latest looks include Kafka
[Release 3.1.1809 · Azure/azure-functions-templates (github.com)](https://github.com/Azure/azure-functions-templates/releases/tag/3.1.1809)

Then we need to release new version of the func command.

michaelpeng36:
Closing, as this is fixed by PR #2862

Issue #2861, opened 2021-12-04
Standardize the way templates are handled across runtimes


michaelpeng36:
Currently, the way templates are handled for projects with a .NET worker runtime is hard-coded
https://github.com/Azure/azure-functions-core-tools/blob/580458df50de10663b2019b040fae34bf2cf5049/src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs#L89
This should be removed and replaced with a method that is standardized across all worker runtimes (for example, getting templates from the extension bundle).

PR #2862, opened 2021-12-04
Add KafkaTrigger template

michaelpeng36:

Hotfix to resolve issue #2819 since this issue is blocking. In general, hard-coding templates for Function projects with .NET runtimes should be removed. This work item is tracked by issue #2861.

### Pull request checklist


test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
```diff
@@ -12,19 +12,21 @@ public class CreateFunctionTests : BaseE2ETest
     {
         public CreateFunctionTests(ITestOutputHelper output) : base(output) { }
 
-        [Fact]
-        public async Task create_template_function_success_message()
+        [Theory]
+        [InlineData("dotnet", "HttpTrigger")]
+        [InlineData("dotnet", "KafkaTrigger")]
```
ejizba, line 17: I don't think this test is meant for every trigger. If it was, it should call `func init` once and then iterate through all the triggers on the same project to save time. For example, the Kafka test as-is took about ~17 seconds and I feel like anything over 2 is long for a unit test

I don't have a super strong preference on how you handle the test for this PR, but if it were me I would skip the unit test for now and investigate a "create function" test for every trigger type later. Depending how much time that added to each test run, I might prioritize trigger types based on usage.

ejizba (approved):

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
michaelpeng36: That makes sense. I will remove the test for now since it seems to be taking a lot of time, then.

PR #2917, opened 2022-01-21
Update VS Code config

ejizba:
The config was super out of date, so I wanted to fix a few annoying prompts and make it so that debugging was as easy as hitting F5.

I plan to port this to v3.x and dev branches.


michaelpeng36 (approved):
LGTM

Issue #2834, opened 2021-11-18
Support running on M1 Macs


ejizba:
Update: Core tools support for M1 macs has been released for both npm and brew.

In addition to work on core tools, each language worker also needs to add support individually:
- Done:
    - Node.js - tracked [here](https://github.com/Azure/azure-functions-nodejs-worker/issues/386)
    - .NET
    - .NET Isolated
    - PowerShell - tracked [here](https://github.com/Azure/azure-functions-powershell-worker/issues/752). The work has been done, but it has not been integrated into core tools yet. It will be included with host version 4.5.0 or greater
- Still to do
    - Python - tracked [here](https://github.com/Azure/azure-functions-python-worker/issues/915). Should be included with the release of support for Python 3.10
    - Java: blocked because the Java worker only supports Java 8 & 11 and AFAIK Java itself doesn't support Apple Silicon until at least Java 17

toklive:
Does this work for .NET 6 and Azure function v4? if not any timeline when M1 support will be available?

RichMercer:
> Does this work for .NET 6 and Azure function v4? if not any timeline when M1 support will be available?

I never got it working but compiled the source myself to target arm64 and copied it over the top of the brew install location. Works lovely now. :)

EdiWang:
Adding related issue https://github.com/Azure/azure-functions-core-tools/issues/2180

toklive:
> > Does this work for .NET 6 and Azure function v4? if not any timeline when M1 support will be available?
> 
> I never got it working but compiled the source myself to target arm64 and copied it over the top of the brew install location. Works lovely now. :)

That is awesome to hear. If it is working fine, any idea when this will be available officially for Dotnet ?

ajonno:
same, we really need this official M1 support too

Arash-Sabet:
Any ETA on when this fix will be released?

natdan:
> > Does this work for .NET 6 and Azure function v4? if not any timeline when M1 support will be available?
> 
> I never got it working but compiled the source myself to target arm64 and copied it over the top of the brew install location. Works lovely now. :)

Do you have any tips how to do it? (to compile it on M1 mac)

[...]

PR #2942, opened 2022-02-03
Add preliminary support for osx-arm64

ejizba:

Related to https://github.com/Azure/azure-functions-core-tools/issues/2834

I was able to do some basic sanity testing (func init, func new, func start) on an M1 Mac for both JS and C# with a build created from this branch. I think that's good enough to get the PR going, but will continue to do more testing.

### Pull request checklist

### DIFF
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -70,7 +70,7 @@ namespace Build
                     { "win-x64", _winPowershellRuntimes },
                     { "linux-x64", new [] { "linux", "linux-x64", "unix", "linux-musl-x64" } },
                     { "osx-x64", new [] { "osx", "osx-x64", "unix" } },
-                    // NOTE: arm support wasn't added until PowerShell 7.2
+                    // NOTE: PowerShell 7.0 does not support arm. First version supporting it is 7.2
                     // https://docs.microsoft.com/powershell/scripting/install/installing-powershell-on-macos?view=powershell-7.2#installation-via-direct-download
                     // That being said, we might as well include "osx" and "unix" since it'll hardly affect package size and should lead to more accurate error messages
                     { "osx-arm64", new [] { "osx", "unix" } }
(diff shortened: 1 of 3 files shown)


ejizba:
Update on language testing - these work:
- Node.js
- Dotnet
- Dotnet Isolated

And these do not:
- Python, tracked here: https://github.com/Azure/azure-functions-python-worker/issues/915
- PowerShell, tracked here: https://github.com/Azure/azure-functions-powershell-worker/issues/752
- Java: I didn't even try this one because it appears M1 support has only been released on Java 17 so far (hard to find docs, but here's [one post](https://www.reddit.com/r/MacOS/comments/ptfflp/oracle_released_native_apple_silicon_java_jdk_17/)) and our Java worker only supports Java 8 & 11

Francisco-Gamino:
Thank you @ejizba for sending this out! Do you happen to know how much space will this change increase? I am not sure if we have any restrictions on the payload size.

/cc @AnatoliB for FYI

ejizba:
@Francisco-Gamino It results in another zip being uploaded to the functionscdn storage account for each core tools release, but that's about the only "space" change I can think of. It has no effect on the host since it's a core-tools-only change. It also shouldn't affect any user's machines because the arm64 build is it's own separate package

michaelpeng36 (approved):
LGTM

build/Settings.cs
```diff
@@ -67,7 +69,11 @@ private static string config(string @default = null, [CallerMemberName] string k
                     { "win-x86", _winPowershellRuntimes },
                     { "win-x64", _winPowershellRuntimes },
                     { "linux-x64", new [] { "linux", "linux-x64", "unix", "linux-musl-x64" } },
-                    { "osx-x64", new [] { "osx", "osx-x64", "unix" } }
+                    { "osx-x64", new [] { "osx", "osx-x64", "unix" } },
+                    // NOTE: arm support wasn't added until PowerShell 7.2
```
Francisco-Gamino, line 73: Could you please update the comment to mention that PowerShell 7.0 does not have ARM support?

Francisco-Gamino (approved):
@ejizba -- One minor comment; otherwise, LGTM.

PR #2950, opened 2022-02-14
Alter the Chocolatey package name

michaelpeng36:
### Issue describing the changes in this PR

Alters the Chocolatey package name to `azure-functions-core-tools`, as all future releases to Chocolatey will be done under that name
### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #2951, opened 2022-02-14
Alter Chocolatey package name

michaelpeng36:

Alters the Chocolatey package name to `azure-functions-core-tools`, as all future releases to Chocolatey will be done under that name. Same changes as done in #2950 . These changes will require documentation changes as described in #2954 
### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR

### DIFF
--- a/publish-scripts/shared/constants.py
+++ b/publish-scripts/shared/constants.py
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
 
 # same for all different OSes
-PACKAGENAME = "azure-functions-core-tools-4"
+PACKAGENAME = "azure-functions-core-tools"
 CMD = "func"
 ARTIFACTFOLDER = "artifact"
 BUILDFOLDER = "build"


publish-scripts/shared/constants.py
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
 
 # same for all different OSes
-PACKAGENAME = "azure-functions-core-tools-4"
+PACKAGENAME = "azure-functions-core-tools"
```
ejizba, line 4: I think this constant is shared with the ubuntu installer, and I don't think that was discussed. Do we also want to change the name for ubuntu? That might be fine, but it would mean more testing/doc changes/etc.

For example: https://github.com/Azure/azure-functions-core-tools#2-install

publish-scripts/shared/constants.py
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
```
ejizba, line 1: I don't think this should be checked: "My changes do not require documentation changes"

publish-scripts/shared/constants.py
michaelpeng36: Thanks for the catch, @ejizba . I will continue this discussion offline with Anatoli.

ejizba (approved):

PR #2957, opened 2022-02-19
Update v4 Python Worker Version to 4.0.0

vrdmr:
Python Worker Release note [4.0.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.0.0)

### DIFF
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -164,7 +164,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.NodeJsWorker" Version="2.1.2" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS6" Version="3.0.630" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS7" Version="3.0.1272" />
-    <PackageReference Include="Microsoft.Azure.Functions.PythonWorker" Version="3.4.0.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.PythonWorker" Version="4.0.0" />
   </ItemGroup>
 
   <Target Name="ExcludeWorkersFromReadyToRun">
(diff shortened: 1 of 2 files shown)


YunchuWang (approved):

gavin-aguiar (approved):

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -164,7 +164,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.NodeJsWorker" Version="2.1.2" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS6" Version="3.0.630" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS7" Version="3.0.1272" />
-    <PackageReference Include="Microsoft.Azure.Functions.PythonWorker" Version="3.1.2.6" />
+    <PackageReference Include="Microsoft.Azure.Functions.PythonWorker" Version="4.0.0" />
```
ejizba, line 167: cc @michaelpeng36 @VpOfEngineering 

My understanding is that the workers should be updated at the same time as the host nuget package, not before. Otherwise we get in weird situations where core tools might accidentally release a worker before the host officially supports it

Docs [here](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-benbyrd/app-service-web-apps/antares-team-documents/functionteamdocs/engineeringprocesses/releaseprocesses/coretoolsreleaseprocess)

ejizba (changes requested):

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
vrdmr: Got it. So - workers should not be updated unless the host has released them?

Python worker got updated in the host: https://github.com/Azure/azure-functions-host/pull/8156. But I can generate the release PR here later.

ejizba: Correct. There's some discussion at what point during the release process we should update the core tools repo, but at a minimum the following should be true:

1. Check the version of "Microsoft.Azure.WebJobs.Script.WebHost" in this file
2. Find the matching github tag on the host repo (i.e. https://github.com/Azure/azure-functions-host/tree/v4.1.3)
3. Make sure the worker versions on that tag match what's in this file

Honestly, the language worker owners shouldn't need to worry about this process. Whoever updates the reference to the host should be the one updating the workers at the same time. I'm thinking about adding some sort of unit test to verify the steps mentioned above to ensure the versions are always in sync.

vrdmr:
Closing this PR as the processes has been automated and the version number is now picked up from the Host itself. Thanks @ejizba for enabling this.

PR #2971, opened 2022-03-02
Add script to validate host/worker versions

ejizba:
Currently, the process for updating workers in this repo is very disjointed and could easily lead to a situation where core tools ships a host/worker combination that is not possible in Azure. I've added a script to ensure that the worker versions match what was shipped with the respective version of the host. The script can be used to either validate the versions (as a part of the dev ops build) or update the versions (run manually before submitting a PR). Here is the criteria used to determine if a worker version matches:
1. Check the version of "Microsoft.Azure.WebJobs.Script.WebHost" in `Azure.Functions.Cli.csproj`
2. Find the matching github tag on the host repo (i.e. https://github.com/Azure/azure-functions-host/tree/v4.1.3)
3. Make sure the worker versions on that tag match what's in `Azure.Functions.Cli.csproj`

### Pull request checklist

  * [x] Otherwise: Just trust me I'll backport it :P

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -85,6 +85,9 @@ jobs:
         $accessToken = (az account get-access-token --query "accessToken" | % { $_.Trim('"') })
         echo "##vso[task.setvariable variable=azure_management_access_token]$accessToken"
   - pwsh: |
+      .\validateWorkerVersions.ps1
+    displayName: 'Validate worker versions'
+  - pwsh: |
       .\build.ps1
     env:
       AzureBlobSigningConnectionString: $(AzureBlobSigningConnectionString)
--- a/validateWorkerVersions.ps1
+++ b/validateWorkerVersions.ps1
@@ -1,0 +1,87 @@
+<#
+    .SYNOPSIS
+        Used to validate and/or update worker package versions
+    .EXAMPLE
+        ./validateWorkerVersions.ps1
+
+        Validates the workers match the existing host version and throws an error if they don't
+    .EXAMPLE
+        ./validateWorkerVersions.ps1 -Update -HostVersion 4.0.0
+
+        Updates the host reference to 4.0.0 and the workers to their matching versions
+#>
+param (
+    [Switch]$Update,
+    
+    # An explicit host version, otherwise the host version from Azure.Functions.Cli.csproj will be used
+    [string]$hostVersion
+)
+
+# the xml will fail to parse if the data is encoded with a bom character
+function removeBomIfExists([string]$data)
+{
+    if ($data.StartsWith(0xFEFF)) {
+        $data = $data.substring(1)
+    }
+    return $data
+}
+
+$cliCsprojPath = "./src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
+$cliCsprojContent = removeBomIfExists(Get-Content $cliCsprojPath)
+$cliCsprojXml = [xml]$cliCsprojContent
+
+function getPackageVersion([string]$packageName, [string]$csprojContent)
+{
+    $version = (Select-Xml -Content $csprojContent -XPath "/Project//PackageReference[@Include='$packageName']/@Version").ToString()
+    if (-Not $version) {
+        throw "Failed to find version for package $packageName"
+    }
+    return $version
+}
+
+function setCliPackageVersion([string]$packageName, [string]$version)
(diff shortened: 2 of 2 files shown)


validateWorkerVersions.ps1
```diff
@@ -0,0 +1,87 @@
...
+        Write-Output "Reference to $worker in the host ($hostWorkerVersion) does not match version in the cli ($cliWorkerVersion)"
+        $failedValidation = $true
+    }
+}
+
+if ($Update) {
+    $cliCsprojXml.Save($cliCsprojPath)
+    Write-Output "Updated worker versions! 🚀"
+} elseif ($failedValidation) {
+    Write-Output "You can run './validateWorkerVersions.ps1 -Update' locally to fix worker versions."
```
VpOfEngineering, line 83: Ah, I see. This runs with every build pipeline, so the pipeline will fail early. Running the script locally will apply the reference changes.

vrdmr: Thanks @ejizba for this.

For fixing worker versions, can we create a GitHub action which - given a host version - fixes the worker versions (sends in a PR for the fix)?

ejizba: Yes, I think it be would a logical next-step to automate this in some way

michaelpeng36 (approved):
LGTM

vrdmr (approved):
:shipit:

azure-pipelines.yml
```diff
@@ -84,6 +84,9 @@ jobs:
         # acquire access token from Azure CLI and export it to AZURE_MANAGEMENT_ACCESS_TOKEN
         $accessToken = (az account get-access-token --query "accessToken" | % { $_.Trim('"') })
         echo "##vso[task.setvariable variable=azure_management_access_token]$accessToken"
+  - pwsh: |
```
Francisco-Gamino, line 87: Should this task only be run for release builds?

validateWorkerVersions.ps1
```diff
@@ -0,0 +1,87 @@
...
+    $version = (Select-Xml -Content $csprojContent -XPath "/Project//PackageReference[@Include='$packageName']/@Version").ToString()
+    if (-Not $version) {
+        throw "Failed to find version for package $packageName"
+    }
+    return $version
+}
+
+function setCliPackageVersion([string]$packageName, [string]$version)
+{
+    $node = $cliCsprojXml.SelectSingleNode("/Project//PackageReference[@Include='$packageName']")
```
Francisco-Gamino, line 44: Should we throw if `$node` is `null`?

validateWorkerVersions.ps1
```diff
@@ -0,0 +1,87 @@
...
+    $node.Version = $version
```
Francisco-Gamino, line 45: Having a log entry to says what `version` the worker is being updated to would be helpful to debug without having to rerun the script.

validateWorkerVersions.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+$hostPackageName = "Microsoft.Azure.WebJobs.Script.WebHost"
+if (-Not $hostVersion) {
+    $hostVersion = getPackageVersion $hostPackageName $cliCsprojContent
+} elseif ($Update) {
+    setCliPackageVersion $hostPackageName $hostVersion
+}
+
+$hostRepoUrlBase = "https://raw.githubusercontent.com/Azure/azure-functions-host/v$hostVersion"
+$hostCsprojContent = removeBomIfExists((Invoke-WebRequest -Uri "$hostRepoUrlBase/src/WebJobs.Script/WebJobs.Script.csproj").Content)
```
Francisco-Gamino, line 56: Consider adding `-MaximumRetryCount` `-RetryIntervalSec` to retry in case the first `Invoke-WebRequest` fails.

Francisco-Gamino (approved):
Thank you @ejizba for sending this out! I left a few minor comments. Otherwise, LGTM.

azure-pipelines.yml
ejizba: My preference is to run during PR builds to prevent language owners from updating just their worker instead of all host/workers at once. I think this keeps our branches in more of a "shippable" state

Francisco-Gamino: Good point. Let's keep it as is.

PR #2987, opened 2022-03-16
Bump app insights package to fix build

ejizba:
Remote builds are currently failing with this:

> /Users/erijiz/Repos/func-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : error NU1605: Detected package downgrade: Microsoft.ApplicationInsights from 2.20.0 to 2.18.0. Reference the package directly from the project to select a different version. 
> /Users/erijiz/Repos/func-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : error NU1605:  Microsoft.Azure.Functions.CoreTools -> Microsoft.Azure.WebJobs.Script.WebHost 4.1.3 -> Microsoft.Azure.WebJobs.Script 4.1.3 -> Microsoft.Azure.WebJobs.Script.Abstractions 1.0.3-preview -> Microsoft.ApplicationInsights (>= 2.20.0) 
> /Users/erijiz/Repos/func-cli/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj : error NU1605:  Microsoft.Azure.Functions.CoreTools -> Microsoft.ApplicationInsights (>= 2.18.0)

Depending on the state of your repo and dotnet cache, the error may not repro locally.

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
+++ b/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
@@ -153,7 +153,7 @@
     <PackageReference Include="Colors.Net" Version="1.1.0" />
     <PackageReference Include="AccentedCommandLineParser" Version="2.0.0" />
     <PackageReference Include="DotNetZip" Version="1.13.3" />
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.18.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.20.0" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="2.2.0" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage.Internal" Version="1.4.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Script.WebHost" Version="4.1.3" />


Francisco-Gamino:
Hi @ejizba -- The Core Tools build for Integration testing was failing for sometime because the Host referenced `Microsoft.ApplicationInsights`  version `2.20.0` while the Core Tools referenced version `2.18.0`. The workaround was to create a new branch and update `Microsoft.ApplicationInsights` to `2.20.0`. 

Going forward, would it make sense to have a similar script like https://github.com/Azure/azure-functions-core-tools/blob/v4.x/validateWorkerVersions.ps1 with all the packages that are referenced in the Host and the Core Tools, and run it every so often to keep the packages updated? Thoughts?

ejizba:
Tbh, as a non-.NET developer I'm not that familiar with how nuget and dependencies are handled. I've heard of "nuget dependency hell", but I have not had much experience with it until now.

In Node.js, people normally specify ranges of a dependency by default instead of a specific version. I'm pretty sure dotnet/nuget supports ranges, though, right? Any issue if I changed "2.20.0" to "2.*"? We care about the worker versions (why the script is necessary), but do we really care which minor version we get for app insights? For some reason, I don't see people use ranges a lot in .NET and I'm not sure why.

ejizba:
Moving this discussion to a separate issue (https://github.com/Azure/azure-functions-core-tools/issues/2990) so that I can unblock the build

PR #2991, opened 2022-03-18
Make null and unsupported Key Vault reference formats non-blocking

### LINKED ISSUE #2989: KeyVaultReferencesManager throws an error when provided a null config item
When KeyVaultReferencesManager attempts to match config items with the Key Vault Reference format it can result in a cryptic error being thrown
`Key Vault Reference format invalid:`

The following is an example from a newly created empty function app 

local.settings.json
```
{
  "IsEncrypted": false,
  "Values": {
    "test": null
  }
}
```
On Debug
```
Azure Functions Core Tools
Core Tools Version:       3.0.4378 Commit hash: N/A  (64-bit)
Function Runtime Version: 3.4.2.0

[2022-03-17T01:43:06.010Z] Found C:\***\NullValueFunction\NullValueFunction.csproj. Using for user secrets file configuration.
Key Vault Reference format invalid:
Press any key to continue....
```

michaelpeng36:

Resolves #2989. This PR corrects the regular expressions used in matching Key Vault references. Furthermore, issues with parsing Key Vault references should be non-blocking for `func start` (a warning is emitted instead). Support for other Key Vault reference formats that have secret version as an optional parameter has also been added.

#### Old Behavior
The KeyVaultReferencesManager attempts to match items in the configuration with regular expressions intended to capture two Key Vault Reference formats:

1. The "primary format": `@Microsoft.KeyVault(SecretUri=https://myvault.vault.azure.net/secrets/mysecret/version)`
2. The "secondary format":  `@Microsoft.KeyVault(VaultName=myvault;SecretName=mysecret)`

While the regular expressions used indeed match with their corresponding formats, they were not entirely correct, leading to matches where there should be none and the cryptic error `Key Vault Reference format invalid: <config-value>` being thrown upon creation of the `ParseSecretResult` object.

#### Current Behavior
1. Item value is `null`: `func start` proceeds, writing the warning to the console `Skipping <config-item> because value is null`
2. Item value resembles the use of a Key Vault reference, i.e. takes on the form `@Microsoft.KeyVault(.*)`, but does not match any of the syntaxes specified in [this doc](https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references): Core Tools writes the warning `Unable to parse the Key Vault reference: <config-item-value>` to the console and proceeds with `func start`
3. Item value does not resemble the use of a Key Vault reference: `func start` proceeds as normal.
4. Item value matches a syntax specified in [this doc](https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references): Core Tools attempts to retrieve the secret value. If successful, `func start` proceeds as normal. Otherwise, an error is thrown.

### Pull request checklist

  * [x] Otherwise: This change will be backported to `v3.x`

### DIFF
--- a/src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
+++ b/src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
@@ -24,17 +24,26 @@ namespace Azure.Functions.Cli.Common
         {
             foreach (var key in settings.Keys.ToList())
             {
-                var keyVaultValue = GetSecretValue(settings[key]);
-                if (keyVaultValue != null)
+                try
                 {
-                    settings[key] = keyVaultValue;
+                    var keyVaultValue = GetSecretValue(key, settings[key]);
+                    if (keyVaultValue != null)
+                    {
+                        settings[key] = keyVaultValue;
+                    }
                 }
+                catch
+                {
+                    // Do not block StartHostAction if secret cannot be resolved: instead, skip it
+                    // and attempt to resolve other secrets
+                    continue;
+                }   
             }
         }
 
-        private string GetSecretValue(string value)
+        private string GetSecretValue(string key, string value)
         {
-            var result = ParseSecret(value);
+            var result = ParseSecret(key, value);
 
             if (result != null)
             {
@@ -46,7 +55,7 @@ namespace Azure.Functions.Cli.Common
             return null;
         }
 
-        private ParseSecretResult ParseSecret(string value)
+        internal ParseSecretResult ParseSecret(string key, string value)
         {
             // If the value is null, then we return nothing, as the subsequent call to
             // UpdateEnvironmentVariables(settings) will log to the user that the setting
@@ -66,14 +75,14 @@ namespace Azure.Functions.Cli.Common
                 // the supported formats, we write a warning to the console.
                 if (result == null)
                 {
-                    ColoredConsole.WriteLine(WarningColor($"Unable to parse the Key Vault reference: {value}"));
+                    ColoredConsole.WriteLine(WarningColor($"Unable to parse the Key Vault reference for setting: {key}"));
                 }
                 return result;
             }
             return null;
         }
 
-        private ParseSecretResult ParsePrimaryRegexSecret(string value)
+        internal ParseSecretResult ParsePrimaryRegexSecret(string value)
         {
             var match = PrimaryReferenceStringRegex.Match(value);
             if (match.Success)
@@ -88,7 +97,7 @@ namespace Azure.Functions.Cli.Common
             return null;
         }
 
-        private ParseSecretResult ParseSecondaryRegexSecret(string value)
+        internal ParseSecretResult ParseSecondaryRegexSecret(string value)
         {
             var altMatch = SecondaryReferenceStringRegex.Match(value);
             if (altMatch.Success)
@@ -107,7 +116,7 @@ namespace Azure.Functions.Cli.Common
             return clients.GetOrAdd(vaultUri.ToString(), _ => new SecretClient(vaultUri, credential));
         }
 
-        private class ParseSecretResult
+        internal class ParseSecretResult
         {
             public Uri Uri { get; set; }
             public string Name { get; set; }
--- a/src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
+++ b/src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
@@ -14,9 +14,9 @@ namespace Azure.Functions.Cli.Common
     class KeyVaultReferencesManager
     {
         private const string vaultUriSuffix = "vault.azure.net";
-        private static readonly Regex KeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\([\S]*\)$", RegexOptions.Compiled);
-        private static readonly Regex PrimaryValidKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\(SecretUri=(?<VaultUri>https://[\S^/]+)/(?<Secrets>[\S^/]+)/(?<SecretName>[\S^/]+)/(?<Version>[\S^/]+)\)$", RegexOptions.Compiled);
-        private static readonly Regex SecondaryValidKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\(VaultName=(?<VaultName>[\S^;]+);SecretName=(?<SecretName>[\S^;]+)\)$", RegexOptions.Compiled); 
+        private static readonly Regex BasicKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\([\S]*\)$", RegexOptions.Compiled);
+        private static readonly Regex PrimaryKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\(SecretUri=(?<VaultUri>https://[\S^/]+)/(?<Secrets>[\S^/]+)/(?<SecretName>[\S^/]+)/(?<Version>[\S^/]+)\)$", RegexOptions.Compiled);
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
```diff
@@ -45,19 +48,34 @@ private string GetSecretValue(string value)
...
+            if (keyVaultReferenceMatch.Success)
             {
-                throw new FormatException($"Key Vault Reference format invalid: {value}");
+                var referenceString = keyVaultReferenceMatch.Groups["ReferenceString"].Value;
+                var result = ParsePrimaryRegexSecret(referenceString) ?? ParseSecondaryRegexSecret(referenceString);
+                // If we detect that a key vault reference was attempted, but did not match any of
+                // the supported formats, we write a warning to the console.
+                if (result == null)
+                {
+                    ColoredConsole.WriteLine(WarningColor($"Unable to parse the Key Vault reference: {value}"));
```
ejizba, line 69: I think this should use `key` in the message, not `value`. People often have sensitive information in app setting values that we should avoid logging

src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
```diff
@@ -45,19 +48,34 @@ private string GetSecretValue(string value)
...
         {
-            try
+            // If the value is null, then we return nothing, as the subsequent call to
+            // UpdateEnvironmentVariables(settings) will log to the user that the setting
+            // is skipped. We check here, because Regex.Match throws when supplied with a
+            // null value.
+            if (value == null)
             {
-                return ParsePrimaryRegexSecret(value) ?? ParseSecondaryRegexSecret(value);
+                return null;
```
ejizba, line 57: I like the extra checks, but I think we still need an overarching try/catch somewhere so that we can be 100% sure this won't block `func start`

I would suggest putting the try/catch either in `GetSecretValue` or within the for loop of `ResolveKeyVaultReferences` so that
- it covers both `ParseSecret` and `client.GetSecret`
- if one secret value is bad, we'll still try the other ones

src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
```diff
@@ -3,17 +3,20 @@
...
+using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Common
 {
     class KeyVaultReferencesManager
     {
         private const string vaultUriSuffix = "vault.azure.net";
-        private static readonly Regex PrimaryKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\(SecretUri=(?<VaultUri>[\S^/]+)/(?<Secrets>[\S^/]+)/(?<SecretName>[\S^/]+)/(?<Version>[\S^/]+)\)$", RegexOptions.Compiled);
-        private static readonly Regex SecondaryKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\(VaultName=(?<VaultName>[\S^;]+);SecretName=(?<SecretName>[\S^;]+)\)$", RegexOptions.Compiled); 
+        private static readonly Regex BasicKeyVaultReferenceRegex = new Regex(@"^@Microsoft.KeyVault\((?<ReferenceString>\S*)\)$", RegexOptions.Compiled);
```
ejizba, line 17: Can we add some unit tests? It would give me a lot more confidence in any regex change

src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
```diff
@@ -3,35 +3,45 @@
...
```
ejizba, line 17:
```suggestion
        private static readonly Regex BasicKeyVaultReferenceRegex = new Regex(@"^@Microsoft\.KeyVault\((?<ReferenceString>.*)\)$", RegexOptions.Compiled);
```

src/Azure.Functions.Cli/Common/KeyVaultReferencesManager.cs
```diff
@@ -43,43 +53,69 @@ private string GetSecretValue(string value)
...
+                return null;
             }
-            catch
+            // Determine if the secret value is attempting to use a key vault reference
+            var keyVaultReferenceMatch = BasicKeyVaultReferenceRegex.Match(value);
+            if (keyVaultReferenceMatch.Success)
             {
-                throw new FormatException($"Key Vault Reference format invalid: {value}");
+                var referenceString = keyVaultReferenceMatch.Groups["ReferenceString"].Value;
+                var result = ParseVaultReference(referenceString);
```
ejizba, line 71:
```suggestion
                ParseSecretResult result = null;
                try
                {
                    result = ParseVaultReference(referenceString);
                }
                catch
                {
                    // ignore and show warning below
                }

```

test/Azure.Functions.Cli.Tests/KeyVaultReferencesManagerTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            exception.Should().BeNull();
+        }
+
+        [Theory]
+        [InlineData("test", null, false)]
+        [InlineData("test", "@Microsoft.KeyVault()", true)]
+        [InlineData("test", "@Microsoft.KeyVault(string)", true)]
+        // Attempted Key Vault references are seen as those matching the regular expression
+        // "^@Microsoft.KeyVault(\S*)$".
```
ejizba, line 49:
```suggestion
        [InlineData("test", "@Microsoft.KeyVault(string)", true)]
        [InlineData("test", "@Microsoft.KeyVault(SecretUri=bad uri)", true)]
        [InlineData("test", "@Microsoft.KeyVault(VaultName=vault;)", true)] // missing secret name
        [InlineData("test", "@Microsoft.KeyVault(SecretName=vault;)", true)] // missing vault name
        [InlineData("test", "@Microsoft-KeyVault()", false)] // hyphen instead of dot
        // Attempted Key Vault references are seen as those matching the regular expression
        // "^@Microsoft.KeyVault(.*)$".
```

test/Azure.Functions.Cli.Tests/KeyVaultReferencesManagerTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            
+        }
+
+        // See https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references
+        // for more detail on supported key vault reference syntax.
+        [Theory]
+        [InlineData("SecretUri=https://sampleurl/secrets/mysecret/version", true, "https://sampleurl/", "mysecret", "version")]
+        [InlineData("SecretUri=https://sampleurl/secrets/mysecret/", true, "https://sampleurl/", "mysecret", null)]
+        [InlineData("VaultName=sampleVault;SecretName=mysecret", true, "https://samplevault.vault.azure.net/", "mysecret", null)]
+        [InlineData("VaultName=sampleVault;SecretName=mysecret;SecretVersion=secretVersion", true, "https://samplevault.vault.azure.net/", "mysecret", "secretVersion")]
```
ejizba, line 79:
```suggestion
        [InlineData("SecretUri=https://sampleurl/secrets/mysecret/version", true, "https://sampleurl/", "mysecret", "version")]
        [InlineData("SecretUri=https://sampleurl/secrets/mysecret/version;", true, "https://sampleurl/", "mysecret", "version")] // with semicolon at the end
        [InlineData("SecretUri=https://sampleurl/secrets/mysecret/", true, "https://sampleurl/", "mysecret", null)]
        [InlineData("VaultName=sampleVault;SecretName=mysecret", true, "https://samplevault.vault.azure.net/", "mysecret", null)]
        [InlineData("VaultName=sampleVault;SecretName=mysecret;", true, "https://samplevault.vault.azure.net/", "mysecret", null)] // with semicolon at the end
        [InlineData("VaultName=sampleVault;SecretName=mysecret;SecretVersion=secretVersion", true, "https://samplevault.vault.azure.net/", "mysecret", "secretVersion")]
        [InlineData("SecretName=mysecret;VaultName=sampleVault;SecretVersion=secretVersion", true, "https://samplevault.vault.azure.net/", "mysecret", "secretVersion")] // different order
```

ejizba (commented):
Some of my suggested tests already worked with your previous changes - some only work with my suggested changes in the src file

PR #3006, opened 2022-04-01
Copy dev branch readme to v4

ejizba:
The "dev" branch has been our default for a while, so that's why it has the most up to date README. However, I want to copy it over to the "v4.x" branch for a few reasons:
1. The npm module gets it's readme from "v4.x" which means right now it looks out of date: https://www.npmjs.com/package/azure-functions-core-tools
2. Soon we will make v4.x the default branch


michaelpeng36 (approved):
LGTM

PR #3010, opened 2022-04-04
Replacing bash with sh

jshcmpbll:

https://github.com/NixOS/nixpkgs/issues/139102

In NixOS bash is not stored in `/usr/bin` I am working on getting this to compile in the derivation but as an alternative I figured I would ask if changing bash out for sh is an acceptable change for this project. Feel free to close if not.

### Pull request checklist


tnichols217:
I would recommend using `/usr/bin/env bash` instead of `/bin/sh` which should keep the shell more consistent if bash features are needed

jshcmpbll:
Totally agree but was trying to reduce friction 😄 Happy to see I have some support here though. Ill update my PR

tnichols217:
I don't work here ;-; I just had the same issue on nixos and was looking if there was already a PR or issue made. Is there a reason they haven't already merged this?

PR #3033, opened 2022-04-20
Sign mac executables

ejizba:
Related to https://github.com/Azure/azure-functions-core-tools/issues/2834, Apple [requires](https://eclecticlight.co/2020/08/22/apple-silicon-macs-will-require-signed-code/) arm64 executables to be signed in order to run. Without signing, any command just silently fails. A few notes:
1. This requirement does not apply to x64 executables, but it's still encouraged to sign binaries when possible and it's trivial for us to sign both x64 and arm64 at the same time
1. The requirement does not extend to dlls or other binaries. My changes are the minimum required to unblock users from running `func`, but we may want to consider signing those other bits in the future
1. The `.exe` file extension is unique to windows. Mac executables have _no_ file extension
1. Mac signatures are completely different than Windows signatures, so I had to reflect that in our build pipeline. For example, they do not use `sigcheck.exe`, rather a tool called `codesign` that ships with XCode on Mac. A few example commands:
    1. List signature information: `codesign -dv <file path>`
    1. Test sign a file: `codesign -s - <file path>` (This is how I originally tested arm64 bits before the last release. At the time, I didn't fully understand what the command was doing and thought it was only necessary because I was testing pre-release bits as opposed to using an official release channel like npm/brew)
1. For some reason, mac signing (through the ESRP dev ops task) requires us to zip up the files before sending them over. If I don't do that, I get an error like `Incorrect file type used for command -z. It needs to be a .zip or .dmg file.`. Importantly, it still signs the individual files _in_ the zip, not the zip itself

### Pull request checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -166,6 +166,27 @@ jobs:
       MaxConcurrency: '50'
       MaxRetryAttempts: '5'
     condition: and(succeeded(), eq(variables['IsReleaseBuild'], 'true'))
+  - task: EsrpCodeSigning@1
+    displayName: 'Mac signing'
+    inputs:
+      ConnectedServiceName: 'ESRP Service'
+      FolderPath: '$(Build.Repository.LocalPath)\artifacts\ToSign\Mac\'
+      Pattern: '*.zip'
+      signConfigType: 'inlineSignParams'
+      inlineOperation: |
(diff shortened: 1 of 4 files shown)


michaelpeng36 (approved):
LGTM

PR #3063, opened 2022-05-25
Added PSScriptRoot to validateWorkerVersions script V4

michaelpeng36:

Alters the path specified for the Core Tools .csproj file in the validateWorkerVersions.ps1 script to use `$PSScriptRoot` rather than `.`. Currently, on a Windows machine, the use of `.` sets the path to be `C:`, which causes issues when attempting to save changes to the .csproj. This change is to be ported over to the `v3.x` branch.

### Pull request checklist

  * [ ] Otherwise: This is to be ported over to the `v3.x` branch

### DIFF
--- a/validateWorkerVersions.ps1
+++ b/validateWorkerVersions.ps1
@@ -26,7 +26,7 @@ function removeBomIfExists([string]$data)
     return $data
 }
 
-$cliCsprojPath = "./src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
+$cliCsprojPath = "$PSScriptRoot/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
 $cliCsprojContent = removeBomIfExists(Get-Content $cliCsprojPath)
 $cliCsprojXml = [xml]$cliCsprojContent
 


validateWorkerVersions.ps1
```diff
@@ -26,7 +26,7 @@ function removeBomIfExists([string]$data)
     return $data
 }
 
-$cliCsprojPath = "./src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
+$cliCsprojPath = "$PSScriptRoot/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
```
ejizba, line 29: Isn't the build machine on windows? How come that wasn't failing?

ejizba (approved):

validateWorkerVersions.ps1
michaelpeng36: It looks like the `XmlDocument.Save` method converts the `.` component of the .csproj path to be `C:\WINDOWS\system32` for some reason. Strangely enough, other methods like `Set-Location` and `Get-ChildItem` do not seem to have this issue. If the issue resides solely in the `XmlDocument.Save` method, then it would make sense this would run appropriately on ADO since we do not add the `-Update` parameter there, and therefore do not execute `XmlDocument.Save`

PR #3078, opened 2022-06-23
Clean up warnings when running dotnet publish V4

michaelpeng36:

Currently, when running `dotnet publish`, several warnings appear. This PR aims to address a few of these, namely

1. `azure-functions-core-tools\src\azure.functions.cli\Extensions\HttpExtension.cs(28,59): warning CS0618: 'HttpRequestMessage.Properties' is obsolete: 'HttpRequestMessage.Properties has been deprecated. Use Options instead.'`
2. `azure-functions-core-tools\src\azure.functions.cli\Helpers\SecurityHelpers.cs(124,35): warning SYSLIB0028: 'X509Certificate2.PrivateKey' is obsolete: 'X509Certificate2.PrivateKey is obsolete. Use the appropriate method to get the private key, such as GetRSAPrivateKey, or use the CopyWithPrivateKey method to create a new instance with a private key.'`
3. `azure-functions-core-tools\src\azure.functions.cli\Kubernetes\FuncKeys\KeyBasedDictionaryComparer.cs(27,33): warning CA2013: Do not pass an argument with value type 'System.Collections.Generic.KeyValuePair<string, string>' to 'ReferenceEquals'. Due to value boxing, this call to 'ReferenceEquals' will always return 'false'.`

### Pull request checklist

  * [ ] Otherwise: The same clean up needs to be done for hte `v3.x` branch


src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
```diff
@@ -452,25 +451,6 @@ private static async Task<string> GetServiceIp(ServiceV1 service, int retryCount
...
-        private static IDictionary<string, string> GetFunctionKeys(IDictionary<string, string> currentImageFuncKeys, IDictionary<string, string> existingFuncKeys)
-        {
-            if ((currentImageFuncKeys == null || !currentImageFuncKeys.Any())
-                || (existingFuncKeys == null || !existingFuncKeys.Any()))
-            {
-                return currentImageFuncKeys;
-            }
-
-            //The function keys that doesn't exist in Kubernetes yet
-            IDictionary<string, string> funcKeys = currentImageFuncKeys.Except(existingFuncKeys, new KeyBasedDictionaryComparer()).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
michaelpeng36, line 464: This line and line 465 seem to be taking the set difference between `currentImageFuncKeys` and `existingFuncKeys` and adding back the elements in their intersection, in essence doing nothing but creating a copy of `currentImageFuncKeys`.

src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
```diff
@@ -11,35 +11,11 @@
...
 using System.Threading.Tasks;
 using System.Security.Cryptography;
-using System.Linq;
-using System.Reflection;
 
 namespace Azure.Functions.Cli.Helpers
 {
     internal static class SecurityHelpers
     {
-        private static bool CheckCurrentUser(dynamic obj)
```
ejizba, line 21: So what this just a bunch of unused methods?

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
ejizba: I don't think the names of the variables are super accurate because it seems like they have both keys _and_ values, and the values are where the difference comes in. So if the same key is defined in both `currentImageFuncKeys` and `existingFuncKeys` - I believe the value from `existingFuncKeys` is chosen over the value from `currentImageFuncKeys`.

src/Azure.Functions.Cli/Helpers/SecurityHelpers.cs
michaelpeng36: Yes. I think in previous iterations of the internal action they were used, but that action has been disabled since quite a while ago

src/Azure.Functions.Cli/Kubernetes/KubernetesHelpers.cs
michaelpeng36: Right, thanks for the catch Eric. An interesting exception to this seems to be when `currentImageFuncKeys` is `null` but `existingFuncKeys` is not `null` or empty. In either case, I'll add this function back and make the changes to the `KeyBasedDictionaryComparer`.

ejizba (approved):

PR #3099, opened 2022-07-13
Add azure-functions-core-tools as a dev dependency to init templates package.json

### LINKED ISSUE #255: Allow to install azure-functions-core-tools as dependency

GravlLift:
Add azure-functions-core-tools as a dev dependency to init templates

Adds the node package of these tools as a dev dependency in package.json by default, thus enabling users to actually run `npm start` right off the bat on a fresh template without needing the core tools globally installed.

Resolves #255 

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/javascriptPackage.json
+++ b/src/Azure.Functions.Cli/StaticResources/javascriptPackage.json
@@ -7,5 +7,7 @@
     "test": "echo \"No tests yet...\""
   },
   "dependencies": {},
-  "devDependencies": {}
+  "devDependencies": {
+    "azure-functions-core-tools": "^4.x"
+  }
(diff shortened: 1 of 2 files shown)


ejizba (approved):
Thank you!

PR #3126, opened 2022-08-16
Check for non-Azure Storage backends

cgillum:
The Durable commands currently only work with the Azure Storage backend, and not when users configure alternate backends, like MSSQL or Netherite. This PR adds detection for those other backend types and returns an error if the user tries to run a command against function apps that use those other backend types.

(I'm not aware of any issue opened for this)

### Pull request checklist

  * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr

### DIFF
--- a/src/Azure.Functions.Cli/Common/DurableManager.cs
+++ b/src/Azure.Functions.Cli/Common/DurableManager.cs
@@ -54,12 +54,14 @@ namespace Azure.Functions.Cli.Common
         public DurableManager(ISecretsManager secretsManager)
         {
             _secretsManager = secretsManager;
-            SetConnectionStringAndTaskHubName();
+            this.IsValid = TrySetConnectionStringAndTaskHubName();
         }
 
         public BackendType BackendType { get; private set; } = BackendType.AzureStorage;
 
-        private void SetConnectionStringAndTaskHubName()
+        public bool IsValid { get; private set; }
+
+        private bool TrySetConnectionStringAndTaskHubName()
         {
             // Set connection string key and task hub name to defaults
             _connectionStringKey = DefaultConnectionStringKey;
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -18,6 +18,13 @@
 
 namespace Azure.Functions.Cli.Common
 {
+    internal enum BackendType
```
cgillum, line 21: I plan on either moving this into a separate file or perhaps making it a private inner class.

src/Azure.Functions.Cli/Common/DurableManager.cs
```diff
@@ -135,6 +154,12 @@ private void SetStorageServiceAndTaskHubClient(out AzureStorageOrchestrationServ
         private void Initialize(out AzureStorageOrchestrationService orchestrationService, out TaskHubClient taskHubClient, string connectionStringKey = null, string taskHubName = null)
         {
             CheckAssemblies();
+
+            if (this.BackendType != BackendType.AzureStorage)
+            {
+                throw new CliException($"The {this.BackendType} storage provider for Durable Functions is not yet supported by this command.");
```
cgillum, line 160: TBD on whether we want to try supporting these other backends as part of this PR.

cgillum (commented):
TBD is adding tests to exercise this new code.

lilyjma:
@cgillum - I'm wondering if we should also add a link to docs in the error message to help users quickly find another way to do what they wanted with the commands.

cgillum:
@lilyjma do you recommend putting the full link in the error message, or do you think it's better to do aka.ms links?

lilyjma:
> @lilyjma Lily Ma FTE do you recommend putting the full link in the error message, or do you think it's better to do aka.ms links?

I'd recommend aka.ms link because the underlying link can be easily updated. A full link change will require a code change.

cgillum:
Hi @michaelpeng36, @AnatoliB, and @lilyjma. This PR is ready for review.

michaelpeng36 (approved):

michaelpeng36:
Thanks for the contribution, @cgillum! Would you be able to port these changes over to the `v3.x` branch as well?

cgillum:
@michaelpeng36 yes, if it's okay with you, I'd like to merge this change first so I can then do a simple cherry-pick PR to the v3.x branch.

michaelpeng36:
Sure, feel free to merge this PR

PR #3128, opened 2022-08-17
Update linuxFxVersion for publishing .NET 7 applications

michaelpeng36:
Currently, Core Tools V4 ensures the `linuxFxVersion` setting remains empty for Linux Consumption applications, updating it only for other SKUs. However, with .NET 7 function apps on Linux, we rely on setting the value to `DOTNET-ISOLATED|7.0`. This PR:

1. Updates the `linuxFxVersion` setting for Linux Consumption publishing. 
2. Automatically detects the target framework version for .NET isolated function apps by parsing the .csproj file.

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -161,13 +161,9 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 {
                     var projectRoot = ProjectHelpers.GetProject(projectFilePath);
                     var targetFramework = ProjectHelpers.GetPropertyValue(projectRoot, Constants.TargetFrameworkElementName);
-                    switch (targetFramework)
+                    if (targetFramework.Equals("net7.0", StringComparison.InvariantCultureIgnoreCase))
                     {
-                        case "net7.0":
-                            _requiredNetFrameworkVersion = "7.0";
-                            break;
-                        default:
-                            break;
+                        _requiredNetFrameworkVersion = "7.0";
                     }
                 }
                 // We do not change the default targetFramework if no .csproj file is found
@@ -373,9 +369,9 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             }
         }
 
-        private static async Task UpdateNetFrameworkVersionWindows(Site functionApp, string dotnetFramworkVersion, AzureHelperService helperService)
+        private static async Task UpdateNetFrameworkVersionWindows(Site functionApp, string dotnetFrameworkVersion, AzureHelperService helperService)
         {
-            string normalizedVersion = NormalizeDotnetFrameworkVersion(dotnetFramworkVersion);
+            string normalizedVersion = NormalizeDotnetFrameworkVersion(dotnetFrameworkVersion);
 
             // Websites ensure it begins with 'v'.
             string version = $"v{normalizedVersion}";
--- a/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -405,47 +405,28 @@ namespace Azure.Functions.Cli.Actions.AzureActions
         {
             string normalizedVersion = NormalizeDotnetFrameworkVersion(dotnetFramworkVersion);
 
-            string linuxFxVersion;
+            string linuxFxVersion = $"DOTNET-ISOLATED|{normalizedVersion}";
 
-            if (functionApp.IsDynamic)
+            // If things are already set, do nothing
+            if (string.Equals(functionApp.LinuxFxVersion, linuxFxVersion, StringComparison.OrdinalIgnoreCase))
             {
-                if (!string.IsNullOrEmpty(functionApp.LinuxFxVersion))
-                {
-                    linuxFxVersion = string.Empty;
-                }
-                else
-                {
-                    return;
-                }
-            }
-            else
-            {
-                linuxFxVersion = $"DOTNET-ISOLATED|{normalizedVersion}";
-
-                // If things are already set, do nothing
-                if (string.Equals(functionApp.LinuxFxVersion, linuxFxVersion, StringComparison.OrdinalIgnoreCase))
-                {
-                    return;
-                }
+                return;
             }
 
-            if (linuxFxVersion != null)
+            ColoredConsole.WriteLine($"Updating '{Constants.LinuxFxVersion}' to '{linuxFxVersion}'.");
+
+            var updatedSettings = new Dictionary<string, string>
             {
-                ColoredConsole.WriteLine($"Updating '{Constants.LinuxFxVersion}' to '{linuxFxVersion}'.");
+                [Constants.LinuxFxVersion] = linuxFxVersion
+            };
 
-                var updatedSettings = new Dictionary<string, string>
-                {
-                    [Constants.LinuxFxVersion] = linuxFxVersion
-                };
+            var settingsResult = await helperService.UpdateWebSettings(functionApp, updatedSettings);
 
-                var settingsResult = await helperService.UpdateWebSettings(functionApp, updatedSettings);
-
-                if (!settingsResult.IsSuccessful)
-                {
-                    ColoredConsole.Error
-                        .WriteLine(ErrorColor("Error updating linux image property:"))
-                        .WriteLine(ErrorColor(settingsResult.ErrorResult));
-                }
+            if (!settingsResult.IsSuccessful)
+            {
+                ColoredConsole.Error
+                    .WriteLine(ErrorColor("Error updating linux image property:"))
+                    .WriteLine(ErrorColor(settingsResult.ErrorResult));
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -151,6 +151,28 @@ public override async Task RunAsync()
...
+                string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
+                if (projectFilePath != null)
+                {
+                    var projectRoot = ProjectHelpers.GetProject(projectFilePath);
+                    var targetFramework = ProjectHelpers.GetPropertyValue(projectRoot, Constants.TargetFrameworkElementName);
+                    switch (targetFramework)
+                    {
+                        case "net7.0":
+                            _requiredNetFrameworkVersion = "7.0";
+                            break;
```
pragnagopa, line 168: please add the same for `dotnet-isolated|6.0`

test/Azure.Functions.Cli.Tests/PublishActionTests.cs
```diff
@@ -14,38 +14,26 @@ public class PublishActionTests
     {
         TestAzureHelperService _helperService = new TestAzureHelperService();
 
-        [Fact]
-        public async Task NetFrameworkVersion_DotnetIsolated_Linux_Consumption_AlreadyEmpty()
+        [Theory]
+        [InlineData(null)]
+        [InlineData("something")]
+        public async Task NetFrameworkVersion_DotnetIsolated_Linux_Consumption_Updated(string initialLinuxFxVersion)
         {
```
pragnagopa, line 21: please add tests for `dotnet-isolated|5.0`, `dotnet-isolated|6.0` and `dotnet-isolated|7.0`,

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -151,6 +151,28 @@ public override async Task RunAsync()
...
+            if (workerRuntime == WorkerRuntime.dotnetIsolated)
+            {
+                string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
+                if (projectFilePath != null)
+                {
+                    var projectRoot = ProjectHelpers.GetProject(projectFilePath);
+                    var targetFramework = ProjectHelpers.GetPropertyValue(projectRoot, Constants.TargetFrameworkElementName);
+                    switch (targetFramework)
+                    {
+                        case "net7.0":
```
kshyju, line 166: Should this be case insensitive check? I think `NET7.0` is valid value to be used when creating a new function app.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -29,7 +29,7 @@ internal class PublishFunctionAppAction : BaseFunctionAppAction
 
         private readonly ISettings _settings;
         private readonly ISecretsManager _secretsManager;
-        private const string _requiredNetFrameworkVersion = "6.0";
+        private static string _requiredNetFrameworkVersion = "6.0";
```
kshyju, line 32: Should this be `static` ? Looks like a private variable used just within this class.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -380,47 +402,28 @@ private static async Task DotnetIsolatedLinuxValidation(Site functionApp, string
         {
             string normalizedVersion = NormalizeDotnetFrameworkVersion(dotnetFramworkVersion);
```
kshyju, line 403: nit: typo in var name.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
kshyju: Can multiple instances of core tools run at same time? It might be better to avoid global variables to store state in that case.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
michaelpeng36: Updating from an offline conversation: From [Supported languages in Azure Functions | Microsoft Docs](https://docs.microsoft.com/en-us/azure/azure-functions/supported-languages#languages-by-runtime-version), it doesn't look like .NET 5 is expected to be supported by Functions V4, and since users should be using Core Tools V4 to publish Functions V4 apps, I think it should be fine to just check for .NET 7.0. The default value for `_requiredNetFrameworkVersion` should also be `"6.0"` already.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
michaelpeng36: Updating from an offlien conversation: we do not expect multiple instances of Core Tools to run simultaneously. The reason the `_requiredNetFrameworkVersion` variable is `static` is to allow it to be used by static methods like `UpdateFrameworkVersions`, which then are easily test-able. I agree in this case it might be better to make these methods just `private`, but it would also require further changes in our tests. I'd be happy to address those changes in a separate PR.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -151,6 +151,24 @@ public override async Task RunAsync()
...
+            // NOTE: .NET 7.0 is only supported on dotnet-isolated
+            // TODO: Include proper steps for publishing a .NET Framework 4.8 application
+            if (workerRuntime == WorkerRuntime.dotnetIsolated)
+            {
+                string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
+                if (projectFilePath != null)
+                {
+                    var projectRoot = ProjectHelpers.GetProject(projectFilePath);
+                    var targetFramework = ProjectHelpers.GetPropertyValue(projectRoot, Constants.TargetFrameworkElementName);
+                    if (targetFramework.Equals("net7.0", StringComparison.InvariantCultureIgnoreCase))
```
kshyju, line 164: For .NET 5 Isolated project (`net5.0`), Would the `_requiredNetFrameworkVersion` value will be incorrect (6.0)? Let's make sure net5 app deployment also work.

src/Azure.Functions.Cli/Helpers/ProjectHelpers.cs
```diff
@@ -40,16 +46,22 @@ public static string FindProjectFile(string path, LoggingFilterHelper loggingFil
                 {
                     foreach (FileInfo file in projectFiles)
                     {
-                        if (string.Equals(file.Name, Constants.ExtenstionsCsProjFile, StringComparison.OrdinalIgnoreCase)) continue;
-                        logger.LogDebug($"Found {file.FullName}. Using for user secrets file configuration.");
+                        if (string.Equals(file.Name, Constants.ExtensionsCsProjFile, StringComparison.OrdinalIgnoreCase)) continue;
+                        if (shouldLog)
```
kshyju, line 50: Did you mean to have this conditional logging change in this PR? Is that relevant to the NET7 publishing change?

test/Azure.Functions.Cli.Tests/PublishActionTests.cs
```diff
@@ -14,38 +14,26 @@ public class PublishActionTests
...
-            await PublishFunctionAppAction.UpdateFrameworkVersions(site, WorkerRuntime.dotnetIsolated, "6.0", false, _helperService);
-
-            // no-op if already null or empty
-            Assert.Null(_helperService.UpdatedSettings);
-        }
-
-        [Fact]
-        public async Task NetFrameworkVersion_DotnetIsolated_Linux_Consumption_Updated()
+        [Theory]
+        [InlineData(null, "6.0")]
```
kshyju, line 18: Also use `net5.0` as input

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
michaelpeng36: We don't plan on supporting .NET 5.0 deployments with Core Tools V4 since Core Tools V4 is tied to Functions V4, and according to [Supported languages in Azure Functions | Microsoft Docs](https://docs.microsoft.com/en-us/azure/azure-functions/supported-languages#languages-by-runtime-version), .NET 5.0 is supported only on Functions V3.

src/Azure.Functions.Cli/Helpers/ProjectHelpers.cs
michaelpeng36: Yes, this change is to avoid unnecessary logging when attempting to find the function app's `.csproj` file, since we parse it to find the appropriate target framework when publishing. An alternative option I considered was making a copy of this method and removing the sections involving logging, but I believe this to be a more concise way to accomplish the same goal since:
1. The primary use of this method is to find the path to the `.csproj` file.
2. Any use of this method that does use logging should not be using `null` values for either `loggingFilterHelper` or `loggerFilterOptions`

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
kshyju: Got it. Thanks!

kshyju (approved):
Take my approval with a grain of salt as I am not super familiar with core tools.

ejizba (approved):

test/Azure.Functions.Cli.Tests/PublishActionTests.cs
michaelpeng36: Resolved, as we do not plan on supporting .NET 5.0 scenarios with Core Tools V4

pragnagopa (approved):

PR #3145, opened 2022-09-07
Improve publishing speed for Typescript apps

wwestrop:
The default template for a Typescript function app (created using `func init --language typescript --worker-runtime node`) will install development tooling into `node_modules` when you run `npm install`.

Publishing this simple app using `func azure functionapp publish myappname` includes all this, and is therefore very slow.

This PR reduces the amount deployed to Azure by about 181 MB (on Windows) so it is substantially faster to publish.

### Pull request checklist

  * [ ] ~Otherwise: Documentation issue linked to PR~
  * [ ] ~Otherwise: Backport tracked by issue/PR #issue_or_pr~
* [ ] ~I have added all required tests (Unit tests, E2E tests)~

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/funcignore
+++ b/src/Azure.Functions.Cli/StaticResources/funcignore
@@ -5,8 +5,6 @@
 local.settings.json
 test
 getting_started.md
-node_modules/.bin/
-node_modules/@azure/
 node_modules/@types/
 node_modules/azure-functions-core-tools/
 node_modules/typescript/
--- a/src/Azure.Functions.Cli/StaticResources/funcignore
+++ b/src/Azure.Functions.Cli/StaticResources/funcignore
@@ -5,3 +5,8 @@
 local.settings.json
 test
 getting_started.md
+node_modules/.bin/
+node_modules/@azure/
+node_modules/@types/
+node_modules/azure-functions-core-tools/
+node_modules/typescript/


src/Azure.Functions.Cli/StaticResources/funcignore
```diff
@@ -5,3 +5,8 @@
 local.settings.json
 test
 getting_started.md
+node_modules/.bin/
+node_modules/@azure/
```
ejizba, line 9: We can't exclude this one because the `@azure` scope has a lot of sdks and whatnot that people use in production.

The "@azure/functions" package is also shifting to a production dependency (More info [here](https://github.com/Azure/azure-functions-nodejs-worker/issues/568) if you're curious).

src/Azure.Functions.Cli/StaticResources/funcignore
```diff
@@ -5,3 +5,8 @@
 local.settings.json
 test
 getting_started.md
+node_modules/.bin/
```
ejizba, line 8: I'm a bit nervous about excluding this one because literally any node module could put something in here, not just dev dependencies. I'd rather leave it off for now.

src/Azure.Functions.Cli/StaticResources/funcignore
```diff
@@ -5,3 +5,8 @@
...
+node_modules/@azure/
+node_modules/@types/
+node_modules/azure-functions-core-tools/
```
ejizba, line 11: I think this is the most impactful line, and I think it makes total sense to add it. Thank you!!!

Fwiw, adding to the funcignore will help with one-off deployments from core tools. However, if you deploy through a CI/CD pipeline I would recommend adding `npm prune --production` before deploying, which will more consistently get rid of _all_ dev dependencies.

Another fwiw, VS Code will run `npm prune --production` by default if you deploy through there.

wwestrop: `npm prune` is a good shout, I was just trying to do a quick-n-dirty functionapp in this case 😉

wwestrop:
I've made the changes requested - `typescript` and `azure-functions-core-tools` between them account for almost the entire payload so it's still a very good saving if you forget to prune 👍

ejizba (approved):
Thank you!

PR #3249, opened 2023-01-19
.NET 6 Testing Update

VpOfEngineering:

With the dotnet 6 SDK templating changed, and the command `dotnet new HttpTrigger` is no longer valid. The new conventioon is `dotnet new http`.

This PR changes dotnet tests that use the old convention. It also removes one test that is no longer valid (as there is no space in http).

### Pull request checklist

### DIFF
--- a/test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
+++ b/test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
@@ -20,11 +20,11 @@ namespace Azure.Functions.Cli.Tests.E2E
                 Commands = new[]
                 {
                     "init . --worker-runtime dotnet",
-                    "new --template HttpTrigger --name testfunc"
+                    "new --template http --name testfunc"
                 },
                 OutputContains = new[]
                 {
-                    "The function \"testfunc\" was created successfully from the \"HttpTrigger\" template."
+                    "The function \"testfunc\" was created successfully from the \"http\" template."
                 }
             }, _output);
         }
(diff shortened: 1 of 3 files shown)


test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
```diff
@@ -175,23 +175,6 @@ public async Task create_template_function_js_no_space_name()
...
-                    "new --template \"http trigger\" --name testfunc2"
-                },
-                OutputContains = new[]
-                {
-                    "The function \"testfunc2\" was created successfully from the \"http trigger\" template."
-                }
-            }, _output);
-        }
-
         [Fact]
```
khkh-ms, line 178: Why did we remove this test?

khkh-ms (approved):

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
VpOfEngineering: This test is no longer valid. It tested when creating a function with a space in the template name (e.g. `http trigger`), and the new format `http` doesn't have a space.

VpOfEngineering: @michaelpeng36 - is this a fair assessment?

michaelpeng36: If having the customer use the .NET 6 CLI with Core Tools V3 is an expected scenario, then I agree this is a fair assessment. However, my understanding was users who want V3 will be expected to use .NET 3.1, in which case, I am unsure of whether it accepts template short names.

michaelpeng36: [Here](https://github.com/Azure/azure-functions-core-tools/blob/e1d7373f2d72f645652460218a780dc451edb4f2/src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs#L36) is the underlying `dotnet` command that we run for .NET and .NET-isolated apps. If we expect users to have .NET 6 with Core Tools V3, there should be some mapping from template names to template short names, as in the V4 version of the `DotnetHelpers.cs` file [here](https://github.com/Azure/azure-functions-core-tools/blob/50d214d8b12dcfa6d7166632cc0e8c2ba68fa63b/src/Azure.Functions.Cli/Helpers/DotnetHelpers.cs#L75).

PR #3319, opened 2023-03-24
Handling a telemetry exception where a Linux file path isn't readable

### LINKED ISSUE #3236: Telemetry docker container check throws an exception when a Linux path isn't readable
On Linux environments where the path '/proc/1/cgroup' can't be read, the telemetry checking code will throw, which doesn't appear to be caught anywhere else.

The code in question is found here:

https://github.com/Azure/azure-functions-core-tools/blob/db2ff784941c3f6241aea996818ba370ea441bd2/src/Azure.Functions.Cli/Telemetry/DockerContainerDetectorForTelemetry.cs#L44-L49

This issue is a little more nuanced as there appears to be a copy of this code in the dotnet SDK, where the same code has already been fixed up to avoid this issue.

You can see the fixed code with try-catch here:

https://github.com/dotnet/sdk/blob/f765d4255f9a4609deda8b1c8ea9a600c8dd52ff/src/Cli/dotnet/Telemetry/DockerContainerDetectorForTelemetry.cs#L35-L50

And the PR which introduces the required fixes:

https://github.com/dotnet/sdk/pull/22927

I don't mind putting together a simple PR to draw the files back in line again and fix this issue, but should we consider a better way to share code between the tools? Or at least keep them in sync (or know when they get out of sync)?

andrwca:
This fix has already been applied to the .NET SDK under PR https://github.com/dotnet/sdk/pull/22927.

Also disabling unnecessary platform compatibility warnings where the the platform specific APIs are in the correct platform specific code path.

Resolves #3236 

### Pull request checklist

Note: I don't know what the backporting rules are. Happy to open other PRs to back port if necessary. 

Unit test wise, I can add some, but it'll require a little refactoring. Let me know if it's needed.

### DIFF
--- a/src/Azure.Functions.Cli/Telemetry/DockerContainerDetectorForTelemetry.cs
+++ b/src/Azure.Functions.Cli/Telemetry/DockerContainerDetectorForTelemetry.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Security;
 using Microsoft.Win32;
 using Microsoft.DotNet.PlatformAbstractions;
+using System;
 
 namespace Azure.Functions.Cli.Telemetry
 {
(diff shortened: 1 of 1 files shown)


kshyju (approved):
Thank you for your contribution @andrwca. The change looks good.

PR #3339, opened 2023-04-21
Update TypeScript package.json templates

ejizba:
- Add a clean script to model v4, related to [this issue](https://github.com/ejizba/func-nodejs-prototype/issues/18) by @hossam-nasr
- Update a few versions

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/package-ts-v4.json
+++ b/src/Azure.Functions.Cli/StaticResources/package-ts-v4.json
@@ -6,16 +6,18 @@
   "scripts": {
     "build": "tsc",
     "watch": "tsc -w",
-    "prestart": "npm run build",
+    "clean": "rimraf dist",
+    "prestart": "npm run clean && npm run build",
     "start": "func start",
     "test": "echo \"No tests yet...\""
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


src/Azure.Functions.Cli/StaticResources/package-ts-v4.json
```diff
@@ -6,16 +6,18 @@
   "scripts": {
     "build": "tsc",
     "watch": "tsc -w",
-    "prestart": "npm run build",
+    "clean": "rimraf dist",
+    "prestart": "npm run clean && npm run build",
```
hossam-nasr, line 10: Doesn't running `npm run clean && npm run build` still run into the same issue of not being compatible on all shells (particularly PowerShell)? I thought that was the main argument against putting `rimraf dist &&` directly in the `prestart` script

ejizba: Yeah that was a concern, but turns out PowerShell added support for those in PowerShell 7 (first released in 2020 I think)

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pipeline_chain_operators

hossam-nasr (approved):

PR #3347, opened 2023-04-27
Adding support for 'func AzureContainerApps deploy'

khkh-ms:

resolves #issue_for_this_pr

### Pull request checklist


src/Azure.Functions.Cli/Actions/ContainerService/AzureContainerAppsDeployAction.cs
```diff
@@ -0,0 +1,191 @@
...
+using Azure.Functions.Cli.ContainerApps.Models;
+using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Interfaces;
+using Colors.Net;
+using Fclp;
+using static Azure.Functions.Cli.Common.OutputTheme;
+
+namespace Azure.Functions.Cli.Actions.ContainerServiceActions
+{
+    [Action(Name = "deploy", Context = Context.AzureContainerApps, HelpText = "")]
```
michaelpeng36, line 19: Please include documentation in the `HelpText` parameter here so that it does not show up as a blank entry when running the `func azurecontainerapps` command for help.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -62,7 +66,15 @@ private static async Task<Site> TryGetFunctionAppFromArg(string name, IEnumerabl
             {
                 string siteId = await GetResourceIDFromArg(subscriptions, query, accessToken, managementURL);
                 var app = new Site(siteId);
-                await LoadFunctionApp(app, accessToken, managementURL);
+                if (loadFunction != null)
+                {
+                    await loadFunction(app, accessToken, managementURL);
```
michaelpeng36, line 71: Could you add a comment here explaining the difference between `loadFunction` and `LoadFunctionApp`? While the usage here is straightforward, I think it would be good to clarify when to use each method. Perhaps we could also rename the parameter `loadFunction` as `loadContainerApp`. In general, however, if another service would like to add support on top of Core Tools, adding an entirely new `loadFunction` parameter to this method doesn't appear to be sustainable. Is there a way we could pass an enumerable here instead to describe which type of site we are trying to load?

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -326,8 +350,8 @@ public static async Task<Site> LoadSiteObjectAsync(Site site, string accessToken
             var url = new Uri($"{managementURL}{site.SiteId}?api-version={ArmUriTemplates.WebsitesApiVersion}");
             var armSite = await ArmHttpAsync<ArmWrapper<ArmWebsite>>(HttpMethod.Get, url, accessToken);
 
-            site.HostName = armSite.properties.enabledHostNames.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) == -1);
-            site.ScmUri = armSite.properties.enabledHostNames.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) != -1);
+            site.HostName = armSite.properties.enabledHostNames?.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) == -1);
```
michaelpeng36, line 353: Clarification point: are there scenarios now where we expect the `enabledHostName` to be `null`? If so, is it acceptable for the `site.HostName` and `site.ScmUri` to be `null`?

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -449,11 +473,63 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+            {
+                try
+                {
+                    var content = response.Content.ReadAsStringAsync().Result;
+                    var errorResponse = JsonConvert.DeserializeObject<ContainerAppsFunctionDeployResponse>(content);
+                    throw new CliException($"The function app deployment to Container Apps failed. {(!string.IsNullOrWhiteSpace(errorResponse?.Message)? $"Error: {errorResponse.Message}" : string.Empty )}");
+                }
+                catch (CliException)
+                {
+                    throw;
```
michaelpeng36, line 491: Why is this try-catch block necessary? It seems the only other places where an exception can occur here are when the content is being read and deserialized, at which point I think the specified methods would throw their own exceptions anyway.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -449,11 +473,63 @@ private static async Task ArmHttpAsync(HttpMethod method, Uri uri, string access
...
+        public static async Task CreateFunctionAppOnContainerService(string accessToken, string managementURL, string subscriptionId, string resourceGroup, ContainerAppsFunctionPayload payload)
+        {
+            var url = new Uri($"{managementURL}/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Web/sites/{payload.Name}?api-version={ArmUriTemplates.FunctionAppOnContainerAppsApiVersion}");
+            ColoredConsole.WriteLine(Constants.FunctionAppDeploymentToContainerAppsMessage);
+            var response = await ArmClient.HttpInvoke(HttpMethod.Put, url, accessToken, payload);
+            if (!response.IsSuccessStatusCode)
+            {
+                try
+                {
+                    var content = response.Content.ReadAsStringAsync().Result;
```
michaelpeng36, line 485: Is there a reason we are using `.Result` here instead of `await`?

michaelpeng36 (commented):
Left a few questions to clarify, but nothing blocking

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
khkh-ms: The null check is probably not needed anymore. The `enabledHostNames` was coming as `null` earlier for function apps on Container Apps. They are continuously working the backend API. It is returning the list now.  It will takeout the null check as it is not needed anymore.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
khkh-ms: I don't understand the statement. ` at which point I think the specified methods would throw their own exceptions anyway.`. 

Yes, the idea is the same. The backend can't guarantee that content will always be JSON when the result is not `success`. Therefore, `JsonConvert.DeserializeObject` can throw exception. We need to catch that. The idea is that if `Message` field cannot be read from the content for any reason/exception then the default message is delivered in the UI. 

I am simplifying the try/catch through.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
khkh-ms: This was an oversight. Thanks for catching it.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
khkh-ms: Added the comment.

src/Azure.Functions.Cli/Actions/ContainerService/AzureContainerAppsDeployAction.cs
khkh-ms: Added the `HelpText`.

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
michaelpeng36: Right, my concern was primarily about rethrowing the same exception since that would obfuscate the stack trace. I think in general, unless we want to throw our own custom `CliException` that it would be best for the exception to originate from the line in which the code failed rather than rethrowing. However, this appears to be addressed with your recent changes. Thanks!

michaelpeng36 (approved):

khkh-ms:
Merging the PR.

PR #3371, opened 2023-05-17
Split 'npm install' and 'npm run build' stages

### LINKED ISSUE #3370: Update nodejs Dockerfile to improve layer caching
The default `Dockerfile` for a nodejs function created via `func init` is currently:

```Dockerfile
# To enable ssh & remote debugging on app service change the base image to the one below
# FROM mcr.microsoft.com/azure-functions/node:4-node16-appservice
FROM mcr.microsoft.com/azure-functions/node:4-node16

ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
    AzureFunctionsJobHost__Logging__Console__IsEnabled=true

COPY . /home/site/wwwroot

RUN cd /home/site/wwwroot && \
    npm install && \
    npm run build
```

This copies the source content and runs `npm install` and `npm run build` as a single build step.

We can leverage layer caching to speed up repeated image builds by splitting this into two steps:
1. Copy `package.json` and run `npm install`
2. Copy full source and run `npm run build`

By doing this, the layer from the first step is only invalidated when `package.json` changes, so making other changes to your code doesn't invalidate this and subsequent builds will skip the `npm install` step. This speeds up repeated builds of the image and also repeated image push/pulls (since only the changed layers need transferring)

stuartleeks:

resolves #3370 

### Pull request checklist

* [X] I have added all required tests (Unit tests, E2E tests) - I don't believe there are any new tests required by this change

### DIFF
--- a/src/Azure.Functions.Cli/StaticResources/Dockerfile.node18
+++ b/src/Azure.Functions.Cli/StaticResources/Dockerfile.node18
@@ -5,7 +5,11 @@ FROM mcr.microsoft.com/azure-functions/node:4-node18
 ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     AzureFunctionsJobHost__Logging__Console__IsEnabled=true
 
-COPY . /home/site/wwwroot
+WORKDIR /home/site/wwwroot
 
-RUN cd /home/site/wwwroot && \
-    npm install
+# Copy package.json and install packages before copying the rest of the code to enable caching
+COPY package.json package.json
+RUN npm install
+
+# Copy the rest of the code
+COPY . .
--- a/src/Azure.Functions.Cli/StaticResources/Dockerfile.node14
+++ b/src/Azure.Functions.Cli/StaticResources/Dockerfile.node14
@@ -5,7 +5,11 @@ FROM mcr.microsoft.com/azure-functions/node:4-node14
 ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     AzureFunctionsJobHost__Logging__Console__IsEnabled=true
 
-COPY . /home/site/wwwroot
+WORKDIR /home/site/wwwroot
 
-RUN cd /home/site/wwwroot && \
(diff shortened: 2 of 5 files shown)


stuartleeks:
Is there any feedback on this?

mattchenderson (dismissed):
This is a good improvement, and these changes seem good to me. My approval should not be considered fully sufficient for a merge, but I think we should be able to take this one as-is. We should still do some quick validation passes though first.

mattchenderson:
@ejizba I'd like to also make sure you're aware of these changes before we complete this. Could you please take a look?

src/Azure.Functions.Cli/StaticResources/Dockerfile.node18
```diff
@@ -5,7 +5,12 @@ FROM mcr.microsoft.com/azure-functions/node:4-node18
...
-RUN cd /home/site/wwwroot && \
-    npm install
\ No newline at end of file
+# Copy package.json and install packages before copying the rest of the code to enable caching
+COPY package.json package.json
+RUN npm install
+
+# Copy the rest of the code
+COPY . .
+RUN  npm run build
```
ejizba, line 16: Hi @stuartleeks sorry for the delay and thanks for the contribution! Please remove the `npm run build` step from the 3 files without `.typescript` in the name. These 3 files should be for JavaScript, which doesn't need a build step and didn't have one before your change

ejizba (changes requested):

src/Azure.Functions.Cli/StaticResources/Dockerfile.node18
stuartleeks: Oops - good catch, thanks!

stuartleeks: Updated and rebased on latest `v4.x`

ejizba (approved):
Thanks!

PR #3458, opened 2023-09-18
Enabling Worker Indexing

khkh-ms:

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -190,6 +190,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
             settings.AddRange(LanguageWorkerHelper.GetWorkerConfiguration(LanguageWorkerSetting));
             _keyVaultReferencesManager.ResolveKeyVaultReferences(settings);
             UpdateEnvironmentVariables(settings);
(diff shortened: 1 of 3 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -299,13 +299,20 @@ private async Task<IWebHost> BuildWebHost(ScriptApplicationHostOptions hostOptio
...
         }
 
         private void EnableDotNetWorkerStartup()
         {
             Environment.SetEnvironmentVariable("DOTNET_STARTUP_HOOKS", "Microsoft.Azure.Functions.Worker.Core");
         }
+        
+        private void EnableWorkerIndexing()
+        {
+            Environment.SetEnvironmentVariable("FunctionsHostingConfig__WORKER_INDEXING_ENABLED", 1.ToString());
```
ejizba, line 314: Can we do this after handling local.settings.json and only if the env var isn't defined already? It would be nice to leave customers a way to turn off worker indexing if necessary, and AFAIK setting this to 0 would be the best way for them

khkh-ms: @ejizba Not setting the env variable anymore if value is set in local.settings.json.

ejizba (approved):

PR #3481, opened 2023-10-10
Khkh/multiple installs checks

khkh-ms:

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HelpAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HelpAction.cs
@@ -58,7 +58,7 @@ namespace Azure.Functions.Cli.Actions
 
         public async override Task RunAsync()
         {
-            var latestVersionMessageTask = VersionHelper.CheckLatestVersionAsync();
+            var latestVersionMessageTask = VersionHelper.IsRunningAnOlderVersion();
             ScriptHostHelpers.SetIsHelpRunning();
             if (!string.IsNullOrEmpty(_context) || !string.IsNullOrEmpty(_subContext))
             {
@@ -97,18 +97,18 @@ namespace Azure.Functions.Cli.Actions
                 DisplayGeneralHelp();
             }
 
-            await ShowLatestVersionMessage(latestVersionMessageTask);
+            await RunVersionCheckTask(latestVersionMessageTask);
             return;
         }
 
-        private static async Task ShowLatestVersionMessage(Task<string> latestVersionMessageTask)
+        private static async Task RunVersionCheckTask(Task<(bool, string)> versionCheckTask)
         {
             try
             {
-                var latestVersionMessage = await latestVersionMessageTask;
-                if (!string.IsNullOrEmpty(latestVersionMessage))
+                var versionCheckMessage = await VersionHelper.RunAsync(versionCheckTask);
+                if (!string.IsNullOrEmpty(versionCheckMessage))
                 {
-                    ColoredConsole.WriteLine(WarningColor($"{latestVersionMessage}{Environment.NewLine}"));
+                    ColoredConsole.WriteLine(WarningColor($"{versionCheckMessage}{Environment.NewLine}"));
                 }
             }
             catch (Exception)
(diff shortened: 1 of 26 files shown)


src/Azure.Functions.Cli/Helpers/VersionHelper.cs
```diff
@@ -13,11 +17,26 @@ namespace Azure.Functions.Cli.Helpers
...
+            }
+
+            return olderVersionWarning;
+        }
 
         // Check that current core tools is the latest version. 
         // To ensure that it doesn't block other tasks. The HTTP Request timeout is only 1 second. 
         // We simply ingnore the exception if for any reason the check fails. 
-        public static async Task<string> CheckLatestVersionAsync()
+        public static async Task<(bool,string)> IsRunningAnOlderVersion()
```
kamperiadis, line 39: Can we just make this function return Task<bool>? The only place where that message seems to be used anyways is for L33 of this file, so couldn't we just replace `olderVersionWarning` with `Constants.OldCoreToolsVersionMessage`?

kamperiadis (changes requested):

src/Azure.Functions.Cli/Helpers/VersionHelper.cs
```diff
@@ -13,11 +17,26 @@ namespace Azure.Functions.Cli.Helpers
 {
     internal class VersionHelper
     {
+        public static async Task<string> RunAsync(Task<(bool, string)> isRunningOlderVersion = null)
+        {
+            isRunningOlderVersion ??= IsRunningAnOlderVersion();
+
+            (var isOlderVersion, var olderVersionWarning) = await isRunningOlderVersion;
+
+            var multipleInstallsWarning = await HaveMultipleInstallations(isOlderVersion);
```
kamperiadis, line 26: Couldn't we just have `HaveMultipleInstallations` not consume the `isOlderVersion` param and just have the `RunAsync` function handle the strings to be shown? That `isOlderVersion` param is only used to determine what message to present customers but it is not related to determining whether the customer has multiple installations.

khkh-ms: I combined the old version and multiple versions message together when necessary. Showing both messages together was too wordy. 

I also started thinking the same as you are thinking but there is a part of the multiple version message being generated within `HaveMultipleInstallations` method. I wanted to generate the whole multiple version message at the same place therefore I decided to that in the method. The mentioned param was needed to help generate the full message as explained above.

src/Azure.Functions.Cli/Helpers/VersionHelper.cs
khkh-ms: Addressing it.

src/Azure.Functions.Cli/Helpers/VersionHelper.cs
kamperiadis: I think it is confusing for the naming convention of the function `HasMultipleInstallations` to serve two purposes (determine if there are multiple installations and then also determining what error message to provide users). In my opinion, it should only be used to determine if there are multiple installations and then the message handling should be determined elsewhere. If we need the directory information, then maybe it should not be named `HasMultipleInstallations` but rather something like `GetInstallationDirectories` and return the list of directories.

src/Azure.Functions.Cli/Helpers/VersionHelper.cs
khkh-ms: I have addressed it.

src/Azure.Functions.Cli/Helpers/VersionHelper.cs
khkh-ms: The output of the "where" command which determines the multiple installations, it is the same output which is used to generate the message. If we do the determine and message generation in the two different places, then we have run the run the "where" command twice which is already expensive. 

I agree with the name. Naming is sometimes tricky when we expected that the name should reflect the behavior the method when there is more going on the method. I am open for the suggestion.

src/Azure.Functions.Cli/Actions/HelpAction.cs
```diff
@@ -97,18 +97,20 @@ public async override Task RunAsync()
                 DisplayGeneralHelp();
             }
 
-            await ShowLatestVersionMessage(latestVersionMessageTask);
+            // var isOldVersion = await latestVersionMessageTask; //await Constants.OldCoreToolsVersionMessage
```
kamperiadis, line 100: LGTM! Just a minor comment. Can you remove this commented out code?

kamperiadis (approved):
LGTM!

PR #3512, opened 2023-11-06
Revert "Host revert to 4.25"

khkh-ms:
Reverts Azure/azure-functions-core-tools#3500


vrdmr:
Can we make it explicit that this change updates the Host to what version?

kamperiadis (approved):

PR #3570, opened 2024-01-11
updating default TFM and help strings

### LINKED ISSUE #3569: Update default version for .NET isolated in Core Tools
Now that the system is optimized around .NET 8, we should change the default in the Core Tools so that if the user doesn't specify the target framework, they get .NET 8.

I believe the following areas need to be updated. In all cases where a list of versions is presented, we should order them descending. So `\"net8.0\", \"net7.0\", \"net6.0\", and \"net48\"` would be ideal. This will also put the default value first in the list.

https://github.com/Azure/azure-functions-core-tools/blob/dd158c50d09a9de6c76fed62199b960c0542b3fc/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs#L334-L340

https://github.com/Azure/azure-functions-core-tools/blob/dd158c50d09a9de6c76fed62199b960c0542b3fc/build/Settings.cs#L180

https://github.com/Azure/azure-functions-core-tools/blob/dd158c50d09a9de6c76fed62199b960c0542b3fc/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs#L119

mattchenderson:
String and enum updates. I tested ad hoc. I don't believe additional unit tests are needed but please correct me on that. Docs already reflect passing the parameter itself.

resolves #3569

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR
  * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr

### DIFF
--- a/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
+++ b/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs
@@ -116,7 +116,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
             Parser
                 .Setup<string>("target-framework")
-                .WithDescription("Initialize a project with the given target framework moniker. Currently supported only when --worker-runtime set to dotnet-isolated. Options are - \"net48\", \"net6.0\", \"net7.0\", and \"net8.0\"")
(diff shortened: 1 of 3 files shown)


mattchenderson:
Ah there was a test update missing. Will update.

mattchenderson:
@khkh-ms is the current error just something transient? It looks like a typescript test timeout.

Issue #3537, opened 2023-11-17
[Function Core Tools] --dotnetIsolated doesn't exist as a worker runtime option to publish from function core tools  


Justrebl:
Hello, 

**Context** 
The **_func azure functionapp publish_** command has 2 options to define the function worker runtime of the target Function App : 
- setting FUNCTION_WORKER_RUNTIME setting in local.settings.json in a local dev environment
- Defining the <language> option as a command parameter as _--csharp --typescript --custom_, etc.

**Issue**
Trying to publish a dotnet isolated function app from a github action isn't an option as we need to specify the FUNCTION_WORKER_RUNTIME to be set in the local.settings.json : _There doesn't seem to be any specific func azure functionapp publish_ <language> option for dotnetisolated (while **_--csharp will default to .Net in-process runtime_**, not .Net Isolated).

**Repro Steps**
1. Create a fresh dotnet isolated function app from the function core tools
2. Make sure the proper .gitignore is set (to avoid pushing local.settings.json with the function app code)
3. Push the function app code to a Github repository
4. Create a Github Action that will login to Azure 
5. Add a step to run the following command : func azure functionapp publish <function-app-name>
6. Running the Github action should fail at the func publish step with the following error : 
```
Your Azure Function App has 'FUNCTIONS_WORKER_RUNTIME' set to 'dotnet-isolated' while your local project is set to 'None'.
You can pass --force to update your Azure app with 'None' as a 'FUNCTIONS_WORKER_RUNTIME'
```

**Workaround we found**
Local.settings.json usually isn't versionned, ruling out the option to keep it in the repo next to the function app code.

While creating a github action to deploy the dotnet isolated function app code after IaC completes, we had no other option than copy a local.settings.json.template file (with only the FUNCTION_WORKER_RUNTIME value set to "dotnet-isolated") to a fresh local.settings.json as per github action step so that the dotnet isolated runtime is recognized by the func azure function app publish at run time. 

Feel free to route me to an article in case i missed it.
Kind regards

mattchenderson:
I'm not aware of the command paramters being described for publish. That certainly doesn't show up when I do a `func azure functionapp publish -h` at least, which could be its own issue.

[...]

PR #3616, opened 2024-03-07
Updated error message to reflect unmentioned language options

jhueppauff:
Extending the error message to reflect the missing language options. Currently if you try to run or (publish) a function app you get the following error message if the local.settings.json isn't present.
image

But this does not mention the other options, like dotnet-isolated. Just by reading this message, it seems to suggest --csharp as the right option. https://github.com/Azure/azure-functions-core-tools/blob/fc93cca9640179f934cefb71f641f5e6243a1ac5/src/Azure.Functions.Cli/Helpers/GlobalCoreToolsSettings.cs#L40 but this will set the workerruntime to dotnet instead of dotnet isolated. The option --dotnet-isolated isn't mentioned here at all. 

Related to #3537 (if not resolving that one)

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/GlobalCoreToolsSettings.cs
+++ b/src/Azure.Functions.Cli/Helpers/GlobalCoreToolsSettings.cs
@@ -16,7 +16,7 @@ namespace Azure.Functions.Cli.Helpers
             {
                 if (_currentWorkerRuntime == WorkerRuntime.None)
                 {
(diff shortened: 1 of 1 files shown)


khkh-ms:
The change was included as part of a different PR.

jhueppauff:
Can you link the PR? I don't see it in the default branch yet? 
When will this be shipped?

jhueppauff:
I have seen this is now in the default branch. But --node seems still to be missing

PR #3625, opened 2024-03-15
Flex Deployment Changes

khkh-ms:

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -423,6 +423,7 @@ namespace Azure.Functions.Cli.Helpers
             site.Kind = armSite.kind;
             site.Sku = armSite.properties.sku;
             site.SiteName = armSite.name;
+            site.FunctionAppConfig = armSite.properties.functionAppConfig;
             return site;
         }
 
--- a/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
+++ b/src/Azure.Functions.Cli/Helpers/AzureHelper.cs
@@ -427,6 +427,37 @@ namespace Azure.Functions.Cli.Helpers
             return site;
         }
 
+        public static async Task<bool> UpdateFlexRuntime(Site site, string runtimeName, string runtimeValue, string accessToken, string managementURL)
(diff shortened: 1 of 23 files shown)


src/Azure.Functions.Cli/Helpers/AzureHelper.cs
```diff
@@ -423,9 +423,41 @@ public static async Task<Site> LoadSiteObjectAsync(Site site, string accessToken
...
+                return false;
+            }
+
+            runtimeConfig.name = runtimeName;
+            runtimeConfig.version = runtimeValue;
+
+            string functionAppString = JsonConvert.SerializeObject(functionApp);
+
+            url = new Uri($"{managementURL}{site.SiteId}?api-version={ArmUriTemplates.FlexApiVersion}");
+            await ArmHttpAsyncForFlex(new HttpMethod("PATCH"), url, accessToken, functionApp);
```
kamperiadis, line 448: AFAIK, we currently can't do a PATCH request to update attributes, we'll need to do a PUT.

khkh-ms: Thanks @kamperiadis!  Can you please share the recommended API. I got it from my local AZ CLI copy which was using Patch. Interestingly, it is actually working in beta :).

kamperiadis: I just confirmed with David today - we did not do any implementation changes for PATCH, which might be missing things like blocking unsupported configs in the Site object. I would not suggest using PATCH. Instead, we should be doing a GET call to the function app and then do a PUT request with that same function app object but with the updated values. 

Not sure what Az CLI copy you have but for flex specifically, the Az CLI makes PUT calls when updating the runtime.

kamperiadis (changes requested):

src/Azure.Functions.Cli/Helpers/AzureHelper.cs
khkh-ms: Changed it to us `PUT` instead of `PATCH`.

kamperiadis (approved):

PR #3730, opened 2024-06-27
Set the in-proc net8 executable name based on operating system

### LINKED ISSUE #3713: .NET8 func start fails when running on Linux host
The below check doesn't work, because it is searching for `func.exe`. However on a Linux host, this should be `func`, without the `.exe` part.

https://github.com/Azure/azure-functions-core-tools/blob/69dabe1d59b714159d7ec142cbeddc89978603e9/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs#L499

kshyju:
Updating the executable name for net8 in-proc based on operating system.

resolves #3713

### Pull request checklist


brettsam (approved):

fabiocav (approved):
Some other concerns with the resolution (how the working/base path is being resolved), but won't block since it's pre-existing. Will follow up directly.

PR #3749, opened 2024-07-09
Kill child process when Ctrl + break key is pressed. 

kshyju:
Kill child process when Ctrl + break key is pressed. 
Also made `IProcessManager` dependency singleton. This is a safe change because the current implementation does not store any state other than the child process count). We want a single instance (which was used to register the child process) so that we have the same child processes when we handle the Ctrl + Break key event handler, where we terminate the child processes.

### Pull request checklist

### DIFF
--- a/test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
+++ b/test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
@@ -93,7 +93,8 @@ namespace Azure.Functions.Cli.Tests.ActionsTests
                 .SingleInstance();
 
             builder.RegisterType<ProcessManager>()
-                .As<IProcessManager>();
+                .As<IProcessManager>()
+                .SingleInstance();
 
             var mockedSecretsManager = Substitute.For<ISecretsManager>();
             builder.RegisterInstance(mockedSecretsManager)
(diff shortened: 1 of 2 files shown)


test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
```diff
@@ -93,7 +93,8 @@ private IContainer InitializeContainerForTests()
                 .SingleInstance();
 
             builder.RegisterType<ProcessManager>()
-                .As<IProcessManager>();
+                .As<IProcessManager>()
+                .SingleInstance();
```
kshyju, line 97: This is a safe change because the [current implementation](https://github.com/Azure/azure-functions-core-tools/blob/6b9a1c051c83aaa33366a55463de6bb92631d24b/src/Azure.Functions.Cli/Common/ProcessManager.cs#L8) does not store any state other than the child process count). We want a single instance (which was used to register the child process) so that we have the same child processes when we handle the Ctrl + Break key event handler, where we terminate the child processes.

kshyju:
Ping @khkh-ms @fabiocav @soninaren

fabiocav (approved):

test/Azure.Functions.Cli.Tests/ActionsTests/ResolveActionTests.cs
```diff
@@ -93,7 +93,8 @@ private IContainer InitializeContainerForTests()
                 .SingleInstance();
 
             builder.RegisterType<ProcessManager>()
-                .As<IProcessManager>();
+                .As<IProcessManager>()
```
fabiocav, line 96: Reviewed usage and this is not currently used by anything outside of the new surface used for .NET 8.

fabiocav (approved):

PR #3769, opened 2024-07-15
Clean up Node.js docker files and update to Node 20

ejizba:

Noticed a few things while reviewing https://github.com/Azure/azure-functions-core-tools/pull/3371

The "InitAction.cs" code only uses one Node.js version at a time, so there's no point in having extra unused dockerfile templates. I deleted all but 2 dockerfile templates (1 for JS and 1 for TS) and updated them to point to Node 20, the latest supported version.

### Pull request checklist


mattchenderson (approved):
Changes themselves look fine, but I did want to ask if we wanted to add logic to have this use the passed-in version. You can see an example of this with .NET isolated a little bit above: https://github.com/Azure/azure-functions-core-tools/blob/9f498f1ec376e60ccc8604bbc81498344f0a12df/src/Azure.Functions.Cli/Actions/LocalActions/InitAction.cs#L422-L436

The concern would be that Node 18 is still supported, but we give you an image that's on Node 20. That may be fine, and it's no worse than the current state today. In fact, this PR makes things strictly better. I just wanted to flag this as a potential improvement. Perhaps we could create a separate issue to track it for a later time.

ejizba:
Yeah I'm not too worried about it. I think Node.js is a lot more forgiving in the sense that most apps works just fine across Node.js versions whereas I assume .NET apps won't even run at all on the wrong version (right?)

mattchenderson:
> I assume .NET apps won't even run at all on the wrong version (right?)

That depends. There's pretty high compatibility between .NET 6 and .NET 8 for example. For that reason, many apps on the in-proc model can switch the config of the app without redeploying code, for example. But.... there are still places where breaking changes occur across major versions. So, keeping things aligned is safer. But if that's not as much a concern for Node, then we're all good.

ejizba:
I still can't merge until someone from @Azure/Azure-functions-core approves 🙃

jviau (approved):

andystaples (approved):

PR #3773, opened 2024-07-21
Fix Issue 3594: Look for csproj and fsproj non-recursively

### LINKED ISSUE #3594: Can't determine Project to build. Expected 1 .csproj or .fsproj but found 3 when running "func start"
The `func start` command used to work flawlessly but it started showing the subject error message. Further info:

- .NET SDK 8.0.101
- `func --version` reveals version 4.0.5455
- There is only one .csproj file in the function app's folder, but the function app is a part of a C# solution
- I launched VS Code to create and initialize a new minimal azure function app. The `func start` works.

ryanski44:
add searchOption parameter to GetFiles to allow for non-recursive searching

use non-recursive searching when looking for .csproj and .fsproj files during validation.

resolves #3594

### Pull request checklist

  * I could not find any existing tests for this function

### DIFF
--- a/src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
+++ b/src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
@@ -170,18 +170,20 @@ namespace Azure.Functions.Cli.Common
                 }
             }
 
-            if (searchOption == SearchOption.AllDirectories)
+            if (searchOption == SearchOption.TopDirectoryOnly)
             {
-                foreach (var directory in Instance.Directory.GetDirectories(directoryPath, "*", SearchOption.TopDirectoryOnly))
+                yield break;
+            }
+                
+            foreach (var directory in Instance.Directory.GetDirectories(directoryPath, "*", SearchOption.TopDirectoryOnly))
+            {
+                var directoryName = Path.GetFileName(directory);
+                if (excludedDirectories == null ||
+                    !excludedDirectories.Any(d => d.Equals(directoryName, StringComparison.OrdinalIgnoreCase)))
                 {
-                    var directoryName = Path.GetFileName(directory);
-                    if (excludedDirectories == null ||
-                        !excludedDirectories.Any(d => d.Equals(directoryName, StringComparison.OrdinalIgnoreCase)))
+                    foreach (var file in GetFiles(directory, excludedDirectories, excludedFiles, searchPattern, searchOption))
                     {
-                        foreach (var file in GetFiles(directory, excludedDirectories, excludedFiles, searchPattern, searchOption))
-                        {
-                            yield return file;
-                        }
+                        yield return file;
                     }
                 }
             }
--- a/src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
+++ b/src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
@@ -158,7 +158,7 @@ namespace Azure.Functions.Cli.Common
             }
         }
 
-        internal static IEnumerable<string> GetFiles(string directoryPath, IEnumerable<string> excludedDirectories = null, IEnumerable<string> excludedFiles = null, string searchPattern = "*")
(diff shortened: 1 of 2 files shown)


ryanski44:
@microsoft-github-policy-service agree

jviau (commented):
Thanks for the PR!

This does address the issue, but it will also break the scenario of running `func` from a parent directory and it finding the csproj. We will need to discuss how we can fix this discovery issue while also avoiding breaking existing scenarios.

ryanski44:
> Thanks for the PR!
> 
> This does address the issue, but it will also break the scenario of running `func` from a parent directory and it finding the csproj. We will need to discuss how we can fix this discovery issue while also avoiding breaking existing scenarios.

I couldn't find anywhere that called CanDotnetBuild() that wasn't already expecting to be in the current working directory of a functions app. I must have missed something. Another fix would be putting "bin" and "obj" as excludedDirectories, but I'll let you all discuss options. I'm just glad this is getting some love.

jviau:
@ryanski44 we discussed this internally and we found the `CanDotnetBuild()` may not even be adding value. Would you mind submitting a new iteration which just removes the `CanDotnetBuild()` calls and method entirely? There is another location to remove the call here: https://github.com/Azure/azure-functions-core-tools/blob/bb61864dc7365002bc1b4df435c6201ea448b73f/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs#L215

ryanski44:
> @ryanski44 we discussed this internally and we found the `CanDotnetBuild()` may not even be adding value. Would you mind submitting a new iteration which just removes the `CanDotnetBuild()` calls and method entirely? There is another location to remove the call here:
> 
> https://github.com/Azure/azure-functions-core-tools/blob/bb61864dc7365002bc1b4df435c6201ea448b73f/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs#L215

I just took a look and I could, but there are a couple places where it seems like having no project files found will result in no errors and a skipped build action (like in KubernetesDeployAction.cs:215). If we remove the CanDotnetBuild check we have to assume to build in these cases, and that might result in an error. So, it seems like a change in behavior to me if it were to be removed completely.

jviau:
> > @ryanski44 we discussed this internally and we found the `CanDotnetBuild()` may not even be adding value. Would you mind submitting a new iteration which just removes the `CanDotnetBuild()` calls and method entirely? There is another location to remove the call here:
> > https://github.com/Azure/azure-functions-core-tools/blob/bb61864dc7365002bc1b4df435c6201ea448b73f/src/Azure.Functions.Cli/Actions/KubernetesActions/KubernetesDeployAction.cs#L215
> 
> I just took a look and I could, but there are a couple places where it seems like having no project files found will result in no errors and a skipped build action (like in KubernetesDeployAction.cs:215). If we remove the CanDotnetBuild check we have to assume to build in these cases, and that might result in an error. So, it seems like a change in behavior to me if it were to be removed completely.

Ah I see what you mean - the skipping of a build when no project is present is intentional. In that case your change looks good.

src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
```diff
@@ -170,15 +170,18 @@ internal static IEnumerable<string> GetFiles(string directoryPath, IEnumerable<s
                 }
             }
 
-            foreach (var directory in Instance.Directory.GetDirectories(directoryPath, "*", SearchOption.TopDirectoryOnly))
+            if (searchOption == SearchOption.AllDirectories)
```
kshyju, line 173: nit: Invert if to reduce nesting.

kshyju (approved):

src/Azure.Functions.Cli/Common/FileSystemHelpers.cs
ryanski44: Done

jviau (approved):

kshyju (approved):
Thanks for your contribution @ryanski44

Issue #3783, opened 2024-08-07
If restore fails, CI script should exit with an error


mattchenderson:
### Version

N/A

### Description

In [`check-vulnerabilities.ps1`](../blob/d76aa7b7dd3d11fa348a302e58ed647b5b4b5a2d/check-vulnerabilities.ps1), we should probably capture the exit code from the `dotnet restore` command and possibly throw an error before we get to the package listing operation. If restore fails partially, the package listing can git tripped on versions in dependencies not getting snapped forward properly.

### Steps to reproduce

This was encountered with early iterations of https://github.com/Azure/azure-functions-core-tools/pull/3780, as the package was not on the feed.

mattchenderson:
Per @jviau, the error might not be the behavior we want. We believe this will be addressed by shifting things toward a standard msbuild process, as everything will be handled implicitly through the build itself.

PR #3786, opened 2024-08-09
Update check-vulnerabilities.ps1 to error out if dotnet command fails and update build project dependencies

Francisco-Gamino:

* Fixes https://github.com/Azure/azure-functions-core-tools/issues/3783
* Updated `check-vulnerabilities.ps1` to validate the packages for the following projects are up-to-date: 
"$PSScriptRoot/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"       "$PSScriptRoot/test/Azure.Functions.Cli.Tests/Azure.Functions.Cli.Tests.csproj"
"$PSScriptRoot/build/Build.csproj"
* Updated project dependencies for `Build.csproj` 

### Pull request checklist

### DIFF
--- a/check-vulnerabilities.ps1
+++ b/check-vulnerabilities.ps1
@@ -1,33 +1,60 @@
-$projectPath = ".\src\Azure.Functions.Cli"
-$projectFileName = ".\Azure.Functions.Cli.csproj"
-$logFilePath = "..\..\build.log"
-if (-not (Test-Path $projectPath))
-{
-    throw "Project path '$projectPath' does not exist."
+param
+(
+    [String[]]
+    $CsprojFilePath
+)
+
+if (-not $CsprojFilePath) {
+    $CsprojFilePath = @(
+        "$PSScriptRoot/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
+        "$PSScriptRoot/test/Azure.Functions.Cli.Tests/Azure.Functions.Cli.Tests.csproj"
+        "$PSScriptRoot/build/Build.csproj"
+    )
 }
 
-cd $projectPath
+$logFilePath = "$PSScriptRoot/build.log"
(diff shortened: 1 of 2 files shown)


check-vulnerabilities.ps1
```diff
@@ -1,33 +1,60 @@
...
+    $CsprojFilePath = @(
+        "$PSScriptRoot/src/Azure.Functions.Cli/Azure.Functions.Cli.csproj"
+        "$PSScriptRoot/test/Azure.Functions.Cli.Tests/Azure.Functions.Cli.Tests.csproj"
+        "$PSScriptRoot/build/Build.csproj"
+    )
 }
 
-cd $projectPath
+$logFilePath = "$PSScriptRoot/build.log"
+foreach ($projectFilePath in $CsprojFilePath) {
```
mattchenderson, line 16: Thinking aloud, I wonder if we would want to parallelize these upstream of the call into the script, I suppose that's supported by the parameter being included here. But we should probably ensure that gets tracked, as it will be much more efficient.

mattchenderson (approved):
One small comment about next work that should be tracked, but that does not need to block this specific PR.

liliankasem (dismissed):

jviau (approved):
I am fine with these changes, but I do want to point out with .NET8 SDK this is built into the restore process (configurable via msbuild properties). Additionally, we have component governance running for this repo which will flag these CVEs.

I'm not a big fan of hard failures for CVEs in the pipeline. I prefer being given some time to address the CVEs before all our pipelines are blocked (thus impacting ongoing work).

mattchenderson:
Per @jviau, the core problem should be addressed by a new set of changes coming in for the build system. Rather than address current conflicts to get this one in, we think it will be best to just focus on the new changes.

PR #3790, opened 2024-08-12
Ensuring execute permissions are present for in-proc8/func executable in unix scenarios

### LINKED ISSUE #3766: Permission denied when running "func start" on .NET 8 in-process Function on Ubuntu 24.04
- .NET SDK: 8.0.107
- Azure Function Core Tools: 4.0.5907
- Ubuntu 24.04 (WSL but same issue was experienced using GitHub Actions)

When I run `func start` I get the below error:

```plaintext
Failed to start the in-process model host. An error occurred trying to start process '/usr/lib/azure-functions-core-tools-4/in-proc8/func' with working directory '/home/gabriel/MyProjFolder/bin/output'. Permission denied
```

Steps taken:

1. `func init MyProjFolder --worker-runtime dotnet`
2. `cd MyProjFolder/`
3. `func new --template "Http Trigger" --name MyHttpTrigger`

I added `FUNCTIONS_INPROC_NET8_ENABLED` to `local.settings.json`:

```json
{
    "IsEncrypted": false,
    "Values": {
        "FUNCTIONS_INPROC_NET8_ENABLED": "1"
    }
}
```

I updated the project to target .NET 8:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.4.0" />
  </ItemGroup>
  <ItemGroup>
    <None Update="host.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
    <None Update="local.settings.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
    </None>
  </ItemGroup>
</Project>
```

I'm able to run an in-process .NET 8 on Windows using the Azure Function Core Tools and on a Windows Function App.

kshyju:
resolves #3766

Ensuring execute permissions are present for in-proc8/func executable in unix scenarios.

#### Validation
I validated the changes locally in WSL/Ubuntu 24 environment. By default my terminal user had executable permission, so I had to delete executable permission first using `chmod a-x in-proc8/func` command.

`myfunc` is the directory where `Azure.Functions.Cli.csproj` publish (`dotnet publish`) output is present.

image

`net8app/app` directory is where net8 in-proc function app is present.

image

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Diagnostics;
@@ -28,6 +27,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Mono.Unix;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Azure.Functions.Cli.Common.OutputTheme;
@@ -511,6 +511,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
             var inProc8FuncExecutablePath = GetInProcNet8ExecutablePath();
 
             EnsureNet8FuncExecutablePresent(inProc8FuncExecutablePath);
+            EnsureUnixExecutePermissions(inProc8FuncExecutablePath);
 
             var inprocNet8ChildProcessInfo = new ProcessStartInfo
             {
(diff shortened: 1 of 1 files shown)


cjaliaga (approved):
Do we know if this is needed for any type of installation? For example If you manually download and unzip a release, the binaries won't maintain permissions, so you would need to manually `chmod +x`, such with `func` binary. 

However, if you install using a package manager, such `apt`, `func` should come with execute permissions, and we should ensure our child executables come already like that.

My only concern with this solution is what would happen if the user running `func start` doesn't have permissions to change file properties, so it cannot change the properties of this other file?

kshyju:
> My only concern with this solution is what would happen if the user running func start doesn't have permissions to change file properties, so it cannot change the properties of this other file?

`UnauthorizedAccessException` will be thrown in that case. I am not sure what we can do in that case ( if the user running func start doesn't have permissions to change file properties). Any thoughts @fabiocav @brettsam ?

aishwaryabh (approved):
LGTM!

kshyju:
> > My only concern with this solution is what would happen if the user running func start doesn't have permissions to change file properties, so it cannot change the properties of this other file?
> 
> `UnauthorizedAccessException` will be thrown in that case. I am not sure what we can do in that case ( if the user running func start doesn't have permissions to change file properties). Any thoughts @fabiocav @brettsam ?

@brettsam pointed me to this snippet code which gets executed during installation. https://github.com/Azure/azure-functions-core-tools/blob/ae9eff234b96a6120d9106a739252aa200476c31/src/Azure.Functions.Cli/npm/lib/install.js#L98-L101

kshyju:
Going to abandon this and will create a new PR with changes to install.js

PR #3801, opened 2024-08-23
Improve speed of func version command

ejizba:

This has been a pet peeve of mine for a while: the `func --version` command often takes several seconds on my machine. It can be random, but for example it took 18 seconds when I tested it just now. In particular, it annoys me because this slows down the VS Code UI which will check the version of core tools installed for various things.

With my changes, I'm consistently seeing `func --version` take <1 second.

### Pull request checklist


mattchenderson (approved):
Noting the side effect of dropping the telemetry event. I'm ok with that here. This also bypasses the first time experience, which again, I'm good with. This seems like a great set of changes.

src/Azure.Functions.Cli/Program.cs
```diff
@@ -4,26 +4,39 @@
...
+                _container = InitializeAutofacContainer();
+                AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
+
+                Console.CancelKeyPress += (s, e) =>
+                {
+                    _container.Resolve<IProcessManager>()?.KillChildProcesses();
+                };
+
+                ConsoleApp.Run<Program>(args, _container);
+            }
```
liliankasem, line 39: nit: return and leave the rest as it is outside of an if-else brace

```suggestion
                Environment.Exit(ExitCodes.Success);
                return;
            }

            FirstTimeCliExperience();
            SetupGlobalExceptionHandler();
            SetCoreToolsEnvironmentVariables(args);
            _container = InitializeAutofacContainer();
            AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;

            Console.CancelKeyPress += (s, e) =>
            {
                _container.Resolve<IProcessManager>()?.KillChildProcesses();
            };

            ConsoleApp.Run<Program>(args, _container);
```

liliankasem (dismissed):
I think there are plans to revamp this to work a lot more like the dotnet cli, but this should be okay for now!

liliankasem (dismissed):

liliankasem (approved):

PR #3802, opened 2024-08-28
Core Tools OOP Host

### LINKED ISSUE #3744: Core Tools updates to use out-of-process host by default
The Core Tools should switch to use the out-of-process host for out-of-process stacks. Today, it still uses the .NET 6 in-process host, unless it knows it needs to use the .NET 8 in-process host.

This item requires additional design. A few quick notes:

- This is a variant of what was done to support .NET 8 in-process. 
  - The out-of-process host should be used unless we have a signal which tells us to instead use the .NET 6 in-process host.
  - We should allow the start operation to be forced to launch the right version via a parameter.
- Special consideration needs to be given for Visual Studio

We should expect this to be at least a minor version bump.

aishwaryabh:

resolves #3744 

This PR creates the feature branch `feature/oop-host`, which has the OOP host build. This PR also includes checks for the user passing in the `--runtime` param specifying the host and infers the host if that is not specified. I've also added tests which checks that we try to execute the correct func.exe (even though those tests technically fail since there is no `inproc-6` or `inproc-8` folder). We also only run on .NET 8 TFM and remove multi-targeting in the feature branch.

For when the user specifies the `--runtime` parameter, these are the following scenarios we support today where the first row is the options the user may put in and the first column are the potential apps a user may have:

| | inproc6|inproc8|default|
|--|:-------:|:-------:|:-------:|
|.NET Isolated app|| |x|
|.NET 6 in-proc app| x| ||
|.NET 8 in-proc app| | x||
|non .NET app| ||x|

Anything that does not have an X within it throws an error to the user with a helpful message on why the passed in host runtime is not valid.

</body>
</html>

For inference the path will be:
| |Sequence|Tool version|
|--|:-------:|:-------:|
|.NET Isolated app| default| default |
|.NET 6 in-proc app| default -> inproc6|inproc6 |
|.NET 8 in-proc app| default -> inproc8|inproc8 |
|non .NET app|default |default|

This is how the logic looks like today for the inference scenario:

- If app is in-proc app
    - If `FUNCTIONS_INPROC_NET8_ENABLED` is set
          - Run inproc8 host
     - Else
           - Run inproc6 host
- else
    - Run OOP host

### Pull request checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ trigger:
     include:
     - v4.x
     - release_4.0
- 
+
 jobs:
 - job: Default
   condition: eq(variables['LinuxPackageBuildTag'], '')
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,15 +5,12 @@ pr:
     include:    
     - v4.x
     - release_4.0
-    - feature/*
 
 trigger:
   branches:
     include:
     - v4.x
     - release_4.0
-    - feature/*
-
 jobs:
 - job: Default
   condition: eq(variables['LinuxPackageBuildTag'], '')
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,6 +5,7 @@ pr:
     include:    
     - v4.x
     - release_4.0
+    - feature/*
 
 trigger:
   branches:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,7 @@ trigger:
     include:
     - v4.x
     - release_4.0
+    
 jobs:
 - job: Default
   condition: eq(variables['LinuxPackageBuildTag'], '')
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,6 +12,7 @@ trigger:
     include:
     - v4.x
     - release_4.0
+    - feature/*
 
 jobs:
 - job: Default
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ trigger:
     include:
     - v4.x
     - release_4.0
-    
+ 
 jobs:
 - job: Default
   condition: eq(variables['LinuxPackageBuildTag'], '')
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,14 +5,14 @@ pr:
     include:    
     - v4.x
     - release_4.0
-    - feature/*
+    - 'feature/*'
 
 trigger:
   branches:
     include:
     - v4.x
     - release_4.0
-    - feature/*
+    - 'feature/*'
 
 jobs:
 - job: Default
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,14 +5,14 @@ pr:
     include:    
     - v4.x
     - release_4.0
-    - 'feature/*'
+    - feature/*
 
 trigger:
   branches:
     include:
     - v4.x
     - release_4.0
-    - 'feature/*'
+    - feature/*
 
 jobs:
 - job: Default
--- a/build.ps1
+++ b/build.ps1
@@ -23,11 +23,11 @@ if ($env:IntegrationBuildNumber)
         throw $errorMessage
     }
 
-    $buildCommand = { dotnet run --integrationTests }
+    $buildCommand = { dotnet run --integrationTests --skipArtifactGeneration}
 }
 else
 {
-    $buildCommand = { dotnet run --ci }
+    $buildCommand = { dotnet run --ci --skipArtifactGeneration}
 }
 
 Write-Host "Running $buildCommand"
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -116,7 +116,7 @@ namespace Build
                 var outputPath = Path.Combine(Settings.OutputDir, runtime);
                 var rid = GetRuntimeId(runtime);
 
-                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: isMinVersion);
                 if (isMinVersion)
                 {
                     RemoveLanguageWorkers(outputPath);
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -116,8 +116,9 @@ namespace Build
                 var outputPath = Path.Combine(Settings.OutputDir, runtime);
                 var rid = GetRuntimeId(runtime);
 
-                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: isMinVersion);
-                RemoveLanguageWorkers(outputPath);
+                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
+                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                RemoveLanguageWorkers(outputPathNet8);
 
             }
 
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -117,7 +117,11 @@ namespace Build
                 var rid = GetRuntimeId(runtime);
 
                 ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
-                RemoveLanguageWorkers(outputPath);
+
+                // Publish net8 version of the artifact as well for VS.
+                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
+                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                RemoveLanguageWorkers(outputPathNet8);
 
             }
 
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -444,10 +444,27 @@ namespace Build
                     : Enumerable.Empty<string>();
                 var toSignThirdPartyPaths = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForInProc8);
 
+                // Add out of proc directory as well
+                var outOfProcDirectory = Path.Combine(targetDir, "out-of-proc");
+                var outOfProcDirectoryExists = Directory.Exists(outOfProcDirectory);
+
+                var toSignPathsForOutOfProc = outOfProcDirectoryExists
+                    ? Settings.SignInfo.authentiCodeBinaries.Select(el => Path.Combine(outOfProcDirectory, el))
+                    : Enumerable.Empty<string>();
+                var toSignPathsOutOfProc = Settings.SignInfo.authentiCodeBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignPathsForOutOfProc);
+
+                var toSignThirdPartyPathsForOutOfProc = outOfProcDirectoryExists
+                    ? Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(outOfProcDirectory, el))
+                    : Enumerable.Empty<string>();
+                var toSignThirdPartyPathsOutOfProc = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForOutOfProc);
+
                 var unSignedFiles = FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignPaths))
                                     .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))).ToList();
 
                 unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPaths))
+                                        .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
+
+                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
                                         .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
 
                 unSignedFiles.ForEach(filePath => File.Delete(filePath));
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -17,7 +17,6 @@ namespace Build
     public static class BuildSteps
     {
         private const string Net8ArtifactNameSuffix = "_net8";
-        private const string OutOfProcDirectoryName = "out-of-proc";
         private static readonly string _wwwroot = Environment.ExpandEnvironmentVariables(@"%HOME%\site\wwwroot");
         private static IntegrationTestBuildManifest _integrationManifest;
 
@@ -103,7 +102,7 @@ namespace Build
             Shell.Run("dotnet", $"pack {Settings.SrcProjectPath} " +
                                 $"/p:BuildNumber=\"{Settings.BuildNumber}\" " +
                                 $"/p:NoWorkers=\"true\" " +
-                                $"/p:TargetFramework=net6.0 " +  // without TargetFramework, the generated nuspec has incorrect path for the copy files operation.
+                                $"/p:TargetFramework=net8.0 " +  // without TargetFramework, the generated nuspec has incorrect path for the copy files operation.
                                 $"/p:CommitHash=\"{Settings.CommitId}\" " +
                                 (string.IsNullOrEmpty(Settings.IntegrationBuildNumber) ? string.Empty : $"/p:IntegrationBuildNumber=\"{Settings.IntegrationBuildNumber}\" ") +
                                 $"-o {outputPath} -c Release --no-build");
@@ -117,17 +116,9 @@ namespace Build
                 var outputPath = Path.Combine(Settings.OutputDir, runtime);
                 var rid = GetRuntimeId(runtime);
 
-                ExecuteDotnetPublish(outputPath, rid, "net6.0", skipLaunchingNet8ChildProcess: isMinVersion);
+                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: isMinVersion);
+                RemoveLanguageWorkers(outputPath);
 
-                if (isMinVersion)
-                {
-                    RemoveLanguageWorkers(outputPath);
-                }
-
-                // Publish net8 version of the artifact as well.
-                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
-                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
-                RemoveLanguageWorkers(outputPathNet8);
             }
 
             if (!string.IsNullOrEmpty(Settings.IntegrationBuildNumber) && (_integrationManifest != null))
@@ -343,7 +334,7 @@ namespace Build
 
             Environment.SetEnvironmentVariable("DURABLE_FUNCTION_PATH", Settings.DurableFolder);
 
-            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net6.0 --logger trx");
+            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net8.0 --logger trx");
         }
 
         public static void CopyBinariesToSign()
@@ -445,30 +436,10 @@ namespace Build
                     : Enumerable.Empty<string>();
                 var toSignThirdPartyPaths = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForInProc8);
 
-                // Add out of proc directory as well
-                var outOfProcDirectory = Path.Combine(targetDir, OutOfProcDirectoryName);
-                var outOfProcDirectoryExists = Directory.Exists(outOfProcDirectory);
-
-                var toSignPathsForOutOfProc = outOfProcDirectoryExists
-                    ? Settings.SignInfo.authentiCodeBinaries.Select(el => Path.Combine(outOfProcDirectory, el))
-                    : Enumerable.Empty<string>();
-                var toSignPathsOutOfProc = Settings.SignInfo.authentiCodeBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignPathsForOutOfProc);
-
-                var toSignThirdPartyPathsForOutOfProc = outOfProcDirectoryExists
-                    ? Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(outOfProcDirectory, el))
-                    : Enumerable.Empty<string>();
-                var toSignThirdPartyPathsOutOfProc = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForOutOfProc);
-
                 var unSignedFiles = FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignPaths))
                                     .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))).ToList();
 
                 unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPaths))
-                                        .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
-
-                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignPathsForOutOfProc))
-                                        .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
-
-                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
                                         .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
 
                 unSignedFiles.ForEach(filePath => File.Delete(filePath));
@@ -664,10 +635,10 @@ namespace Build
             Shell.Run("dotnet", $"publish {Settings.ProjectFile} " +
                                 $"/p:BuildNumber=\"{Settings.BuildNumber}\" " +
                                 $"/p:NoWorkers=\"true\" " +
-                                $"/p:TargetFramework=net6.0 " +
+                                $"/p:TargetFramework=net8.0 " +
                                 $"/p:CommitHash=\"{Settings.CommitId}\" " +
                                 (string.IsNullOrEmpty(Settings.IntegrationBuildNumber) ? string.Empty : $"/p:IntegrationBuildNumber=\"{Settings.IntegrationBuildNumber}\" ") +
-                                $"-c Release -f net6.0");
+                                $"-c Release -f net8.0");
         }
 
         public static void GenerateSBOMManifestForNupkg()
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -342,7 +342,7 @@ namespace Build
 
             Environment.SetEnvironmentVariable("DURABLE_FUNCTION_PATH", Settings.DurableFolder);
 
-            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net6.0 --logger trx");
+            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net6.0 --logger trx --filter \"(FullyQualifiedName~start_nodejs_with_inspect | FullyQualifiedName~start_dotnet_isolated_csharp_net9 | FullyQualifiedName~start_dotnet_isolated_csharp_with_oop_host_with_runtime_specified | FullyQualifiedName~start_dotnet6_inproc_without_specifying_runtime)\"");
         }
 
         public static void CopyBinariesToSign()
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -122,6 +122,10 @@ namespace Build
                     RemoveLanguageWorkers(outputPath);
                 }
 
+                // Publish net8 version of the artifact as well.
+                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
+                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                RemoveLanguageWorkers(outputPathNet8);
             }
 
             if (!string.IsNullOrEmpty(Settings.IntegrationBuildNumber) && (_integrationManifest != null))
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -116,9 +116,8 @@ namespace Build
                 var outputPath = Path.Combine(Settings.OutputDir, runtime);
                 var rid = GetRuntimeId(runtime);
 
-                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
-                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
-                RemoveLanguageWorkers(outputPathNet8);
+                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                RemoveLanguageWorkers(outputPath);
 
             }
 
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -464,7 +464,7 @@ namespace Build
                 unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPaths))
                                         .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
 
-                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
+                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignPathsForOutOfProc))
                                         .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
 
                 unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -467,6 +467,9 @@ namespace Build
                 unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
                                         .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
 
+                unSignedFiles.AddRange(FileHelpers.GetAllFilesFromFilesAndDirs(FileHelpers.ExpandFileWildCardEntries(toSignThirdPartyPathsOutOfProc))
+                                        .Where(file => !filterExtensionsSignSet.Any(ext => file.EndsWith(ext))));
+
                 unSignedFiles.ForEach(filePath => File.Delete(filePath));
 
                 var unSignedPackages = GetUnsignedBinaries(targetDir);
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -122,11 +122,6 @@ namespace Build
                     RemoveLanguageWorkers(outputPath);
                 }
 
-                // Publish net8 version of the artifact as well for VS.
-                var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
-                ExecuteDotnetPublish(outputPathNet8, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
-                RemoveLanguageWorkers(outputPathNet8);
-
             }
 
             if (!string.IsNullOrEmpty(Settings.IntegrationBuildNumber) && (_integrationManifest != null))
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -17,6 +17,7 @@ namespace Build
     public static class BuildSteps
     {
         private const string Net8ArtifactNameSuffix = "_net8";
+        private const string OutOfProcDirectoryName = "out-of-proc";
         private static readonly string _wwwroot = Environment.ExpandEnvironmentVariables(@"%HOME%\site\wwwroot");
         private static IntegrationTestBuildManifest _integrationManifest;
 
@@ -342,7 +343,7 @@ namespace Build
 
             Environment.SetEnvironmentVariable("DURABLE_FUNCTION_PATH", Settings.DurableFolder);
 
-            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net6.0 --logger trx --filter \"(FullyQualifiedName~start_nodejs_with_inspect | FullyQualifiedName~start_dotnet_isolated_csharp_net9 | FullyQualifiedName~start_dotnet_isolated_csharp_with_oop_host_with_runtime_specified | FullyQualifiedName~start_dotnet6_inproc_without_specifying_runtime)\"");
+            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net6.0 --logger trx");
         }
 
         public static void CopyBinariesToSign()
@@ -445,7 +446,7 @@ namespace Build
                 var toSignThirdPartyPaths = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForInProc8);
 
                 // Add out of proc directory as well
-                var outOfProcDirectory = Path.Combine(targetDir, "out-of-proc");
+                var outOfProcDirectory = Path.Combine(targetDir, OutOfProcDirectoryName);
                 var outOfProcDirectoryExists = Directory.Exists(outOfProcDirectory);
 
                 var toSignPathsForOutOfProc = outOfProcDirectoryExists
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -117,6 +117,10 @@ namespace Build
                 var rid = GetRuntimeId(runtime);
 
                 ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: true);
+                if (isMinVersion)
+                {
+                    RemoveLanguageWorkers(outputPath);
+                }
 
                 // Publish net8 version of the artifact as well for VS.
                 var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -40,7 +40,15 @@ namespace Build
         public static readonly string DurableFolder = Path.Combine(TestProjectPath, "Resources", "DurableTestFolder");
 
         public static readonly string[] TargetRuntimes = new[] {
-            "min.win-arm64" };
+             "min.win-arm64",
+            "min.win-x86",
+            "min.win-x64",
+            "linux-x64",
+            "osx-x64",
+            "osx-arm64",
+            "win-x86",
+            "win-x64",
+            "win-arm64" };
 
         public static readonly Dictionary<string, string> RuntimesToOS = new Dictionary<string, string>
         {
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -40,7 +40,7 @@ namespace Build
         public static readonly string DurableFolder = Path.Combine(TestProjectPath, "Resources", "DurableTestFolder");
 
         public static readonly string[] TargetRuntimes = new[] {
-             "min.win-arm64",
+            "min.win-arm64",
             "min.win-x86",
             "min.win-x64",
             "linux-x64",
@@ -341,7 +341,7 @@ namespace Build
                 "Microsoft.OData.Edm.dll",
                 "Microsoft.Spatial.dll",
                 "Mono.Posix.NETStandard.dll",
-                "OpenTelemetry.*dll",
+                //"OpenTelemetry.*dll",
                 Path.Combine("tools", "python", "packapp", "distlib")
             };
         }
--- a/build/Settings.cs
+++ b/build/Settings.cs
@@ -40,15 +40,7 @@ namespace Build
         public static readonly string DurableFolder = Path.Combine(TestProjectPath, "Resources", "DurableTestFolder");
 
         public static readonly string[] TargetRuntimes = new[] {
-            "min.win-arm64",
-            "min.win-x86",
-            "min.win-x64",
-            "linux-x64",
-            "osx-x64",
-            "osx-arm64",
-            "win-x86",
-            "win-x64",
-            "win-arm64" };
+            "min.win-arm64" };
 
         public static readonly Dictionary<string, string> RuntimesToOS = new Dictionary<string, string>
         {
@@ -341,6 +333,7 @@ namespace Build
                 "Microsoft.OData.Edm.dll",
                 "Microsoft.Spatial.dll",
                 "Mono.Posix.NETStandard.dll",
+                "OpenTelemetry.*dll",
                 Path.Combine("tools", "python", "packapp", "distlib")
             };
         }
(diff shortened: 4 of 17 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -381,6 +392,73 @@ private async Task<bool> IsInProcNet8Enabled()
             return isInProcNet8Enabled;
         }
 
+        /// <summary>
+        /// Check local.settings.json to determine what value of FUNCTIONS_WORKER_RUNTIME is.
+        /// </summary>
+        private async Task<string> GetFunctionsWorkerRuntime()
```
kshyju, line 398: Don't we have an existing helper method to do this ? `DetermineTargetFramework` ?

aishwaryabh: Replaced with existing helper method `DetermineTargetFramework`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +429,104 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
+            {
+                if (SetHostRuntime == "default")
+                {
+                    if (isCurrentProcessNet6Build)
+                    {
+                        if (isVerbose)
+                        {
+                            ColoredConsole.WriteLine(VerboseColor("Selected out-of-process host"));
+                        }
+                        await StartHostAsChildProcessAsync(true);
```
kshyju, line 445: Consider adding the argument name when passing boolean values directly, to improve readability.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +429,104 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
 
             var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            if (SetHostRuntime != null)
```
kshyju, line 435: This method has become too large, affecting its readability. Could we refactor it by extracting parts into smaller, more manageable methods?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +429,104 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
+                    {
+                        if (isVerbose)
+                        {
+                            ColoredConsole.WriteLine(VerboseColor("Selected out-of-process host"));
+                        }
+                        await StartHostAsChildProcessAsync(true);
+                        return;
+                    }
+                }
+                else if (SetHostRuntime == "inproc8")
```
kshyju, line 449: consider using some constants for these string values.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -176,6 +182,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                .WithDescription("If provided, a path to the file that will be used to write the output when using --enable-json-output.")
                .Callback(jsonOutputFile => JsonOutputFile = jsonOutputFile);
 
+            Parser
+               .Setup<string>("runtime")
+               .WithDescription("If provided, determines which version of the host to start.")
```
kshyju, line 187: Should the description also provide allowed values for this option?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +429,104 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
 
             var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            if (SetHostRuntime != null)
+            {
+                if (SetHostRuntime == "default")
```
kshyju, line 437: We should make this case insensitive check.

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -40,6 +40,9 @@
   <PropertyGroup Condition="'$(SkipInProcessHost)' == 'True'">
     <DefineConstants>SkipInProcessHost</DefineConstants>
   </PropertyGroup>
+	<PropertyGroup>
```
kshyju, line 43: Indentation is off.

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -274,46 +277,65 @@
...
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage.Internal" Version="1.4.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Script.WebHost" Version="4.35.4" />
     <PackageReference Include="Microsoft.Build" Version="17.0.0" />
     <PackageReference Include="Microsoft.Identity.Client" Version="4.61.3" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="NuGet.Packaging" Version="5.11.6" />
     <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
     <PackageReference Include="WindowsAzure.Storage" Version="9.3.1" />
-    <PackageReference Include="YamlDotNet" Version="6.0.0" />
+	<PackageReference Include="YamlDotNet" Version="6.0.0" />
```
kshyju, line 293: Indentation.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: added!

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: done

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: added

build/BuildSteps.cs
```diff
@@ -444,12 +444,32 @@ public static void TestPreSignedArtifacts()
                     : Enumerable.Empty<string>();
                 var toSignThirdPartyPaths = Settings.SignInfo.thirdPartyBinaries.Select(el => Path.Combine(targetDir, el)).Concat(toSignThirdPartyPathsForInProc8);
 
+                // Add out of proc directory as well
+                var outOfProcDirectory = Path.Combine(targetDir, "out-of-proc");
```
liliankasem, line 448: Can you use the "OutOfProcDirectoryName" const here instead of hardcoding

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -176,6 +184,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                .WithDescription("If provided, a path to the file that will be used to write the output when using --enable-json-output.")
                .Callback(jsonOutputFile => JsonOutputFile = jsonOutputFile);
 
+            Parser
+               .Setup<string>("runtime")
+               .WithDescription("If provided, determines which version of the host to start. Allowed values are inproc6, inproc8, and default.")
```
liliankasem, line 189: nit: use consts for allowed values (noticed InProc8HostRuntime and 6 defined above)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +505,103 @@ public override async Task RunAsync()
             await runTask;
         }
 
+        private async Task<bool> HandleRuntimeParam(bool isCurrentProcessNet6Build, bool isVerbose)
```
liliankasem, line 508: Given this returns a bool, "handle" isn't the right keyword to use here. Ask yourself, what is this method saying true/false to? And choose a name that best describes the purpose and behavior of the method.

Does perhaps "IsRuntimeParamValid" make more sense? Or "ShouldProcessRuntimeParam"?

If the answer isn't clear, is this method perhaps doing too much? Could this be further broken down into different methods that make more sense?

This is also an async method and should end in "Async"

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +505,103 @@ public override async Task RunAsync()
...
+                    {
+                        ColoredConsole.WriteLine(VerboseColor($"Selected {InProc8HostRuntime} host"));
+                    }
+                    await StartHostAsChildProcessAsync(isOutOfProc: false);
+                    return true;
+                }
+            }
+            else if (string.Equals(SetHostRuntime, InProc6HostRuntime, StringComparison.OrdinalIgnoreCase))
+            {
+                if (isVerbose)
```
liliankasem, line 536: Seeing this pattern used a lot, could this be it's own function? "LogVerboseMessage"

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +505,103 @@ public override async Task RunAsync()
...
+                    ColoredConsole.WriteLine(VerboseColor($"Selected {InProc6HostRuntime} host"));
+                }
+                if (!isCurrentProcessNet6Build)
+                {
+                    throw new CliException($"Cannot set host runtime to '{SetHostRuntime}' for the current process. The current process is not a .NET 6 build.");
+                }
+            }
+            else
+            {
+                throw new CliException($"Invalid host runtime '{SetHostRuntime}'. Valid values are 'default', 'in-proc8', 'in-proc6'.");
```
liliankasem, line 547: Same nit here about using const for known values such as in-proc8 etc.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +505,103 @@ public override async Task RunAsync()
...
+            }
+            return false;
+        }
+
+        private async Task<bool> InferHostToLaunchOnStartup(bool isCurrentProcessNet6Build, bool isVerbose)
```
liliankasem, line 552: Similar comment about this being a bool method and async. Thinking of naming guidelines, what could make more sense for this method name?

- Be Specific About the Boolean Result: The name should help the caller understand what the boolean result represents. For instance, if the method checks if an operation is successful or if a condition is met, reflect that in the name.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,27 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            // If --runtime param is set, handle runtime param logic; otherwise we infer the host to launch on startup
+            if (SetHostRuntime != null)
             {
-                await StartInProc8AsChildProcessAsync();
-                return;
+                var shouldReturn = await ShouldReturnAfterHandlingRuntimeAsync(isCurrentProcessNet6Build, isVerbose);
+                if (shouldReturn)
+                {
+                    return;
```
liliankasem, line 443: What happens if we return here? What is the behaviour?

aishwaryabh: updated naming to make it more accurate!

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,33 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
             var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            // If --runtime param is set, handle runtime param logic; otherwise we infer the host to launch on startup
+            if (SetHostRuntime != null)
             {
-                await StartInProc8AsChildProcessAsync();
-                return;
+                // Check if we should start child process from user specified host runtime and return
+                var shouldStartChildProcess = await ShouldStartChildProcessFromHostRuntime(isCurrentProcessNet6Build, isVerbose);
+                var isOutOfProcSpecified = IsOutOfProcSpecifiedFromHostRuntime();
```
liliankasem, line 442: This isn't used outside of the if statement so should be defined right before calling await StartHostAsChildProcessAsync

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,33 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
+                if (shouldStartChildProcess)
+                {
+                    await StartHostAsChildProcessAsync(isOutOfProcSpecified);
+                    return;
+                }
+            }
+            else
+            {
+                // Infer host runtime and check if we should launch child process
+                var shouldLaunchOutOfProc = ShouldLaunchOutOfProcFromWorkerRuntime();
```
liliankasem, line 452: Same here, this is only used in the if statement so should be created there

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +511,111 @@ public override async Task RunAsync()
             await runTask;
         }
 
+        private bool IsOutOfProcSpecifiedFromHostRuntime()
```
liliankasem, line 514: How often do we have to evaluate this?

liliankasem (approved):

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: removed this

build/Settings.cs
```diff
@@ -341,6 +341,7 @@ public class SignInfo
                 "Microsoft.OData.Edm.dll",
                 "Microsoft.Spatial.dll",
                 "Mono.Posix.NETStandard.dll",
+                "OpenTelemetry.*dll",
```
kshyju, line 344: Why do we need to add assembly?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +514,128 @@ public override async Task RunAsync()
...
+        private async Task<string> GetTargetFrameworkAsync()
+        {
+            var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
+
+            string targetFramework = "";
+
+            string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
+            if (projectFilePath != null)
+            {
+                targetFramework = await DotnetHelpers.DetermineTargetFramework(Path.GetDirectoryName(projectFilePath));
```
kshyju, line 578: Can't we pass the `functionAppRoot` here ? Otherwise we are executing the logic to find profile file twice (once here and once inside the method)

build/Settings.cs
aishwaryabh: the TestPresignedArtifacts test fails since the OpenTelemetry files are not signed (I'm assuming they are added with the new reference to the OOP host)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: I tried doing that but it fails to find the TFM so I do believe this is needed

build/BuildSteps.cs
```diff
@@ -116,17 +116,17 @@ public static void DotnetPublishForZips()
...
-
+                ExecuteDotnetPublish(outputPath, rid, "net8.0", skipLaunchingNet8ChildProcess: isMinVersion);
                 if (isMinVersion)
                 {
                     RemoveLanguageWorkers(outputPath);
                 }
 
-                // Publish net8 version of the artifact as well.
+                // Publish net8 version of the artifact as well for VS.
                 var outputPathNet8 = BuildNet8ArtifactFullPath(runtime);
```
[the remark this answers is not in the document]
aishwaryabh: Let me try out the tests without this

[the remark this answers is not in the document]
aishwaryabh: Tests do not work without this so we need to keep this for VS

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -27,14 +27,11 @@
     <NuspecFile>Azure.Functions.Cli.nuspec</NuspecFile>
     <NuspecProperties>configuration=$(Configuration);targetFramework=$(TargetFramework);version=$(Version)</NuspecProperties>
   </PropertyGroup>
-  <!-- Pack only the net6.0 version (which will include the bits needed for net8.0 usecase) -->
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <IsPackable>true</IsPackable>
-    <PackAsTool>true</PackAsTool>
-    <ToolCommandName>func</ToolCommandName>
-  </PropertyGroup>
+  <!-- Pack only the net8.0 version -->
```
fabiocav, line 30: This comment is no longer applicable (no longer multi-targeting)

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -27,14 +27,11 @@
...
   <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
fabiocav, line 31: You can now remove this condition.

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
```diff
@@ -214,7 +214,7 @@ await CliTester.Run(new RunConfiguration
             }, _output);
         }
 
-        [Fact]
+        [Fact(Skip="Flaky test")]
```
fabiocav, line 217: Can we please create an issue to track all skipped tests in this PR?

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -309,7 +309,7 @@ await CliTester.Run(new RunConfiguration
                         result.Should().Be("Hello, Test. This HTTP triggered function executed successfully.", because: "response from default function should be 'Hello, {name}. This HTTP triggered function executed successfully.'");
                     }
                 },
-                CommandTimeout = TimeSpan.FromSeconds(300),
+                CommandTimeout = TimeSpan.FromSeconds(900),
```
fabiocav, line 312: Please revert as the test is being skipped.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -184,14 +184,14 @@ await CliTester.Run(new RunConfiguration[]
                     Test = async (_, p) =>
                     {
                         // give the host time to load functions and print any errors
-                        await Task.Delay(TimeSpan.FromSeconds(10));
+                        await Task.Delay(TimeSpan.FromSeconds(60));
```
fabiocav, line 187: Revert the change as test is being skipped.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -338,42 +338,172 @@ await CliTester.Run(new RunConfiguration
...
+                        result.Should().Be("Welcome to Azure Functions!", because: "response from default function should be 'Welcome to Azure Functions!'");
+
+                        if (_output is Xunit.Sdk.TestOutputHelper testOutputHelper)
+                        {
+                            testOutputHelper.Output.Should().Contain("4.10");
+                            testOutputHelper.Output.Should().Contain("Selected out-of-process host");
+                        }
+                    }
+                },
+                CommandTimeout = TimeSpan.FromSeconds(900),
```
fabiocav, line 506: These are some high timeouts on all the tests. Is this required? If possible, let's avoid this, but this is certainly something we want to track and fix.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -32,13 +32,13 @@ await CliTester.Run(new RunConfiguration
                 {
                     "init . --worker-runtime node",
                     "new --template \"Http trigger\" --name HttpTrigger",
-                    "start"
+                    "start --port 7077"
```
fabiocav, line 35: Can you add some information describing why the port is now being passed in?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,36 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
 
-            var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            var isCurrentProcessNet8Build = RuntimeInformation.FrameworkDescription.Contains(Net8FrameworkDescriptionPrefix);
```
fabiocav, line 436: This check shouldn't be required anymore.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +514,128 @@ public override async Task RunAsync()
             await runTask;
         }
 
+        private async Task<bool> ShouldStartChildProcessFromHostRuntime(bool isCurrentProcessNet8Build, bool isVerbose)
```
fabiocav, line 517: Can we please have a unit test for this logic covering all the expected scenarios? I think that will help highlight some opportunities for simplification and some gaps.

Removing the, no longer needed, `isCurrentProcessNet8Build` argument will help identify how this may be simplified.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +514,128 @@ public override async Task RunAsync()
...
+
+            string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
+            if (projectFilePath != null)
+            {
+                targetFramework = await DotnetHelpers.DetermineTargetFramework(Path.GetDirectoryName(projectFilePath));
+            }
+            return targetFramework;
+        }
+
+        private bool ShouldLaunchOutOfProcFromWorkerRuntime()
```
fabiocav, line 583: We shouldn't have any scenarios where out of proc would be launched from this instance.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -40,10 +43,13 @@ internal class StartHostAction : BaseAction
     {
         private const int DefaultPort = 7071;
         private const int DefaultTimeout = 20;
-        private const string Net6FrameworkDescriptionPrefix = ".NET 6.0";
+        private const string Net8FrameworkDescriptionPrefix = ".NET 8.0";
         private const string WindowsExecutableName = "func.exe";
         private const string LinuxExecutableName = "func";
         private const string InProc8DirectoryName = "in-proc8";
+        private const string InProc6DirectoryName = "in-proc6";
```
fabiocav, line 50: Can we move this constants to a class scoping them to .NET (e.g., `DotnetConstants` or something along those lines)?

fabiocav (commented):
Initial review iteration.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -176,6 +184,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                .WithDescription("If provided, a path to the file that will be used to write the output when using --enable-json-output.")
                .Callback(jsonOutputFile => JsonOutputFile = jsonOutputFile);
 
+            Parser
+               .Setup<string>("runtime")
+               .WithDescription($"If provided, determines which version of the host to start. Allowed values are {InProc6HostRuntime}, {InProc8HostRuntime}, and default.")
+               .Callback(startHostAction => SetHostRuntime = startHostAction);
```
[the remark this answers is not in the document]
aishwaryabh: For `runtime`?

test/Azure.Functions.Cli.Tests/E2E/CreateFunctionTests.cs
aishwaryabh: Added https://github.com/Azure/azure-functions-core-tools/issues/3828

build/Settings.cs
aishwaryabh: Added each manually

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
aishwaryabh: Removed this; was a previous change

azure-pipelines.yml
```diff
@@ -11,7 +11,7 @@ trigger:
     include:
     - v4.x
     - release_4.0
-
+
```
liliankasem, line 14:
```suggestion

```

liliankasem: The new line is fine, but should remove empty space at very least. Or undo all changes to this file!

build.ps1
```diff
@@ -23,11 +23,11 @@ if ($env:IntegrationBuildNumber)
         throw $errorMessage
     }
 
-    $buildCommand = { dotnet run --integrationTests }
+    $buildCommand = { dotnet run --integrationTests --skipArtifactGeneration}
 }
 else
 {
-    $buildCommand = { dotnet run --ci }
+    $buildCommand = { dotnet run --ci --skipArtifactGeneration}
```
liliankasem, line 30: Why are we now skipping the artifact generation for all the build items? (in this file and in build/Program.cs)

eng/ci/public-build.yml
```diff
@@ -13,13 +13,15 @@ pr:
     include:
     - v4.x
     - release_4.0
+    - feature/*
 
 trigger:
```
liliankasem, line 18: These changes are fine for now, but in future it's cleaner to separate the CI changes from the code changes. It's easier to maintain and review and ensures changes like this don't get undone if we ever need to revert this PR for any reason.

pipelineUtilities.psm1
```diff
@@ -150,6 +150,7 @@ function Install-Dotnet {
         $listRuntimesOutput = dotnet --list-runtimes
         $installedDotnetRuntimes = $listRuntimesOutput | ForEach-Object { $_.Split(" ")[1] }
         Write-Host "Detected dotnet Runtimes: $($installedDotnetRuntimes -join ', ')"
+        dotnet --info
```
liliankasem, line 153: Curious why this change was made?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -81,6 +87,8 @@ internal class StartHostAction : BaseAction
 
         public string JsonOutputFile { get; set; }
 
+        public string? HostRuntime { get; set; }
```
liliankasem, line 90: Wondering why this property is nullable; is it okay for HostRuntime to be null?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -176,6 +184,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                .WithDescription("If provided, a path to the file that will be used to write the output when using --enable-json-output.")
                .Callback(jsonOutputFile => JsonOutputFile = jsonOutputFile);
 
+            Parser
+               .Setup<string>("runtime")
+               .WithDescription($"If provided, determines which version of the host to start. Allowed values are {InProc6HostRuntime}, {InProc8HostRuntime}, and default.")
```
liliankasem, line 189:
```suggestion
               .WithDescription($"If provided, determines which version of the host to start. Allowed values are '{InProc6HostRuntime}', '{InProc8HostRuntime}', and 'default'.")
```

Might also be good to clarify what the default is as well.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -176,6 +184,11 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                .WithDescription("If provided, a path to the file that will be used to write the output when using --enable-json-output.")
                .Callback(jsonOutputFile => JsonOutputFile = jsonOutputFile);
 
+            Parser
+               .Setup<string>("runtime")
```
liliankasem, line 188: Which runtime?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,36 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
 
-            var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            var isCurrentProcessNet8Build = RuntimeInformation.FrameworkDescription.Contains(Net8FrameworkDescriptionPrefix);
+            // If --runtime param is set, handle runtime param logic; otherwise we infer the host to launch on startup
+            if (HostRuntime != null)
```
liliankasem, line 438: nit: 
```suggestion
            if (HostRuntime is not null)
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +431,36 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
             {
-                await StartInProc8AsChildProcessAsync();
-                return;
+                // Check if we should start child process from user specified host runtime and return
+                var shouldStartChildProcess = await ShouldStartChildProcessFromHostRuntime(isCurrentProcessNet8Build, isVerbose);
+
+                if (shouldStartChildProcess)
+                {
+                    var isNet8InProcSpecified = (string.Equals(HostRuntime, InProc8HostRuntime, StringComparison.OrdinalIgnoreCase)) ? true : false;
```
liliankasem, line 445: `string.Equals` already returns a bool so no need to do a conditional check here

```suggestion
                    var isNet8InProcSpecified = string.Equals(HostRuntime, InProc8HostRuntime, StringComparison.OrdinalIgnoreCase);
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,6 +514,128 @@ public override async Task RunAsync()
...
+            {
+                ColoredConsole.WriteLine(VerboseColor($"Selected {hostRuntime} host"));
+            }
+        }
+
+        private async Task<string> GetTargetFrameworkAsync()
+        {
+            var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
+
+            string targetFramework = "";
```
liliankasem, line 573: nit:
```suggestion
            string targetFramework = string.Empty;
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: Added unit tests for all the happy paths (to check if the func.exe is located in the correct directory)

build.ps1
aishwaryabh: For the OOP branch we don't want any artifacts generated since the artifact generation/packaging will happen in the third and final pipeline, so I added the param so we wouldn't need to generate it here

eng/ci/public-build.yml
aishwaryabh: sounds good!

pipelineUtilities.psm1
aishwaryabh: I forgot to remove from before; this is not needed

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: The host runtime. We decided to have that be the parameter that users can pass in

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: yes! sometimes the user won't pass in the property and we can infer the host runtime by that :)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+                    throw new CliException($"Invalid config for running {DotnetConstants.InProc8HostRuntime} host.");
+                }
+            }
+            else if (string.Equals(HostRuntime, DotnetConstants.InProc6HostRuntime, StringComparison.OrdinalIgnoreCase))
+            {
+                if (isVerbose)
+                {
+                    PrintVerboseForHostSelection(isVerbose, DotnetConstants.InProc6HostRuntime);
+                }
+                return true;
```
kshyju, line 535: What happens when user has a net8 in-proc project and they pass --runtime inproc6 ?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+            }
+            else
```
kshyju, line 537: nit: Remove the `else { ` and just keep the throw. (reduces some nesting and improve readability)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +420,34 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
 
-            var isCurrentProcessNet6Build = RuntimeInformation.FrameworkDescription.Contains(Net6FrameworkDescriptionPrefix);
-            if (isCurrentProcessNet6Build && ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
+            // If --runtime param is set, handle runtime param logic; otherwise we infer the host to launch on startup
+            if (HostRuntime is not null)
```
kshyju, line 426: All the logic inside this `if` block should be done only if the CurrentWorkerRuntime is "dotnet". Should we check that here first and move on?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+            if (isVerbose)
+            {
+                ColoredConsole.WriteLine(VerboseColor($"Selected {hostRuntime} host"));
+            }
+        }
+
+        private async Task<bool> ShouldNet8InProcBeLaunched()
+        {
+            // Start .NET 8 child process if InProc8 is enabled
+            if (ShouldLaunchInProcNet8AsChildProcess() && await IsInProcNet8Enabled())
```
kshyju, line 554: Do we need this `ShouldLaunchInProcNet8AsChildProcess` method anymore? It was needed when we had a single branch, and we were producing artifacts for 2 versions (net6,net8). For this branch (oop core tools), I do not think we need that method anymore.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+            return false;
+        }
+
+        private bool ShouldLaunchChildProcessAfterInferringHostRuntimeAsync(bool shouldLaunchNet8, bool isVerbose)
+        {
+            // We should try to infer if we run inproc6 host, inproc8 host, or OOP host (default)
+            var isOutOfProc = GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.dotnet ? false : true;
+
+            // Check if the app is in-proc
+            if (!isOutOfProc)
```
kshyju, line 567: See my other comment about doing this check earlier.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+            var funcExecutablePath = shouldStartNet8ChildProcess ? GetInProcNet8ExecutablePath() : GetInProcNet6ExecutablePath();
 
-            EnsureNet8FuncExecutablePresent(inProc8FuncExecutablePath);
+            EnsureFuncExecutablePresent(funcExecutablePath, shouldStartNet8ChildProcess);
 
-            var inprocNet8ChildProcessInfo = new ProcessStartInfo
+            var childProcessInfo = new ProcessStartInfo
             {
-                FileName = inProc8FuncExecutablePath,
+                FileName = funcExecutablePath,
```
kshyju, line 625: Where do we use the `--no-build` argument?

test/Azure.Functions.Cli.Tests/Azure.Functions.Cli.Tests.csproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="Xunit.SkippableFact" Version="1.3.6" />
+	<PackageReference Include="System.IO.Abstractions" Version="2.1.0.227" />
```
kshyju, line 27: nit: indentation.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +501,128 @@ public override async Task RunAsync()
...
+                {
+                    if (isVerbose)
+                    {
+                        PrintVerboseForHostSelection(isVerbose, DotnetConstants.InProc8HostRuntime);
+                    }
+                    return true;
+                }
+                else
+                {
+                    throw new CliException($"Invalid config for running {DotnetConstants.InProc8HostRuntime} host.");
```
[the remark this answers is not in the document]
aishwaryabh: I updated the error message to be more specific so the user knows what's wrong

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: removed method

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: defined isOutOfProc earlier (still kept it in method for the logging though)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: We use it in line 626?

fabiocav:
The table shared above doesn't seem to capture the expected supported scenarios.

Shouldn't this reflect what is supported when `--runtime` is explicitly set?

| | inproc6|inproc8|default|
|--|:-------:|:-------:|:-------:|
|.NET Isolated app|| |x|
|.NET 6 in-proc app| x| ||
|.NET 8 in-proc app| | x||
|non .NET app| ||x|


For inference, the path would always be:
| |Sequence|Tool version|
|--|:-------:|:-------:|
|.NET Isolated app| default| default |
|.NET 6 in-proc app| default -> inproc6|inproc6 |
|.NET 6 in-proc app| default -> inproc8|inproc8 |
|non .NET app|default |default|

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: Added all the unit tests now (for not happy path as well)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +421,47 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
```
fabiocav, line 424: Can we move this logic that is specific to .NET to a different method and out of `RunAsync`? That will help declutter this method and make the code easier to navigate.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +421,47 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
...
+                    await StartHostAsChildProcessAsync(isNet8InProcSpecified);
+                    return;
+                }
+            }
+            else
             {
-                await StartInProc8AsChildProcessAsync();
-                return;
+                // Infer host runtime and check if we should launch child process
+                var shouldNet8InProcBeLaunched = await IsInProcNet8Enabled();
```
fabiocav, line 455: This all should be conditional on using `dotnet` (would be addressed if refactored into another method.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
...
+
+        private void PrintVerboseForHostSelection(bool isVerbose, string hostRuntime)
+        {
+            if (isVerbose)
+            {
+                ColoredConsole.WriteLine(VerboseColor($"Selected {hostRuntime} host"));
+            }
+        }
+
+        private bool ShouldLaunchChildProcessAfterInferringHostRuntimeAsync(bool shouldLaunchNet8, bool isInProc, bool isVerbose)
```
fabiocav, line 559: Non async/`Task` returning method. Drop `Async` suffix.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
...
+            }
+            // Throw an exception if HostRuntime is set to none of the expected values
+            throw new CliException($"Invalid host runtime '{HostRuntime}'. Valid values are 'default', '{DotnetConstants.InProc8HostRuntime}', '{DotnetConstants.InProc6HostRuntime}'.");
+        }
+
+        private void PrintVerboseForHostSelection(bool isVerbose, string hostRuntime)
+        {
+            if (isVerbose)
+            {
+                ColoredConsole.WriteLine(VerboseColor($"Selected {hostRuntime} host"));
```
fabiocav, line 555: nit: End with period.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
...
+            }
+        }
+
+        private bool ShouldLaunchChildProcessAfterInferringHostRuntimeAsync(bool shouldLaunchNet8, bool isInProc, bool isVerbose)
+        {
+            // Check if the app is in-proc
+            if (isInProc)
+            {
+                // Start .NET 8 child process if InProc8 is enabled
+                if (shouldLaunchNet8)
```
fabiocav, line 565: Please simplify to avoid code duplication.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
fabiocav: This will also enable you to reduce the variables as you'll be able to drive the logic without needing to have an additional flag for OOP.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -418,15 +421,47 @@ private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
         public override async Task RunAsync()
         {
             await PreRunConditions();
+            var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
+            var isOutOfProc = GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.dotnetIsolated ? true : false;
+            var isInProc = GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.dotnet ? true : false;
```
fabiocav, line 426: Simplify this assignments (i.e., `var isInProc = GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.dotnet;`)

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
             await runTask;
         }
 
+        private async Task<bool> ShouldStartChildProcessFromHostRuntime(bool isInProc, bool isVerbose)
+        {
+            if (string.Equals(HostRuntime, "default", StringComparison.OrdinalIgnoreCase))
+            {
+                if (isInProc)
+                {
+                    throw new CliException($"The runtime host value passed in, default, is not a valid host version for your project. For the default host runtime, the worker runtime must be set to {WorkerRuntime.dotnetIsolated}.");
```
fabiocav, line 524: There's a mix of `HostRuntime` validation in `RunAsync` and this method. Can we consolidate all this (moving this to a dedicated method?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
...
+            PrintVerboseForHostSelection(isVerbose, "out-of-process");
+            return false;
+          
+        }
+
         private static string GetInProcNet8ExecutablePath()
         {
             var funcExecutableDirectory = Path.GetDirectoryName(typeof(StartHostAction).Assembly.Location)!;
-            var executableName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? WindowsExecutableName : LinuxExecutableName;
+            var executableName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? DotnetConstants.WindowsExecutableName : DotnetConstants.LinuxExecutableName;
```
fabiocav, line 584: `GetInProcNet8ExecutablePath` and `GetInProcNet6ExecutablePath` are nearly identical, differing only on the path resolved.

Can we please consolidate these to avoid duplication?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -480,31 +515,102 @@ public override async Task RunAsync()
...
-        private Task StartInProc8AsChildProcessAsync()
+        private static string GetInProcNet6ExecutablePath()
+        {
+            var funcExecutableDirectory = Path.GetDirectoryName(typeof(StartHostAction).Assembly.Location)!;
+            var executableName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? DotnetConstants.WindowsExecutableName : DotnetConstants.LinuxExecutableName;
+
+            return Path.Combine(funcExecutableDirectory, DotnetConstants.InProc6DirectoryName, executableName);
+        }
+
+        private async Task StartHostAsChildProcessAsync(bool shouldStartNet8ChildProcess)
```
fabiocav, line 597: Nothing is being awaited on, we can remove `async` here.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: addressed use case

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: removed

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: added a code block that we only run this logic if we are running an inproc app

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: Added a statement where we only run this logic if running on dotnet

fabiocav (approved):
There are a few changes/improvements I'm tracking, but let's get this merged into the feature branch and collaborate there.

Issue #3817, opened 2024-09-06
Create documentation for Core tools requirement on minimum in-proc SDK version


mattchenderson:
Documentation should reflect the behavior of #3816

mattchenderson:
This was published this morning.

PR #3825, opened 2024-09-17
Updating coretools to detect unsupported inproc6 sdk scenario

### LINKED ISSUE #3816: Update Core Tools to detect unsupported in-proc workloads
The Core Tools version using the OOP host should be updated to detect unsupported in-proc scenarios and provide information about the required SDK version (recommend update to latest) if that is triggered.

surgupta-msft:

resolves #3816, https://github.com/Azure/azure-functions-core-tools/issues/3854

Our goal is to detect if OOP host is getting inproc payload and log the message to update the sdk to the latest and exit. Discussion in comment - https://github.com/Azure/azure-functions-core-tools/pull/3825#discussion_r1767697268

File `artifactsconfig.json` - It's a new config file we are introducing with this PR - it may or may not be present in coretools. Any changes that should go to visual studio and not to cli/vscode can have this check - `IsMinifiedVersion() = true`. In BuildSteps.cs, we create minified version of coretools for Visual Studio case - [link](https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/build/BuildSteps.cs#L121).

For now, artifactsconfig.json has only below content -

```json
{
    "minifiedVersion" : true
}
```

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR - https://github.com/Azure/azure-functions-core-tools/issues/3817


liliankasem (approved):

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -652,6 +659,31 @@ private static string GetInProcNet6ExecutablePath()
             return Path.Combine(funcExecutableDirectory, InProc6DirectoryName, executableName);
         }
 
+        private bool IsInvalidInProc6SdkVersion()
+        {
+            var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
+            string projectFilePath = ProjectHelpers.FindProjectFile(functionAppRoot);
```
mattchenderson, line 665: Should we be relying on the project file like this? Will this resolve properly in VS? I wonder if we should instead be using the built assets.

mattchenderson (changes requested):
Should this PR be targeting the other PR branch? We should land this, as it will release ahead of the OOP Core Tools changes, right?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
fabiocav: Yes, the goal here should be to evaluate the application state (inspect whether you're running an in-proc payload against the OOP Core Tools). Relying on the parsing the project to perform an SDK check here will be unreliable (e.g., using Central Package Management, importing packages from other projects/files, etc.)

@surgupta-msft we should be able to inspect the runtime configuration and detect if it's an in-proc payload.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -50,6 +50,9 @@ internal class StartHostAction : BaseAction
         private const string InProc6DirectoryName = "in-proc6";
         private const string InProc8HostRuntime = "inproc8";
         private const string InProc6HostRuntime = "inproc6";
+        private const string InProcFunctionsSdk = "Microsoft.NET.Sdk.Functions";
```
fabiocav, line 53: Commented below, but this information shouldn't be required.

fabiocav:
> Should this PR be targeting the other PR branch? We should land this, as it will release ahead of the OOP Core Tools changes, right?

Discussed this with @mattchenderson and we're good with the changes against the feature branch. This will give us the ability to update the SDK first while retaining support for in-proc with the default version of Core Tools.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: Do we only want to detect if OOP coretools is getting inproc6 payload? or do we also want to detect if that inproc6 payload is referencing to the older version of `Microsoft.NET.Sdk.Functions` which does not support custom host v0?

If it is option1 then I don't need to search for sdk and its version in the payload. But if it is option2 and we dont want to rely on project file and then may be I can parse `deps.json` file to find out the sdk version that will be loaded. Let me know if there is a better way to know the specific package version

surgupta-msft: Ah, my apologies, Fabio's comment here got deleted by mistake :( Not able to undo it. Had an offline discussion with Fabio. Our goal is to detect if OOP host is getting inproc payload and log the message to update the sdk to the latest and exit.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: Using sdk name just to print in the log message. Let me know if any correction is needed in the log.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -677,31 +693,6 @@ await CliTester.Run(new RunConfiguration
             }, _output);
         }
 
-        [Fact]
-        public async Task start_dotnet_isolated_inproc_with_specifying_runtime()
```
surgupta-msft, line 681: Duplicate of test - `start_dotnet6_inproc_with_specifying_runtime()`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -494,8 +494,7 @@ private async Task<bool> ShouldExitAfterDeterminingHostRuntime(bool isVerbose)
 
                 if (!string.Equals(HostRuntime, "default", StringComparison.OrdinalIgnoreCase))
                 {
-                    var isNet8InProcSpecified = string.Equals(HostRuntime, DotnetConstants.InProc8HostRuntime, StringComparison.OrdinalIgnoreCase);
-                    StartHostAsChildProcess(isNet8InProcSpecified);
+                    PrintMessageForInProcSdk();
```
aishwaryabh, line 497: do we always want this message printed out? Or only if we see it's not the lastest version?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -494,8 +494,7 @@ private async Task<bool> ShouldExitAfterDeterminingHostRuntime(bool isVerbose)
 
                 if (!string.Equals(HostRuntime, "default", StringComparison.OrdinalIgnoreCase))
                 {
-                    var isNet8InProcSpecified = string.Equals(HostRuntime, DotnetConstants.InProc8HostRuntime, StringComparison.OrdinalIgnoreCase);
-                    StartHostAsChildProcess(isNet8InProcSpecified);
```
aishwaryabh, line 498: also do we need to  remove this?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -514,7 +513,7 @@ private async Task<bool> ShouldExitAfterDeterminingHostRuntime(bool isVerbose)
                     PrintVerboseForHostSelection(isVerbose, DotnetConstants.InProc6HostRuntime);
                 }
 
-                StartHostAsChildProcess(shouldNet8InProcBeLaunched);
```
aishwaryabh, line 517: same with this; does this need to be removed?

src/Azure.Functions.Cli/Azure.Functions.Cli.csproj
```diff
@@ -62,9 +62,9 @@
     <EmbeddedResource Include="StaticResources\Dockerfile.dotnet8Isolated">
       <LogicalName>$(AssemblyName).Dockerfile.dotnet8Isolated</LogicalName>
     </EmbeddedResource>
-	<EmbeddedResource Include="StaticResources\Dockerfile.dotnet9Isolated">
+    <EmbeddedResource Include="StaticResources\Dockerfile.dotnet9Isolated">
```
aishwaryabh, line 65: nit indentation

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
aishwaryabh: otherwise the child process won't launch correct?

surgupta-msft: yeah that's my understanding that we need to just log and exit if an inproc payload is coming to OOP host.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: yeah that's my understanding that we need to just log and exit if an inproc payload is coming to OOP host.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: If they are using the latest version, the payload would not reach here.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -571,6 +570,10 @@ private async Task ValidateHostRuntime(bool isInProc, bool isVerbose)
             throw new CliException($"Invalid host runtime '{HostRuntime}'. Valid values are 'default', '{DotnetConstants.InProc8HostRuntime}', '{DotnetConstants.InProc6HostRuntime}'.");
         }
 
+        private void PrintMessageForInProcSdk()
+        {
+            Console.WriteLine($"This version of the Azure Functions Core Tools requires your project to reference version {DotnetConstants.InProcFunctionsMinSdkVersion} or later of {DotnetConstants.InProcFunctionsSdk}. Please update to the latest version. For more information, see: {DotnetConstants.InProcFunctionsDocsLink}");
```
mattchenderson, line 575: Will this print to the desired output stream? In earlier iterations, we had it as an exception that should go to stderr. This would just show up in stdout, right? I think this should be error level, as it is a terminal state.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
mattchenderson: Can you help me understand that? It seems like this is a code path that comes in naturally by simply running `func start --runtime inproc8`.

Maybe I'm tracing through it wrong, but it seems like that would be `RunAsync()` -> `ShouldExitAfterDeterminingHostRuntime()`. The `ValidateHostRuntime` method would return without throwing if I was indeed in the context of a .NET 8 in-process project. And then I would fail the string equality check against "default" and end up here.

That means this would break standard command line usage. I can't find the preconditions which assert this to be the VS variant or determine if the payload is missing the required SDK version.

surgupta-msft: Yeah so based on my discussion with Fabio, there does not seem to be any branching in code that one path will for VS scenario and other part will go to command line. Its the same code and during release, we use some configuration to prepare coretools for VS scenario.

surgupta-msft: I was able to confirm in our runtime sync. It is during the release we make the distinction.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: In my discussion with Fabio, we talked about not throwing an exception here, instead just log and exit. But we can discuss if you prefer an error here.  @fabiocav any thoughts/preference on this?

surgupta-msft: Updating to throw cli exception here

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: yes, will add this back

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: yes, will add this back

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: Will add more changes here which checks for VS variant

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -454,7 +454,7 @@ await CliTester.Run(new RunConfiguration
                 {
                     using (var client = new HttpClient() { BaseAddress = new Uri("http://localhost:7073") })
                     {
-                        await Task.Delay(TimeSpan.FromSeconds(2));
+                        await Task.Delay(TimeSpan.FromSeconds(5));
```
aishwaryabh, line 457: does the test fail if the delay isn't log enough?

aishwaryabh (approved):
LGTM!

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -494,8 +494,16 @@ private async Task<bool> ShouldExitAfterDeterminingHostRuntime(bool isVerbose)
 
                 if (!string.Equals(HostRuntime, "default", StringComparison.OrdinalIgnoreCase))
                 {
-                    var isNet8InProcSpecified = string.Equals(HostRuntime, DotnetConstants.InProc8HostRuntime, StringComparison.OrdinalIgnoreCase);
-                    StartHostAsChildProcess(isNet8InProcSpecified);
+                    if (Utilities.IsMinifiedVersion())
+                    {
+                        ThrowIfInProc();
+                    }
+                    else
```
aishwaryabh, line 501: nit: don't need else since an exception would've been thrown already

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
surgupta-msft: yes

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: good point, missed it before

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -571,6 +583,10 @@ private async Task ValidateHostRuntime(bool isInProc, bool isVerbose)
             throw new CliException($"Invalid host runtime '{HostRuntime}'. Valid values are 'default', '{DotnetConstants.InProc8HostRuntime}', '{DotnetConstants.InProc6HostRuntime}'.");
         }
 
+        private void ThrowIfInProc()
```
jviau, line 586: Using a "Throw**If**" naming is typically when the function itself will perform the conditional check to throw/not throw. Considering that is how it is used, maybe you could move the `if (Utilities.IsMinifiedVersion())` check to within this function?

src/Azure.Functions.Cli/ArtifactsConfigurationBuilder.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Functions.Cli
+{
+    internal class ArtifactsConfigurationBuilder : IConfigureBuilder<IConfigurationBuilder>
```
jviau, line 12: Is your goal with this class just to assist with registering that `builder.AddJsonFile` call on line 16? If so, this can be accomplished without needing to use an interface / instantiate this type.

Option 1: extension method
``` CSharp
internal static class ConfigurationBuilderExtensions
{
    public static IConfigurationBuilder AddArtifactsConfigFile(this IConfigurationBuilder builder)
    {
        return builder.AddJsonFile(Constants.ArtifactsConfigFileName, optional: true);   
    }
}
```

Option 2: Just inline it

This is only used once, so it wouldn't be the worst thing to just put `builder.AddJsonFile(Constants.ArtifactsConfigFileName, optional: true);` directly where it is used.

src/Azure.Functions.Cli/Common/Utilities.cs
```diff
@@ -263,5 +263,33 @@ internal static IConfigurationRoot BuildHostJsonConfigutation(ScriptApplicationH
...
+                {
+                    string value = section.Value;
+                    return bool.TryParse(value, out bool isValue) && isValue;
+                }
+            }
+            catch { }
+            return false;
+        }
+
+        internal static IConfigurationRoot BuildArtifactsConfiguration()
```
jviau, line 285: Do we not have an existing `IConfiguration` to use? This seems overkill to use config builder to extract a single value from a json file (whose name is hard coded).

I am not familiar with what is all in `artifactsconfig.json` or its uses outside of this check are, but I would recommend the following:

1. Is it already loaded somewhere? Can we reuse that?
2. If not loaded, can we just use `System.Text.Json` to load _and cache_ that single value we need to a static variable?

jviau (changes requested):

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
surgupta-msft: How about changing the name to something like `ThrowForInProc()` or `ThrowForInProcCase()`. To provide more context, `minified version` means the coretools that goes with Visual studio and it is slightly different than the cli version (it is minified by removing workers to trim its size). I feel in future, other parts of code can also use this check, hence keeping it in utilities seems right.

src/Azure.Functions.Cli/ArtifactsConfigurationBuilder.cs
surgupta-msft: Inline version looks good

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
jviau: `ThrowForInProc()` works for me.

src/Azure.Functions.Cli/Common/Utilities.cs
surgupta-msft: This class itself has configuration builder used a few times and its also there in [startup.cs](https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/src/Azure.Functions.Cli/Actions/HostActions/Startup.cs#L97). Any suggestions on how can I combine/reuse them?

For now, `artifactsconfig.json` has only below content -

```json
{
    "minifiedVersion" : true
}
```

jviau: Where does `artifactsconfig.json` come from?

src/Azure.Functions.Cli/ArtifactsConfigurationBuilder.cs
jviau, line 1: Can this file be removed now?

src/Azure.Functions.Cli/Common/Utilities.cs
surgupta-msft: Its a new config file we are introducing with this PR - it may or may not be present in coretools. Any changes that should go to visual studio and not to cli/vscode can have this check - `IsMinifiedVersion() = true`. In BuildSteps.cs, we create minified version of coretools for Visual Studio case - [link](https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/build/BuildSteps.cs#L121). We will need another PR to update this check in buildsteps.cs - probably create file `artifactsconfig.json` and make sure artifacts for both the scenarios vs and cli are being generated correctly.

jviau (approved):
Won't block this on the details of using `IConfigurationBuilder` here. Maybe we can revisit later.

src/Azure.Functions.Cli/ArtifactsConfigurationBuilder.cs
surgupta-msft: Yes, missed it

src/Azure.Functions.Cli/Common/Utilities.cs
jviau: Oh, so this introduces that file. Let me think on this for a minute.

jviau:
I'd like to discuss why we are using a config file for something that should be constant for a given set of artifacts. Curious why this path was chosen over embedding constants into the assembly for the min versions?

@fabiocav @surgupta-msft

mattchenderson (commented):
Why is a separate issue and PR needed for generation of the new configuration file? Shouldn't that be a part of this PR as well?

surgupta-msft:
> Why is a separate issue and PR needed for generation of the new configuration file? Shouldn't that be a part of this PR as well?

Thought the PR would be merged in the last sprint and did not want to increase the scope of it but since we are already halfway in the current sprint, it makes sense to include build side changes as well. I have updated the PR with changes in `BuildSteps.cs`

mattchenderson (approved):

PR #3835, opened 2024-09-25
Small refactor and new tests

fabiocav:
Small refactoring and some unit tests.

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
+++ b/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
@@ -23,7 +23,6 @@ using Microsoft.Azure.WebJobs.Script;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.WebHost;
-using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
@@ -357,42 +356,21 @@ namespace Azure.Functions.Cli.Actions.HostActions
         /// <summary>
         /// Check local.settings.json to determine whether in-proc .NET8 is enabled.
         /// </summary>
-        private async Task<bool> IsInProcNet8Enabled()
+        private async Task<bool> IsInProcDotNet8Enabled()
         {
             var localSettingsJObject = await GetLocalSettingsJsonAsJObjectAsync();
-            var inProcNet8EnabledSettingValue = localSettingsJObject?["Values"]?[Constants.FunctionsInProcNet8Enabled]?.Value<string>();
-            var isInProcNet8Enabled = string.Equals("1", inProcNet8EnabledSettingValue);
+            var inProcDotNet8EnabledValue = localSettingsJObject?["Values"]?[Constants.InProcDotNet8EnabledSetting]?.Value<string>();
+            var isInProcDotNet8Enabled = string.Equals("1", inProcDotNet8EnabledValue, StringComparison.Ordinal);
 
             if (VerboseLogging == true)
             {
-                var message = isInProcNet8Enabled
-                    ? $"{Constants.FunctionsInProcNet8Enabled} app setting enabled in local.settings.json"
-                    : $"{Constants.FunctionsInProcNet8Enabled} app setting is not enabled in local.settings.json";
+                var message = isInProcDotNet8Enabled
+                    ? $"{Constants.InProcDotNet8EnabledSetting} app setting enabled in local.settings.json"
+                    : $"{Constants.InProcDotNet8EnabledSetting} app setting is not enabled in local.settings.json";
                 ColoredConsole.WriteLine(VerboseColor(message));
             }
 
-            return isInProcNet8Enabled;
-        }
-
-        // We launch the in-proc .NET8 application as a child process only if the SkipInProcessHost conditional compilation symbol is not defined.
-        // During build, we pass SkipInProcessHost=True only for artifacts used by our feed (we don't want to launch child process in that case).
-        private bool ShouldLaunchInProcNet8AsChildProcess()
-        {
-#if SkipInProcessHost
-            if (VerboseLogging == true)
-            {
-                ColoredConsole.WriteLine(VerboseColor("SkipInProcessHost compilation symbol is defined."));
-            }
-
-            return false;
-#else
-            if (VerboseLogging == true)
-            {
-                ColoredConsole.WriteLine(VerboseColor("SkipInProcessHost compilation symbol is not defined."));
-            }
-
-            return true;
-#endif
+            return isInProcDotNet8Enabled;
         }
 
         private async Task<JObject> GetLocalSettingsJsonAsJObjectAsync()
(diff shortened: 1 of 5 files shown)


src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -482,98 +461,92 @@ public override async Task RunAsync()
...
-                    throw new CliException($"The runtime host value passed in, {DotnetConstants.InProc8HostRuntime}, is not a valid host version for your project. For the {DotnetConstants.InProc8HostRuntime} runtime, the {Constants.FunctionsInProcNet8Enabled} variable must be set while running a .NET 8 in-proc app.");
-                }
-                PrintVerboseForHostSelection(isVerbose, DotnetConstants.InProc8HostRuntime);
-                return;
+                ThrowCliException($"The runtime argument value provided, '{HostRuntime}', is not valid. Valid values are '{string.Join("', '", DotnetConstants.ValidRuntimeValues)}'.");
             }
-            else if (isInProc6)
+
+            var isInproc6Runtime = string.Equals(HostRuntime, DotnetConstants.InProc6HostRuntime, StringComparison.OrdinalIgnoreCase);
+            var isInproc8Runtime = string.Equals(HostRuntime, DotnetConstants.InProc8HostRuntime, StringComparison.OrdinalIgnoreCase);
```
mattchenderson, line 514: worth renaming these to `userSpecifiedInproc6Runtime` or something? "Is" feels ambiguous here.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -482,98 +461,92 @@ public override async Task RunAsync()
...
+                    break;
+                case WorkerRuntime.dotnetIsolated:
+                    if (isInproc8Runtime || isInproc6Runtime)
+                    {
+                        ThrowCliException($"The provided value is only valid for the worker runtime '{WorkerRuntime.dotnet}'");
+                    }
+                    break;
+                default:
+                    ThrowCliException($"The runtime argument is only valid for runtimes '{WorkerRuntime.dotnetIsolated}' and '{WorkerRuntime.dotnet}'");
+                    break;
```
mattchenderson, line 541: This may just be in the logic we're refactoring here, but what is the valid case for specifying the runtime argument for the isolated worker model? Should we just limit this flag to in-proc fully?

test/Azure.Functions.Cli.Tests/ActionsTests/StartHostActionTests.cs
```diff
@@ -186,6 +190,53 @@ private IFileSystem GetFakeFileSystem(IEnumerable<(string folder, string functio
...
+        // Isolated target,'default' argument, project configured for .NET 8. Succeeds.
+        [InlineData(WorkerRuntime.dotnetIsolated, "default", false, false)]
+        // Isolated target,in-proc 8 argument, project configured for .NET 8. Fails.
+        [InlineData(WorkerRuntime.dotnetIsolated, DotnetConstants.InProc8HostRuntime, true, true)]
+        // Isolated target,in-proc 6 argument, project not configured for .NET 8. Fails.
+        [InlineData(WorkerRuntime.dotnetIsolated, DotnetConstants.InProc6HostRuntime, false, true)]
+        // Unsupported runtime targets
+        [InlineData(WorkerRuntime.dotnetIsolated, "somevalue", false, true)]
+        [InlineData(WorkerRuntime.dotnet, "somevalue", false, true)]
+        public async Task ValidateHostRuntimeAsync_MatchesExpectedResults(WorkerRuntime currentRuntime, string hostRuntimeArgument, bool validNet8Configuration, bool expectException)
```
mattchenderson, line 213: Given the default branch of the switch statement I commented on separately, should there also be test for some other worker runtime?

mattchenderson (commented):
I think this looks good, but I'll want to scan over it again with fresh eyes tomorrow. A few small comments/questions, but none blocking I don't think.

mattchenderson:
I also just noticed the CI is failing with the following
> Expected error "The runtime argument value provided, 'inproc8', is not valid for your project. The runtime argument is only valid for runtimes 'dotnetIsolated' and 'dotnet'" to contain the strings: {"The runtime host value passed in, inproc8, is not a valid host version for your project. The runtime is only valid for dotnetIsolated and dotnet"}.

It seems like it's primarily upset about the quotes. I'm not 100% sure where that's coming from. Regardless, I would also expect the runtime string to be properly sausage-cased as "dotnet-isolated". I recall we had to work around this in help text previously as well.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
fabiocav: This logic was present in the initial implementation to support an explicit `default` value. The primary goal of these changes is to refactor the code for better readability and maintainability, while also adding test coverage, preserving the original behavior.

The use case this covers was not defined in previous designs but covers the somewhat unlikely scenario where inference would incorrectly map to an in-proc runtime, enabling an explicit override. Such a case would be a bug, but again, re-evaluating that was not the focus of the current changes, and having the option to avoid inference doesn't case any problems here.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -357,42 +357,21 @@ private void UpdateEnvironmentVariables(IDictionary<string, string> secrets)
...
-            var inProcNet8EnabledSettingValue = localSettingsJObject?["Values"]?[Constants.FunctionsInProcNet8Enabled]?.Value<string>();
-            var isInProcNet8Enabled = string.Equals("1", inProcNet8EnabledSettingValue);
+            var inProcDotNet8EnabledValue = localSettingsJObject?["Values"]?[Constants.FunctionsInProcNet8Enabled]?.Value<string>();
+            var isInProcDotNet8Enabled = string.Equals("1", inProcDotNet8EnabledValue, StringComparison.Ordinal);
 
             if (VerboseLogging == true)
             {
-                var message = isInProcNet8Enabled
+                var message = isInProcDotNet8Enabled
                     ? $"{Constants.FunctionsInProcNet8Enabled} app setting enabled in local.settings.json"
```
aishwaryabh, line 369: nit: change constant name also to be `FunctionsInProcDotNet8Enabled` just to be consistent with the rest of the changes

fabiocav:
There was a small update to the actual verbiage as well, so just updating the tests to deal with them for now:

This:
`The runtime host value passed in, inproc6, is not a valid host version for your project. The runtime is only valid for dotnetIsolated and dotnet`
Became:
`The runtime argument value provided, 'inproc6', is not valid for your project. The runtime argument is only valid for runtimes 'dotnetIsolated' and 'dotnet'`

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
fabiocav: I debated on this on making the changes, but decided to keep the existing name because they map to the constant (setting) `FUNCTIONS_INPROC_NET8_ENABLED`, which I can't change.

I'll move forward with the change.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
fabiocav: I actually ended up collapsing the handling for isolated and non .NET worker cases, simplifying the flow and expanding the runtime support to other stacks (that's something I had explicitly blocked with the changes, differing from the original). This let me clean things up even further.

aishwaryabh (approved):
LGTM!

liliankasem (approved):

PR #3836, opened 2024-09-25
When local.settings.json is absent, runnig func host start does not work for newer worker models #3544

### LINKED ISSUE #3544: When local.settings.json is absent, runnig `func host start` does not work for newer worker models
With worker indexing enabled, when local.settings.json is absent, runnnig `func host start` does not work. The terminal becomes stuck.

See https://github.com/Azure/azure-functions-dotnet-worker/issues/2107#issuecomment-1836736389

VineethReyya:
-> when local.settings.json  is absent.
Scenario1:  running 'func host start' will Display Available WorkerRuntime on the screen and user can select the Runtime from the list so the FUNCTION_WORKER_RUNTIME setting is added and start hosting the function app
Scenario2: running 'func host start --{WorkerRuntime}' will add FUNCTION_WORKER_RUNTIME setting and start hosting the function app.

resolves #3544 

### Pull request checklist


sinedied:
Thanks for working on this!
Could you also check if this approach works when the core tools are started by the Static Web Apps CLI?
I'm afraid the terminal might not be TTY in that case, and would then not allow user interactions.

SWA CLI is used in many of our samples along with the Functions emulators, so this would still remain an issue.

VineethReyya:
> Thanks for working on this! Could you also check if this approach works when the core tools are started by the Static Web Apps CLI? I'm afraid the terminal might not be TTY in that case, and would then not allow user interactions.
> 
> SWA CLI is used in many of our samples along with the Functions emulators, so this would still remain an issue.

Thanks for the suggestion. will look into it

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
                 : null;
             return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
         }
+        private void CheckSettingFunctionWorkerRuntime()
```
liliankasem, line 731: new line:

```suggestion
        
        private void CheckSettingFunctionWorkerRuntime()
```

src/Azure.Functions.Cli/Common/AppSettingsFile.cs
```diff
@@ -29,7 +29,7 @@ public AppSettingsFile(string filePath)
             {
                 Values = new Dictionary<string, string>();
                 ConnectionStrings = new Dictionary<string, JToken>();
-                IsEncrypted = true;
+                IsEncrypted = false;
```
liliankasem, line 32: Why is this being changed?

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -913,5 +913,60 @@ private void SetUserSecrets(string workingDir, Dictionary<string, string> userSe
                 _output.WriteLine(procOutput);
             }
         }
+
+        [Theory]
+        [InlineData("dotnet")]
+        [InlineData("node")]
+        [InlineData("javascript")]
+        public async Task start_on_missing_local_settings_json(string language)
```
liliankasem, line 921:
```suggestion
        public async Task Start_MissingLocalSettingsJson_{expectedResult}(string language)
```

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -913,5 +913,60 @@ private void SetUserSecrets(string workingDir, Dictionary<string, string> userSe
                 _output.WriteLine(procOutput);
             }
         }
+
+        [Theory]
```
liliankasem, line 917: Please move this test above the utility methods in this class

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -585,6 +585,7 @@ private void ValidateAndBuildHostJsonConfigurationIfFileExists(ScriptApplication
 
         private async Task PreRunConditions()
         {
+            CheckSettingFunctionWorkerRuntime();
```
liliankasem, line 588: Why is this the first thing that needs to happen?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
liliankasem, line 731: Check function worker runtime setting for what? Can the method name be more specific of what it is actually doing?

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
                 : null;
             return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
         }
+        private void CheckSettingFunctionWorkerRuntime()
+        {
+            var setting = _secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
+            if (setting == null)
```
liliankasem, line 734:
```suggestion
            
            if (setting is not null)
            {
                // exit early, cli message about what the value is set to
            }
            
            // rest of code
                 
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
...
+            {
+                if (GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone == WorkerRuntime.None)
+                {
+                    SelectionMenuHelper.DisplaySelectionWizardPrompt("worker runtime");
+                    IDictionary<WorkerRuntime, string> workerRuntimeToDisplayString = WorkerRuntimeLanguageHelper.GetWorkerToDisplayStrings();
+                    string workerRuntimedisplay = SelectionMenuHelper.DisplaySelectionWizard(workerRuntimeToDisplayString.Values);
+                    GlobalCoreToolsSettings.CurrentWorkerRuntime = workerRuntimeToDisplayString.FirstOrDefault(wr => wr.Value.Equals(workerRuntimedisplay)).Key;
+
```
liliankasem, line 742:
```suggestion
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
...
+                }
+                var workerRuntime = WorkerRuntimeLanguageHelper.GetRuntimeMoniker(GlobalCoreToolsSettings.CurrentWorkerRuntime);
```
liliankasem, line 744:
```suggestion
                
                var workerRuntime = WorkerRuntimeLanguageHelper.GetRuntimeMoniker(GlobalCoreToolsSettings.CurrentWorkerRuntime);
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
...
+        private void CheckSettingFunctionWorkerRuntime()
+        {
+            var setting = _secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
+            if (setting == null)
+            {
+                if (GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone == WorkerRuntime.None)
+                {
+                    SelectionMenuHelper.DisplaySelectionWizardPrompt("worker runtime");
+                    IDictionary<WorkerRuntime, string> workerRuntimeToDisplayString = WorkerRuntimeLanguageHelper.GetWorkerToDisplayStrings();
+                    string workerRuntimedisplay = SelectionMenuHelper.DisplaySelectionWizard(workerRuntimeToDisplayString.Values);
```
liliankasem, line 740:
```suggestion
                    string workerRuntimeDisplay = SelectionMenuHelper.DisplaySelectionWizard(workerRuntimeToDisplayString.Values);
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,24 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
                 : null;
             return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
         }
+        private void CheckSettingFunctionWorkerRuntime()
+        {
+            var setting = _secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
```
liliankasem, line 733: Can you be more specific about the setting name? i.e. `workerRuntimeSettingValue`

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -913,5 +913,60 @@ private void SetUserSecrets(string workingDir, Dictionary<string, string> userSe
...
+        [InlineData("javascript")]
+        public async Task start_on_missing_local_settings_json(string language)
+        {
+            await CliTester.Run(new RunConfiguration[]
+            {
+                new RunConfiguration
+                {
+                    Commands = new[]
+                    {
+                        $"init . --worker-runtime {language}",
```
liliankasem, line 929: Can you write a test for when --worker-runtime is not provided and is None

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
VineethReyya: @liliankasem when local.settings.json file is missing and if we start hosting the function app . we just get the error 
image
To Avoid this  CheckSettingFunctionWorkerRuntime() needs to run first to add setting for Function_Worker_Runtime

src/Azure.Functions.Cli/Common/AppSettingsFile.cs
VineethReyya: @liliankasem Changed it because of the below warning:
image
the value of setting (FUNCTION_WORKER_RUNTIME) is encrypted.

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
VineethReyya: @liliankasem start to host function app when  local.settings.json is missing ,it will show 
image
then we can select the worker runtime

src/Azure.Functions.Cli/Common/AppSettingsFile.cs
liliankasem: I think this is something you can do to decrypt it? I don't think we should be changing this setting

@fabiocav fyi

VineethReyya: @liliankasem Yes, we can decrypt it and there is no issue with that. it will work in either way

liliankasem: Great, then let's do that and keep `IsEncrypted` = true

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
VineethReyya: @liliankasem could you please review overall changes after review. also added new test cases

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
liliankasem: Right so it's not just checking, it's taking an action too. `EnsureWorkerRuntimeIsSet` might be a better name here.

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,26 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
                 : null;
             return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
         }
+        
+        private void CheckSettingFunctionWorkerRuntime()
+        {
+            var workerRuntimeSettingValue = _secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
```
liliankasem, line 734: `_secretsManager.GetSecrets()` could be null as well

Either use null-conditional

```suggestion
            var workerRuntimeSettingValue = _secretsManager?.GetSecrets()?.FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase))?.Value;
```

Or consider breaking this out into a smaller method with sufficient null checks:

```
private string GetWorkerRuntimeFromSecrets()
{
  var secrets = _secretsManager?.GetSecrets();
  
  if (secrets is null)
  {
      throw new InvalidOperationException("Secrets Manager did not return any secrets.");
  }
  
  var workerRuntimeSettingValue = secrets
      .FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase))
      ?.Value;
  
  if (workerRuntimeSettingValue == null)
  {
      // Handle the case where the worker runtime setting is missing (e.g., log it, set default, etc.)
  }
}
```

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
```diff
@@ -727,5 +728,26 @@ internal static bool ConnectionExists(IEnumerable<KeyValuePair<string, string>>
...
+            {
+                SelectionMenuHelper.DisplaySelectionWizardPrompt("worker runtime");
+                IDictionary<WorkerRuntime, string> workerRuntimeToDisplayString = WorkerRuntimeLanguageHelper.GetWorkerToDisplayStrings();
+                string workerRuntimeDisplay = SelectionMenuHelper.DisplaySelectionWizard(workerRuntimeToDisplayString.Values);
+                GlobalCoreToolsSettings.CurrentWorkerRuntime = workerRuntimeToDisplayString.FirstOrDefault(wr => wr.Value.Equals(workerRuntimeDisplay)).Key;
+            }
+
+            var workerRuntime = WorkerRuntimeLanguageHelper.GetRuntimeMoniker(GlobalCoreToolsSettings.CurrentWorkerRuntime);
+            _secretsManager.SetSecret(Constants.FunctionsWorkerRuntime, workerRuntime);
+            ColoredConsole.WriteLine(WarningColor($"'{workerRuntime}' has been set in your local.settings.json"));
```
liliankasem, line 749: Consider other items that may result in a null result in this function

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
VineethReyya: @liliankasem the current implementation is throwing an exception for the missing setting in the secret file and
image

 user does not want to stop hosting without the secret file by throwing an exception for missing setting. Instead, user want to add when none is found.

->solution I preferred to choose is to check the secretfile(local.settings.json). if its null it will add the setting FUNCTION_WORKER_RUNTIME. (While adding the setting it will add new secret file if it's doesn't exist and if it exits it will overwrite the setting which was added.)

VineethReyya: @liliankasem in both cases (either secret file is missing, or secrets are missing). I am adding setting (while adding setting it will check if we have secret file then it will add new setting else it will add new secret file with settings in it).

src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs
VineethReyya: @liliankasem it's already handling null in the catch block by assigning [values](https://github.com/Azure/azure-functions-core-tools/blob/v4.x/src/Azure.Functions.Cli/Common/AppSettingsFile.cs#L29) when its null

liliankasem (dismissed):

VineethReyya:
this PR is refreshed with [new PR](https://github.com/Azure/azure-functions-core-tools/pull/4142)

Issue #3830, opened 2024-09-19
Update tests in OOP feature branch to have trait to run E2E tests


aishwaryabh:
Any E2E tests should have the following trait attached to it so that integration tests are skipped in the OOP host pipeline, but should be run in the final pipeline:
```
[Fact]
        [Trait(TraitName.Category, TestCategory.Integration)]
        public async Task start_dotnet8_inproc()
        {
```

PR #3841, opened 2024-09-25
Add traits for e2e tests to use in final pipeline

aishwaryabh:

resolves [#3830](https://github.com/Azure/azure-functions-core-tools/issues/3830)

### Pull request checklist


test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -389,6 +389,42 @@ await CliTester.Run(new RunConfiguration
             }, _output);
         }
 
+        [Fact]
+        [Trait(TestingTraits.TraitName.Category, TestingTraits.TestCategory.FinalIntegration)]
+        public async Task start_dotnet8_inproc_with_specifying_runtime_e2e()
+        {
+            await CliTester.Run(new RunConfiguration
```
kshyju, line 396: I think we also need a slight adjustment to use the consolidated zip (which contains all the needed artifacts from other builds) to be used for these E2E test. Also, we don't need to run it through the Shell class. We can do `dotnet` task in that pipeline to run the test project. will follow up with you offline.

build/BuildSteps.cs
```diff
@@ -341,7 +341,7 @@ public static void Test()
 
             Environment.SetEnvironmentVariable("DURABLE_FUNCTION_PATH", Settings.DurableFolder);
 
-            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net8.0 --logger trx");
+            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net8.0 --logger trx --filter \"{TestingTraits.TraitName.Category}~{TestingTraits.TestCategory.FinalIntegrationTests}\"");
```
liliankasem, line 344: Why do we need to add traits? And why would it be called "final"

aishwaryabh: so we need to add traits because we want to test the E2E functionality for when we have the core tools OOP host as the root and inproc6 and inproc8 host's as child folders. Before I had integration tests in the `feature/oop-host` branch where we try to run the inproc6/inproc8 host runtimes, but we ultimately fail since the artifact hasn't been assembled with the child folders. Technically those are still integration tests (but they fail since the artifacts have not been generated).

So that's why I added the name `FinalIntegrationTests`. Please let me know if you have better suggestions for the name haha

aishwaryabh: maybe `FinalArtifactsIntegrationTests` would be more clear?

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
aishwaryabh: yeah this adjustment will be done in the pipeline itself. Let me know if this looks ok

build/BuildSteps.cs
```diff
@@ -341,7 +341,7 @@ public static void Test()
 
             Environment.SetEnvironmentVariable("DURABLE_FUNCTION_PATH", Settings.DurableFolder);
 
-            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net8.0 --logger trx");
+            Shell.Run("dotnet", $"test {Settings.TestProjectFile} -f net8.0 --logger trx --filter \"{TestingTraits.TraitName.Category}!={TestingTraits.TestCategory.FinalIntegrationTests}\"");
```
mattchenderson, line 344: Note that this only fixes this for this specific build tool and will have problems elsewhere, such as when someone uses Test Explorer or does a `dotnet run` themselves.

I would instead recommend adjusting the test project default filters appropriately to exclude the trait, leaving the existing line here, and then adding a line to have a filter just to the new trait. You can see an example of this in the Worker - For SDK container builds, I had to adjust the configuration to avoid introducing a docker daemon dependency in normal test paths.

mattchenderson: Here's the PR for the worker changeset I was referring to: https://github.com/Azure/azure-functions-dotnet-worker/pull/2671/files

build/BuildSteps.cs
liliankasem: I think `FinalArtifactsIntegrationTests` is better; or even just `ArtifactsIntegrationTests` since these are integration tests that will run with artifacts

aishwaryabh: Changed the name to be `RequiresNestedInProcArtifacts`

build/TestingTraits.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Build
+{
+    internal static class TestingTraits
+    {
+        internal class TestCategory
+        {
+            public const string FinalIntegrationTests = "FinalIntegrationTests";
```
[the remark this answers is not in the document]
aishwaryabh: Changed the name to be RequiresNestedInProcArtifacts

build/BuildSteps.cs
aishwaryabh: thanks! I made the according changes

build/TestingTraits.cs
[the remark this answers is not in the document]
aishwaryabh: Let me know if this looks ok

test/Azure.Functions.Cli.Tests/E2E/Helpers/TestingTraits.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Azure.Functions.Cli.Tests.E2E.Helpers
+{
+    internal static class TestingTraits
```
liliankasem, line 9: nit: TestTraits

test/Azure.Functions.Cli.Tests/E2E/Helpers/TestingTraits.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        internal class TestCategory
```
liliankasem, line 11: I don't think we need these extra internal classes; I also have a change that introduces test traits here:

https://github.com/Azure/azure-functions-core-tools/blob/liliankasem/test-collections/test/Azure.Functions.Cli.Tests/Utilities/TestTraits.cs

So I wonder if we can follow a similar pattern so it would be easy to combine these two later when my PR is open

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -389,6 +391,76 @@ await CliTester.Run(new RunConfiguration
             }, _output);
         }
 
+        [Fact]
+        [Trait(TestingTraits.TraitName.Category, TestingTraits.TestCategory.RequiresNestedInProcArtifacts)]
```
liliankasem, line 395: With the suggested change mentioned in the other PR, it would look like this:

`[Trait(TestTraits.Category, TestTraits.RequiresNestedInProcArtifacts)]`

test/Azure.Functions.Cli.Tests/E2E/StartTests_default.runsettings
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <TestCaseFilter>(Category != RequiresNestedInProcArtifacts)</TestCaseFilter>
+  </RunConfiguration>
```
liliankasem, line 5: curious why we're using .runsettings when the filters can be set via dotnet test --filter? Is this just easier because of how the build.csproj works?

aishwaryabh: yeah so I changed it to use .runsettings since it might be confusing for users once they are running the tests locally (using test explorer or running `dotnet test` themselves). Matthew talked about this in the comment here: https://github.com/Azure/azure-functions-core-tools/pull/3841#discussion_r1778854154

test/Azure.Functions.Cli.Tests/E2E/Helpers/TestingTraits.cs
aishwaryabh: changed it to follow your pattern!

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
liliankasem: Actually, one more suggestion on this: can you switch to using groups for this? I want to have these all in the E2E category @aishwaryabh

aishwaryabh: Switched to using groups!

liliankasem (approved):

Issue #3826, opened 2024-09-17
Release OOP Core Tools


mattchenderson:
Release the artifacts generated from changes tracked by https://github.com/Azure/azure-functions-core-tools/issues/3744

fabiocav:
This has been released for VS (minified versions on CDN).

aishwaryabh:
This has been done

PR #3848, opened 2024-09-27
Create Artifacts for releasing OOP core tools

aishwaryabh:

resolves [#3826](https://github.com/Azure/azure-functions-core-tools/issues/3826)

This PR modifies the `ArtifactAssembler` and zips up the the different core tools versions (OOP, inproc6, and inproc8). The staging directory looks as follows:
image

Within `coretools-cli`, each zip file contains the following structure (assuming that this is a windows generated artifact):
- Func.exe (supports OOP host)
     - Child folder: In-proc6 
         - Contains func.exe for running inproc6 host
     - Child folder: In-proc8
         - Contains func.exe for running inproc8 host

image

After creating the artifacts, we generate the MSI files and perform ESRP code signing on the MSI files. After this, we run E2E tests based on the traits to ensure that the scenarios are working as expected for `win-x64` and `win-x86` platforms. 

Then we finally zip up the CLI artifacts and upload both the VS and CLI artifacts to a blob with the version the as the name of the blob.

Here is the core tools consolidation package creation pipeline we use:
https://azfunc.visualstudio.com/internal/_release?definitionId=4&view=mine&_a=releases

### Pull request checklist


build/BuildSteps.cs
```diff
@@ -134,13 +128,6 @@ public static void DotnetPublishForZips()
...
         }
 
-        private static string BuildNet8ArtifactFullPath(string runtime)
-        {
-            return Path.Combine(Settings.OutputDir, BuildNet8ArtifactDirectory(runtime));
-        }
-
-        private static string BuildNet8ArtifactDirectory(string runtime) => $"{runtime}{Net8ArtifactNameSuffix}";
-
         private static void ExecuteDotnetPublish(string outputPath, string rid, string targetFramework, bool skipLaunchingNet8ChildProcess)
```
kshyju, line 131: Could we remove the code path which relies on `skipLaunchingNet8ChildProcess`? (here and the CLI project) I do not think that is needed anymore since this branch is specifically for oop use cases.

build/BuildSteps.cs
```diff
@@ -121,11 +120,6 @@ public static void DotnetPublishForZips()
                 {
                     RemoveLanguageWorkers(outputPath);
```
kshyju, line 121: Do we have any use case where we need to remove language workers in this version/branch of core tools? If not, let's remove it.

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -161,12 +198,150 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
             // Delete the extracted directories
             Directory.Delete(_inProc6ExtractedRootDir, true);
             Directory.Delete(_inProc8ExtractedRootDir, true);
             Directory.Delete(_coreToolsHostExtractedRootDir, true);
+            Directory.Delete(_outOfProcExtractedRootDir, true);
+
+            Console.WriteLine($"Finished assembling CLI Core Tools artifacts");
+        }
+
+        private string[] GetArtifactDirectoryAndVersionNumber(string extractedRootDirectory, string artifactName)
```
kshyju, line 299: Can you rebase? We have a method to get this information now.

https://github.com/Azure/azure-functions-core-tools/blob/df4c712092e425fff63bb30b7e4231984bef5a02/src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs#L127-L136

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -161,12 +198,150 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+                    var artifactNameWithVersion = $"{artifactName}.{outOfProcVersion}";
+                    outOfProcArtifactDirPath = Path.Combine(_outOfProcExtractedRootDir, artifactNameWithVersion);
+                }
+
+                // Create a new directory to store the oop core tools with in-proc8 and in-proc6 files.
+                var outOfProcArtifactName = Path.GetFileName(outOfProcArtifactDirPath);
+                var consolidatedArtifactDirPath = Path.Combine(cliCoreToolsTargetArtifactDir, outOfProcArtifactName);
+                Directory.CreateDirectory(consolidatedArtifactDirPath);
+
+                // Copy oop host files
```
kshyju, line 237: Comment is incorrect. There is no host files involved in this method. correct?

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -161,12 +198,150 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+                    outOfProcVersion = outOfProcResponse[1];
+                }
+                else
+                {
+                    var artifactNameWithVersion = $"{artifactName}.{outOfProcVersion}";
+                    outOfProcArtifactDirPath = Path.Combine(_outOfProcExtractedRootDir, artifactNameWithVersion);
+                }
+
+                // Create a new directory to store the oop core tools with in-proc8 and in-proc6 files.
+                var outOfProcArtifactName = Path.GetFileName(outOfProcArtifactDirPath);
```
kshyju, line 233: wonder we could move this to a small method which can be called from here and from `CreateVisualStudioCoreToolsAsync` method?

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -161,12 +198,150 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+                    inProcVersion = inProc8Response[1];
+                }
+                else
+                {
+                    var artifactNameWithVersion = $"{artifactName}.{inProcVersion}";
+                    inProc8ArtifactDirPath = Path.Combine(_inProc8ExtractedRootDir, artifactNameWithVersion);
+                }
+
+                // Rename inproc8 directory to have the same version as the out-of-proc artifact before copying
+                string newInProc8ArtifactDirPath = RenameInProcDirectory(inProc8ArtifactDirPath, outOfProcVersion);
```
kshyju, line 257: Why do we need this rename operation ? Aren't we eventually copying it to a directory with static name (`InProc8DirectoryName` in line 260) ?

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
kshyju: If you still need to use this method, `string[]` may not be the appropriate return type here. Either use named tuple or a POCO to improve readability.

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: we need this method so that we can extract the version number from the inproc8 version (since we need the full name/version to copy the files over from one directory to another)

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: the logic is a bit different so I decided to separate them to keep it easier to parse/read

build/BuildSteps.cs
fabiocav: The minified version is used by VS and the OOP version used to run .NET Isolated shouldn't have any workers, so that is still a valid use case.

code-mirror.yml
```diff
@@ -9,6 +9,7 @@ trigger:
     - release_3.0
     - release_4.0_hotfix
     - feature/*
+    - aibhandari/*
```
aishwaryabh, line 12: TODO remove this

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -105,7 +129,8 @@ private static string CreateStagingDirectory(string rootWorkingDirectory)
 
             if (Directory.Exists(stagingDirectory))
             {
-                Directory.Delete(stagingDirectory, true);
+                Console.WriteLine($"Directory already exists");
+                Directory.Delete(stagingDirectory, true );
```
liliankasem, line 133:
```suggestion
                Directory.Delete(stagingDirectory, true);
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -105,7 +129,8 @@ private static string CreateStagingDirectory(string rootWorkingDirectory)
 
             if (Directory.Exists(stagingDirectory))
             {
-                Directory.Delete(stagingDirectory, true);
+                Console.WriteLine($"Directory already exists");
```
liliankasem, line 132:
```suggestion
                Console.WriteLine($"Directory '{stagingDirectory}' already exists, deleting...");
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -137,8 +162,9 @@ private static string GetCoreToolsProductVersion(string artifactDirectoryName)
 
         private async Task CreateVisualStudioCoreToolsAsync()
         {
+            Console.WriteLine($"Starting to assemble Visual Studio Core Tools artifacts");
```
liliankasem, line 165: You only need the "$" for string interpolation

```suggestion
            Console.WriteLine("Starting to assemble Visual Studio Core Tools artifacts");
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+        }
+
+        private async Task CreateCliCoreToolsAsync()
+        {
+            Console.WriteLine($"Starting to assemble CLI Core Tools artifacts");
+
+            // Create a directory to store the assembled artifacts.
+            var cliCoreToolsTargetArtifactDir = Path.Combine(_stagingDirectory, CliOutputArtifactDirectoryName);
+            Directory.CreateDirectory(cliCoreToolsTargetArtifactDir);
+            string outOfProcVersion = string.Empty;
```
liliankasem, line 220: nit: doesn't matter either way, but can be simplified like this:

```csharp
string outOfProcVersion = string.Empty, 
       inProcVersion = string.Empty, 
       outOfProcArtifactDirPath = string.Empty, 
       inProc8ArtifactDirPath = string.Empty;
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
             Directory.Delete(_coreToolsHostExtractedRootDir, true);
+            Directory.Delete(_outOfProcExtractedRootDir, true);
+
+            Console.WriteLine($"Finished assembling CLI Core Tools artifacts");
+        }
+
+        private (string artifactDirectory, string version) GetArtifactDirectoryAndVersionNumber(string extractedRootDirectory, string artifactName)
+        {
+            var artifactDirPath = Directory.EnumerateDirectories(extractedRootDirectory)
+                                          .FirstOrDefault(dir => dir.Contains(artifactName));
```
liliankasem, line 294: nit: line up `.`

```suggestion
            var artifactDirPath = Directory.EnumerateDirectories(extractedRootDirectory)
                                           .FirstOrDefault(dir => dir.Contains(artifactName));
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+            if (artifactDirPath == null)
```
liliankasem, line 295: nit: using `is`

```suggestion
            if (artifactDirPath is null)
```

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+                throw new InvalidOperationException($"Artifact directory '{artifactDirPath}' not found!");
+            }
+
+            var version = GetCoreToolsProductVersion(artifactDirPath);
+            return (artifactDirPath, version);
+        }
+
+        private string RenameInProcDirectory(string oldArtifactDirPath, string newVersion)
+        {
+            string pattern = @"^(.*?)(\d+\.\d+\.\d+)$";  // Capture everything before the version
```
liliankasem, line 306: I think it's better to define at class level like this if this method is called more than once

`private readonly Regex ArtifactNameRegex = new Regex(@"^(.*?)(\d+\.\d+\.\d+)$");`

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+
+            Match match = Regex.Match(oldArtifactDirPath, pattern);
+
+            if (!match.Success)
+            {
+                throw new InvalidOperationException($"Unable to extract content before version number from '{oldArtifactDirPath}'.");
+            }
+
+            var contentBeforeVersion = match.Groups[1];
+            var newDirectoryName =  $"{contentBeforeVersion}{newVersion}";
```
liliankasem, line 316: is content before version just the artifact name? can we rename the variable to something more descriptive?

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +207,125 @@ private async Task CreateVisualStudioCoreToolsAsync()
...
+
+            // Rename (move) the directory
+            try
+            {
+                Directory.Move(oldArtifactDirPath, newDirectoryName);
+            }
+            catch (Exception ex)
+            {
+                Console.Write(ex.ToString());
```
liliankasem, line 325: What happens if we are unable to move the directory? Will this be a break that we can't recover from?

src/Azure.Functions.ArtifactAssembler/Program.cs
```diff
@@ -11,8 +13,8 @@ static async Task<int> Main(string[] args)
             {
                 var currentWorkingDirectory = Environment.CurrentDirectory;
                 var artifactAssembler = new ArtifactAssembler(currentWorkingDirectory);
-                await artifactAssembler.AssembleArtifactsAsync();
 
+                await artifactAssembler.AssembleArtifactsAsync();
```
liliankasem, line 17: Not sure I understand why a change was made here? Looks the same to me - can we undo?

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: done!

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: done!

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: yeah I realized I added that try/catch when implementing but we don't need this anymore. Removing try /catch

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
aishwaryabh: renamed!

mattchenderson (commented):
I think this is looking good based on a quick overview - I admittedly have not traced through all of it. I would want to see the pipeline drop the personal branches prior to approval (already commented upon). I do feel like `BuildSteps.cs` should also be running the assembler at some level, though that's probably out-of-scope future work.

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -11,9 +11,12 @@ internal sealed class ArtifactAssembler
         private const string InProc8DirectoryName = "in-proc8";
         private const string InProc6DirectoryName = "in-proc6";
         private const string CoreToolsHostDirectoryName = "host";
-        private const string OutputArtifactDirectoryName = "coretools-visualstudio";
+        private const string VisualStudioOutputArtifactDirectoryName = "coretools-visualstudio";
         private const string _InProcOutputArtifactNameSuffix = "_inproc";
         private const string _coreToolsProductVersionPattern = @"(\d+\.\d+\.\d+)$";
+        private const string _artifactNameRegexPattern = @"^(.*?)(\d+\.\d+\.\d+)$";
+        private const string OutOfProcDirectoryName = "out-of-proc";
```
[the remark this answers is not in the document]
aishwaryabh: changed name to default

src/Azure.Functions.ArtifactAssembler/ArtifactAssembler.cs
```diff
@@ -181,11 +208,117 @@ private async Task CreateVisualStudioCoreToolsAsync()
             });
 
             await Task.WhenAll(packTasks);
+            Console.WriteLine($"Finished assembling Visual Studio Core Tools artifacts");
+        }
+
+        private async Task CreateCliCoreToolsAsync()
+        {
+            Console.WriteLine($"Starting to assemble CLI Core Tools artifacts");
```
liliankasem, line 216: Can you remove the $ from all strings that are not doing string interpolation?

aishwaryabh: done!

mattchenderson (approved):
Based on the executions of test runs through the pipelines, this seems to be in good shape.

PR #3847, opened 2024-09-27
StartTests: Rename tests & seperate func start commands from init

liliankasem:
Rename tests to follow csharp naming patterns & separate func start commands from func init commands.

There should be no functional changes in this PR other than fixing one path mistake (I'll point this out below)

Chain of PRs:
1. #3847
2. #3850 
3. #3846

PR #3846, opened 2024-09-26
Update UserSecret E2E Tests

liliankasem:
### Issue describing the changes in this PR

resolves #3833

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

- Added new queue helper
- Added storage emulator script
- Refactored user secret tests to make sure a function can still execute successfully with user secrets

> I refactored a lot of the CliTester/StartTests but the PR was getting too big and complicated so I'm splitting this out into multiple PRs.

Chain of PRs:
1. #3847
2. #3850 
3. #3846

PR #3850, opened 2024-09-28
Implement measures to fix flakey tests

liliankasem:
- Implement test cleanup measure to kill processes on a given port 
- Make `Executable` class disposable by implementing `IAsyncDisposable` interface
- Add job registry to clean up child processes
- Add log watcher with timeout to check for a specific log message
- Add logic to wait for function host to be in a running state

Chain of PRs:
1. #3847
2. #3850 
3. #3846

### DIFF
--- a/src/Azure.Functions.Cli/Common/Executable.cs
+++ b/src/Azure.Functions.Cli/Common/Executable.cs
@@ -9,7 +9,7 @@ using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Common
 {
-    internal class Executable
+    internal class Executable : IAsyncDisposable
     {
         private readonly string _arguments;
         private readonly string _exeName;
@@ -18,6 +18,7 @@ namespace Azure.Functions.Cli.Common
         private readonly bool _visibleProcess;
         private readonly string _workingDirectory;
         private readonly IDictionary<string, string> _environmentVariables;
+        private bool _disposed;
 
         public Executable(
             string exeName,
@@ -28,7 +29,7 @@ namespace Azure.Functions.Cli.Common
             string workingDirectory = null,
             IDictionary<string, string> environmentVariables = null)
         {
-            _exeName = exeName;
+            _exeName = exeName ?? throw new ArgumentNullException(nameof(exeName));
             _arguments = arguments;
             _streamOutput = streamOutput;
             _shareConsole = shareConsole;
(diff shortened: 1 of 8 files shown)


kshyju:
We also have parallel tests disabled.

https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/test/Azure.Functions.Cli.Tests/xunit.runner.json#L2

Could we investigate what can be done so that it can be enabled? Could we run multiple func.exe process at same time , with unique port numbers? I think that would decrease our build time a lot.

liliankasem:
> We also have parallel tests disabled.
> 
> https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/test/Azure.Functions.Cli.Tests/xunit.runner.json#L2
> 
> Could we investigate what can be done so that it can be enabled? Could we run multiple func.exe process at same time , with unique port numbers? I think that would decrease our build time a lot.

I like this idea but I'm keeping that out of scope for this PR. It would be great if you can create a new GH issue to address this.

test/Azure.Functions.Cli.Tests/E2E/Helpers/CliTester.cs
```diff
@@ -103,36 +113,48 @@ private static async Task InternalRun(string workingDir, RunConfiguration[] runC
 
                     if (runConfiguration.ExpectExit || (i + 1) < runConfiguration.Commands.Length)
                     {
-                        var exitCode = await exe.RunAsync(logStd, logErr, timeout: runConfiguration.CommandTimeout);
-                        exitError |= exitCode != 0;
+                        exitCode = await exe.RunAsync(logStd, logErr, timeout: runConfiguration.CommandTimeout);
```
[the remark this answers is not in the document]
liliankasem: I can give it a shot. It seems like it's a delay to let some commands execute rather than waiting for the host to be "running"? I also don't know what this 'startHost' logic exists so I may look into how tests are using it and replace it with my new "WaitForHostRunning" flag in run config

test/Azure.Functions.Cli.Tests/E2E/Helpers/RunConfiguration.cs
```diff
@@ -17,8 +18,10 @@ public class RunConfiguration
         public string[] OutputDoesntContain { get; set; } = Array.Empty<string>();
         public string[] ErrorDoesntContain { get; set; } = Array.Empty<string>();
         public Action<string> PreTest { get; set; }
-        public Func<string, Process, Task> Test { get; set; }
+        public Func<string, Process, StringBuilder, Task> Test { get; set; }
         public TimeSpan CommandTimeout { get; set; } = TimeSpan.FromSeconds(40);
         public string CommandsStr => $"{string.Join(", ", Commands)}";
+        public bool WaitForRunningHostState = false;
+        public string HostProcessPort = "7071";
```
aishwaryabh, line 25: is 7071 going to be the default when running the E2E tests?

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -210,7 +207,7 @@ await CliTester.Run(new RunConfiguration[]
                         "start --port 7073 --verbose"
                     },
                     ExpectExit = false,
-                    Test = async (workingDir, p) =>
+                    Test = async (workingDir, p, _) =>
                     {
                         using (var client = new HttpClient() { BaseAddress = new Uri("http://localhost:7073") })
```
aishwaryabh, line 212: should we also try to kill 7073 when disposing? Or whatever port the process is running on?

aishwaryabh (approved):
LGTM! Just left a few questions on port process cleanup

test/Azure.Functions.Cli.Tests/E2E/Helpers/RunConfiguration.cs
liliankasem: 7071 is the default port when you run func start, unless you specify "--port" in the start command. I believe 90% of our tests use the default and the remaining 10ish % provide a different port

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
liliankasem: Yeah, it looks like I forgot to utilize "HostProcessPort" for the tests not using the default - I'll go through and update them all

aishwaryabh (approved):

satvu (dismissed):

src/Azure.Functions.Cli/Common/Executable.cs
```diff
@@ -94,48 +97,78 @@ public async Task<int> RunAsync(Action<string> outputCallback = null, Action<str
...
-                if (ex.Message == "The system cannot find the file specified")
+
+                if (timeout is null)
+                {
+                    return await exitCodeTask.ConfigureAwait(false);
+                }
+                else
                 {
-                    throw new FileNotFoundException(ex.Message, ex);
+                    var completedTask = await Task.WhenAny(exitCodeTask, Task.Delay(timeout.Value)).ConfigureAwait(false);
```
jviau, line 129: what version of .net are we on? Do you have access to [WaitAsync](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-8.0)?

test/Azure.Functions.Cli.Tests/E2E/Helpers/LogWatcher.cs
```diff
@@ -0,0 +1,43 @@
...
+using Xunit.Abstractions;
+
+namespace Azure.Functions.Cli.Tests.E2E.Helpers
+{
+    public static class LogWatcher
+    {
+        public static async Task WaitForLogOutput(StringBuilder stdout, string expectedOutput, TimeSpan timeout)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            var cancellationTokenSource = new CancellationTokenSource(timeout);
```
jviau, line 16:
```suggestion
            using var cancellationTokenSource = new CancellationTokenSource(timeout);
```

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -12,18 +11,25 @@
...
+        private int _port;
         private const string _serverNotReady = "Host was not ready after 10 seconds";
+
         public StartTests(ITestOutputHelper output) : base(output) { }
 
-        [Fact(Skip = "Flaky test")]
+        public async Task InitializeAsync()
+        {
+            // Set as default port for majority of tests
+            _port = 7071;
```
jviau, line 28: Will using static ports throughout cause issues? What if something else is holding that port open?

Maybe we can instead search for a random available port?

jviau (approved):

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
liliankasem: I was also wondering if this would be an issue - I can look into using a random open port, but I will need to update all the tests here that are referencing 7071 directly and it might be cleaner to open a new PR for that 🤔

src/Azure.Functions.Cli/Common/Executable.cs
liliankasem: I think on .net 8, let me double check and will update if i can!

jviau (approved):

PR #3861, opened 2024-10-04
Restructuring custom host to improve error handling

### LINKED ISSUE #3858: [CoreToolsHost] VS2022 ver.17.11.4 - Debug FunctionApp .NET8 -  An error occurred while running CoreToolsHost.System.NullReferenceException: Object reference not set to an instance of an object
### Version

4.90.0-inprocess

### Description

While debugging Azure Function locally I encountered an issue. Please refer attached file.

image

### Steps to reproduce

Update the Azure Function Core Toolset and then debug, application will refuse to work as normal

mattchenderson:
Resolves #3858

The existing setup allows for early exits with an error code of 0. This should be 1 to reflect the error state. These changes also add defensiveness and reduce the set of scenarios that can cause us to exit. The logic is more tolerant of the project lacking or having a malformed `local.settings.json`. My justification for this is that we should not care if `FUNCTIONS_WORKER_RUNTIME` is present or not from the custom host's perspective. That state simply implies that .NET 6 should be used. For .NET 8 on the in-process model, we hard-require the two settings are provided through `local.settings.json`. The custom host logic just needs to determine if those .NET 8 criteria are met.

Also included are some minor changes to the custom host logger to make it more consistent with the rest of the Core Tools. We should unify instead of copying, but that is outside the scope of this PR.

I have manually validated these changes for the in-process model with .NET 6 and .NET 8, both with and without `local.settings.json`. In the case that was previously early exiting with an opaque error, we now see traditional symptoms of a .NET 8 project being loaded from a .NET 6 process. The communication of that state also needs some work, but that is also outside the scope of this PR.

### Pull request checklist

### DIFF
--- a/host/src/CoreToolsHost/Logger.cs
+++ b/host/src/CoreToolsHost/Logger.cs
@@ -10,17 +10,17 @@ namespace CoreToolsHost
         /// <summary>
         /// Logs a message.
         /// </summary>
-        internal static void Log(string message)
+        internal static void Log(string message, bool includeTimeStamp = true)
         {
-            var ts = DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss.fff", CultureInfo.InvariantCulture);
-            Console.WriteLine($"[{ts}] [CoreToolsHost] {message}");
+            var timeStampPrefix = includeTimeStamp ? $"[{DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ", CultureInfo.InvariantCulture)}] " : string.Empty;
+            Console.WriteLine($"{timeStampPrefix}[CoreToolsHost] {message}");
         }
 
         internal static void LogVerbose(bool isVerbose, string message)
         {
             if (isVerbose)
             {
-                Console.WriteLine($"{message}");
+                Log(message);
             }
         }
     }
--- a/host/src/CoreToolsHost/Program.cs
+++ b/host/src/CoreToolsHost/Program.cs
@@ -9,66 +9,58 @@ namespace CoreToolsHost
     internal class Program
     {
         static bool isVerbose = false;
+        
         static async Task Main(string[] args)
         {
+            isVerbose = args.Contains(DotnetConstants.Verbose);
+
+            Logger.LogVerbose(isVerbose, "Starting CoreToolsHost");
+
+            var localSettingsJson = await LocalSettingsJsonParser.GetLocalSettingsJsonAsJObjectAsync();
+
+            if (localSettingsJson is null)
+            {
+                Logger.LogVerbose(isVerbose, "No local.settings.json file was found");
+            }
+
+            bool projectOptsIntoDotnet8 = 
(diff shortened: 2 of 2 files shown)


aishwaryabh (approved):
Thanks for making these changes! lgtm

host/src/CoreToolsHost/Logger.cs
```diff
@@ -10,17 +10,17 @@ internal static class Logger
         /// <summary>
         /// Logs a message.
         /// </summary>
-        internal static void Log(string message)
+        internal static void Log(string message, bool includeTimeStamp = true)
```
aishwaryabh, line 13: is there any scenario where we set includeTimeStamp to false? if not, maybe we should just always have it include timestamp as it's a good metric to have as well

mattchenderson: Yeah, I ended up without a scenario for overriding this. My original thinking had been:
- Ssequencing, as it looked kind of odd to me to have timestamps before some of our non-verbose startup stuff. Example pasted below. But then I waffled on it since it seems like it could be useful for those debugging setups, and you won't see it if you aren't setting the verbose flag, anyway.
- Alignment with [`Azure.Functions.Cli.Diagnostics.ColoredConsoleLogger.LogToConsole()`](https://github.com/Azure/azure-functions-core-tools/blob/a88b2c4611d25c7e0f9b33ad0f98f46219b2371b/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs#L124), which I hope this can eventually merge with. Ultimately, we can defer that alignment to when they actually are using the same logic. I did still want to change the timestamp to match, though.

With the default value set, I don't think this hurts to have the extra param here, but if we feel I should remove it, I certainly can.

```
[2024-10-04T15:55:24.361Z] [CoreToolsHost] Starting CoreToolsHost
[2024-10-04T15:55:24.375Z] [CoreToolsHost] Loading .NET 6 host
[2024-10-04T15:55:24.376Z] [CoreToolsHost] File path to load: [REDACTED FOR GITHUB]
[2024-10-04T15:55:24.380Z] [CoreToolsHost] hostfxr path:[REDACTED FOR GITHUB]
[2024-10-04T15:55:24.381Z] [CoreToolsHost] hostfxr loaded.
[2024-10-04T15:55:24.394Z] [CoreToolsHost] hostfxr initialized with [REDACTED FOR GITHUB]

                  %%%%%%
                 %%%%%%
            @   %%%%%%    @
          @@   %%%%%%      @@
       @@@    %%%%%%%%%%%    @@@
     @@      %%%%%%%%%%        @@
       @@         %%%%       @@
         @@      %%%       @@
           @@    %%      @@
                %%
                %


Azure Functions Core Tools
Core Tools Version:       4.0.6409 Commit hash: N/A +4aa43469380b37fd9ca0b7f32c3a9134b4091ed1 (64-bit)
Function Runtime Version: 4.35.4.23179
```

aishwaryabh: ok that makes sense then to include the pre-defined arg!

Issue #3864, opened 2024-10-05
Customers reporting .NET 6 being loaded for .NET 8 in-proc from within IDE tools


mattchenderson:
### Version

4.90-inprocess - specifically in IDEs

### Description

This is a spinoff of https://github.com/Azure/azure-functions-core-tools/issues/3856

Some customers are reporting that they are seeing the .NET 6 host being loaded from the latest version of the Core Tools used by IDEs. This is most easily detected by including the `--verbose` flag as an additional argument to the IDE's run command.

Please note that in general, .NET 6 is treated as the default, and moving to .NET 8 requires all of the following to be true:

- `local.settings.json` is present and contains `FUNCTIONS_WORKER_RUNTIME` set to "dotnet"
- `local.settings.json` is present and contains `FUNCTIONS_INPROC_NET8_ENABLED` set to "1"
- The project needs to reference [version 4.4.0 or later of Microsoft.NET.Sdk.Functions](https://www.nuget.org/packages/Microsoft.NET.Sdk.Functions/4.4.0)

[Here's the doc reference for that.](https://learn.microsoft.com/azure/azure-functions/functions-dotnet-class-library#updating-to-target-net-8)

### Steps to reproduce

If you are impacted by this, please reply to this issue with repro steps. At time of this posting, we do not have an active repro for the team to analyze.

When providing steps to reproduce, please confirm if the criteria described in the above description have been met.

mattchenderson:
Per [this comment](https://github.com/Azure/azure-functions-core-tools/issues/3856#issuecomment-2395064725) on the other thread, it seems one possible cause is if FUNCTIONS_INPROC_NET8_ENABLED is set to an integer 1 instead of the string "1".

mattchenderson:
Closing in favor of the parent issue. Known issues with this have been addressed and included in the latest version. For any further issues of this type, please open a new issue instead of using this one.

Issue #3856, opened 2024-10-02
VS2022 ver.17.11.4 - Debug FunctionApp .NET8 - An error occurred while running CoreToolsHost.System.Text.Json.JsonReaderException


ripamontigiorgio:
### Version

4.90-inprocess

### Description

With 4.90-inprocess version i receive this error while starting debug a Function App running on .NET8
![image](https://github.com/user-attachments/assets/9e9e1306-fb15-41d8-b7c9-379a6a7ec9c0)

The problem is related to this file "<user>\AppData\Local\AzureFunctionsTools\Releases\4.90.0-inprocess\cli_x64\func.exe".
Copying the func.exe from 4.89 folder resolve, at the moment, the problem. 

### Steps to reproduce

Use VS2022 and try to debug a Function App wiht .NET8 inprocess

izodmungo:
+1

Different symptom same fix.

```
[2024-10-03T14:19:47.582Z] Error configuring services in an external startup class.
[2024-10-03T14:19:47.584Z] Error configuring services in an external startup class. : Could not load file or assembly 'Microsoft.Extensions.Configuration, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
[2024-10-03T14:19:47.617Z] A host error has occurred during startup operation '15e70146-e57f-4a11-af78-fe7ed653bd86'.
[2024-10-03T14:19:47.619Z] Microsoft.Azure.WebJobs.Script: Error configuring services in an external startup class. : Could not load file or assembly 'Microsoft.Extensions.Configuration, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
Value cannot be null. (Parameter 'provider')
```

The directory at `%userprofile%\appdata\Local\AzureFunctionsTools\Releases\4.90.0-inprocess\cli_x64` wasn't bare but appears to be missing most of the required files. Bad deploy of 4.90.0-inprocess? 

Another dev on my team got OP's error due to comments in local.settings.json removing and got the error I was seeing.

Skymogul:
We are experiencing a variation of this issue. When debugging a .Net 8 in-process function in Visual Studio, it appears to call `func.exe` from the `in-proc6` folder instead of the `in-proc8` folder, resulting in the same error. If we rename `in-proc8` to `in-proc6` it works as expected.

tiwilson7:
+1

[...]

PR #3866, opened 2024-10-06
Support comments and int values of FUNCTIONS_INPROC_NET8_ENABLED

mattchenderson:
This addresses two known issues with the custom host version of the Core Tools.

Previous versions of the Core Tools tolerated JSON comments just fine, but the custom host would break on them. This PR adds tolerance for JSON comments.

It also appears that FUNCTIONS_INPROC_NET8_ENABLED could be set with an integer value rather than a string in previous versions. Although we document this as being a string, the system did not enforce this until the custom host version of the Core Tools. This PR removes that enforcement and re-allows integer values.

I've verified these through local testing.

These changes are connected to https://github.com/Azure/azure-functions-core-tools/issues/3856 and https://github.com/Azure/azure-functions-core-tools/issues/3864. It is not known at present if this is sufficient to resolve them. We are awaiting other known error cases to be reported with repro steps.

### Pull request checklist

### DIFF
--- a/host/src/CoreToolsHost/LocalSettingsJsonParser.cs
+++ b/host/src/CoreToolsHost/LocalSettingsJsonParser.cs
@@ -23,7 +23,7 @@ namespace CoreToolsHost
                 }
                 if (!string.IsNullOrEmpty(fileContent))
                 {
-                    var localSettingsJObject = JsonDocument.Parse(fileContent);
+                    var localSettingsJObject = JsonDocument.Parse(fileContent, new JsonDocumentOptions { CommentHandling = JsonCommentHandling.Skip });
                     return localSettingsJObject;
                 }
             }
--- a/host/src/CoreToolsHost/Program.cs
+++ b/host/src/CoreToolsHost/Program.cs
@@ -47,11 +47,10 @@ namespace CoreToolsHost
 
         private static bool ElementExistsWithValue(JsonElement element, string key, string value)
         {
-
             return !string.IsNullOrEmpty(key) && !string.IsNullOrEmpty(value)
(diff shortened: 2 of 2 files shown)


host/src/CoreToolsHost/Program.cs
```diff
@@ -47,11 +47,10 @@ static async Task Main(string[] args)
 
         private static bool ElementExistsWithValue(JsonElement element, string key, string value)
         {
-
             return !string.IsNullOrEmpty(key) && !string.IsNullOrEmpty(value)
                 && element.TryGetProperty(key, out JsonElement property) 
-                && !string.IsNullOrEmpty(property.GetString()) 
-                && property.GetString() == value;
+                && !string.IsNullOrEmpty(property.ToString())
```
mattchenderson, line 52: This line could probably be removed, but I'm happy to leave it here just for extra defensiveness.

host/src/CoreToolsHost/LocalSettingsJsonParser.cs
```diff
@@ -23,7 +23,7 @@ internal static async Task<JsonDocument> GetLocalSettingsJsonAsJObjectAsync()
                 }
                 if (!string.IsNullOrEmpty(fileContent))
                 {
-                    var localSettingsJObject = JsonDocument.Parse(fileContent);
+                    var localSettingsJObject = JsonDocument.Parse(fileContent, new JsonDocumentOptions { CommentHandling = JsonCommentHandling.Skip });
```
drdamour, line 26: can we get AllowTrailingCommas = true on this too to match prior behaviour plz? @mattchenderson

PR #3879, opened 2024-10-10
Core tools release updates (trigger custom host after code mirror, sign func.exe and nethost.dll, and remove inproc subfolders for minified default version)

aishwaryabh:
This PR addresses the following changes:
   - We want the custom host pipeline to trigger after code mirror automatically,
   - Pivot to 1ES template for custom host pipeline
   - Sign func.exe and nethost.dll in the custom host as well,
   - Remove inproc6/inproc8 subfolders in minified default versions

### Pull request checklist


eng/ci/build-core-tools-host-artifacts-windows.yml
```diff
@@ -1,18 +1,7 @@
...
-      - ./host/src/**
-
-pr:
-  branches:
-    include:
-    - feature/oop-host
-  paths:
-    include:
-      - ./host/src/**
+    - feature/*
```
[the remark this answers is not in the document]
aishwaryabh: yeah I thought so too but when I had the previous changes the pipeline wasn't triggering automatically which is why I removed this to see if it would work. Any suggestions for why the custom host build wasn't trigerred?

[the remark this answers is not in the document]
mattchenderson: I wonder if this might just be a syntax issue on the path. Wildcards seem to be limited to a single asterisk from this doc, but I don't know the double asterisk is defined elsewhere: https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml#wildcards

I'm also not sure about using `feature/*` here. We really only need this for `feature/oop-host`, so we should leave this with the full name, I think.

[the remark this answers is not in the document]
mattchenderson: Also should we include an explicit `pr: none` similar to the official build?

eng/ci/build-core-tools-host-artifacts-windows.yml
jviau, line 1: This pipeline needs to extend 1ES templates otherwise it will be flagged as non-compliant.

eng/ci/build-core-tools-host-artifacts-windows.yml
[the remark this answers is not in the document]
aishwaryabh: Ok I removed all `**` so hopefully the build will get triggered here. I've instead added all the possible paths that we can modify to have the build trigger.

I've also only restricted this to `feature/oop-host` and added an explicit `pr: none`

eng/ci/build-core-tools-host-artifacts-windows.yml
```diff
@@ -1,89 +1,42 @@
...
+trigger:
   branches:
     include:
     - feature/oop-host
   paths:
     include:
-      - ./host/src/**
+      - /host/src/*
+      - /host/src/*/*
+      - /host/src/*/*/*
```
jviau, line 11: does `/host/src/**` not work?

eng/ci/build-core-tools-host-artifacts-windows.yml
```diff
@@ -1,89 +1,42 @@
...
+      os: windows
+    sdl:
+      codeql:
+         compiled:
+           enabled: true
+         runSourceLanguagesInSourceAnalysis: true
+    stages:
+      - stage: BuildCoreToolsHost
+        jobs:
+        - template: /eng/ci/templates/official/jobs/build-core-tools-host.yml@self
```
jviau, line 42: You don't _have_ to split this into a separate template. But also fine if you do. It is done that way in other pipelines for organization.

eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -0,0 +1,57 @@
...
+  displayName: '[Windows] Build CoreToolsHost'
+  pool:
+    name: 1es-pool-azfunc
+    image: 1es-windows-2022
+    os: windows
+
+  steps:
+  - pwsh: |
+      Import-Module "./pipelineUtilities.psm1" -Force
+      Install-Dotnet
```
jviau, line 12: Use the `UseDotnet` task

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines

code-mirror.yml
```diff
@@ -9,6 +9,7 @@ trigger:
     - release_3.0
     - release_4.0_hotfix
     - feature/*
+    - aibhandari/*
```
jviau, line 12: why do you need personal branches mirrored? If you need to test your branches in CI, you can always add the ADO mirror as a remote and manually push to it:

```
git remote add ado <ado_repo_uri>
git checkout <branch>
git push ado <branch>
```

eng/ci/build-core-tools-host-artifacts-windows.yml
aishwaryabh: no it wasn't working before :( I had it has `./host/src/**` before though so maybe that might be the reason why? I wouldn't think that the . in the beginning would change anything

eng/ci/build-core-tools-host-artifacts-windows.yml
aishwaryabh: done

eng/ci/build-core-tools-host-artifacts-windows.yml
aishwaryabh: Changed to `/host/src/**`

code-mirror.yml
aishwaryabh: done

eng/ci/templates/official/jobs/build-core-tools-host.yml
aishwaryabh: done

eng/ci/build-core-tools-host-artifacts-windows.yml
```diff
@@ -1,89 +1,40 @@
...
 
-pr:
+trigger:
   branches:
     include:
     - feature/oop-host
   paths:
     include:
-      - ./host/src/**
+      - /host/src/**
```
fabiocav, line 9: 👍🏼

eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -0,0 +1,123 @@
...
+      version: 6.x
+    displayName: Install .NET 6
+
+  - task: DotnetCoreCLI@2
+    displayName: Dotnet Publish (win-x64)
+    inputs:
+      command: publish
+      publishWebProjects: false
+      zipAfterPublish: false
+      arguments: -c Release -r win-x64 -o $(Build.SourcesDirectory)/pkg_output/windows/win-x64
```
fabiocav, line 26: We want those builds to be self-contained. Starting with .NET 8, providing a RID will no longer trigger a self-contained build by default, so we should explicitly pass that in here and below.

aishwaryabh: added `--self-contained`

fabiocav (approved):

PR #4199, opened 2024-11-21
Adding additional command line param of sync-access-token

### LINKED ISSUE #4196: Inability to refresh triggers behind authenticated function app post deploy
### Version

4.0.6610

### Description

Unable to sync triggers on Function App protected by OpenID auth (Azure AD B2C tenant) - 401

Unable to pass a seperate access token for the refresh trigger step, as the access token is used for publish as well as the refresh step. Unable to split the refresh step off and call it seperately with a different access token. (Both would be potential solutions, although I guess so would adding another auth provider!)

### Steps to reproduce

Deploy function app with OpenID "required" auth configured. 
Tooling is unable to refresh the triggers, as is unable to auth.

Log:

[SyncTriggerStep] starting.
Waiting 60 seconds for the workers to recycle with deployed content.
[CleanUpStep] starting.
Cleaned the source packages directory.
Cleaned the result artifact directory.
Finished deployment pipeline.
FunctionHostSyncTrigger, statusCode = Unauthorized
Deployment was successful but the app appears to be unhealthy, please check the app logs.
Checking the app health...................

aneillans:

Related to the issue linked below, adds the ability to pass a different access token to be used for the sync triggers call.

resolves #4196

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -1,9 +1,7 @@
 ﻿using System;
-using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
-using Azure.Functions.Cli.Arm;
 using Azure.Functions.Cli.Common;
 using static Azure.Functions.Cli.Common.OutputTheme;
 using Azure.Functions.Cli.Interfaces;
@@ -11,9 +9,6 @@ using Colors.Net;
 using Fclp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
-using static Colors.Net.StringStaticMethods;
-using Azure.Functions.Cli.Helpers;
-using Microsoft.Identity.Client;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
--- a/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
+++ b/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
@@ -13,6 +13,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using static Colors.Net.StringStaticMethods;
 using Azure.Functions.Cli.Helpers;
+using Microsoft.Identity.Client;
 
 namespace Azure.Functions.Cli.Actions.AzureActions
 {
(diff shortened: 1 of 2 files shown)


liliankasem:
@soninaren do you mind taking a look at this

src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs
```diff
@@ -43,6 +40,10 @@ public override ICommandLineParserResult ParseArgs(string[] args)
                 .Setup<string>("access-token")
                 .WithDescription("Access token to use for performing authenticated azure actions")
                 .Callback(t => AccessToken = t);
+            Parser
+                .Setup<string>("sync-access-token")
+                .WithDescription("Access token to use for performing trigger sync azure action (if different to normal)")
```
soninaren, line 45: nit: remove `(if different to normal)`. This might cause some confusion for customers as what makes access token `different`.

src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -648,7 +648,11 @@ await RetryHelper.Retry(async () =>
                 ColoredConsole.WriteLine(GetLogMessage("Syncing triggers..."));
                 HttpResponseMessage response = null;
                 // This SyncTriggers function calls the endpoint for linux syncTriggers
-                response = await AzureHelper.SyncTriggers(functionApp, AccessToken, ManagementURL);
+                if (string.IsNullOrEmpty(SyncAccessToken))
```
soninaren, line 651: Can we do the inverse instead and leave rest of the code as is.

```c#
if (!string.IsNullOrEmpty(SyncAccessToken))
{
    AccessToken = SyncAccessToken;
}
```

liliankasem:
Sorry it's taken so long to get to this. 

The App Service Authentication layer shouldn't come into play here. The Core Tools calls ARM, which in turn calls an app admin endpoint which is not subject to user-defined auth checks.

aneillans:
> Sorry it's taken so long to get to this. 
> 
> The App Service Authentication layer shouldn't come into play here. The Core Tools calls ARM, which in turn calls an app admin endpoint which is not subject to user-defined auth checks.
>  

If that's the case, then there are some issues with the kudu deployment process (would need to go back and check my notes on the issue I was hitting at the time). 

Deployments behind fully authentication enabled do fail to correctly sync due to these calls not having the correct Auth passed.

PR #4212, opened 2024-12-13
Updating package zipping

brettsam:
The `DotNetZip` nuget package is deprecated and we need to remove support for it. On top of that, we've had a `gozip` utility that we build and maintain and that ships alongside Core Tools for the express reason of zipping packages that are mountable in Linux using `fuse-zip`, which is what is used in production.

The goal here is to move the default zipping to use a new in-proc zip utility rather than maintaining the `gozip` build and deployment long term to make things simpler.

The reason for `gozip` in the first place was that Linux executable file permissions were not being correctly applied when the zip file was created in Windows when using `System.IO.Compression.ZipArchive`. This is because of some uniqueness in the way `gozip` creates these zips and the way `fuse-zip` mounts them:

- it's explicitly [setting the file's mode to rwx](https://github.com/Azure/azure-functions-core-tools/blob/d59b84bb7d45b946722fa18c19dbd57058101270/tools/go/gozip/main.go#L146-L147) (even on windows)
- when this happens in go, it [auto-flips the "created by" byte in the zip entry header from "FAT" to "Unix"](https://github.com/golang/go/blob/master/src/archive/zip/struct.go#L328)
- when unzipped, unzippers typically (although not always?) look at this header to know whether to honor the zip entry's mode. fuse-zip, the one we use to mount zips, [looks at this to determine if it should care about the file's mode](https://bitbucket.org/agalanin/fuse-zip/annotate/89545b47f19801ca298d37fcf460e4ed311dc6c0/lib/fuseZipData.cpp?at=master#fuseZipData.cpp-202:224). If it's "FAT", it treats the modes as windows and sets it to 444 (unless it's explicitly read-only).
- the .Net zip lib doesn't do this -- or allow you access to that "created by" setting at all. It does let you set attributes... but the "created by" byte is always [dependent on the OS you're zipping from](https://github.com/dotnet/runtime/blob/main/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.Windows.cs).
- the library i [pulled into the worker](https://github.com/Azure/azure-functions-dotnet-worker/pull/529) a while ago, [SharpZibLib, does let you](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Tasks/ZipDeploy/CreateZipFileTask.cs#L154)

However, rather than pull in yet another external dependency, I decided to use `ZipArchive`, but modify the resulting stream when running on Windows to flip each file to be marked as `CreatedByUnix`.

I've also added a full test that, when running on CI, takes a zip created on Windows and mounts it on Linux using `fuse-zip` to ensure that it is runnable.

As an escape-hatch, I'm leaving `gozip` as part of the build and using the setting of `FUNCTIONS_CORE_TOOLS_USE_GOZIP` the local environment will change the zipping mechanism back. I also added telemetry so we can see usage. Once we're confident, we can remove this entirely in a future release.


src/Azure.Functions.Cli/Common/Constants.cs
```diff
@@ -63,6 +63,7 @@ internal static class Constants
         public const string UserSecretsIdElementName = "UserSecretsId";
         public const string TargetFrameworkElementName = "TargetFramework";
         public const string DisplayLogo = "FUNCTIONS_CORE_TOOLS_DISPLAY_LOGO";
+        public const string UseGoZip = "FUNCTIONS_CORE_TOOLS_USE_GOZIP";
```
aishwaryabh, line 66: should we add some documentation about this new env variable that the user can set?

test/Azure.Functions.Cli.Tests/ZipHelperTests.cs
```diff
@@ -0,0 +1,197 @@
...
+
+            // reset to default in case other tests have set this up with mocks
+            FileSystemHelpers.Instance = null;
+        }
+
+        [Fact]
+        public async Task CreateZip_Succeeds()
+        {
+            var windowsZip = await BuildAndCopyFileToZipAsync("win-x64");
+            var linuxZip = await BuildAndCopyFileToZipAsync("linux-x64");
```
liliankasem, line 31: can we include macos in these tests?

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+        public static Stream CreateDotNetZip(IEnumerable<string> files, string rootPath, IEnumerable<string> executables)
         {
-            const int defaultBufferSize = 4096;
-            var fileStream = new FileStream(zipFilePath, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite, defaultBufferSize, FileOptions.DeleteOnClose);
-            using (ZipFile zip = new ZipFile())
+            // See section 4.4.2.2 in the zip spec: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+            const byte CreatedByUnix = 3;
+
+            // Signature that defines beginning of each file metadata. See section 4.3.12 in the zip spec above.
+            const uint centralDirectorySignature = 0x02014B50;
```
aishwaryabh, line 103: nit: have these consts in a separate file to make it easier to modify/read

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+
+            // Unix file permissions
+            int UnixExecutablePermissions = Convert.ToInt32("100777", 8) << 16;
```
aishwaryabh, line 106: should this be a const too?

liliankasem (approved):
lgtm, I like that you included a feature flag as a fail safe via the environment variable + included telemetry! Woild like to see more tests for macos as well

aishwaryabh (dismissed):
LGTM! Left some nit comments :)

src/Azure.Functions.Cli/Common/Constants.cs
brettsam: I was only going to add it to release notes -- it's going to go away soon and I'd rather not go through the full documentation flow if we see it's not needed and then gets pulled away. It's basically an escape-hatch if we see an issue so customers aren't blocked.

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
brettsam: I was doing this for the opposite reason -- since they're only used in this one method, I didn't want them confusing things in another file (plus they're actually quicker to reference here).

test/Azure.Functions.Cli.Tests/ZipHelperTests.cs
brettsam: good call -- let me sync with you on that

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
brettsam: it was hard to read as a const -- i was trying to write this up in a more readable/understandable way for folks (i.e. 777 means "executable" to folks that know linux permissions already. Plus, consts aren't a huge deal in a short-lived command like this from a perf standpoint, so i was trying to make it a little more readable..

test/Azure.Functions.Cli.Tests/ProcessWrapper.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace Azure.Functions.Cli.Tests;
+
+internal static class ProcessWrapper
+{
+    public static void RunProcess(string fileName, string arguments, string workingDirectory, Action<string> writeOutput = null, Action<string> writeError = null)
+    {
+        TimeSpan procTimeout = TimeSpan.FromMinutes(3);
```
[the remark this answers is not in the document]
brettsam: Well, I didn't realize `ProcessHelper` existed :-)

Let me consolidate this so it can do what I need it to.

surgupta-msft (dismissed):

test/Azure.Functions.Cli.Tests/ZipHelperTests.cs
liliankasem: discussed resolving this in a new PR

aishwaryabh (approved):

surgupta-msft (approved):

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+            }
+        }
+
+        // Returns true if we are out of bytes
+        private static int SeekBackwardsAndRead(Stream stream, byte[] buffer)
+        {
+            if (stream.Position >= buffer.Length)
+            {
+                stream.Seek(-buffer.Length, SeekOrigin.Current);
+                stream.Read(buffer.AsSpan());
```
fabiocav, line 201: Let's switch this to `ReadExactly` to make the intent clear.

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+        }
+
+        // Assumes all bytes of signatureToFind are non zero.
+        // If the signature is found then returns true and positions stream at first byte of signature
+        // If the signature is not found, returns false
+        private static bool SeekBackwardsToSignature(Stream stream, uint signatureToFind)
+        {
+            int bufferPointer = 0;
+            uint currentSignature = 0;
+            byte[] buffer = new byte[32];
```
fabiocav, line 158: Can we add a comment here to understand why we're using this buffer size?

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+        }
+
+        // Returns true if we are out of bytes
+        private static int SeekBackwardsAndRead(Stream stream, byte[] buffer)
+        {
+            if (stream.Position >= buffer.Length)
+            {
+                stream.Seek(-buffer.Length, SeekOrigin.Current);
+                stream.Read(buffer.AsSpan());
+                stream.Seek(-buffer.Length, SeekOrigin.Current);
```
fabiocav, line 202: nit: Add a comment here to indicate we're just resetting the stream position after the read

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
fabiocav: Any real benefit of using a span here?

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
```diff
@@ -87,21 +94,117 @@ public static bool GoZipExists(out string fileLocation)
...
+                {
+                    break;
+                }
+
+                while (bufferPointer >= 0 && !signatureFound)
+                {
+                    currentSignature = (currentSignature << 8) | ((uint)buffer[bufferPointer]);
+                    if (currentSignature == signatureToFind)
+                    {
+                        signatureFound = true;
```
fabiocav, line 175: Can we bail out of the loop here?

fabiocav (approved):
Approving, but left some small comments.

fabiocav:
Would also be good to add a note to the release notes for this change.

src/Azure.Functions.Cli/Helpers/ZipHelper.cs
brettsam: We chatted about it, but yeah, `Read` and `ReadExactly` need a `Span`

fabiocav (approved):

PR #4245, opened 2025-01-31
Fix validateWorkersVersions.ps1 script

aishwaryabh:
### Issue describing the changes in this PR

With the refactor of the host, each Worker.props file has been split into `eng/build/Worker.[WorkerName].props`, rather than having the worker references directly in the csproj. The `validateWorkersVersions.ps1` script has been updated to reflect this change.

Example of running `validateWorkerVersions.ps1 -Update -HostVersion 4.1037.0`:
image
 

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **do not** need to be backported to a previous version
  * [x] Otherwise: Backport tracked by #4250 
* [x] I have added all required tests (Unit tests, E2E tests)

PR #4250, opened 2025-02-03
Fixing in-proc pipeline

aishwaryabh:

The in-proc builds for the core tools pipeline breaks, due to the architecture of the build not being specified like how it is for the main branch. This PR also addresses the CVE for DotnetZip which will be addressed by Brett's PR [here](https://github.com/Azure/azure-functions-core-tools/pull/4212)

This PR also addresses bumping up the host and worker versions as well which will be used in the next release.

### Pull request checklist

### DIFF
--- a/check-vulnerabilities.ps1
+++ b/check-vulnerabilities.ps1
@@ -1,13 +1,7 @@
-$projectPath = ".\src\Azure.Functions.Cli"
 $projectFileName = ".\Azure.Functions.Cli.csproj"
 $logFilePath = "..\..\build.log"
 $skipCveFilePath = "..\..\skipPackagesCve.json"
-if (-not (Test-Path $projectPath))
-{
-    throw "Project path '$projectPath' does not exist."
-}
 
-cd $projectPath
 
 $cmd = "restore"
 Write-Host "dotnet $cmd"
@@ -63,5 +57,3 @@ if ($logFileExists)
 {
   Remove-Item $logFilePath
 }
-
-cd ../..
--- a/check-vulnerabilities.ps1
+++ b/check-vulnerabilities.ps1
@@ -1,6 +1,7 @@
 $projectPath = ".\src\Azure.Functions.Cli"
 $projectFileName = ".\Azure.Functions.Cli.csproj"
 $logFilePath = "..\..\build.log"
+$skipCveFilePath = "..\..\skipPackagesCve.json"
 if (-not (Test-Path $projectPath))
 {
     throw "Project path '$projectPath' does not exist."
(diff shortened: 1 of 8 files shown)


surgupta-msft (dismissed):

validateWorkerVersions.ps1
```diff
@@ -59,7 +59,7 @@ if (-Not $hostVersion) {
 
 function getHostFileContent([string]$filePath) {
     $uri = "https://raw.githubusercontent.com/Azure/azure-functions-host/v$hostVersion/$filePath"
-    return removeBomIfExists((Invoke-WebRequest -Uri $uri -MaximumRetryCount 5 -RetryIntervalSec 2).Content)
+    return removeBomIfExists((Invoke-WebRequest -Uri $uri).Content)
```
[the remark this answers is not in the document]
aishwaryabh: I don't think `MaximumRetryCount` and `RetryIntervalSec` are valid parameters anymore. I have removed them in this PR [here ](https://github.com/Azure/azure-functions-core-tools/pull/4245/files#diff-4df1460f28b1b3796ba764aace917bc1aebdde2aa79089fb78ffbb8a8033b0cfL62) as well, since it fails locally :)

eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -96,9 +96,6 @@ jobs:
       TELEMETRY_INSTRUMENTATION_KEY: $(TELEMETRY_INSTRUMENTATION_KEY)
       IntegrationBuildNumber: $(INTEGRATIONBUILDNUMBER)
     displayName: 'Executing build script'
-  - pwsh: |
-      .\check-vulnerabilities.ps1
-    displayName: "Check for security vulnerabilities"
```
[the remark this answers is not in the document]
aishwaryabh: We already had the check above here: https://github.com/Azure/azure-functions-core-tools/pull/4250/files/d6f1fa3da420ea38ed29a56bac0af827f57d0a38#diff-18a89dd6b9df852b6bd8d36b0cbf980cc4aaa2ca1c4f04f4247fcd1bc0b7d3a0L86

I just removed it since we don't need double :)

surgupta-msft (approved):

check-vulnerabilities.ps1
```diff
@@ -12,22 +13,55 @@ $cmd = "restore"
...
 
-cd ../..
-
-if (!$result)
-{
-  Write-Host "Vulnerabilities found" 
-  Exit 1
-}
\ No newline at end of file
+cd ../..
```
liliankasem, line 67: Why does this need to happen at the end of the script?

liliankasem (approved):

check-vulnerabilities.ps1
aishwaryabh: I think it's because we are in the $projectPath before [here](https://github.com/Azure/azure-functions-core-tools/pull/4250/files#diff-a5e3d80ba6be3dccdb27a502f7f1a8506385ca8cd8ca94711348e7da7ca23a9eR10), so it resets the current working directory back to root of the project

liliankasem: Hmm I would argue that is not the responsability of this script, but the user of the script? The script should act in isolation and the consumer should position themselves where they need to be and reset as needed.

i.e. I should be able to invoke this script where I am and should stay where I am when the script is done.

You can also make it so that the script takes a workingDirectory param and it can use that to execute its work in that directory

aishwaryabh: Added the working directory and removing the `cd ../..` from the script

PR #4256, opened 2025-02-04
Update the debian pack script

liliankasem:
Use the debian pack script

### DIFF
--- a/publish-scripts/ubuntu/buildDEB.py
+++ b/publish-scripts/ubuntu/buildDEB.py
@@ -8,11 +8,15 @@ from string import Template
 from shared import constants
 from shared import helper
 
+# version used in url is provided from user input
+# version used for packaging .deb package needs a slight modification
+# for beta, change to tilde, so it will be placed before rtm versions in apt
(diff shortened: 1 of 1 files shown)


surgupta-msft (approved):

aishwaryabh (approved):

eng/ci/templates/official/jobs/linux-package.yml
```diff
@@ -29,33 +42,40 @@ jobs:
...
-        [
-          {
-            "keyCode": "CP-450779-Pgp",
-            "operationSetCode": "LinuxSign",
-            "parameters": [],
-            "toolName": "signtool.exe",
-            "toolVersion": "1.0"
-          }
-        ]
+    condition: eq(variables['BuildOnly'], 'false')
```
aishwaryabh, line 59: should the default value for this variable be true so that the pipeline doesn't try to sign and publish the artifact to storage everytime there's a new commit on main?

surgupta-msft (approved):

eng/ci/templates/official/jobs/linux-package.yml
liliankasem: moving script to its own PR

PR #4257, opened 2025-02-05
Suppress codeql alert for 'buildNUPKG.py'

liliankasem:
Suppress codeql alert for publish-scripts/chocolatey/buildNUPKG.py

### DIFF
--- a/publish-scripts/chocolatey/buildNUPKG.py
+++ b/publish-scripts/chocolatey/buildNUPKG.py
@@ -55,7 +55,7 @@ def preparePackage():
             wget.download(url)
 
         # get the checksums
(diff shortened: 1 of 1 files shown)


publish-scripts/chocolatey/buildNUPKG.py
```diff
@@ -55,6 +55,7 @@ def preparePackage():
             wget.download(url)
 
         # get the checksums
+        # CodeQL [SM02167] False Positive: CodeQL wrongly detected. Authorized hash algorithm (SHA-512) is used; this is defined on line 13 of this script.
```
surgupta-msft, line 58: Can we provide permalink to the code as the line number can change in future?

liliankasem: done!

PR #4201, opened 2024-11-25
Custom Host E2E Tests

aishwaryabh:
### Issue describing the changes in this PR

resolves [#3869](https://github.com/Azure/azure-functions-core-tools/issues/3869)

This PR adds E2E tests for the consolidation pipeline for the custom host. To test this out, I created two inproc 6 and inproc8 azure functions projects and I run the minified windows custom host's `func.exe` on starting the respective projects.

Here is an example of the e2e tests running in the consolidation pipeline:https://dev.azure.com/azfunc/internal/_releaseProgress?_a=release-environment-logs&releaseId=193&environmentId=195

After the test are run, we then call the ArtifactAssembler to zip up the artifacts.

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #4289, opened 2025-03-06
Fix metadata.json file

aishwaryabh:
The `metadata.json` file was not being generated correctly due to a bug in the script for parsing the visual studio files to have a `.zip` ending in the name, which is not valid anymore. This PR fixes the bug by updating the regular expression to match the correct file name.

### Pull request checklist


soninaren (approved):

kshyju:
>which is not valid anymore

Can you remind me why this is not valid anymore? since when? May be an example of the file name?

aishwaryabh:
> > which is not valid anymore
> 
> Can you remind me why this is not valid anymore? since when? May be an example of the file name?

Yeah so in [this PR](https://github.com/Azure/azure-functions-core-tools/pull/4201), I added a step in the consolidated pipeline where only zip up the VS artifacts after the testing has been done. Therefore, when this step is run, which is before the zipping up of the files, the script fails to find a file that ends with `.zip`. I just removed that filter for finding a file that ends with `.zip` to account for this change.

kshyju (approved):

PR #4318, opened 2025-03-18
Allow for the UserLogDefaultLogLevel to be set independent of system logs

### LINKED ISSUE #4295: Allow for the `UserLogDefaultLogLevel` to be set independent of system logs
### Description

In Core Tools today, the log level for user logs can only be set as the same value of system logs via the "default" property in host.json:

https://github.com/Azure/azure-functions-core-tools/blob/bb4c949899cd5659d6bfe8b92cc923453a2e8f88/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs#L34-L38

https://github.com/Azure/azure-functions-core-tools/blob/bb4c949899cd5659d6bfe8b92cc923453a2e8f88/src/Azure.Functions.Cli/Common/Utilities.cs#L237

And this is the only way to control the logs that user's can see when they run their application.

This is frustrating as you won't be able to see your debug logs without also seeing all of the debug logs coming from the system, and there is no way to filter those out.

### Proposal

We can fix this either by 

a) exposing a way for customers to set `UserLogDefaultLogLevel` without having to set the default log level, and/or without having to also set the `SystemLogDefaultLogLevel` to the same value. This could be through something like a flag i.e. `--userLogLevel Debug`, or environment setting. I would advise against this being a host.json setting.

b) implement a way for us to respect the settings coming from workers. For example, in dotnet-isolated, you can set the min log level in your Program.cs:

`builder.Logging.SetMinimumLevel(LogLevel.Debug);`

Is there a way we can respect this setting without having to make customers manually configure `UserLogDefaultLogLevel` as well?

I did also find that instead of setting the log level in Program.cs - you can also use appsettings.json; Core Tools(?) is respecting any appsettings.json file that is provided without any additional config, however `logLevel` in host.json is still required.

> This might be future work rather than immediate fix, but with the Core Tools refactor we are thinking about tools for each stack, and part of that could include logging management per stack.

#### Considerations

1. We often tell customers, and document, that host.json is only used to control host logs, and not worker logs. So it is confusing that you have to set the host log level to control your worker logs. I think we need a mechanism outside of host.json to control this.

2. My understanding is that this is an issue that only impacts local development so there shouldn't be need for changes in the workers

3. How do other workers outside of dni allow customers to configure their logging setttings? Does that need to be taken into account?

VineethReyya:
Added --flag userLogLevel to set UserLogDefaultLogLevel without having to set the default log level, and without having to also set the SystemLogDefaultLogLevel to the same value.

resolves #4295 

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
@@ -18,7 +18,6 @@ namespace Azure.Functions.Cli
         public const string Ci_Continuous_Integration = "CONTINUOUS_INTEGRATION";  // Travis CI, Cirrus CI
         public const string Ci_Build_Number = "BUILD_NUMBER";  // Travis CI, Cirrus CI
         public const string Ci_Run_Id = "RUN_ID"; // TaskCluster, dsari
-        public static readonly string[] ValidUserLogLevels = ["Trace", "Debug", "Information", "Warning", "Error","Critical", "None"];
 
         public LoggingFilterHelper(IConfigurationRoot hostJsonConfig, bool? verboseLogging, string userLogLevel = null)
         {
--- a/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
@@ -32,18 +32,22 @@ namespace Azure.Functions.Cli
             {
                 SystemLogDefaultLogLevel = LogLevel.Information;
             }
-            if (Utilities.LogLevelExists(hostJsonConfig, Utilities.LogLevelDefaultSection, out LogLevel logLevel))
+            if (Utilities.LogLevelExists(hostJsonConfig, Utilities.LogLevelDefaultSection, out LogLevel defaultLogLevel))
             {
-                SystemLogDefaultLogLevel = logLevel;
+                SystemLogDefaultLogLevel = defaultLogLevel;
+                UserLogDefaultLogLevel = defaultLogLevel;
             }
 
-            // Check for user log level
-            if (!string.IsNullOrEmpty(userLogLevel))
+            // set user log value to Function
+            if ( Utilities.LogLevelExists(hostJsonConfig, Utilities.LogLevelFunctionSection, out LogLevel functionLogLevel))
             {
-                if (Enum.TryParse(userLogLevel, true, out LogLevel UserLogLevel))
-                {
-                    UserLogDefaultLogLevel = UserLogLevel;
-                }
+                UserLogDefaultLogLevel = functionLogLevel;
+            }
+
+            // set user log value to --userLoglevel Flag
+            if (!string.IsNullOrEmpty(userLogLevel) && Enum.TryParse(userLogLevel, true, out LogLevel UserLogLevel))
+            {
+                UserLogDefaultLogLevel = UserLogLevel;
             }
         }
 
--- a/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
+++ b/src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
@@ -18,8 +18,9 @@ namespace Azure.Functions.Cli
         public const string Ci_Continuous_Integration = "CONTINUOUS_INTEGRATION";  // Travis CI, Cirrus CI
         public const string Ci_Build_Number = "BUILD_NUMBER";  // Travis CI, Cirrus CI
         public const string Ci_Run_Id = "RUN_ID"; // TaskCluster, dsari
+        public static readonly string[] ValidUserLogLevels = ["Trace", "Debug", "Information", "Warning", "Error","Critical", "None"];
 
-        public LoggingFilterHelper(IConfigurationRoot hostJsonConfig, bool? verboseLogging)
+        public LoggingFilterHelper(IConfigurationRoot hostJsonConfig, bool? verboseLogging, string userLogLevel = null)
         {
             VerboseLogging = verboseLogging.HasValue && verboseLogging.Value;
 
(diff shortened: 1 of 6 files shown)


VineethReyya:
@liliankasem could you review the changes when have a moment.

Thanks

test/Azure.Functions.Cli.Tests/E2E/StartTests.cs
```diff
@@ -1857,6 +1857,80 @@ await CliTester.Run(new RunConfiguration[]
...
+                    CommandTimeout = TimeSpan.FromSeconds(300),
+                },
+                new RunConfiguration
+                {
+                    Commands = new[]
+                    {
+                        $"start  --port {_funcHostPort} --userLogLevel {UserLogLevel}"
+                    },
+                    ExpectExit = false,
+                    Test = async (workingDir, p, _) =>
```
liliankasem, line 1883: I don't think this test is testing the log level - you should check to see if a debug log is present

src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
```diff
@@ -68,5 +78,12 @@ internal bool IsCiEnvironment(bool verboseLoggingArgExists)
             }
             return false;
         }
+        private void ValidateUserLogLevel(string UserLogLevel)
```
liliankasem, line 81: always new line between functions and methods

src/Azure.Functions.Cli/Diagnostics/LoggingFilterHelper.cs
```diff
@@ -68,5 +78,12 @@ internal bool IsCiEnvironment(bool verboseLoggingArgExists)
...
+        {
+            if (LoggingFilterHelper.ValidUserLogLevels.Contains(UserLogLevel, StringComparer.OrdinalIgnoreCase) == false)
+            {
+                throw new CliException($"The userLogLevel value provided, '{UserLogLevel}', is invalid. Valid values are '{string.Join("', '", LoggingFilterHelper.ValidUserLogLevels)}'.");
```
liliankasem, line 85: Maybe add that the default level for user logs is Information

VineethReyya:
> Thanks, did you also investigate option (b) in my proposal? It would be good to see if that works?
> 
> Second, can you validate this works for all language workers?
> 
> I am still on the fence between a flag or env var for this

@liliankasem 
Though the changes were made to setup the value of UserLoglevel and systemLogLevel be different by using --userLoglevel flag. It's still depending on host.json settings( logging {loglevel : debug}) and  builder.Logging.SetMinimumLevel(LogLevel.Debug); to send the debug logs from worker to coretool.

-> my observation is:  the debuglogs is filtering out before hitting to coretool. (core tool is writing the logs which are hitting).

VineethReyya:
> > Thanks, did you also investigate option (b) in my proposal? It would be good to see if that works?
> > Second, can you validate this works for all language workers?
> > I am still on the fence between a flag or env var for this
> 
> @liliankasem Though the changes were made to setup the value of UserLoglevel and systemLogLevel be different by using --userLoglevel flag. It's still depending on host.json settings( logging {loglevel : debug}) and builder.Logging.SetMinimumLevel(LogLevel.Debug); to send the debug logs from worker to coretool.
> 
> -> My observation is: the debuglogs is filtering out before hitting to coretool. (core tool is writing the logs which are hitting).
> 
> scenario2 : function app with dotnet worker runtime will start host as childprocess [lineNo 575](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs#L575) which doesnt have any logging configurations. still, it is depending on host.json file writing errors and logs to console.
> 
> Any suggestions on this?

VineethReyya:
@liliankasem 

If we set "AzureFunctionsJobHost__Logging__LogLevel__Function" to 'debug' in local.settings.json or also use as an environment variable: Its showing debug logs for dotnet worker (if we set minimum level in program.cs), dotnet, Node without depending on host.json.
but doesn't work for Python, Powershell

The changes were made to setup the value of UserLoglevel and systemLogLevel be different by using --userLoglevel flag to show logs based on loglevel is not working. it's still depending on host.json

VineethReyya:
@liliankasem added new changes and tests. could you provide review.
Thanks

liliankasem:
> @liliankasem added new changes and tests. could you provide review. Thanks

Sorry I didn't mean for you to remove the implementation for the --userLogLevel flag, just that we should decide which approach. 

@aishwaryabh @mattchenderson @soninaren - can you share thoughts on what you think here?

a) We can provide a flag for users to set on func start => `--userLogLevel`

b) We can provide an environment variable that can be set in local.settings.json or in local env => ``

c) We can use the existing host.json logging properties. We know that a `Function` log category exists already, we can utilize that here. This might be a little more complicated in code as we would want to do something like:

- if `default` is set, and `Function` is not set == set `default` value for both user and system
- if `defailt` is not set, and `Function` is set == set user log value to `Function`
- if `default is set, and `Function` is set == use `default` for system and `Function` for user logs

```json
    "logLevel": {
      "default": "Warning",
      "Host.Aggregator": "Trace",
      "Host.Results": "Information",
      "Function": "Information" // use this to set user log level
    }
```

I am thinking we do both `a` and `c` - where users can utilize the host config file or they can set the flag with a func start command as a one off debug situation. Thoughts?

@VineethReyya lets hold off on more changes until we have consensus with the team, thanks!

mattchenderson (changes requested):
I think this needs way more discussion and design before we try and land a PR here. There need to be lines around config expectations and how that config translates across environments. My personal preference at this moment would be controls as a CLI flags to suppress the system logs, actually.

VineethReyya:
> > @liliankasem added new changes and tests. could you provide review. Thanks
> 
> Sorry I didn't mean for you to remove the implementation for the --userLogLevel flag, just that we should decide which approach.
> 
> @aishwaryabh @mattchenderson @soninaren - can you share thoughts on what you think here?
> 
> a) We can provide a flag for users to set on func start => `--userLogLevel`
> 
> b) We can provide an environment variable that can be set in local.settings.json or in local env => ``
> 
> c) We can use the existing host.json logging properties. We know that a `Function` log category exists already, we can utilize that here. This might be a little more complicated in code as we would want to do something like:
> 
> * if `default` is set, and `Function` is not set == set `default` value for both user and system
> * if `defailt` is not set, and `Function` is set == set user log value to `Function`
> * if `default is set, and `Function`is set == use`default`for system and`Function` for user logs
> 
> ```json
>     "logLevel": {
>       "default": "Warning",
>       "Host.Aggregator": "Trace",
>       "Host.Results": "Information",
>       "Function": "Information" // use this to set user log level
>     }
> ```
> 
> I am thinking we do both `a` and `c` - where users can utilize the host config file or they can set the flag with a func start command as a one off debug situation. Thoughts?
> 
> @VineethReyya lets hold off on more changes until we have consensus with the team, thanks!

@liliankasem Based on your inputs, I made changes work for both 'a' and 'c'. please review when you have a moment 
Thanks for the help.

liliankasem:
> @liliankasem @aishwaryabh Following up on my previous message—could you please take a moment to review the changes? I’d appreciate your feedback when convenient

Lets put this on pause for now whilst we dicuss internally how we want to do this.

PR #4319, opened 2025-03-19
Add docker & .devcontainer

liliankasem:
- Add dockerfile + compose to make it easy to build and run the CLI for testing
- Add .devcontainer to make it easier to setup and contribute to this project

These are initial changes, we can make more as needed.


docker/Dockerfile
jviau, line 1: Dotnet SDK has native docker publish integration. Can we use that instead?

https://learn.microsoft.com/en-us/dotnet/core/containers/sdk-publish

liliankasem: We can use that, but without a custom dockerfile we don't get a long standing container, and we wont have the components we need installed to create and run dotnet apps (likely the same for other stacks too).

I'm down to use it in combination with a Dockerfile or something else to configure the image with everything you need to test and validate function apps and cli features, but as a stand alone I am not sold?

Will look into `launchProfile`

src/Cli/func/Azure.Functions.Cli.csproj
```diff
@@ -11,6 +11,7 @@
     <PackAsTool>true</PackAsTool>
     <ToolCommandName>func</ToolCommandName>
     <PackageId>Microsoft.Azure.Functions.CoreTools</PackageId>
```
liliankasem, line 13: I'd love to move away from CoreTools and change this to Cli as well

docker/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+# Run: docker-compose up --build -d && docker exec -it func-cli bash
+services:
+  func-cli:
+    build:
+      context: ..  # Go up one level to repo root
+      dockerfile: docker/Dockerfile
+      args:
+        TARGET_RUNTIME: linux-x64
+    # platform: linux/amd64 # Uncomment if you are on an M1/M2 Mac
```
liliankasem, line 9: Using the apple silicone chips AND trying to run an x64 build. No need for this is running arm64

PR #4358, opened 2025-04-10
Allow File.Open to create the logging file with the correct sharing permissions

### LINKED ISSUE #4263: Cannot access a disposed object: IServiceProvider when outputting to json
### Version

4.104.0 (is the version in the path where `func.exe` is located)
Azure Functions Core Tools
Core Tools Version:       4.0.6821 Commit hash: N/A +c09a2033faa7ecf51b3773308283af0ca9a99f83 (64-bit)
Function Runtime Version: 4.1036.1.23224

### Description

When running `func.exe host start --dotnet-isolated-debug --enable-json-output --json-output-file c:\somePath\some-file-json.pid` it fails with the following error (with `CLI_DEBUG=1`):
```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'IServiceProvider'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier serviceIdentifier, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Azure.Functions.Cli.Actions.HostActions.StartHostAction.RunAsync() in D:\a\_work\1\s\src\Azure.Functions.Cli\Actions\HostActions\StartHostAction.cs:line 437
   at Azure.Functions.Cli.ConsoleApp.RunAsync[T](String[] args, IContainer container) in D:\a\_work\1\s\src\Azure.Functions.Cli\ConsoleApp.cs:line 63
```

When the file specified exists, it works without issues. It is only when this file does not yet exist that this error occurs.
When not running in debug, so with `--dotnet-isolated` instead, this works without issues even if the file does not exist.

This is problematic, as the IDE I use (Rider) uses this flag to obtain the PID of the ~~function host~~ worker process, in order to attach the debugger. From what I can tell, the filename is generated semi-randomly by Rider, so I also don't have a workaround of ensuring this file already exists.

### Steps to reproduce

- Have an isolated worker azure function project.
- Run `func.exe host start --dotnet-isolated-debug --enable-json-output --json-output-file c:\somePath\some-file-json.pid` when that output file does not exist

weavc:
Seems to be throwing an exception when opening the json file when the provider creates new instances, File.Create seems to lock the file. 

Unsure what's changed recently to cause this, will try do some additional investigation later, but for now a custom build of this has unblocked me

resolves #4263 

### Pull request checklist

### DIFF
--- a/src/Cli/func/Diagnostics/ColoredConsoleLogger.cs
+++ b/src/Cli/func/Diagnostics/ColoredConsoleLogger.cs
@@ -41,14 +41,7 @@ namespace Azure.Functions.Cli.Diagnostics
 
             if (jsonOutputFilePath != null)
             {
-                if (!File.Exists(jsonOutputFilePath))
-                {
-                    _jsonOutputFileStream = File.Create(jsonOutputFilePath);
-                }
-                else
-                {
-                    _jsonOutputFileStream = File.Open(jsonOutputFilePath, FileMode.Append, FileAccess.Write, FileShare.ReadWrite);
-                }
+                _jsonOutputFileStream = File.Open(jsonOutputFilePath, FileMode.Append, FileAccess.Write, FileShare.ReadWrite);
 
                 _logJsonOutput = LogJsonToFile;
 


weavc:
I cant find any changes that have caused this issue, I wonder if its been there a while and we've only started to see this due to the updates in the Azure toolkit for Rider plugin. 

I would say the change is valid as `File.Create` does lock the file as its meant to, however this prevents any new instances of the `ColoredConsoleLogger` being instantiated due being unable to open the file. If the file is created before starting the host it works as intended, but if func is responsible for creating the file we see the errors outlined in #4263. 

I did try adding additional test coverage for this into the E2E tests but they seem rather sketchy at the moment, not sure if they are being reworked or this is them working as intended and I've misunderstood how they work. Mine are waiting for the timeout on each test (many of the start tests are 5 minutes each) and they ignore exceptions from asserts in the tests you pass through to `CliTester`, it just proceeds to pass the tests. Please advise if you want additional tests for this change.

jviau:
@weavc thank you for your contribution. This change looks good to me. Will wait on some input from others on the team to see if they raise any concerns.

PR #4384, opened 2025-04-20
Ycode/improve cs template error

### LINKED ISSUE #2955: Command `func templates list` returns invalid templates
The templates returned from `func templates list` are not valid as input for `func new`'s `--template` parameter.

`func templates list --language dotnet` returns (amongst others) the template 'Azure Service Bus Topic trigger'.

Using the output in `func new --template 'Azure Service Bus Topic trigger'` will result in an error `Unknown template 'AzureServiceBusTopictrigger' (Parameter 'templateName')`.

Instead `Service Bus Topic trigger` needs to be used.

Version `4.0.3971`

y-code:
The template list provided by `func templates list` confused me when I used the template names with `func function new --template` command because neither of them worked. After looking into the tool's code, I learnt they can be used only within in-process model Functions projects, and we need to specify the `--csx` option. I also learnt that the tool uses `dotnet new` templates for the newer C# templates although the `func templates list` command does list the templates from `dotnet new`.

As the tool has complexity due to the two template management systems behind the scenes, it should be more self-explanatory and provide clear guidance to users.

This PR adds notes to the tool's outputs, assuming that the most recommended approach is to start simply with `func function new` without specifying the `--template` option. There are no functional changes.

It should effectively resolve #2955

### Command 'func templates list'

The command will have two lines of notes under the C# template list.

> (Use the templates above with 'func function new ls--csx --template' command within in-process model projects.)
> (More templates are available for C#. To list those run 'func funciton new' command without '--template'.)

### Command 'func function new --template' with Template Name from 'func templates list'

The command will output the error message:

> Template 'Azure Cosmos DB trigger' is for C# script in in-process model. Retry with --csx option to use the template. Or, retry without --template option to see available templates.

### Command 'func function new --template' with Invalid Template Name

The command will output the error message:

> Unknown template 'AzureDBtrigger'. Retry without --template option to see available templates.

## Pull request checklist

### DIFF
--- a/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
@@ -1,19 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.IO.Abstractions;
-using System.Linq;
-using System.Runtime.CompilerServices;
-using System.Text.RegularExpressions;
-using System.Threading.Tasks;
+﻿using System.Text.RegularExpressions;
 using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Exceptions;
 using Azure.Functions.Cli.ExtensionBundle;
 using Azure.Functions.Cli.Extensions;
 using Azure.Functions.Cli.Helpers;
 using Azure.Functions.Cli.Interfaces;
 using Colors.Net;
 using Fclp;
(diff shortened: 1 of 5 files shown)


liliankasem:
Thanks for taking the time to work on this! We’ve decided to go in a different direction because dotnet new is now the recommended approach. It provides more flexibility, better tooling support, and aligns with our long-term plans. Introducing a new flag for csx would draw attention to a path we’re moving away from. We also have planned improvements coming that will address this issue as part of aligning list templates with dotnet new templates.

PR #4398, opened 2025-04-30
Print warning if preview version

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### Details

- Refactored the CliVersion prop so we get the full version and not the assembley version. This ensures that when `func --version` is called, the user should see 4.0.8050-preview1` and not just `4.0.8050`.
- Updated project version to add `-preview1` suffix
- Implemented `WarnIfPreviewVersion` method to notify users that they are on a preview version of the CLI. This should print when in preview when `func`, `func --version`, `func init`, `func new`, or `func start` is invoked.

Screenshot 2025-04-29 at 5 30 24 PM

### DIFF
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -400,12 +400,14 @@ namespace Azure.Functions.Cli.Actions.HostActions
         {
             await PreRunConditions();
             var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
-            
+
             // Return if running is delegated to another version of Core Tools
             if (await TryHandleInProcDotNetLaunchAsync())
             {
                 return;
             }
+
+            Utilities.WarnIfPreviewVersion();
 
             if (isVerbose || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.DisplayLogo))
             {
--- a/src/Cli/func/Azure.Functions.Cli.csproj
+++ b/src/Cli/func/Azure.Functions.Cli.csproj
@@ -20,14 +20,18 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
+    <!-- Base version components -->
     <MajorMinorProductVersion>4.0</MajorMinorProductVersion>
-    <Version>$(MajorMinorProductVersion).$(BuildNumber)</Version>
+    <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
+    <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
+
+    <!-- Final version properties -->
+    <Version>$(MajorMinorProductVersion).$(BuildNumber)-preview1</Version>
     <AssemblyVersion>$(MajorMinorProductVersion).$(BuildNumber)</AssemblyVersion>
     <FileVersion>$(MajorMinorProductVersion).$(BuildNumber)</FileVersion>
-    <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
+
+    <InformationalVersion>$(Version) Commit hash: $(CommitHash) $(IntegrationBuildNumberInfo)</InformationalVersion>
     <IntegrationBuildNumberInfo Condition="$(IntegrationBuildNumber) != ''">Integration build number: $(IntegrationBuildNumber)</IntegrationBuildNumberInfo>
-    <InformationalVersion>$(FileVersion) Commit hash: $(CommitHash) $(IntegrationBuildNumberInfo)</InformationalVersion>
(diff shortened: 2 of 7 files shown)


src/Cli/func/Common/Utilities.cs
```diff
@@ -53,6 +53,19 @@ internal static void PrintVersion()
...
+        internal static void WarnIfPreviewVersion()
+        {
+            if (!Constants.CliVersion.Contains(Constants.PreviewVersionSuffixLabel, StringComparison.OrdinalIgnoreCase))
+            {
+                return;
+            }
+
+            ColoredConsole
+                .WriteLine("You are running a preview version of Azure Functions Core Tools.".DarkYellow())
+                .WriteLine("In this preview 'python' and 'powershell' worker support is not available for linux-arm64.".DarkYellow())
```
liliankasem, line 65: maybe: `In this preview, worker support for 'python' and 'powerShell' is not available on linux-arm64.`

mattchenderson: How about something like this? "This version currently doesn't support linux-arm64 with Python or PowerShell workers."

Do we have any ability to determine if we're on that architecture for this line? I think we always want the preview line in this state, but if I use the preview on another architecture, I don't think we need the second line.

liliankasem: Hmm that's a good idea - let me see if that's possible.

src/Cli/func/Azure.Functions.Cli.csproj
```diff
@@ -20,14 +20,18 @@
...
   <PropertyGroup>
-    <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
+    <!-- Base version components -->
     <MajorMinorProductVersion>4.0</MajorMinorProductVersion>
-    <Version>$(MajorMinorProductVersion).$(BuildNumber)</Version>
+    <BuildNumber Condition=" '$(BuildNumber)' == '' ">1</BuildNumber>
+    <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
+
+    <!-- Final version properties -->
+    <Version>$(MajorMinorProductVersion).$(BuildNumber)-preview1</Version>
```
aishwaryabh, line 29: does this always append preview1 to the version number?

liliankasem: Yes, this is so that we can be intentional about it being a preview build or not. This is only going to be set in the [feature/linux-arm64](https://github.com/Azure/azure-functions-core-tools/tree/feature/linux-arm64) branch, and should be removed when we merge to main

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -400,13 +400,15 @@ public override async Task RunAsync()
...
             var isVerbose = VerboseLogging.HasValue && VerboseLogging.Value;
-            
+
             // Return if running is delegated to another version of Core Tools
             if (await TryHandleInProcDotNetLaunchAsync())
             {
                 return;
             }
 
+            Utilities.WarnIfPreviewVersion();
```
aishwaryabh, line 410: should this happen after we print the version?

liliankasem: We could, I personally preferred it being the first thing we print so it would not be missed so easily

aishwaryabh: ok! can we make it consistent then for `StartHostAction` and `HelpAction` so that it's warn if preview version first and then print out version?

aishwaryabh (approved):

src/Cli/func/Common/Utilities.cs
```diff
@@ -53,6 +53,28 @@ internal static void PrintVersion()
...
+            ColoredConsole
+                .WriteLine("You are running a preview version of Azure Functions Core Tools.".DarkYellow());
+
+            bool isLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+            Architecture arch = RuntimeInformation.ProcessArchitecture;
+
+            if (isLinux && arch == Architecture.Arm64)
+            {
+                ColoredConsole
+                    .WriteLine("This version currently doesn't support linux-arm64 with Python or PowerShell workers, or with in-process dotnet applications.".DarkYellow());
```
mattchenderson, line 72:
```suggestion
                    .WriteLine("This version of the Azure Functions Core Tools currently doesn't support linux-arm64 with Python workers, with PowerShell workers, or with .NET applications using the in-process model.".DarkYellow());
```

mattchenderson (changes requested):
I do wonder about the warning being before the actual printing of the version in effect, but I see the comment thread there. I think it is more consistent if always the first thing, which overrides my personal preference for it coming after the version is actually printed.

Small wording change requested just to make things a bit clearer for some of these actions.

mattchenderson (approved):

PR #4403, opened 2025-05-02
Update CI to package linux arm64

### LINKED ISSUE #4287: Update pipelines to package and build linux-arm64
Update build pipelines to build and pack linux-arm64.

Changes should go into the feature branch:

- https://github.com/Azure/azure-functions-core-tools/tree/feature/linux-arm64

liliankasem:

resolves #4287

### Pull request checklist

### DIFF
--- a/eng/ci/consolidate-artifacts.yml
+++ b/eng/ci/consolidate-artifacts.yml
@@ -49,7 +49,7 @@ resources:
 variables:
 - template: /ci/variables/cfs.yml@eng
 - name: archs
-  value: 'win-x64,win-x86,min.win-x64,osx-x64,linux-x64,osx-arm64,win-arm64,min.win-arm64,min.win-x86'
+  value: 'win-x64,win-x86,min.win-x64,osx-x64,linux-x64,linux-arm64,osx-arm64,win-arm64,min.win-arm64,min.win-x86'
 
 extends:
   template: v1/1ES.Official.PipelineTemplate.yml@1es
--- a/eng/scripts/ArtifactAssemblerHelpers/generateMetadataFile.ps1
+++ b/eng/scripts/ArtifactAssemblerHelpers/generateMetadataFile.ps1
@@ -8,10 +8,18 @@ $stagingCoreToolsCli = Join-Path $StagingDirectory "coretools-cli"
 $stagingCoreToolsVisualStudio = Join-Path $StagingDirectory "coretools-visualstudio"
 
 # Get OOP Artifact Version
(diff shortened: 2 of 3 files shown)


eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -146,6 +146,16 @@ jobs:
       arguments: -c Release -r linux-x64 -o $(Build.SourcesDirectory)/pkg_output/linux/linux-x64
       workingDirectory: $(Build.SourcesDirectory)/src/CoreToolsHost
 
+  # do we need to build a linux-arm64 version?
```
aishwaryabh, line 149: Is it a requirement to create a custom host with linux-arm64? If so we probably would need to publish it similar to linux-x64

Today we build win-x64, win-arm64, and linux-x64, so it would make sense to also include linux-arm64 but just wanted to double check!

eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -146,6 +146,16 @@ jobs:
       arguments: -c Release -r linux-x64 -o $(Build.SourcesDirectory)/pkg_output/linux/linux-x64
```
aishwaryabh, line 146: along with these changes, we would need to also update the artifact assembler [here](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Cli/ArtifactAssembler/ArtifactAssembler.cs#L16) to ensure that the custom host artifact is actually built

I would also just test on docker to make sure that the created artifact also works as expected!

eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -156,14 +166,23 @@ jobs:
...
+      TargetFolder: $(Build.ArtifactStagingDirectory)/_coreToolsHostPackages/linux-arm64
+
   templateContext:
     outputParentDirectory: $(Build.ArtifactStagingDirectory)
     outputs:
     - output: pipelineArtifact
       path: $(Build.ArtifactStagingDirectory)/_coreToolsHostPackages
       artifact: drop-coretools-host-linux
 
-- job: SignCoreToolsHostLinux
```
aishwaryabh, line 166: I think we can sign in the same job; I'd just created it like this to add some separation iirc

liliankasem: For windows the sign and build were in the same job so was wondering why it was separated only for linux. Not something we need to figure out in this PR but just a thought

eng/ci/templates/official/jobs/build-core-tools-host.yml
aishwaryabh: hmm ok I see. Maybe we should check with Fabio first to see if this makes sense to add, since VS uses the custom host but I'm not sure if linux-arm64 is a valid use case or not

aishwaryabh (approved):

PR #4421, opened 2025-05-15
Update the in-proc pipeline to upload artifacts to internal feed for nightly builds

### LINKED ISSUE #4420: Upload nightly builds to an internal feed and have consolidated pipeline trigger on CI build completion
We want to fix the consolidated pipeline to automatically trigger when a build is done on either the core-tools.official pipeline or the core-tools-host pipeline, rather than having to manually input which artifacts we want to run the consolidated pipeline on. This allows less room for error as well.

However we want to distinguish between nightly builds and regular builds that were triggered by CI. Therefore, we can separate the nightly build artifacts from the CI triggered artifacts by uploading the nightly build artifacts to an internal feed.

aishwaryabh:
Resolves #4420

I'm fixing the consolidated pipeline to automatically trigger when a build is done on either the core-tools.official pipeline or the core-tools-host pipeline. However we want to distinguish between nightly builds and regular builds that were triggered by CI. Therefore, we can seperate the nightly build artifacts from the CI trigerred artifacts by uploading the nightly build artifacts to an internal feed.

The format of the artifact version will be: YYYY.MMDD.buildId (where MMDD has no leading zeros).

Here is an example of the feed:
https://azfunc.visualstudio.com/internal/_artifacts/feed/core-tools-nightly-build

### Pull request checklist


eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     devops_buildNumber: $[counter(format('1'), 6100)]
     DEVOPS_REPO_BRANCH: $[coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranchName'])]
     DEVOPS_REPO_COMMIT: $(Build.SourceVersion)
+    SimulateReleaseBuild: true
```
liliankasem, line 18: What is this foor? should it come from variable or other setting?

eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -250,6 +251,25 @@ jobs:
...
+      displayName: Publish
+      inputs:
+        command: publish
+        publishDirectory: '$(Build.ArtifactStagingDirectory)'
+        vstsFeedPublish: 'internal/core-tools-nightly-build'
+        vstsFeedPackagePublish: 'drop-${{ parameters.artifactTargetFramework }}'
+        versionOption: custom
+        versionPublish: '$(NightlyBuildVersion)'
+        packagePublishDescription: 'Core tools ${{ parameters.artifactTargetFramework }} nightly build'
   - task: 1ES.PublishPipelineArtifact@1
```
liliankasem, line 273: New line between tasks

eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -250,6 +251,25 @@ jobs:
...
     displayName: 'Move artifacts'
+
+  - ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+    - task: PowerShell@2
+      displayName: 'Generate nightly build package version for feed'
+      inputs:
+        targetType: filePath
+        filePath: '$(Build.SourcesDirectory)/generateNightlyBuildPackageVersion.ps1'
+
+  - ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
```
liliankasem, line 262: Given these two tasks are after each other, i dont think you need to repeat `${{ if eq(variables['Build.Reason'], 'Schedule') }}:`

They can be both indented under the first instance of this if condition

eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -250,6 +251,25 @@ jobs:
...
+    - task: UniversalPackages@0
+      displayName: Publish
+      inputs:
+        command: publish
+        publishDirectory: '$(Build.ArtifactStagingDirectory)'
+        vstsFeedPublish: 'internal/core-tools-nightly-build'
+        vstsFeedPackagePublish: 'drop-${{ parameters.artifactTargetFramework }}'
```
liliankasem, line 269: does drop need to be in the name here? we already know its an artifact?

generateNightlyBuildPackageVersion.ps1
```diff
@@ -0,0 +1,15 @@
+# Generate a SemVer 2.0 compliant version number
+# Format: 4.0.YYYYMMDD (where YYYYMMDD has no leading zeros)
```
liliankasem, line 2: Does this need to be a YYYYMMDD - can we not use the build id of the build that generated this artifact so we have a way to back track?

aishwaryabh: I was thinking that we wanted all of the different artifacts to have the same build id to associate it with the correct nightly build that day, so I thought it would be easier to have the same version rather than different ones. Let me know if you think build id from each pipeline makes more sense though!

eng/ci/templates/official/jobs/build-test.yml
aishwaryabh: yeah so ideally we want the artifact name to be the same as whatever the consolidated pipeline uses to make sure we can store the downloaded artifacts in the same folder, so I just kept it the same structure.

eng/ci/templates/official/jobs/build-test.yml
aishwaryabh: ah good catch forgot to remove this!

generateNightlyBuildPackageVersion.ps1
liliankasem: That makes sense, but the build ID usually has the date as well no? e.g. `20250515.1`

https://azfunc.visualstudio.com/internal/_build/results?buildId=218669&view=results 

Maybe its not the build "id" but has another name?

eng/ci/templates/official/jobs/build-test.yml
liliankasem: I think it makes more sense to update the pipeline to remove "drop_" - but either way is ok

generateNightlyBuildPackageVersion.ps1
aishwaryabh: addressed to be YYYY.MMDD.buildId

eng/ci/templates/official/jobs/build-test.yml
aishwaryabh: removed drop!

generateNightlyBuildPackageVersion.ps1
```diff
@@ -0,0 +1,19 @@
...
+$year = Get-Date -Format "yyyy"
+$month = (Get-Date).Month  # No leading zero for single digit months
+$day = Get-Date -Format "dd"  # Always double digit for days
+
+# Create version: YYYY.MMDD.buildId
+$semVerVersion = "$year.$month$day.$buildId"
+Write-Host "Generated SemVer version: $semVerVersion"
+
+# Set as pipeline variables for use in subsequent tasks
+Write-Host "##vso[task.setvariable variable=NightlyBuildVersion]$semVerVersion"
```
kshyju, line 19: nit: Add one newline at the end of the file.

aishwaryabh: added!

PR #4422, opened 2025-05-15
Update the default pipeline to upload artifacts to internal feed for nightly builds and consolidated pipeline updates

aishwaryabh:

Resolves https://github.com/Azure/azure-functions-core-tools/issues/4420

I'm fixing the consolidated pipeline to automatically trigger when a build is done on either the core-tools.official pipeline or the core-tools-host pipeline. However we want to distinguish between nightly builds and regular builds that were triggered by CI. Therefore, we can seperate the nightly build artifacts from the CI trigerred artifacts by uploading the nightly build artifacts to an internal feed.

The format of the artifact version will be: 4.0.YYYYMMDD (where YYYYMMDD has no leading zeros). We assume that nightly builds only run once a day, so each day there will be a unique version that is uploaded to the feed.

Here is an example of the feed:
https://azfunc.visualstudio.com/internal/_artifacts/feed/core-tools-nightly-build

This PR also conditionally downloads the artifacts from the nightly build feed in the consolidated pipeline, if the consolidated pipeline run is also only a nightly build. Otherwise, we just take whatever was triggered by CI and use that build.

### Pull request checklist


eng/ci/templates/official/jobs/build-core-tools-host.yml
```diff
@@ -116,6 +116,25 @@ jobs:
...
       TargetFolder: $(Build.ArtifactStagingDirectory)/_coreToolsHostPackages/win-arm64
 
+  - ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+    - task: PowerShell@2
+      displayName: 'Generate nightly build package version for feed'
+      inputs:
+        targetType: filePath
+        filePath: '$(Build.SourcesDirectory)/eng/scripts/generateNightlyBuildPackageVersionForFeed.ps1'
+
+  - ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
```
liliankasem, line 126: Same comment as the other pr about this being no required

eng/scripts/generateNightlyBuildPackageVersionForFeed.ps1
```diff
@@ -0,0 +1,15 @@
+# Generate a SemVer 2.0 compliant version number
```
liliankasem, line 1: i think you have this script in the other pr?

aishwaryabh: yeah this is the in-proc branch so it won't show up in the other PR :)

liliankasem: Ah sorry missed it was in-proc. But yeah same comment about the version name + build #

eng/ci/templates/official/jobs/assemble-artifacts.yml
```diff
@@ -3,32 +3,35 @@ parameters:
...
+          pipeline: core-tools-host
+          artifactName: drop-coretools-host-linux-signed
+          targetPath: $(Pipeline.Workspace)/core-tools-host/drop-coretools-host-linux-signed
+        - input: pipelineArtifact
+          pipeline: core-tools-host
+          artifactName: drop-coretools-host-windows
+          targetPath: $(Pipeline.Workspace)/core-tools-host/drop-coretools-host-windows
+        - input: pipelineArtifact
+          pipeline: core-tools-inproc
+          artifactName: drop-inproc6
```
liliankasem, line 25: Thoughts on this naming pattern?

##### host

Is there a linux unsigned for us to have `signed` in the name? is windows unsigned? if yes, why are these 2 inconsistent?

`coretools-host-linux` 
`coretools-host-windows`

##### cli

`coretools-cli`
`coretools-cli-inproc6`
`coretools-cli-inproc8`

I actually have hopes of us removing CoreTools and replacing it with CLI so we end up with something like `func-cli` / `func-cli-inproc6` and `func-cli-host-windows` but maybe it's too soon to start doing that? if not we should follow the cli name pattern :)

aishwaryabh: yeah so linux has an unsigned version which is why I specified signed. I can update the final artifact names for core tools host to be the following:
`coretools-host-linux-signed` and `coretools-host-windows-signed`

liliankasem: > linux has an unsigned version 

Why is this the case? Can we sign before we publish like we do with windows?

eng/ci/build-core-tools-host-artifacts.yml
```diff
@@ -1,3 +1,11 @@
+schedules:
```
liliankasem, line 1: Wondering why this was renamed to add -windows? Can't we run this pipeline with a windows job and linux job like we do today?

aishwaryabh: oh I removed -windows? This pipeline was initially named `build-core-tools-host-artifacts-windows.yml` but I removed the windows to make it less confusing?

kshyju (approved):

PR #4423, opened 2025-05-15
Exclude python worker for win-arm64 builds & remove ArmAuthenticationOptions schema

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/build/BuildSteps.cs
+++ b/build/BuildSteps.cs
@@ -193,7 +193,9 @@ namespace Build
         public static void FilterPythonRuntimes()
         {
             var minifiedRuntimes = Settings.TargetRuntimes.Where(r => r.StartsWith(Settings.MinifiedVersionPrefix));
-            foreach (var runtime in Settings.TargetRuntimes.Except(minifiedRuntimes))
+            var unsupportedPythonRuntimes = Settings.UnsupportedPythonRuntimes;
+
+            foreach (var runtime in Settings.TargetRuntimes.Except(minifiedRuntimes).Except(unsupportedPythonRuntimes))
             {
                 var pythonWorkerPath = Path.Combine(Settings.OutputDir, runtime, "workers", "python");
                 var allPythonVersions = Directory.GetDirectories(pythonWorkerPath);
--- a/src/Cli/func/Actions/HostActions/Startup.cs
+++ b/src/Cli/func/Actions/HostActions/Startup.cs
@@ -68,10 +68,10 @@ namespace Azure.Functions.Cli.Actions.HostActions
             }
             else
             {
+                // ArmAuthenticationOptions was removed with host v4.1038.300 - we need to double check the impact of this.
                 services.AddAuthentication()
                     .AddScriptJwtBearer()
-                    .AddScheme<AuthenticationLevelOptions, CliAuthenticationHandler<AuthenticationLevelOptions>>(AuthLevelAuthenticationDefaults.AuthenticationScheme, configureOptions: _ => { })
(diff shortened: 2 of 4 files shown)


build/BuildSteps.cs
```diff
@@ -193,7 +193,9 @@ public static void FilterPowershellRuntimes()
         public static void FilterPythonRuntimes()
         {
             var minifiedRuntimes = Settings.TargetRuntimes.Where(r => r.StartsWith(Settings.MinifiedVersionPrefix));
-            foreach (var runtime in Settings.TargetRuntimes.Except(minifiedRuntimes))
+            var unsupportedPythonRuntimes = Settings.UnsupportedPythonRuntimes;
```
liliankasem, line 196: I think this is a safe change, win-arm64 is not supported in python so not having a directory for this is OK.

src/Cli/func/Actions/HostActions/Startup.cs
```diff
@@ -68,10 +68,10 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             }
             else
             {
+                // ArmAuthenticationOptions was removed with host v4.1038.300 - we need to double check the impact of this.
                 services.AddAuthentication()
                     .AddScriptJwtBearer()
-                    .AddScheme<AuthenticationLevelOptions, CliAuthenticationHandler<AuthenticationLevelOptions>>(AuthLevelAuthenticationDefaults.AuthenticationScheme, configureOptions: _ => { })
-                    .AddScheme<ArmAuthenticationOptions, CliAuthenticationHandler<ArmAuthenticationOptions>>(ArmAuthenticationDefaults.AuthenticationScheme, _ => { });
```
liliankasem, line 74: I think need to understand the implications of this

aishwaryabh (approved):

src/Cli/func/Actions/HostActions/Startup.cs
```diff
@@ -68,10 +68,10 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
+                    .AddScheme<AuthenticationLevelOptions, CliAuthenticationHandler<AuthenticationLevelOptions>>(AuthLevelAuthenticationDefaults.AuthenticationScheme, configureOptions: _ => { });
```
mathewc, line 74: Well, in https://github.com/Azure/azure-functions-host/pull/10654 we removed this auth mechanism from the host. Looks like this line was overriding the previous default ArmAuthenticationHandler that used to exist, which validated tokens passed via x-ms-site-restricted-token. The CLI override this apparently and was always always returning an Admin principal ([here](https://github.com/Azure/azure-functions-core-tools/blob/0e387d75242b291b436027f1115f894bc9f369be/src/Cli/func/Actions/HostActions/WebHost/Security/CliAuthenticationHandler.cs#L18));

liliankasem: Thanks for calling this out! 

> Likely as before, the intent is for that default handler to be overridden such that Admin level is granted locally?

Yes that's correct. We should also already have E2E tests for this flow so I would expect the CI to fail if this is broken now. 

Let me test this out and validate with a local build just in case.

mathewc: Actually, it's probable that this entire else block may be dead code, since below on line 78 you can see there's a catch all handler that is granting Admin to all requests when _enableAuth==false. That would take precedence I believe.

liliankasem: I've validated that this is still the behaviour we have with this change

castrodd (approved):

PR #4426, opened 2025-05-16
Add job attempt # to test log artifact name

liliankasem:
Add job attempt # to test log artifact name to avoid this error on a build retry:

`##[error]Artifact TestLogs already exists for build 218830.`

### DIFF
--- a/eng/ci/templates/public/jobs/build-test-public.yml
+++ b/eng/ci/templates/public/jobs/build-test-public.yml
@@ -66,7 +66,7 @@ jobs:
     condition: succeededOrFailed()
     inputs:
       targetPath: '$(Build.SourcesDirectory)/TestLogs'
-      artifactName: 'TestLogs'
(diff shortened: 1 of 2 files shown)


eng/ci/templates/public/jobs/build-test-public.yml
```diff
@@ -66,7 +66,7 @@ jobs:
     condition: succeededOrFailed()
     inputs:
       targetPath: '$(Build.SourcesDirectory)/TestLogs'
-      artifactName: 'TestLogs'
+      artifactName: 'TestLogs_$(System.JobAttempt)'
```
aishwaryabh, line 69: should we also add:
 continueOnError: true
 
just to ensure that this job failing doesn't block pipeline runs

aishwaryabh (approved):

aishwaryabh (approved):

PR #4432, opened 2025-05-21
Update CI build names to include timestamp and build rev #

liliankasem:
Update CI build names to include timestamp and build rev #

### DIFF
--- a/eng/ci/consolidate-artifacts.yml
+++ b/eng/ci/consolidate-artifacts.yml
@@ -6,7 +6,7 @@ schedules:
     include:
       - main
 
-name: $(Build.SourceBranchName)_$(Build.Reason)
+name: $(Date:yyyyMMdd)$(Rev:.r)-$(Build.SourceBranchName)-$(Build.Reason)
 
 trigger: none
 
--- a/eng/ci/linux-build.yml
+++ b/eng/ci/linux-build.yml
@@ -1,4 +1,4 @@
-name: $(Build.SourceBranchName)_$(Build.Reason)
+name: $(Date:yyyyMMdd)$(Rev:.r)-$(Build.SourceBranchName)-$(Build.Reason)
 
 pr: none
 
--- a/eng/ci/official-build.yml
+++ b/eng/ci/official-build.yml
@@ -6,7 +6,7 @@ schedules:
       - main
   always: true
 
-name: $(Build.SourceBranchName)_$(Build.Reason)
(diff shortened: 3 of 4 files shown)


publish-scripts/npm/package.json
```diff
@@ -12,7 +12,7 @@
   "preferGlobal": true,
   "repository": {
     "type": "git",
-    "url": "https://github.com/Azure/azure-functions-core-tools.git"
+    "url": "git+https://github.com/Azure/azure-functions-core-tools.git"
```
aishwaryabh, line 15: why do we need git+ in front of the url?

aishwaryabh (approved):

publish-scripts/npm/package.json
liliankasem: I think thats just the format npm wants. When publishing in NPM it gets corrected every time, so I just fixed it permenantly:

```log
2025-05-21T23:17:45.4582051Z npm verbose publish [ '.' ]
2025-05-21T23:17:45.4582689Z npm warn publish npm auto-corrected some errors in your package.json when publishing.  Please run "npm pkg fix" to address these errors.
2025-05-21T23:17:45.4583048Z npm warn publish errors corrected:
2025-05-21T23:17:45.4583401Z npm warn publish "repository.url" was normalized to "git+https://github.com/Azure/azure-functions-core-tools.git"
2025-05-21T23:17:45.4583726Z npm notice
```

liliankasem:
> Not sure if you want to port the MSBuild targets I have in the host and other repos we own. Those targets will set the build name for you.

Yeah I do want to bring those in at some point. I also really want to include the version in the name e.g.

`4.0.7030-20250522.1`

The other thing I am questioning is the versioning mechanism here, it's automated and different for every CI. We end up with in-proc being a different version to out of proc e.g.:

```
// default version
Core Tools Version:       4.0.7030 Commit hash: N/A +bb4c949899cd5659d6bfe8b92cc923453a2e8f88 (64-bit)
Function Runtime Version: 4.1037.0.23568

// version for running inproc host
Core Tools Version:       4.0.7032 Commit hash: N/A +ec5981453def131abe6005327168b918fac8d426 (64-bit)
Function Runtime Version: 4.634.3.22875
```

And I am not sure that these need to be different. Would love to brainstorm some ideas with you both on how we can make versioning make sense, and easily coordinate between the many different pipelines we run (core /main; core /in-proc; host; consolidated; linux build etc.)

@aishwaryabh @jviau

PR #4364, opened 2025-04-14
Adding `func start` tests

aishwaryabh:
### Issue describing the changes in this PR

resolves #4328

This PR is the last part of a series of PRs to set up a new testing framework for the Azure Functions CLI. This is the [larger PR](https://github.com/Azure/azure-functions-core-tools/pull/4360), which contains all the changes for the testing framework and the updated func start E2E test, if you would like to take a look at how it all fits in! I just decided to split it up since it would be easier to review and get feedback this way.

`Func.E2ETests.csproj` contains the func start E2E tests that are refactored to use the new test framework. The tests utilize fixtures whenever possible, with the base class being `BaseFunctionAppFixture`. The rest of the tests don't use a fixture and have to set up a function app with func init and func new for each test. Each test is also in its own unique working directory and using a unique port to avoid port conflicts.

**CI changes**:
The yaml files have been updated to publish the TestLogs (the logs are dumped into a txt file and will show up even if the tests timeout). The Build.csproj has been updated to run the new Func.E2ETests.csproj as well.

Note the func start tests in the old framework have been removed in this PR as well. I've also added a timeout of 10min so that if a test is hanging, CI will timeout and publish the logs to the TestLogs folder.

**Performance Impact**
After running these tests on CI, the test time for func start E2E tests have reduced by **more than 50%**! With the testing method, the tests would take 19 min and the newer testing method takes 7-8 min!

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #4447, opened 2025-05-28
Update Execute method for ProcessStartedHandler to prevent stalled tests

### LINKED ISSUE #4392: Add a timeout to `ProcessStartedHandler`
After merging in [this PR](https://github.com/Azure/azure-functions-core-tools/pull/4364), when the tests are run locally today for core tools for `func start`, if the process starts up and is not interrupted within the test by the processStartedHandler (example: we expect the test to fail and throw an error before the host starts up, but it doesn't so the test is stuck since the host starts up successfully and is waiting for the user to manually kill the process), the test stalls, which is a flaw in the design.

To workaround this, a timeout would need to be added to the processTask [here](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Cli/Abstractions/Command/Command.cs#L61). If the timeout is reached before `processStarted` is finished, we have to manually kill the process by doing `_process.Kill(true)`

aishwaryabh:

resolves #4392 

After merging in https://github.com/Azure/azure-functions-core-tools/pull/4364, when the tests are run locally today for core tools for func start, if the process starts up and is not interrupted within the test by the processStartedHandler (example: we expect the test to fail and throw an error before the host starts up, but it doesn't so the test is stuck since the host starts up successfully and is waiting for the user to manually kill the process), the test stalls, which is a flaw in the design.

To workaround this, a timeout would need to be added to the processTask [here](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Cli/Abstractions/Command/Command.cs#L61). If the timeout is reached before processStarted is finished, we have to manually kill the process by doing _process.Kill(true)

### Pull request checklist


src/Cli/Abstractions/Command/Command.cs
```diff
@@ -62,14 +62,33 @@ public CommandResult Execute(Func<Process, Task>? processStarted, StreamWriter?
                     {
                         try
                         {
-                            await processStarted(_process);
+                            var processStartedTask = processStarted(_process);
+                            var timeoutTask = Task.Delay(TimeSpan.FromMinutes(2));
```
liliankasem, line 66: Can we use a retry or timeout helper here instead? for example something like what we do with  `await RetryHelper.Retry(`

aishwaryabh: added a timeout helper! lmk if it looks ok

liliankasem (approved):

src/Cli/Abstractions/Command/Command.cs
```diff
@@ -62,13 +63,37 @@ public CommandResult Execute(Func<Process, Task>? processStarted, StreamWriter?
                     {
                         try
                         {
-                            await processStarted(_process);
+                            // Try to run the process started handler with a timeout of 2 minutes
+                            await TimeoutHelper.RunWithTimeoutAsync(
+                                processStarted(_process),
```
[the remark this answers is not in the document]
aishwaryabh: I am using a cancellation token in the timeout helper already

PR #4453, opened 2025-06-02
 Wrap every command in a retry helper to improve test runs in CI

aishwaryabh:
#4458 

This PR wraps every command in the E2E test framework in a retry helper to improve test runs in CI.
### Pull request checklist


liliankasem:
Why do they need to be wrapped in a retry? After how many retries are these tests passing? I think this points to an underlying issue we should investigate - it shouldn't be flaky to this extend no?

jviau:
Not sure if you have looked into Polly: https://github.com/App-vNext/Polly -- it is a great library for building retry pipelines. But also agree with @liliankasem about understanding the underlying reason these tests are flaky.

PR #4463, opened 2025-06-06
Implement GoZipTool as NoTargets csproj

liliankasem:
Implement GoZipTool as NoTargets csproj


src/GoZipTool/GoZipTool.csproj
```diff
@@ -0,0 +1,54 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ResolvedRuntimeIdentifier Condition="'$(RuntimeIdentifier)' != ''">$(RuntimeIdentifier)</ResolvedRuntimeIdentifier>
+  </PropertyGroup>
+
+  <Target Name="SetGoZipPath">
```
jviau, line 12: Consider adding some `Inputs` and `Outputs` here to support incremental build.

src/GoZipTool/GoZipTool.csproj
```diff
@@ -0,0 +1,54 @@
...
+      <GoZipSource>gozip/main.go</GoZipSource>
+    </PropertyGroup>
+
+    <Exec
+      Command="go build -o $(GoZipPath) $(GoZipSource)"
+      EnvironmentVariables="GOOS=$(GoOS);GOARCH=$(GoArch);CGO_ENABLED=0" />
+  </Target>
+
+  <Target Name="CollectGoZipOutput" DependsOnTargets="SetGoZipPath;BuildGoZip" BeforeTargets="AssignTargetPaths" Condition="'$(ResolvedRuntimeIdentifier)' != ''">
+    <Message Importance="high" Text="Collecting GoZip output: $(GoZipPath)" />
```
jviau: Small polish nit: how does this work with terminal logger? Does it pretty print?

image

jviau (approved):

src/GoZipTool/GoZipTool.csproj
liliankasem: Is not pretty, and only shows on `-v:D`

Screenshot 2025-06-09 at 10 45 55 AM

Which is interesting because with High importance, I shouldn't need to set verbosity to detailed to see it

>Typically, a message set to high is sent when logger verbosity is set to [LoggerVerbosity](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).Minimal or higher. A message set to low is sent when logger verbosity is set to [LoggerVerbosity](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).Detailed.

liliankasem: Let me know if any action is needed beside changing importance to low

jviau: Nothing needed, looks fine to me. I am slightly curious on how to get the GoZipTool.csproj to output a line like:

`GoZipTool succeeded (3.0s) -> /<path>/GoZipTool/debug/gozip`

But that isn't important, or critical to this project. Definitely don't need to spend time figuring that out.

eye twitch

PR #4466, opened 2025-06-06
Move publish scripts to tools/publish-tools

liliankasem:
Moving publish scritps to eng/ fodler where this makes more sense.

We need to update the release pipeline to reflect the new path.


eng/tools/publish-tools/chocolatey/nuspec_template
```diff
@@ -7,7 +7,7 @@
     <authors>Microsoft</authors>
     <owners>nugetazurefunctions</owners>
     <projectUrl>https://github.com/Azure/azure-functions-core-tools</projectUrl>
-    <iconUrl>https://raw.githubusercontent.com/Azure/azure-functions-core-tools/refs/heads/main/publish-scripts/npm/assets/azure-functions-logo-color-raster.png</iconUrl>
+    <iconUrl>https://raw.githubusercontent.com/Azure/azure-functions-core-tools/refs/heads/main/eng/res/functions.png</iconUrl>
```
jviau: The icon url is going to break for old packages because you are moving this file. Can you include the icon in the nupkg itself? That is the recommended practice now. This will prevent this issue from occurring in the future.

liliankasem: Good point, we do include the icon in the nupkg already I think, so it should be safe to just update this reference. Will look into it

jviau (approved):

aishwaryabh (approved):

PR #4467, opened 2025-06-06
Implement new CI pipelines that will build, sign, test & publish the CLI

liliankasem:
Implement new CI pipelines that will build, sign, test & publish the CLI.

This PR makes it so that we no longer need to use the build.csproj to build test and publish the CLI. Instead, this is achieved through build targets, and updated CI pipelines. The new CI pipelines also parallelize the building and testing of each runtime that we support, making the build process x3 faster than it was before.

Some of the CI pipelines have been renamed so we will need to update their reference in the Azure DevOps UI.

Passing CI:

- https://dev.azure.com/azfunc/public/_build/results?buildId=222529&view=results
- https://azfunc.visualstudio.com/internal/_build/results?buildId=222528&view=results


eng/ci/templates/jobs/test-e2e-windows.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  pool:
+    name: 1es-pool-azfunc-public
+    image: 1es-windows-2022
+    os: windows
+
+  # Need to find an arm64 image/agent for windows
+  strategy:
+    matrix:
+      dotnet_win_x64:
```
aishwaryabh, line 17: should we also include win-x86 in the matrix too?

eng/ci/templates/official/jobs/publish-cli.yml
```diff
@@ -0,0 +1,137 @@
...
+    image: 1es-windows-2022
+    os: windows
+
+  variables:
+  - template: /eng/ci/templates/official/variables/authenticode-binaries-tosign.yml@self
+  - template: /eng/ci/templates/official/variables/thirdparty-binaries-tosign.yml@self
+  - template: /eng/ci/templates/official/variables/mac-binaries-tosign.yml@self
+
+  steps:
+  - ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
```
aishwaryabh: the first few steps for publish-cli and pack-cli are very similar. Is it possible to combine the templates for them (just for the first few steps)

eng/ci/templates/official/jobs/publish-cli.yml
```diff
@@ -0,0 +1,137 @@
...
+    displayName: 'Publish & Zip CLI'
+
+  - ${{ if contains(parameters.runtime, 'osx') }}:
+    - template: /eng/ci/templates/official/steps/sign-mac.yml@self
+      parameters:
+        folderPath: '$(Build.Repository.LocalPath)/artifacts'
+        signPatterns: '*.osx-*.zip'
+
+  # arm64 MSI signing is not supported yet, so we skip it for now.
+  - ${{ if and(contains(parameters.runtime, 'win'), not(contains(parameters.runtime, 'arm64'))) }}:
```
aishwaryabh: will this also try to create MSI for minified versions too for windows?

liliankasem: Yeah it will for any windows build that is not arm64; should we not be?

aishwaryabh: I don't think that we create MSI for minified versions:
https://azfunc.visualstudio.com/internal/_build/results?buildId=223061&view=artifacts&pathAsName=false&type=publishedArtifacts

For this build it's only created for win-x64 and win-x86

liliankasem: Sweet, I will update to ignore .min

PR #4468, opened 2025-06-06
Impement build props and targets

liliankasem:
This PR implements the build props and targets that will directly replace each step in the build.csproj. Steps include:

- Downloading templates into the output directory
- Downloading distlib into the tools/python directory
- Filtering PowerShell workers by RID
- Removing the workers folder on .min builds

Second this PR also brings in the same repo props, and release targets as the Host repository to keep us consistent across our repos.

Refer to https://github.com/Azure/azure-functions-host/tree/dev/eng/build for the original files

Finally, we are updating the versioning of the core tools CLI. We will still follow the model we have today for now (4.0.xxxx) - but we no longer want this version set by the CI as we need more control of the version across our many pipelines. This PR sets a new version `4.0.10000` that is higher than any version we have released today to avoid issues. We can discuss the finally version we want to land on when this feature goes into main.


eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
...
+    <DistLibUnzipped>$(ToolsDir)distlib-$(DistLibVersion)</DistLibUnzipped>
+  </PropertyGroup>
+
+  <Target Name="SetupDistlib" AfterTargets="Build">
+    <Message Text="Setting up 'distlib' tool in: $(DestDistLibDir)" Importance="high" />
+
+    <!-- Download distlib -->
+    <DownloadFile
+      SourceUrl="$(DistLibUrl)"
+      DestinationFolder="$(ToolsDir)"
```
jviau: Recommend putting all intermediate / staging files in `$(IntermediateOutputPath)`, not `OutDir` or `OutputPath`. Those are for final output only.

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
...
+    <!-- Customize these properties as needed -->
+    <DistLibVersion>0.3.9</DistLibVersion>
+    <DistLibUrl>https://github.com/vsajip/distlib/archive/$(DistLibVersion).zip</DistLibUrl>
+    <ToolsDir>$(OutDir)tools\</ToolsDir>
+    <DestDistLibDir>$(ToolsDir)python\packapp\distlib\</DestDistLibDir>
+    <DistLibZip>$(ToolsDir)distlib.zip</DistLibZip>
+    <DistLibUnzipped>$(ToolsDir)distlib-$(DistLibVersion)</DistLibUnzipped>
+  </PropertyGroup>
+
+  <Target Name="SetupDistlib" AfterTargets="Build">
```
jviau: Can optimize incremental build by downloading only if the zip is missing.

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
...
+    </PropertyGroup>
+
+    <!-- Validate existence of source distlib folder -->
+    <Error
+      Condition="!Exists('$(SourceDistLib)')"
+      Text="Expected distlib folder not found in $(SourceDistLib)" />
+
+    <!-- Copy all files inside the distlib folder to the destination -->
+    <ItemGroup>
+      <DistLibFiles Include="$(SourceDistLib)\**\*" />
```
jviau: Small nit: using just `**` is equivalent to `**/*`

```suggestion
      <DistLibFiles Include="$(SourceDistLib)/**" />
```

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <!-- Customize these properties as needed -->
+    <DistLibVersion>0.3.9</DistLibVersion>
+    <DistLibUrl>https://github.com/vsajip/distlib/archive/$(DistLibVersion).zip</DistLibUrl>
+    <ToolsDir>$(OutDir)tools\</ToolsDir>
```
jviau: nit: recommend using `/` and not `\` for consistency

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
...
+      <DistLibFiles Include="$(SourceDistLib)\**\*" />
+    </ItemGroup>
+
+    <Copy
+      SourceFiles="@(DistlibFiles)"
+      DestinationFiles="@(DistlibFiles->'$(DestDistLibDir)%(RecursiveDir)%(Filename)%(Extension)')"
+      SkipUnchangedFiles="true" />
+
+    <!-- Cleanup -->
+    <Delete Files="$(DistLibZip)" />
```
jviau: If you put these to `IntermediateOutputPath`, then no need to cleanup - you can re-use them between builds.

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
...
+    <RemoveDir Directories="$(DistLibUnzipped)" />
+  </Target>
+
+  <Target Name="CopyDistlibOnPublish" BeforeTargets="Publish">
+    <Message Text="Copy 'distlib/' to publish directory..." Importance="high" />
```
jviau: Avoid `high` importance messages unless necessary

```suggestion
    <Message Text="Copy 'distlib/' to publish directory..." Importance="low" />
```

eng/build/Minified.targets
```diff
@@ -0,0 +1,32 @@
...
+  <ItemGroup Condition="'$(IsMinified)' == 'true'">
+    <_LanguageWorkers Include="$(PublishDir)\workers\python" />
+    <_LanguageWorkers Include="$(PublishDir)\workers\java" />
+    <_LanguageWorkers Include="$(PublishDir)\workers\powershell" />
+    <_LanguageWorkers Include="$(PublishDir)\workers\node" />
+  </ItemGroup>
+
+  <!-- Remove worker directories from minified builds -->
+  <!-- Cannot remove entire directory due to this bug: https://github.com/Azure/azure-functions-core-tools/issues/2380 -->
+  <Target Name="RemoveLanguageWorkersForMin" AfterTargets="Publish" Condition="'$(RuntimeIdentifier)' != '' AND '$(IsMinified)' == 'true'">
```
jviau: Instead of having to check `IsMinified` in each target, you could condition importing of this entire file on `'$(IsMinified)' == 'true'`

eng/build/Telemetry.targets
```diff
@@ -0,0 +1,24 @@
+<Project>
+
+    <PropertyGroup Condition="'$(PublicRelease)' == 'true'">
+        <DefineConstants>$(DefineConstants);IS_PUBLIC_RELEASE</DefineConstants>
+    </PropertyGroup>
+
+    <Target Name="ReplaceTelemetryKey" BeforeTargets="BeforeCompile" Condition="'$(PublicRelease)' == 'true'">
```
jviau: See the other PR where I commented we can avoid writing to source files by using an assembly attribute.

That, OR you could write the entire `Constants.Build.cs` file to intermediate output directory, then include that in `Compile` item group. (And not include it in the src folder at all).

The goal is to avoid having `dotnet build` update any files in source control.

eng/build/Templates.targets
```diff
@@ -0,0 +1,113 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <TemplatesVersion>4.0.5086</TemplatesVersion>
+    <TemplatesJsonVersion>3.1.1648</TemplatesJsonVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TemplatesOutputDir>$(OutDir)templates\</TemplatesOutputDir>
```
jviau: nit: use `$(IntermediateOutputPath)` instead of `$(OutDir)` for this staging work. You don't need to clean it up then and can skip the download steps if they are already on disk.

eng/build/Templates.targets
```diff
@@ -0,0 +1,113 @@
...
+    <TemplatesNetIsolatedDir>$(TemplatesOutputDir)net-isolated\</TemplatesNetIsolatedDir>
+    <TemplatesJsonZip>$(TemplatesOutputDir)templates.zip</TemplatesJsonZip>
+  </PropertyGroup>
+
+  <!-- Download NuGet template packages -->
+  <ItemGroup>
+    <PackageDownload Include="Microsoft.Azure.Functions.Worker.ItemTemplates" Version="[$(TemplatesVersion)]" />
```
jviau, line 17: `PackageReference` has a property `GeneratePathProperty` you can set to `true` which will instruct nuget to add the path to the downloaded package into the generated props. I wonder if `PackageDownload` does also? You don't need manually set up that property then.

https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#generatepathproperty

eng/build/Workers.Python.Workaround.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+
+  <!-- PythonWorker v4.35 introduced an error when being build for win-arm64 (or any unsupported RID).
```
jviau, line 3: Optional suggestion: you could add a safeguard here to error out if this file gets imported for the incorrect RID.

``` xml
<Target Name="_IncorrectWorkaroundImport" BeforeTargets="Build" Condition="'$(RuntimeIdentifier)' != 'win-arm64'">
    <Error Text="File '$(MSBuildThisFile)' must only be imported for runtime identifier 'win-arm64'. Actual runtime: '$(RuntimeIdentifier)'" />
</Target>
```

test/Directory.Build.props
jviau, line 1: Looks like this file can be removed?

eng/build/Distlib.targets
liliankasem: Looks like there is a PR from Gavin to remove distlib so will likely just delete this target

eng/build/Distlib.targets
```diff
@@ -0,0 +1,61 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <!-- Customize these properties as needed -->
+    <DistLibVersion>0.3.9</DistLibVersion>
+    <DistLibUrl>https://github.com/vsajip/distlib/archive/$(DistLibVersion).zip</DistLibUrl>
+    <ToolsDir>$(OutDir)tools\</ToolsDir>
+    <DestDistLibDir>$(ToolsDir)python\packapp\distlib\</DestDistLibDir>
```
aishwaryabh: I saw gavin was removing distlib for python?

aishwaryabh: ah I just saw your other comment

eng/build/Distlib.targets
liliankasem: It looks like most of the targets are using `\` so wouldnt that be the consistent one?

eng/build/Templates.targets
liliankasem: Looks like not :( https://github.com/NuGet/Home/issues/8476

liliankasem:
@jviau ready for a second pass!

eng/build/Engineering.props
```diff
@@ -43,6 +18,15 @@
     <NuGetAuditMode>all</NuGetAuditMode> <!-- audit transitive dependencies. -->
   </PropertyGroup>
 
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <GenerateRuntimeConfigDevFile>true</GenerateRuntimeConfigDevFile><!-- https://github.com/dotnet/runtime/issues/54684 -->
+    <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
```
Frulfump, line 26: Is there a VB project somewhere? It seems `DisableImplicitNamespaceImports` is only for VB projects
"The DisableImplicitNamespaceImports property can be used to disable implicit namespace imports in Visual Basic projects that target .NET 6 or a later version. Implicit namespaces are the default namespaces that are imported globally in a Visual Basic project. Set this property to true to disable implicit namespace imports."
https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#disableimplicitnamespaceimports

eng/build/Distlib.targets
jviau: Up to you. I try to use `/` throughout msbuild files. ('try' is the keyword, I most likely miss some places)

Directory.Build.props
jviau, line 1: Tiny nit throughout: add empty newlines at end of files.

eng/build/Templates.targets
```diff
@@ -0,0 +1,128 @@
...
+    <_TemplatePackages Include="microsoft.azure.functions.worker.projecttemplates" Version="$(TemplatesVersion)" />
+    <_TemplatePackages Include="microsoft.azure.webjobs.itemtemplates" Version="$(TemplatesVersion)" />
+    <_TemplatePackages Include="microsoft.azure.webjobs.projecttemplates" Version="$(TemplatesVersion)" />
+  </ItemGroup>
+
+  <!-- Target to download and extract templates.json zip -->
+  <Target Name="DownloadAndExtractTemplatesJson" BeforeTargets="Build" Outputs="$(TemplatesJsonZip)">
+    <Message Importance="low" Text="Downloading templates.json file to '$(TemplatesOutputDir)'..." />
+
+    <DownloadFile
```
jviau, line 36: Maybe add a condition to download only if the zip doesn't exist on disk?

jviau (approved):

eng/build/Version.targets
```diff
@@ -0,0 +1,23 @@
...
+  <!-- CI versioning steps -->
+  <PropertyGroup Condition="'$(CI)' == 'true' AND '$(PublicRelease)' != 'true'">
+    <VersionSuffix Condition="'$(VersionSuffix)' != ''">$(VersionSuffix).$(BuildReason).$(BuildNumber).$(BuildCounter)</VersionSuffix>
+    <VersionSuffix Condition="'$(VersionSuffix)' == ''">$(BuildReason).$(BuildNumber).$(BuildCounter)</VersionSuffix>
+    <Version>$(VersionPrefix)-$(VersionSuffix)</Version>
+  </PropertyGroup>
+
+  <!-- When building in Azure pipelines, update the build number. -->
+  <!-- Specifically use '$(TF_BUILD)' and not '$(CI)' to ensure we are in Azure pipelines. -->
+  <Target Name="UpdateAzDoBuildNumber" Condition="'$(TF_BUILD)' == 'true' AND '$(UpdateBuildNumber)' == 'true'" BeforeTargets="BeforeBuild">
```
aishwaryabh, line 19: Do we also need to update metadata.json script so it includes this build id as well?

jviau: Can you point me to where metadata.json is generated? Does it currently use `$(Version)`?

aishwaryabh: We get the current release number from ADO: https://github.com/Azure/azure-functions-core-tools/blob/main/eng/scripts/artifact-assembler/generate-metadata-file.ps1#L17

Is this value saved as $env:BUILD_BUILDID as well?

aishwaryabh: I do think we need to update the parsing though for getting the out-of-proc version number since today in the script we look for `$_.Name -match "^Azure\.Functions\.Cli\..*\.(\d+\.\d+\.\d+)$"`

aishwaryabh:
Should we update the in-proc branch to have similar targets for the versions? It might be strange if the in-proc vs default files have different names/versioning schemas but we can work with it technically (since we package everything in the consolidated pipeline).

I would also just run this through the consolidated pipeline once with the official build on the feature branch just to make sure it packages everything correctly :)

eng/build/Version.targets
jviau: We are setting `BUILD_BUILDNUMBER` here, not `BUILD_BUILDID`. I don't believe anything is changing for that script.

liliankasem: I updated zip artifacts so we're still using the same naming convention we have today `Azure.Functions.Cli` - I'll defo do an consolidated pipeline run once the feature branch is ready and we can fix anything that doesnt work

liliankasem:
> Should we update the in-proc branch to have similar targets for the versions? It might be strange if the in-proc vs default files have different names/versioning schemas but we can work with it technically (since we package everything in the consolidated pipeline).
> 
> I would also just run this through the consolidated pipeline once with the official build on the feature branch just to make sure it packages everything correctly :)

Probably a good idea to update the in-proc versioning if nothing else; I can start a feature branch with in-proc changes once we're happy with main.

aishwaryabh (approved):

PR #4469, opened 2025-06-06
Update versioning logic & move TelemetryInstrumentationKey to its own Constants file

liliankasem:
This PR introduces an AssembleyExtensions.cs file to help us get the CLI version. The `InformationalVersion` looks like this:

- `4.0.10000-dev+67bd99a8ce2ec3cf833f25c039f60222caf44573 (64-bit)`

And the `GetCliVersion` method allows us extract the first part if it i.e. `4.0.10000-dev`

(Note: on release -dev or -ci will not be part of the CLI version)

Second, we convert constants to a partial class so that we can override the value of TelemetryInstrumentationKey as part of the build process.

### DIFF
--- a/src/Cli/func/Common/Constants.Build.cs
+++ b/src/Cli/func/Common/Constants.Build.cs
@@ -1,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+namespace Azure.Functions.Cli.Common
+{
+    internal static partial class Constants
+    {
+#if IS_PUBLIC_RELEASE
+        public const string TelemetryInstrumentationKey = "__TELEMETRY_KEY__";
+#else
+        public const string TelemetryInstrumentationKey = "00000000-0000-0000-0000-000000000000";
+#endif
+    }
+}
--- a/src/Cli/func/Common/Constants.cs
+++ b/src/Cli/func/Common/Constants.cs
@@ -2,12 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections.ObjectModel;
-using System.Reflection;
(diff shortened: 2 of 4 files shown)


src/Cli/func/Common/Constants.Build.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+namespace Azure.Functions.Cli.Common
+{
+    internal static partial class Constants
+    {
+#if IS_PUBLIC_RELEASE
+        public const string TelemetryInstrumentationKey = "__TELEMETRY_KEY__";
```
jviau: Oh one way to get this configured via msbuild is to use a custom attribute and the [AssemblyAttribute](https://learn.microsoft.com/en-us/dotnet/standard/assembly/set-attributes-project-file#set-arbitrary-attributes) item group. You can then read that in at runtime.

``` xml
<PropertyGroup>
  <!-- overridden in CI builds -->
  <TelemetryInstrumentationKey>00000000-0000-0000-0000-000000000000</TelemetryInstrumentationKey>
</PropertyGroup>

<ItemGroup>
  <AssemblyAttribute Include="Azure.Functions.Cli.TelemetryInstrumentationKeyAttribute" Parameter1="$(TelemetryInstrumentationKey)" />
</ItemGroup>
```

``` csharp
internal static partial class Constants
{
    public static readonly TelemetryInstrumentationKey = typeof(Constants).Assembly.GetCustomAttribute<TelemetryInstrumentationKeyAttribute>().Value;
}

internal class TelemetryInstrumentationKeyAttribute(string value) : Attribute
{
    public string Value => value;
}
```

jviau (approved):

jviau (approved):

PR #4482, opened 2025-06-20
Implement file locking mechanism for templates/bundle installation

### LINKED ISSUE #4458: Make func new and func init tests less flakey by changing when templates are installed/uninstalled
`Func init` calls `dotnet new template` but sometimes it doesn't fetch the correct template due to a race condition when tests are run in parallel, and typically retrying the command results in a success

Here is an example of the run that does this: 
https://azfunc.visualstudio.com/internal/_build/results?buildId=221679&view=logs&j=6ce172a1-8e70-5920-fc7b-153ca736e7e6&t=60d2697b-8955-5045-82c8-6cb6e9f12161&l=11139

We need to modify the code within `DotnetHelper` to only uninstall/re-install templates if needed.

aishwaryabh:

resolves #4458

When we run dotnet new func, we are expecting the project and item templates to already be installed since they happen right before that  command for func new and func init. For some reason, when the tests run in parallel, the templateconfig.json somehow gets messed up when we have multiple instances of func new and func init running on the same machine, so it doesn't find the template for func. If we isolate each test so it isn't run in parallel, we don't see this error anymore.

I've come up with a fix where we install web jobs/isolated templates only if we need it, given that the template isn't already installed.

There also is a race condition when running E2E func new tests where `templatesManager.Templates.Result` is not set, so fetching the templates fail. This change adds more logs and forces `templatesManager.Templates.Result` to populate if it hasn't already, since the value is lazily loaded.

 The error message that is shown today to the user is: 

`One or more errors occurred. (Value cannot be null. (Parameter 'value'))`

This error message should still show up but with more logs on what exactly it's doing before, in case this error pops up again.

Example of race condition: https://dev.azure.com/azfunc/public/_build/results?buildId=224520&view=logs&j=da439e82-8265-5962-85f8-172d6600cb6f&t=17f2655d-abad-5398-83c4-ac47d6b8119e&l=162

### Pull request checklist

### DIFF
--- a/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
@@ -115,7 +115,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
             await UpdateLanguageAndRuntime();
 
-            if (IsNewPythonProgrammingModel()) // depends on UpdateLanguageAndRuntime to set 'Language'
+            // Depends on UpdateLanguageAndRuntime to set 'Language'
+            if (IsNewPythonProgrammingModel())
             {
                 _newTemplates = await _templatesManager.NewTemplates;
                 _userPrompts = await _templatesManager.UserPrompts;
--- a/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
@@ -40,9 +40,9 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             _contextHelpManager = contextHelpManager;
             _initAction = new InitAction(_templatesManager, _secretsManager);
             _userInputHandler = new UserInputHandler(_templatesManager);
-            _templates = new Lazy<IEnumerable<Template>>(() => { return _templatesManager.Templates.Result; });
-            _newTemplates = new Lazy<IEnumerable<NewTemplate>>(() => { return _templatesManager.NewTemplates.Result; });
-            _userPrompts = new Lazy<IEnumerable<UserPrompt>>(() => { return _templatesManager.UserPrompts.Result; });
+            _templates = new Lazy<IEnumerable<Template>>(() => EnsureTemplatesLoaded(tm => tm.Templates, "Templates"));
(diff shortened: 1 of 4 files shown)


src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -309,12 +309,23 @@ public async Task UpdateLanguageAndRuntime()
...
                     var languages = WorkerRuntimeLanguageHelper.LanguagesForWorker(_workerRuntime);
-                    var displayList = _templates.Value
+
+                    // If templates are not loaded yet, we need to check the task status and force it to populate if it hasn't already
+                    if (!_templates.IsValueCreated)
+                    {
+                        ColoredConsole.WriteLine("Templates not loaded yet, checking task status...");
+                        var taskStatus = _templatesManager.Templates.Status;
+                        ColoredConsole.WriteLine($"Templates task status: {taskStatus}");
+                        _ = _templatesManager.Templates.Result; // This should force templates to load in templatesManager
```
aishwaryabh: Lmk if this makes sense! Not sure how else to force templatesManager.Templates.Result to populate

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -291,30 +292,106 @@ private static Task TemplateOperation(Func<Task> action, WorkerRuntime workerRun
 
         private static async Task IsolatedTemplateOperation(Func<Task> action)
         {
-            try
+            await EnsureIsolatedTemplatesInstalled();
```
aishwaryabh: Lmk if this looks ok too. I figured we don't need to always install/uninstall if the user already has the template on their machine, so this scenario should be applicable to a regular environment (not just CI when we run things in parallel)

src/Cli/func/Common/TemplatesManager.cs
```diff
@@ -20,7 +22,7 @@ public TemplatesManager(ISecretsManager secretsManager)
         }
 
         /// <summary>
-        /// Gets get new templates.
+        /// Gets new (v2) templates.
```
satvu, line 25: what does v2 mean here?

liliankasem: This refers to the v2 templates for the python model and node (v4?)

satvu: Offline: This refers to v2 python programming model templates

satvu (approved):

PR #4495, opened 2025-07-01
Fix Console Encoding for Proper Display of Non-ASCII Characters for out of proc

### LINKED ISSUE #4429: Console encoding might be limited to ASCII?
### Version

4.0.7030

### Description

It appears that console output when running locally is not displaying all characters correctly. It might be limited to ASCII.

### Steps to reproduce

From a dotnet isolated app write a log with Japanese characters, IE:

``` CSharp
logger.LogError("Test String: こんにちは");
```

Console output will not be correct:

Image

VineethReyya:
Azure Functions Core Tools was not correctly displaying non-ASCII characters in console output. Japanese characters (and other non-Latin scripts) were showing as question marks (?????) when logging from a function.

Root Cause
The console output encoding was not explicitly set to UTF-8 at application startup, causing the console to use the default encoding of the system, which often doesn't support the full range of Unicode characters.

Solution
Added a single line at the start of the application to configure the console output encoding to UTF-8:

Console.OutputEncoding = Encoding.UTF8;
This ensures that all Unicode characters, including Japanese and other non-Latin scripts, are properly displayed in the console when running functions locally.

[Release Notes Document ](https://microsoftapc-my.sharepoint.com/:w:/g/personal/v-vreyya_microsoft_com/EWrINPUfIftFkjPYvAX1GnIB0ThttUi5JR75HimCLoVi5Q?e=8CfZmC)

resolves #4429 
### Pull request checklist

### DIFF
--- a/src/Cli/func/Program.cs
+++ b/src/Cli/func/Program.cs
@@ -17,8 +17,8 @@ namespace Azure.Functions.Cli
 
         internal static void Main(string[] args)
         {
-            // Set console encoding
-            SetConsoleEncoding();
+            // Configure console encoding
+            ConsoleHelper.ConfigureConsoleOutputEncoding();
 
             // Check for version arg up front and prioritize speed over all else
             // Tools like VS Code may call this often and we want their UI to be responsive
@@ -112,20 +112,6 @@ namespace Azure.Functions.Cli
                 .As<IContextHelpManager>();
 
             return builder.Build();
-        }
-
-        private static void SetConsoleEncoding()
-        {
-            // Set console encoding to UTF-8 to properly display international characters
-            try
-            {
-                Console.OutputEncoding = Encoding.UTF8;
-            }
-            catch
-            {
-                // Silently fall back to default encoding if UTF-8 isn't supported
-                // International characters may not display correctly but CLI will still function
-            }
         }
     }
 }
--- a/src/Cli/func/Program.cs
+++ b/src/Cli/func/Program.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Text;
 using Autofac;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Helpers;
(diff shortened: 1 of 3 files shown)


VineethReyya:
@liliankasem @aishwaryabh Could you please review the recent changes when you have a moment and let me know if anything needs adjustment?

Thanks!

jviau:
@VineethReyya do we need to consider the scenario that the users terminal doesn't support UTF8?

VineethReyya:
> do we need to consider the scenario that the users terminal doesn't support UTF8?

@jviau Yes, it's important to consider the possibility that a user's terminal might not support UTF-8. To account for that, we're implementing a try-catch block around the console encoding setup as a safeguard.

VineethReyya:
@liliankasem @aishwaryabh @jviau Could you please review the recent changes when you have a moment.

src/Cli/func/Program.cs
```diff
@@ -109,5 +113,19 @@ internal static IContainer InitializeAutofacContainer()
...
+        private static void SetConsoleEncoding()
+        {
+            // Set console encoding to UTF-8 to properly display international characters
+            try
+            {
+                Console.OutputEncoding = Encoding.UTF8;
+            }
+            catch
+            {
+                // Silently fall back to default encoding if UTF-8 isn't supported
```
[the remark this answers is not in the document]
liliankasem: ignore

src/Cli/func/Program.cs
```diff
@@ -109,5 +113,19 @@ internal static IContainer InitializeAutofacContainer()
 
             return builder.Build();
         }
+
+        private static void SetConsoleEncoding()
```
liliankasem: Can we move this into a helper class? And rename + add doc

```csharp
/// <summary>
/// Attempts to switch the console to UTF-8.  
/// Falls back silently if the code page isn’t registered or the host refuses.
/// </summary>
private static void ConfigureConsoleOutputEncoding()
{
    try
    {
        Console.OutputEncoding = Encoding.UTF8;
    }
    catch
    {
        // UTF-8 encoding not available, international characters may not display correctly.
    }
}
```

liliankasem: Let's also update the release notes to mention this change.

VineethReyya:
Hi @liliankasem, @aishwaryabh 

I’ve made changes that are ready for review and included the release notes in the body of the PR for now — when you get a chance, could you please take a look and let me know your feedback?

Also, I’m not entirely sure where we typically document release notes for this project. Could you point me to the right place or let me know the usual process?

Thanks in advance!

liliankasem:
> Hi @liliankasem, @aishwaryabh
> 
> I’ve made changes that are ready for review and included the release notes in the body of the PR for now — when you get a chance, could you please take a look and let me know your feedback?
> 
> Also, I’m not entirely sure where we typically document release notes for this project. Could you point me to the right place or let me know the usual process?
> 
> Thanks in advance!

Thanks! Release notes live here now: https://github.com/Azure/azure-functions-core-tools/blob/main/release_notes.md

VineethReyya:
closing this PR as it needs rebase. created new PR : https://github.com/Azure/azure-functions-core-tools/pull/4515

PR #4496, opened 2025-07-01
Fix Console Encoding for Proper Display of Non-ASCII Characters for Inproc

### LINKED ISSUE #4429: Console encoding might be limited to ASCII?

VineethReyya:
Azure Functions Core Tools was not correctly displaying non-ASCII characters in console output. Japanese characters (and other non-Latin scripts) were showing as question marks (?????) when logging from a function.

Root Cause
The console output encoding was not explicitly set to UTF-8 at application startup, causing the console to use the default encoding of the system, which often doesn't support the full range of Unicode characters.

Solution
Added a single line at the start of the application to configure the console output encoding to UTF-8:

Console.OutputEncoding = Encoding.UTF8;
This ensures that all Unicode characters, including Japanese and other non-Latin scripts, are properly displayed in the console when running functions locally.
feature_inproc

Partially resolves #4429

### Pull request checklist


VineethReyya:
After creating PR it appears to have changes other than in program.cs . so, closing it

PR #4520, opened 2025-07-09
Fix consolidated pipeline & nightly feed versioning for linux and host CI

liliankasem:

n/a

### Pull request checklist

### Additional information

Fixing bugs in the consolidation pipeline. Pipeline validated here: https://azfunc.visualstudio.com/internal/_build/results?buildId=227154&view=results . Also noticed the reference to the metadata file was wrong in the linux CI so fixing that as well.

### DIFF
--- a/eng/ci/templates/official/jobs/consolidate-cli-artifacts.yml
+++ b/eng/ci/templates/official/jobs/consolidate-cli-artifacts.yml
@@ -72,6 +72,8 @@ jobs:
         version: '*'
         downloadPath: '$(Pipeline.Workspace)/func-cli-default/func-cli'
 
+  - template: /eng/ci/templates/steps/install-tools.yml@self
+
   - task: DotNetCoreCLI@2
     displayName: "Run ArtifactAssembler"
     inputs:
@@ -98,7 +100,7 @@ jobs:
         targetType: filePath
         filePath: '$(Build.SourcesDirectory)/eng/scripts/generate-msi-files.ps1'
         arguments: '-ArtifactsPath "$(Pipeline.Workspace)/staging/func-cli"'
-        workingDirectory: '$(Build.SourcesDirectory)/build'
+        workingDirectory: '$(Build.SourcesDirectory)'
 
     - template: ci/sign-files.yml@eng
       parameters:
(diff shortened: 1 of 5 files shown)


eng/ci/templates/official/jobs/consolidate-cli-artifacts.yml
```diff
@@ -98,7 +100,7 @@ jobs:
         targetType: filePath
         filePath: '$(Build.SourcesDirectory)/eng/scripts/generate-msi-files.ps1'
         arguments: '-ArtifactsPath "$(Pipeline.Workspace)/staging/func-cli"'
```
liliankasem, line 102: @aishwaryabh or @soninaren - do you know why we only generate MSI for x architectures? Can we generate an MSI build for arm?

eng/ci/templates/official/jobs/consolidate-cli-artifacts.yml
```diff
@@ -165,9 +167,11 @@ jobs:
     displayName: 'Copy Files to Artifact Staging Directory'
     inputs:
       SourceFolder: '$(Pipeline.Workspace)/staging'
-      Contents: 'func-cli-*/**'
+      Contents: 'func-cli*/**'
       TargetFolder: '$(Build.ArtifactStagingDirectory)/core-tools'
 
+  # The only reason this step occurs only on min.win-x64 is because the metadata.json file only needs to be generated once.
+  # This file is used by the release pipeline to determine the version of the CLI & other metadata needed for release.
```
liliankasem, line 174: We don't have to deal with this now, but we should make this its own job at some point, though I think when we stop shipping in-proc we might not need to generate this file anymore anywho.

eng/ci/templates/official/jobs/consolidate-cli-artifacts.yml
liliankasem: Also wondering why we only run the test step for visual studio but only for 'min.win-x64'?

liliankasem: Spoke to Naren about this offline, MSI for arm64 is not possible so we can address that in another issue when we have capacity. Visual studio uses .min versions so we can update this to test all .min versions.

eng/scripts/artifact-assembler/test-artifacts.ps1
```diff
@@ -1,51 +1,75 @@
...
+    if ($null -ne $inprocFuncExe) {
+        Write-Host "Including $inprocVersion func executable: $inprocFuncExe"
+        $funcExecutables += $inprocFuncExe
+    } else {
+        Write-Host "$inprocVersion func executable not found in: $inprocDir"
+    }
+}
+
+# Loop through each func executable and run tests
+foreach ($func in $funcExecutables) {
```
aishwaryabh: do we have to run tests for each executable separately? I thought we're just testing the root one and after everything is packaged with inproc6 and inproc8 as subdirectories, we only have to run the tests on root one in theory (and the inproc6 and inproc8 func.exe's should execute by themselves without manually having to run the tests on them)

so we are testing functionality for both inproc and default but we only need to test one executable since that's what the customer will do too. does that make sense?

liliankasem: That's what I was trying to ask you earlier! Thanks for clarifying, will update this to just get the root func. Also I've removed my changes from this PR and will be making a separate PR for the testing changes.

satvu (approved):

PR #4526, opened 2025-07-11
Break out consolidated artifact tests into their own stage with more rid coverage

liliankasem:

n/a

### Pull request checklist

### Additional information

- Break out the consolidated artifact tests into their own stage and test osx and linux RIDs as well
- Fix test results publish task unit
- Fix consolidated pipeline scheduled run: implemented a script to get the latest artifact version


eng/ci/templates/official/jobs/test-consolidated-cli-artifacts.yml
```diff
@@ -0,0 +1,107 @@
...
+        displayName: 'Make CLI binaries executable'
+
+    - task: PowerShell@2
+      displayName: 'Test Artifacts'
+      inputs:
+        targetType: filePath
+        filePath: '$(Build.SourcesDirectory)/eng/scripts/artifact-assembler/test-artifacts.ps1'
+        arguments: '-FuncCliPath "$(Pipeline.Workspace)/staging/func-cli"'
+
+  - task: PowerShell@2
```
aishwaryabh, line 71: could we potentially have this as a separate parallel job to execute the VS tests?

liliankasem: I think there are a couple reasons not to:

1. We already have a job running per RID, and I think we already use too many jobs and I want to be a bit more concervative - it will cause bottlenecks else where in the CI process
2. Given the parameters/conditions we're using, we will only ever run the VS tests or the non-VS tests and not both at the same time, so no reason to make them parallel because they kind of already are?

aishwaryabh: that's valid the number of jobs would increase significantly. Let's keep it as it is then :)

satvu (approved):

PR #4535, opened 2025-07-16
Handling Functions Runtime Environment env var when set by user

### LINKED ISSUE #2465: An item with the same key has already been added. Key: AZURE_FUNCTIONS_ENVIRONMENT
Core Tools version: 3.0.3233
Function Runtime Version: 3.0.15193.0

#### Repro steps

1. Fetch function app configuration from Azure function in azure portal which has AZURE_FUNCTIONS_ENVIRONMENT set
2. Start Azure function locally via Rider run

#### Expected behavior

Azure function should start up normally and read the AZURE_FUNCTIONS_ENVIRONMENT variable from previously fetched local.settings.json

#### Actual behavior

Function does not start locally, as the fails to add the AZURE_FUNCTIONS_ENVIRONMENT variable again.

"An item with the same key has already been added. Key: AZURE_FUNCTIONS_ENVIRONMENT"

#### Known workarounds

Workarund is to remove the AZURE_FUNCTIONS_ENVIRONMENT variable from local.settings.json after fetching the settings from the portal to continue developing it locally.

#### Related information 

Using Rider 2020.3.2 and developing in C#/dotnet

AZURE_FUNCTIONS_ENVIRONMENT should work as any other env variable in my opinion. When it's set in a system/user level, it should get it from there, if not, grab it from local.settings.json. Or at least, it should not stop the function from starting.

I am wondering that this problem isn't more common, as I suspect everyone has set AZURE_FUNCTIONS_ENVIRONMENT in his azure function appsettings in Azure(?) and at least a few use the fetch app settings method to generate local.settings.json to work locally

Maybe I am doing something wrong, please let me know

brandonh-msft:
with accompanying test
Fixes #2465

resolves #2465

### Pull request checklist

### DIFF
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -296,7 +296,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
             // when running locally in CLI we want the host to run in debug mode
             // which optimizes host responsiveness
-            settings.Add("AZURE_FUNCTIONS_ENVIRONMENT", "Development");
+            settings.TryAdd("AZURE_FUNCTIONS_ENVIRONMENT", "Development");
 
             // Inject the .NET Worker startup hook if debugging the worker
             if (DotNetIsolatedDebug != null && DotNetIsolatedDebug.Value)
--- a/test/Azure.Functions.Cli.Tests/ActionsTests/StartHostActionTests.cs
+++ b/test/Azure.Functions.Cli.Tests/ActionsTests/StartHostActionTests.cs
@@ -6,15 +6,23 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
+
 using Azure.Functions.Cli.Actions.HostActions;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Helpers;
 using Azure.Functions.Cli.Interfaces;
+
 using Colors.Net;
+
 using FluentAssertions;
+using FluentAssertions.Execution;
+
 using Moq;
+
 using NSubstitute;
+
 using Xunit;
+
 using YamlDotNet.Core;
 
 namespace Azure.Functions.Cli.Tests.ActionsTests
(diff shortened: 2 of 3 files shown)


test/Azure.Functions.Cli.Tests/ActionsTests/StartHostActionTests.cs
```diff
@@ -247,6 +255,42 @@ public async Task ValidateHostRuntimeAsync_MatchesExpectedResults(WorkerRuntime
             Assert.False(expectException, "Expected validation failure.");
         }
 
+        [InlineData(WorkerRuntime.DotnetIsolated, "default", true)]
+        [InlineData(WorkerRuntime.DotnetIsolated, "default", false)]
+        [InlineData(WorkerRuntime.Python, "default", false)]
+        [InlineData(WorkerRuntime.Java, "default", false)]
+        [InlineData(WorkerRuntime.Node, "default", false)]
+        [Theory]
+        public async Task AzureFunctionsEnvironment_EnvironmentVariable_SetByUser_DoesNotThrow(WorkerRuntime currentRuntime, string hostRuntimeArgument, bool validNet8Configuration)
```
liliankasem: The Azure.Functions.Cli.Tests project is being deprecated - can you move this to the UnitTest project instead? (StartHostActionTests.cs should be there now)

liliankasem (approved):

test/Cli/Func.UnitTests/ActionsTests/StartHostActionTests.cs
```diff
@@ -261,6 +268,42 @@ public async Task ValidateHostRuntimeAsync_MatchesExpectedResults(WorkerRuntime
...
+
+            Mock<IProcessManager> processManager = new();
+            Mock<ISecretsManager> secretsManager = new();
+
+            var settings = new Dictionary<string, string> { ["AZURE_FUNCTIONS_ENVIRONMENT"] = "MyEnvironment" };
+            secretsManager.Setup(s => s.GetSecrets()).Returns(() => settings);
+
+            GlobalCoreToolsSettings.Init(secretsManager.Object, []);
+
+            var startHostAction = new StartHostAction(secretsManager.Object, processManager.Object)
```
liliankasem: My bad, I just noticed that you're actually running the CLI with this test. So it's not a unit test but would need to use the E2E test framework we have. Here is an example of how to write one of those instead:

https://github.com/Azure/azure-functions-core-tools/blob/e963a7607b71a06d8cd2780833c5487fa50c511c/test/Cli/Func.E2ETests/Commands/FuncStart/MissingConfigTests.cs#L50-L83

I believe there is a [doc](https://github.com/Azure/azure-functions-core-tools/blob/main/test/Cli/Func.E2ETests/README.md) too

liliankasem: Refactored this as a unit test for GetConfigurationSettings

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -300,7 +300,7 @@ private async Task<IDictionary<string, string>> GetConfigurationSettings(string
 
             // when running locally in CLI we want the host to run in debug mode
             // which optimizes host responsiveness
-            settings.Add("AZURE_FUNCTIONS_ENVIRONMENT", "Development");
+            settings.TryAdd("AZURE_FUNCTIONS_ENVIRONMENT", "Development");
```
fabiocav, line 303: There are a lot of side effects if we allow for this to be overridden. The WebJobs SDK and host behaviors for polling intervals, logging, and other things will change, and this could significantly impact the local debugging experience in surprising ways.

liliankasem:
Closing in favor of #4563

PR #4540, opened 2025-07-18
Implement force exit on second cancel key

### LINKED ISSUE #4355: Core tools not responding to ctrl-C
### Version

4.0.6821 on Windows (and 4.0.6610 on Mac)

### Description

The core tools doesn't seem to be responding properly to terminate commands anymore. Observed both on Windows and Mac. Instead, it just hangs.

This might be a duplicate of https://github.com/Azure/azure-functions-core-tools/issues/4268, but the attempted connection to Azurite is not present in my repros.

I don't know if other environment factors are present that would influence this, but seeing it across users and OSes makes me think there is something surprising at play.

### Steps to reproduce

`Func start` -> `Ctrl+C`

liliankasem:

resolves #4355

### Pull request checklist

### Additional information

- Add handler so that we force exit on a second console cancel key
- Fix bug where we get a null ref error if we shut down the host during startup
- Added docs for how to debug CLI with the host

### DIFF
--- a/docs/debug-with-host.md
+++ b/docs/debug-with-host.md
@@ -1,0 +1,55 @@
+# Guide on debugging Core Tools with the Azure Functions Host
+
+1. Clone the host repo: `azure-functions-host`
+2. In `src/Cli/func/Azure.Functions.Cli.csproj`, replace the WebHost package reference with a local project reference:
+
+    Comment out:
+
+   ```xml
+   <!-- <PackageReference Include="Microsoft.Azure.WebJobs.Script.WebHost" /> -->
+   ```
+
+   Add:
+
+   ```xml
+    <ItemGroup>
+        <ProjectReference Include="<path to host repo>/src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj" />
+    </ItemGroup>
+   ```
+
+> [!NOTE]
+> You may also need to update the worker package references
+> in `eng/build/Packages.props` to match the versions references by the host.
+
+You should now be able to debug the host via the `func start` command in the CLI.
+
+#### VS Code
+
+1. Press F5
+2. Provide the command you want to run i.e. `start --verbose`
+3. Provide the path to your test app, e.g. `<path>/MyTestFuncApp`
+
+This will build the CLI and start the host with the provided command and test app path, and automatically attach the debugger.
+
+#### Visual Studio
+
+//todo: Add instructions for Visual Studio
+
+#### Other
+
+You can also run the CLI directly from the repo:
+
+`dotnet run --project src/Cli/func -- start --script-root <path-to-test-app>`
+
+Or, you can also publish the CLI and run via the path:
+
+```bash
+# Publish the CLI for your platform
+dotnet publish src/Cli/func -r osx-arm64 --sc
+
+# Typically '<path>/out/pub/Azure.Functions.Cli/release_osx-arm64/func'
+export FUNC_CLI=<path-to-published-cli>
+
+# Start the host inside the test app directory
+$FUNC_CLI start
+```
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -438,22 +438,25 @@ namespace Azure.Functions.Cli.Actions.HostActions
             var runTask = host.RunAsync();
             var hostService = host.Services.GetRequiredService<WebJobsScriptHostService>();
 
-            await hostService.DelayUntilHostReady();
-
-            var scriptHost = hostService.Services.GetRequiredService<IScriptJobHost>();
-            var httpOptions = hostService.Services.GetRequiredService<IOptions<HttpOptions>>();
-            if (scriptHost != null && scriptHost.Functions.Any())
+            if (hostService.State is not ScriptHostState.Stopping && hostService.State is not ScriptHostState.Stopped)
             {
-                // Checking if in Limelight - it should have a `AzureDevSessionsRemoteHostName` value in local.settings.json.
-                var forwardedHttpUrl = _secretsManager.GetSecrets().FirstOrDefault(
-                    s => s.Key.Equals(Constants.AzureDevSessionsRemoteHostName, StringComparison.OrdinalIgnoreCase)).Value;
-                if (forwardedHttpUrl != null)
-                {
-                    var baseUrl = forwardedHttpUrl.Replace(Constants.AzureDevSessionsPortSuffixPlaceholder, Port.ToString(), StringComparison.OrdinalIgnoreCase);
-                    baseUri = new Uri(baseUrl);
-                }
+                await hostService.DelayUntilHostReady();
 
-                DisplayFunctionsInfoUtilities.DisplayFunctionsInfo(scriptHost.Functions, httpOptions.Value, baseUri);
+                var scriptHost = hostService.Services.GetRequiredService<IScriptJobHost>();
+                var httpOptions = hostService.Services.GetRequiredService<IOptions<HttpOptions>>();
+                if (scriptHost != null && scriptHost.Functions.Any())
+                {
+                    // Checking if in Limelight - it should have a `AzureDevSessionsRemoteHostName` value in local.settings.json.
+                    var forwardedHttpUrl = _secretsManager.GetSecrets().FirstOrDefault(
+                        s => s.Key.Equals(Constants.AzureDevSessionsRemoteHostName, StringComparison.OrdinalIgnoreCase)).Value;
+                    if (forwardedHttpUrl != null)
+                    {
+                        var baseUrl = forwardedHttpUrl.Replace(Constants.AzureDevSessionsPortSuffixPlaceholder, Port.ToString(), StringComparison.OrdinalIgnoreCase);
+                        baseUri = new Uri(baseUrl);
+                    }
+
+                    DisplayFunctionsInfoUtilities.DisplayFunctionsInfo(scriptHost.Functions, httpOptions.Value, baseUri);
+                }
             }
 
             if (VerboseLogging == null || !VerboseLogging.Value)
--- a/src/Cli/func/Program.cs
+++ b/src/Cli/func/Program.cs
@@ -34,10 +34,10 @@ namespace Azure.Functions.Cli
             _container = InitializeAutofacContainer();
             AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
 
-            Console.CancelKeyPress += (s, e) =>
+            CancelKeyHandler.Register(() =>
             {
                 _container.Resolve<IProcessManager>()?.KillChildProcesses();
-            };
+            });
 
             ConsoleApp.Run<Program>(args, _container);
         }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,7 +13,7 @@
             },
             "args": "${input:funcArgs} --script-root ${input:scriptRootArg}",
             "cwd": "${workspaceFolder}/src/Cli/func",
-            "console": "internalConsole",
+            "console": "integratedTerminal",
             "stopAtEntry": false
         },
         {
--- a/src/Cli/func/Helpers/CancelKeyHandler.cs
+++ b/src/Cli/func/Helpers/CancelKeyHandler.cs
@@ -1,0 +1,47 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Diagnostics;
+using Colors.Net;
+
+namespace Azure.Functions.Cli.Helpers;
+
+public static class CancelKeyHandler
+{
+    private static readonly object _lock = new object();
+    private static int _cancelKeyPressCount = 0;
+    private static Action _onFirstCancel;
+
+    public static void Register(Action onFirstCancel, int exitCode = 1)
+    {
+        _onFirstCancel = onFirstCancel ?? (() => { });
+
+        Console.CancelKeyPress += HandleCancelKeyPress;
+    }
+
+    private static void HandleCancelKeyPress(object sender, ConsoleCancelEventArgs e)
+    {
+        lock (_lock)
+        {
+            _cancelKeyPressCount++;
+
+            if (_cancelKeyPressCount == 1)
+            {
+                ColoredConsole.WriteLine("Press Ctrl+C again to force exit.");
+
+                // Cancel first Ctrl+C to allow graceful cleanup
+                e.Cancel = true;
+                _onFirstCancel?.Invoke();
+            }
+            else if (_cancelKeyPressCount >= 2)
+            {
+                ColoredConsole.WriteLine("Forcing exit...");
+                Console.Out.Flush();
+                Console.Error.Flush();
+
+                // Hard exit
+                Process.GetCurrentProcess().Kill();
+            }
+        }
+    }
+}


src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+    internal static void HandleCancelKeyPress(object sender, ConsoleCancelEventArgs e)
+    {
+        lock (_lock)
+        {
+            _cancelKeyPressCount++;
+
+            if (_cancelKeyPressCount == 1)
+            {
+                var message = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                            ? "Press 'q' to force exit."
```
liliankasem: @mattchenderson I'm not sure how I feel about this but it works, and we can't depend on Ctrl+C again (thank you windows).

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -438,22 +438,25 @@ public override async Task RunAsync()
...
-            var scriptHost = hostService.Services.GetRequiredService<IScriptJobHost>();
-            var httpOptions = hostService.Services.GetRequiredService<IOptions<HttpOptions>>();
-            if (scriptHost != null && scriptHost.Functions.Any())
+            if (hostService.State is not ScriptHostState.Stopping && hostService.State is not ScriptHostState.Stopped)
             {
-                // Checking if in Limelight - it should have a `AzureDevSessionsRemoteHostName` value in local.settings.json.
-                var forwardedHttpUrl = _secretsManager.GetSecrets().FirstOrDefault(
-                    s => s.Key.Equals(Constants.AzureDevSessionsRemoteHostName, StringComparison.OrdinalIgnoreCase)).Value;
-                if (forwardedHttpUrl != null)
+                await hostService.DelayUntilHostReady();
```
jviau, line 447: Do we have a shutdown cancellation token available here? If so, recommend the following:


```suggestion
                await hostService.DelayUntilHostReady().WaitAsync(shutdownToken);
```

src/Cli/func/Common/ProcessManager.cs
```diff
@@ -26,6 +26,11 @@ public IEnumerable<IProcessInfo> GetProcessesByName(string processName)
                 .Select(p => new ProcessInfo(p));
         }
 
+        public void KillMainProcess()
+        {
+            Process.GetCurrentProcess().Kill();
```
mattchenderson, line 31: should this use the overload accepting a boolean? I think if we call this method, we want to go for the whole process tree.

Same could be true for the KillChildProcesses method maybe?

src/Cli/func/CancelKeyHandler.cs
mattchenderson: I'm similarly not sure how I feel about this. It feels fine for non-Windows; Ctrl-C again is a understandable gesture in response to it not taking effect the first time. The console message helps.

What is the Windows behavior here that makes it different?

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+                        {
+                            var key = _consoleReader.ReadKey(true);
+                            if (key.Key == ConsoleKey.Q)
+                            {
+                                ForceQuit();
+                            }
+                        }
+                    });
+                }
+            }
```
mattchenderson: Doesn't this prevent the process from exiting _until_ the user presses the key? It seems like it won't exit naturally?

docs/debug-with-host.md
```diff
@@ -0,0 +1,55 @@
+# Guide on debugging Core Tools with the Azure Functions Host
+
```
fabiocav: The changes here don't seem to be directly related to this PR.

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+                            : "Press Ctrl+C again to force exit.";
+
+                ColoredConsole.WriteLine(message);
+
+                _onFirstCancel?.Invoke();
+
+                // Cancel first Ctrl+C to allow graceful cleanup
+                e.Cancel = true;
+
+                _processManager.KillChildProcesses();
```
fabiocav: This is only dealing with the delegated in-proc executions (spinning up the process for the in-proc host). Is that where we're seeing our issues/impact? If so, I think we can part of the issue we're having here is the fact the child process is never getting notified, and perhaps we can handle that notification instead.

src/Cli/func/CancelKeyHandler.cs
mattchenderson: Per offline thread with Lilian, the issue there is that the second Ctrl+C wouldn't register, which seemed intentional.

But the more I think about it, the more I'm unsure about introducing the extra key command here. There's something to the repeat being an extra signal of intent on its own. I know I said the console message helps in my last comment, but I can't say I love it. It feels like it changes the expectations, and in general it feels awkward as a CLI user.

Also, per that offline conversation, it sounds like the underlying hang getting in the way of a single Ctrl+C is being addressed, so I think I'm inclined to wait for that instead. We can revisit if needed.

fabiocav: @liliankasem that's the default behavior. There are properties you can set in `Console` that would give you the ability to handle `Ctrl + C` multiple times, but it would be your responsibility to provide the termination handling.

src/Cli/func/CancelKeyHandler.cs
fabiocav: Yes, this logic would. @liliankasem can you clarify the intent?

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                // Windows locks up the main thread processing the first Ctrl+C.
```
fabiocav: Did you confirm this? This handler should be invoked on a thread pool thread. Is the problem here caused by the console reader?

src/Cli/func/CancelKeyHandler.cs
liliankasem: Resolving this thread as the scope of the PR has changed.

src/Cli/func/Actions/HostActions/StartHostAction.cs
liliankasem: We do not, but I think we should have a different work item for all IActions taking in a cancellation token. We can do this as part of the parser/command refactor

src/Cli/func/CancelKeyHandler.cs
liliankasem: Resolving this thread as the scope of the PR has changed.

src/Cli/func/CancelKeyHandler.cs
liliankasem: Resolving this thread as the scope of the PR has changed.

docs/debug-with-host.md
liliankasem: Moved to a new PR: #4546

src/Cli/func/CancelKeyHandler.cs
liliankasem: Discussed offline (this is not where we are seeing the impact)

liliankasem:
@mattchenderson @fabiocav - as per @jviau suggestion, I've switched over to a grace period approach. We will give the host 2 seconds to gracefully shut down, if it does not we kill the process. The behaviour on pressing `Ctrl+C` remains the same where the host shutdown event is signalled and it starts its graceful shutdown process. 

After we address the bug in the host, the host should shutdown with the 2 seconds pretty much every time so there shouldn't be an major impact. On the plus side, if anything ever occurs where the host does take longer, thats not something we need to worry about for local dev with the graceful timeout period.

src/Cli/func/Program.cs
```diff
@@ -34,12 +35,22 @@ internal static void Main(string[] args)
             _container = InitializeAutofacContainer();
             AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
 
-            Console.CancelKeyPress += (s, e) =>
+            try
             {
-                _container.Resolve<IProcessManager>()?.KillChildProcesses();
-            };
+                CancelKeyHandler.Register(
```
jviau: This is going to immediately cancel the CTS on CTRL+C, and then kill the process after a delay. Given you have a `.WaitAsync(_shutdownCts.Token)`, will that delay even matter here?

I would expect graceful shutdown to be handled as follows:

1. Have two cancellation tokens.
    - One cancelled immediately on CTRL+C (the "shutting down token").
    - Another cancelled ~2 seconds after CTRL+C (the "force shutdown token").
1. Pass the first token to `ConsoleApp.RunAsync`.
2. Pass the second token to `.WaitAsync()`.
3. Finally, not sure if you need any of the `ProcessManager` changes.

This would give:

``` CSharp
await ConsoleApp.RunAsync<Program>(args, _container, shuttingDownToken).WaitAsync(forceShutdownToken);
```

liliankasem: >This is going to immediately cancel the CTS on CTRL+C, and then kill the process after a delay. Given you have a .WaitAsync(_shutdownCts.Token), will that delay even matter here?

With how we run the host today, the delay does matter. On the first Ctrl+C, we start the host shut down process - the host automatically picks up the cancellation without us needing to pass anything through. After the timeout delay, we exit (if we haven't already). If the app is up and running, the host exists in ms so we never hit the timeout delay. But when we exit during init of a worker, the shut down here takes 60seconds and so we do hit the timeout delay and shut down at 2 seconds.

`Pass the first token to ConsoleApp.RunAsync.`

We can do that, but right now we do not handle cancellation tokens at all so this would not do anything. I think we do eventually want to update the IAction interface so that `RunAsync` does take in a CT, but I think that is a huge change and out of scope of this PR.

>await ConsoleApp.RunAsync<Program>(args, _container, shuttingDownToken).WaitAsync(forceShutdownToken);

This does make sense to me and I think with the IAction changes mentioned above, we can do that. Maybe I should make that PR ahead of this one so to not bloat this.

src/Cli/func/Common/ProcessManager.cs
liliankasem: Reverted this change, turns out no we don't want to pass 'true' here as we end up with the process not exiting

src/Cli/func/Program.cs
liliankasem: Discussed offline, feedback addressed

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        _onShuttingDown = onShuttingDown ?? throw new ArgumentNullException(nameof(onShuttingDown));
+        _onGracePeriodTimeout = onGracePeriodTimeout ?? (() => { });
+
+        Console.CancelKeyPress += HandleCancelKeyPress;
+        _registered = true;
+    }
+
+    internal static void HandleCancelKeyPress(object sender, ConsoleCancelEventArgs e)
+    {
+        _processManager.KillChildProcesses();
```
liliankasem: I wonder if we want to do this after the 2 second grace period as well?

src/Cli/func/Program.cs
```diff
@@ -32,14 +34,26 @@ internal static void Main(string[] args)
...
+                onGracePeriodTimeout: _forceShutdownCts.Cancel);
 
-            ConsoleApp.Run<Program>(args, _container);
+            try
+            {
+                await ConsoleApp.RunAsync<Program>(args, _container, _shuttingDownCts.Token).WaitAsync(_forceShutdownCts.Token);
+            }
+            catch (OperationCanceledException)
+            {
+                processManager.KillMainProcess();
```
liliankasem: We can also do this as part of our `onGracePeriodTimeout` action, but I figured it was better to keep the `CancelKeyHandler` focused on cancelling the CTs

src/Cli/func/CancelKeyHandler.cs
jviau: Probably. I would expect all the individual `IHost`'s to respond to the CTRL+C themselves (I hope) and exit gracefully if possible. We then only force terminate after 2 seconds.

src/Cli/func/Program.cs
jviau: Agreed on having it focused on canceling CTs. I'd even consider not passing the `processManager` to `CancelKeyHandler` at all, and having all usages of it in this context done in the catch block.

src/Cli/func/Program.cs
```diff
@@ -32,14 +34,26 @@ internal static void Main(string[] args)
...
+                processManager: processManager,
+                onShuttingDown: _shuttingDownCts.Cancel,
+                onGracePeriodTimeout: _forceShutdownCts.Cancel);
 
-            ConsoleApp.Run<Program>(args, _container);
+            try
+            {
+                await ConsoleApp.RunAsync<Program>(args, _container, _shuttingDownCts.Token).WaitAsync(_forceShutdownCts.Token);
+            }
+            catch (OperationCanceledException)
```
jviau: We may only want to call `processManager` when it is specifically the `_forceShutdownCts` that is cancelled.

Consider:


```suggestion
            catch (OperationCanceledException ex)
            {
                if (ex.CancellationToken == _forceShutdownCts.Token)
                {
                    // processManager stuff
                }
            }
```

jviau (approved):

src/Cli/func/Common/ProcessManager.cs
aishwaryabh, line 31: Not blocking comment: just read the other comment Matthew left but just wanted to ask why does passing in `Kill(true)` cause the process to not terminate? I thought that would kill all child processes too so just wondering!

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        _ = Task.Run(async () =>
+        {
+            await Task.Delay(_gracefulShutdownPeriod);
+            _onGracePeriodTimeout?.Invoke();
+        });
+    }
+
+    // For testing purposes, we need to ensure that the handler can be disposed of properly.
+    internal static void Dispose()
```
fabiocav, line 47: nit: use a different method name here to avoid confusion with an actual `IDisposable` implementation.

src/Cli/func/CancelKeyHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+    private static readonly TimeSpan _gracefulShutdownPeriod = TimeSpan.FromSeconds(2);
+    private static readonly ConsoleCancelEventHandler _handlerDelegate = HandleCancelKeyPress;
+    private static Action _onShuttingDown;
+    private static Action _onGracePeriodTimeout;
+    private static bool _registered = false;
+    private static bool _shutdownStarted = false;
+
+    public static void Register(Action onShuttingDown = null, Action onGracePeriodTimeout = null)
+    {
+        if (_registered)
```
fabiocav, line 17: This logic may lead a caller to believe their callback was registered when it wasn't. If you're not supporting multiple registration, either return a boolean from this method or throw. 

Currently, calling this method multiple times with different shutdown handlers would result in surprising behavior.

src/Cli/func/Program.cs
```diff
@@ -29,17 +31,28 @@ internal static void Main(string[] args)
...
             {
-                _container.Resolve<IProcessManager>()?.KillChildProcesses();
-            };
-
-            ConsoleApp.Run<Program>(args, _container);
+                await ConsoleApp.RunAsync<Program>(args, _container, _shuttingDownCts.Token).WaitAsync(_forceShutdownCts.Token);
+            }
+            catch (OperationCanceledException ex)
+            {
+                if (ex.CancellationToken == _forceShutdownCts.Token)
```
fabiocav, line 46: What behavior difference are you observing from this calls? Are you seeing hangs after `RunAsync` returns? I would expect this to simply shutdown if we ever reach this point.

src/Cli/func/Common/ProcessManager.cs
liliankasem: Honestly not 100% sure what was going on, but passing in true lead to the process hanging and ever exiting.

src/Cli/func/Common/ProcessManager.cs
```diff
@@ -26,6 +26,11 @@ public IEnumerable<IProcessInfo> GetProcessesByName(string processName)
                 .Select(p => new ProcessInfo(p));
         }
 
+        public void KillMainProcess()
```
fabiocav, line 29: Still not a fan of this as an API on `IProcessManager`.

liliankasem: Can you share more? I can just call `Process.GetCurrentProcess().Kill();` directly without it being in this interface, but would like to understand why.

Context: I originally added it to the interface here because I was using the process manager within the CancelKeyHandler, but the code has been refactored since, leaving the control with the registered actions instead. I was also mocking the behaviour for testing which is why I wanted it to be part of an interface.

src/Cli/func/Program.cs
liliankasem: Yes, so before these changes, it would still wait for the host to shut down - even after we exit here. I would also expect it to shut down at this point but Jacob and I could not figure out why it was not exiting like we would expect on the main process exiting. So we pivoted to a graceful shutdown period instead as this felt safe and predictable. On a healthy host shut down, it shuts down in ms and the timeout is never reached. If some thing weird is happening (like the worker init bug in the host), we are still giving the host plenty of time to gracefully shut down before we force exit.

fabiocav: Are you able to repro this? Can you capture what is still running when you reach this point in the code? Ideally, your call to `RunAsync` wouldn't return until the host has completely shut down.

src/Cli/func/Common/ProcessManager.cs
fabiocav: `IProcessManager` does not manage the lifetime of the main process. I looked for this being mocked in tests, which would have helped better understand the motivation, but didn't see anything leveraging this.

Ultimately, it comes down to the role of `IProcessManager`: This is an abstraction for the service responsible for managing child processes launched by Core Tools, not for the main application lifecycle/process. Looks like there is some general process logic that was already bleeding into that interface, but we shouldn't expand on that.

liliankasem: > but didn't see anything leveraging this.

must have also been removed with the refactor - I can revert this change

src/Cli/func/Program.cs
liliankasem: Before or after these changes?  But yes, it is simple to repro:

1. Create a dni app
2. Run `func start --verbose`
3. When you see the log `Fetching metadata for workerRuntime: dotnet-isolated` immediately hit Ctrl+C

`RunAsync` does not return until the host fully shuts down. 

Before the changes, `RunAsync` does not return until the host fully shuts down. And we end up waiting a minute for the timeout period. Then we end up in the finally block here and exit: 

https://github.com/Azure/azure-functions-core-tools/blob/16b9e2baccf97cc6d8f86f772f016d8d72571cbb/src/Cli/func/ConsoleApp.cs#L116

With this change, `RunAsync` doesn't return because the cancellation token (with the timeout period) sends us here:

https://github.com/Azure/azure-functions-core-tools/blob/16b9e2baccf97cc6d8f86f772f016d8d72571cbb/src/Cli/func/Program.cs#L46

Where we kill the process. If I comment out the code to kill the process, we simple exit the main prcoess, but the host process is still running and trying to shut down. This leads to us still waiting a minute before we exit. In a happy case, `RunAsync` does not return until the host fully shuts down.  And we end up here:

https://github.com/Azure/azure-functions-core-tools/blob/16b9e2baccf97cc6d8f86f772f016d8d72571cbb/src/Cli/func/ConsoleApp.cs#L116

fabiocav: > we simple exit the main prcoess, but the host process is still running and trying to shut down.

To clarify:

- Are you seeing this in the in-proc delegation scenarios only (where Core Tools launches the in-proc version)? Otherwise, there should be no separate host process. The Core Tools (main) process and the host process are the same.
- When you mention the main process exits, is that truly the process (the Core Tools process exits) or are you referring to exiting from the main method with the process still running and in a hung state?

liliankasem: 1. No, I only see this for out-of-proc. For in-proc we spin up a new process for the in-proc host/func cli so this issue doesn't occur, we simply just kill the child proces.
2. Yeah this: `exiting from the main method with the process still running and in a hung state`

fabiocav: >No, I only see this for out-of-proc. For in-proc we spin up a new process for the in-proc host/func cli so this issue doesn't occur, we simply just kill the child proces.

OK, if that's for OOP only, then we need more clarity here, as there are no distinct processes for Core Tools and the Host. Let's follow up and update this. It would be helpful to look at what is still running (active threads, particularly foreground) are still running when you reach this condition.

PR #4550, opened 2025-07-28
Add `--build-local option` for `func pack` and default to local build for Python apps on windows

aishwaryabh:

Resolves [#4551](https://github.com/Azure/azure-functions-core-tools/issues/4551)
After having a discussion with the python team, we decided that it makes sense for the default build option to be remote, as this was done in [this PR](https://github.com/Azure/azure-functions-core-tools/pull/4530/files#diff-7a6ec14c3ba2de03fa733467285f0c19c45b456051b7ecaf03f9d7abec4afe45). However this may be a regression in Windows for those using GH Actions or ADO pipelines, as the customer may want to keep the default build to be local. 

This PR therefore reverts the behavior back to defaulting to a local build if the customer the function app is a python runtime with a requirements.txt and is on a windows machine. We warn the customer that in the next release, we will have all apps default to using a remote build and they may enable a local build by using the `--build-local` flag. 

This PR also adds additional logging as specified by this doc:
image

We now also have a deferred build which is ready for remote deployment when running func pack as well.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -11,5 +11,8 @@
 - Disable diagnostic events in local development by replacing the `IDiagnosticEventRepository` with a `DiagnosticEventNullRepository` (#4542)
 - Add `func pack` support for in-proc functions (#4529)
 - Default to remote build for `func pack` for python apps (#4530)
-- Add   build-local` option for `func pack` and default to local buidl for Python apps on windows with warning (#4550)
+- Add `--build-local` option for `func pack` for Python apps only. (#4550)
+  - Default to local build for Python apps on windows to avoid breaking existing behavior
+  - Update logging for `func pack` to include more details
+  - Include a 'Deferred' build option for `func pack` that is ready for remote deployment
 
--- a/release_notes.md
+++ b/release_notes.md
@@ -11,3 +11,5 @@
 - Disable diagnostic events in local development by replacing the `IDiagnosticEventRepository` with a `DiagnosticEventNullRepository` (#4542)
 - Add `func pack` support for in-proc functions (#4529)
 - Default to remote build for `func pack` for python apps (#4530)
+- Add   build-local` option for `func pack` and default to local buidl for Python apps on windows with warning (#4550)
+
--- a/src/Cli/func/Actions/LocalActions/PackAction.cs
+++ b/src/Cli/func/Actions/LocalActions/PackAction.cs
@@ -101,10 +101,12 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 if (workerRuntime != WorkerRuntime.Python)
                 {
-                    throw new CliException("The --build-local option is only applicable for Python function apps.");
+                    ColoredConsole.WriteLine(WarningColor("The --build-local option is only applicable for Python function apps."));
                 }
-
-                defaultBuildOption = BuildOption.Local;
+                else
+                {
+                    defaultBuildOption = BuildOption.Local;
+                }
             }
 
             // Resolve build option to detect if remote build is needed
--- a/src/Cli/func/Actions/LocalActions/PackAction.cs
+++ b/src/Cli/func/Actions/LocalActions/PackAction.cs
@@ -31,6 +31,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
         public bool Squashfs { get; private set; }
 
+        public bool BuildLocal { get; private set; } = false;
+
         public override ICommandLineParserResult ParseArgs(string[] args)
         {
             Parser
@@ -53,6 +55,10 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             Parser
                 .Setup<bool>("squashfs")
                 .Callback(f => Squashfs = f);
+            Parser.
+                Setup<bool>("build-local")
+                .WithDescription("Builds the function app locally instead of remotely. This is only applicable for Python function apps.")
+                .Callback(f => BuildLocal = f);
 
             if (args.Any() && !args.First().StartsWith("-"))
             {
@@ -89,8 +95,21 @@ namespace Azure.Functions.Cli.Actions.LocalActions
 
             var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
 
-            // Resolve build option to detect if remote build is needed (e.g., Python app on Windows)
-            var buildOption = ResolveBuildOptionHelper.ResolveBuildOption(BuildOption.Default, workerRuntime, site: null, BuildNativeDeps, noBuild: false);
+            var defaultBuildOption = BuildOption.Default;
+
+            if (BuildLocal)
+            {
+                if (workerRuntime != WorkerRuntime.Python)
+                {
+                    throw new CliException("The --build-local option is only applicable for Python function apps.");
+                }
+
+                defaultBuildOption = BuildOption.Local;
+            }
+
+            // Resolve build option to detect if remote build is needed
+            // IncludeLocalBuildForWindows is set to true to ensure that local builds are used for Python on Windows
+            var buildOption = ResolveBuildOptionHelper.ResolveBuildOption(defaultBuildOption, workerRuntime, site: null, BuildNativeDeps, noBuild: false, includeLocalBuildForWindows: true);
 
             outputPath += Squashfs ? ".squashfs" : ".zip";
             if (FileSystemHelpers.FileExists(outputPath))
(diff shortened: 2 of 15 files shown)


src/Cli/func/Actions/LocalActions/PackAction.cs
```diff
@@ -89,8 +95,21 @@ public override async Task RunAsync()
...
 
-            // Resolve build option to detect if remote build is needed (e.g., Python app on Windows)
-            var buildOption = ResolveBuildOptionHelper.ResolveBuildOption(BuildOption.Default, workerRuntime, site: null, BuildNativeDeps, noBuild: false);
+            var defaultBuildOption = BuildOption.Default;
+
+            if (BuildLocal)
+            {
+                if (workerRuntime != WorkerRuntime.Python)
+                {
+                    throw new CliException("The --build-local option is only applicable for Python function apps.");
```
liliankasem: Do we need to throw? Would a warning be sufficient here if we are able to build still with whatever the default is for that language?

aishwaryabh: hmm that's valid. I just didn't want to confuse the user by proceeding with the build, but I think if we add a warning that should be ok

gavin-aguiar (approved):

src/Cli/func/Actions/LocalActions/PackAction.cs
```diff
@@ -89,8 +95,23 @@ public override async Task RunAsync()
...
 
-            // Resolve build option to detect if remote build is needed (e.g., Python app on Windows)
-            var buildOption = ResolveBuildOptionHelper.ResolveBuildOption(BuildOption.Default, workerRuntime, site: null, BuildNativeDeps, noBuild: false);
+            var defaultBuildOption = BuildOption.Default;
+
+            if (BuildLocal)
+            {
+                if (workerRuntime != WorkerRuntime.Python)
+                {
+                    ColoredConsole.WriteLine(WarningColor("The --build-local option is only applicable for Python function apps."));
```
liliankasem: Should we say something like `defaulting to ??`

src/Cli/func/Actions/LocalActions/PackAction.cs
```diff
@@ -89,8 +95,23 @@ public override async Task RunAsync()
 
             var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager);
 
-            // Resolve build option to detect if remote build is needed (e.g., Python app on Windows)
-            var buildOption = ResolveBuildOptionHelper.ResolveBuildOption(BuildOption.Default, workerRuntime, site: null, BuildNativeDeps, noBuild: false);
+            var defaultBuildOption = BuildOption.Default;
+
+            if (BuildLocal)
```
liliankasem: can also be:

```csharp
if (BuildLocal && workerRuntime == WorkerRuntime.Python)
{
    defaultBuildOption = BuildOption.Local;
}
else if (BuildLocal)
{
    ColoredConsole.WriteLine(WarningColor("The --build-local option is only applicable for Python function apps."));
}
```

src/Cli/func/Helpers/ResolveBuildOptionHelper.cs
```diff
@@ -33,6 +30,17 @@ public static BuildOption ResolveBuildOption(BuildOption currentBuildOption, Wor
                     FileSystemHelpers.FileExists(Constants.RequirementsTxt) &&
                     new FileInfo(Path.Combine(Environment.CurrentDirectory, Constants.RequirementsTxt)).Length > 0)
                 {
+                    // Include default (local) build option for Windows as some customers may be using a local build for testing purposes
+                    // Note that this will be deprecated in the future after 4.1.1 is released
```
liliankasem: What is being deprecated? Defaulting to local build for windows?

src/Cli/func/Helpers/ResolveBuildOptionHelper.cs
```diff
@@ -33,6 +30,17 @@ public static BuildOption ResolveBuildOption(BuildOption currentBuildOption, Wor
...
+                    if (includeLocalBuildForWindows && OperatingSystem.IsWindows())
+                    {
+                        ColoredConsole.WriteLine("Python runtime detected on Windows. Using local build option.");
+                        ColoredConsole.WriteLine(OutputTheme.WarningColor($"The default build option for python function apps with a valid requirements.txt " +
+                            $"will be switched over to use remote builds after version 4.1.1.\n" +
+                            "If a local build is still needed, please use the `--build-local` flag when running `func pack`."));
+                        return BuildOption.Default;
```
liliankasem: I'm not sure about this messaging and the discoverability of it. @mattchenderson do you have any thoughts or ideas on this?

src/Cli/func/Helpers/ZipHelper.cs
```diff
@@ -26,9 +26,16 @@ public static async Task<Stream> GetAppZipFile(string functionAppRoot, bool buil
             }
             else if (buildOption == BuildOption.Remote)
             {
-                ColoredConsole.WriteLine(DarkYellow("Performing remote build for functions project."));
+                if (isPackAction)
```
liliankasem: Are we only adding this for a log satement? Does this flag have any impact on how this method runs?

test/Cli/Func.E2ETests/Commands/FuncPack/BasePackTests.cs
```diff
@@ -11,8 +12,10 @@ namespace Azure.Functions.Cli.E2ETests.Commands.FuncPack
 {
     internal static class BasePackTests
     {
-        internal static void TestBasicPackFunctionality(string workingDir, string testName, string funcPath, ITestOutputHelper log, string[] filesToValidate)
+        internal static void TestBasicPackFunctionality(string workingDir, string testName, string funcPath, ITestOutputHelper log, string[] filesToValidate, bool shouldHaveLocalBuildLogs = false)
         {
+            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
liliankasem: `isWindows` is never used?

release_notes.md
```diff
@@ -10,4 +10,6 @@
 - Fix dotnet templates installation (#4538)
 - Disable diagnostic events in local development by replacing the `IDiagnosticEventRepository` with a `DiagnosticEventNullRepository` (#4542)
 - Add `func pack` support for in-proc functions (#4529)
-- Default to remote build for `func pack` for python apps (#4530)
\ No newline at end of file
+- Default to remote build for `func pack` for python apps (#4530)
+- Add   build-local` option for `func pack` and default to local buidl for Python apps on windows with warning (#4550)
```
```suggestion
- Add  `--build-local` option for `func pack` for Python apps only. (#4550)
  - Default to local build for Python apps on windows to avoid breaking existing behaviour. 
```

If this is going to change, we should mention here that this is temporary? I'm not sure how we can deprecate this in a clean way

src/Cli/func/Helpers/ResolveBuildOptionHelper.cs
aishwaryabh: yup let me make the comment more clear

src/Cli/func/Helpers/ResolveBuildOptionHelper.cs
mattchenderson: This feels like a break, so I'm going to need to consider this. I'll review in more detail later. One question I'd have is if this is only in `func pack`, or if it's possible to encounter it through other commands, given that this is in a helper.

Regardless, I like making a forward-looking statement if we are going to change something here. I would just want to have it running for longer than one release.

aishwaryabh: This would only impact `func pack` given the flag `isFuncPackAction`. This has been validated in unit tests too

Sure I think it makes sense to have this be more than one release. Should I remove the specific version now for now?

src/Cli/func/Actions/LocalActions/PackAction.cs
aishwaryabh: Discussed offline and decided to say `Local build setting not applied.`

src/Cli/func/Helpers/ZipHelper.cs
aishwaryabh: removed flag

release_notes.md
aishwaryabh: Added more details! Agreed on this point; will set up some time to discuss with the team

src/Cli/func/Helpers/ResolveBuildOptionHelper.cs
jviau: As discussed in our meeting, this is definitely a breaking change, and we should reconsider this. Customers which have an existing CI/CD split between `func pack` and az cli deployment that upgrade their core tools version and leave their build steps as is will see deployment fail, as they will not be passing in the remote build flag to az cli.

aishwaryabh:
Closing this PR until we have clarity on the direction we want to go for `func pack`

PR #4563, opened 2025-07-31
Overwrite AZURE_FUNCTIONS_ENVIRONMENT if it exists

### LINKED ISSUE #2465: An item with the same key has already been added. Key: AZURE_FUNCTIONS_ENVIRONMENT

liliankasem:

resolves #2465
### Pull request checklist

### Additional information

Due to potential unexpected side effects, we cannot allow users to set `AZURE_FUNCTIONS_ENVIRONMENT`, but I've added a check in place to avoid the exception when there is a duplicate value, ensuring we set it to Development.

### DIFF
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -273,7 +273,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
                 .Build();
         }
 
-        private async Task<IDictionary<string, string>> GetConfigurationSettings(string scriptPath, Uri uri)
+        internal async Task<IDictionary<string, string>> GetConfigurationSettings(string scriptPath, Uri uri)
         {
             var settings = _secretsManager.GetSecrets();
             settings.Add(Constants.WebsiteHostname, uri.Authority);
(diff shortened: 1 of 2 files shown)


fabiocav (approved):

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -298,9 +298,14 @@ private async Task<IDictionary<string, string>> GetConfigurationSettings(string
...
-            // when running locally in CLI we want the host to run in debug mode
-            // which optimizes host responsiveness
-            settings.Add("AZURE_FUNCTIONS_ENVIRONMENT", "Development");
+            // When running locally in CLI we want the host to run in debug mode which optimizes host responsiveness
+            // We intentionally override the value of AZURE_FUNCTIONS_ENVIRONMENT to Development if it is already set to something else.
+            if (settings.TryGetValue("AZURE_FUNCTIONS_ENVIRONMENT", out var oldValue))
+            {
+                ColoredConsole.WriteLine(WarningColor($"AZURE_FUNCTIONS_ENVIRONMENT already exists with value '{oldValue}', overriding to 'Development'."));
+            }
+            settings["AZURE_FUNCTIONS_ENVIRONMENT"] = "Development";
```
jviau: Do customers have any way of using non-development `AZURE_FUNCTIONS_ENVIRONMENT` with core tools? This is useful for testing.

liliankasem: Not that I am aware of, we can open an issue to track that if we think it's a desired feature but going to call it out of scope for this pr.

satvu (approved):

PR #4565, opened 2025-07-31
Fix flakey tests

### LINKED ISSUE #4545: Fix flakey tests
There are a couple flakey tests that should be addressed.

**Unit**

- [HostHelperTests](https://github.com/Azure/azure-functions-core-tools/blob/main/test/Cli/Func.UnitTests/HelperTests/HostHelperTests.cs)
  - My guess is it's something to do with the file system access 

**E2E**

- Any `func new` test for workers that use bundles
  - The failure is on: `Templates JSON is empty. Please check the templates location.`
  - This is coming from the templates manager [here](https://github.com/Azure/azure-functions-core-tools/blob/e963a7607b71a06d8cd2780833c5487fa50c511c/src/Cli/func/Common/TemplatesManager.cs#L88) or [here](https://github.com/Azure/azure-functions-core-tools/blob/e963a7607b71a06d8cd2780833c5487fa50c511c/src/Cli/func/Common/TemplatesManager.cs#L297)

[Here's](https://azfunc.visualstudio.com/public/_build/results?buildId=228472&view=logs&j=da439e82-8265-5962-85f8-172d6600cb6f&t=17f2655d-abad-5398-83c4-ac47d6b8119e) a CI run that shows both of these failures.

liliankasem:

resolves #4545

### Pull request checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,7 +13,7 @@
             },
             "args": "${input:funcArgs} --script-root ${input:scriptRootArg}",
             "cwd": "${workspaceFolder}/src/Cli/func",
-            "console": "internalConsole",
+            "console": "integratedTerminal",
             "stopAtEntry": false
         },
         {
--- a/src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
+++ b/src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
@@ -18,7 +18,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             return base.ParseArgs(args);
         }
 
-        public override Task RunAsync()
+        public override async Task RunAsync()
         {
             var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
             if (extensionBundleManager.IsExtensionBundleConfigured())
@@ -26,7 +26,9 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 try
                 {
                     var options = ExtensionBundleHelper.GetExtensionBundleOptions();
-                    var bundlePath = ExtensionBundleHelper.GetBundleDownloadPath(options.Id);
+                    var bundleBasePath = ExtensionBundleHelper.GetBundleDownloadPath(options.Id);
+                    var bundleDetails = await extensionBundleManager.GetExtensionBundleDetails();
+                    var bundlePath = Path.Combine(bundleBasePath, bundleDetails.Version);
 
                     if (string.IsNullOrEmpty(bundlePath))
                     {
@@ -46,8 +48,6 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 ColoredConsole.WriteLine("Extension bundle not configured.");
             }
-
-            return Task.CompletedTask;
         }
     }
 }
--- a/src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
+++ b/src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
@@ -18,14 +18,16 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             return base.ParseArgs(args);
         }
 
-        public override async Task RunAsync()
+        public override Task RunAsync()
         {
             var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
             if (extensionBundleManager.IsExtensionBundleConfigured())
             {
                 try
                 {
-                    string bundlePath = await extensionBundleManager.GetExtensionBundlePath();
+                    var options = ExtensionBundleHelper.GetExtensionBundleOptions();
+                    var bundlePath = ExtensionBundleHelper.GetBundleDownloadPath(options.Id);
+
                     if (string.IsNullOrEmpty(bundlePath))
                     {
                         throw new CliException("Unable to locate extension bundle.");
@@ -44,6 +46,8 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 ColoredConsole.WriteLine("Extension bundle not configured.");
             }
+
+            return Task.CompletedTask;
         }
     }
 }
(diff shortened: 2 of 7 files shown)


src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -59,6 +60,8 @@ public static async Task GetExtensionBundle()
             }
             catch (Exception)
             {
+                ColoredConsole.WriteLine("Failed to download the Extension Bundle. Please check your network connection and try again.");
```
kshyju: Should we wrap the network call in a retry mechanism so that a transient network issue is not throwing the error?

src/Cli/func/Common/TemplatesManager.cs
```diff
@@ -69,7 +69,10 @@ private static async Task<IEnumerable<Template>> GetTemplates()
             var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
             string templatesJson;
 
-            if (extensionBundleManager.IsExtensionBundleConfigured())
+            var configured = extensionBundleManager.IsExtensionBundleConfigured();
+            ColoredConsole.WriteLine("extensionBundleManager.IsExtensionBundleConfigured() = " + configured);
```
kshyju: Will users see this message in their terminal? Should this be behind the verbose flag?

liliankasem: Sorry this is just for me debugging the issue, not intended to be merged.

aishwaryabh (approved):

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -49,13 +55,23 @@ public static string GetBundleDownloadPath(string bundleId)
         public static async Task GetExtensionBundle()
         {
             var extensionBundleManager = GetExtensionBundleManager();
+            var bundlePath = await extensionBundleManager.GetExtensionBundlePath();
+
+            if (Directory.Exists(bundlePath) && Directory.GetFiles(bundlePath, "*", SearchOption.AllDirectories).Any())
```
aishwaryabh: should we write a quick unit test for this use case?

liliankasem: done!

liliankasem: Actualy removing this code because `await extensionBundleManager.GetExtensionBundlePath();` already does this

test/Cli/Func.UnitTests/HelperTests/ExtensionBundleHelperTests.cs
```diff
@@ -1,7 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.IO.Abstractions;
```
aishwaryabh: do we need these dependencies?

liliankasem: nope, leftovers from the changes I undid - I'll remove them!

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -49,13 +55,17 @@ public static string GetBundleDownloadPath(string bundleId)
...
             var extensionBundleManager = GetExtensionBundleManager();
+
             try
             {
-                using (var httpClient = new HttpClient())
-                {
-                    httpClient.Timeout = TimeSpan.FromMinutes(10);
-                    await extensionBundleManager.GetExtensionBundlePath(httpClient);
-                }
+                using var httpClient = new HttpClient { Timeout = _httpTimeout };
```
kshyju, line 61: The `GetExtensionBundle` method is called from `StartHostAction.RunAsync`. I am wondering that a 10 minute HTTP timeout might be too long. If the request hangs, users running `func host start` command could be left waiting without feedback and may think the CLI is unresponsive. Since we already have retries in place, maybe we could consider reducing the timeout to fail faster and improve the user experience?

src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
```diff
@@ -25,7 +25,11 @@ public override async Task RunAsync()
             {
                 try
                 {
-                    string bundlePath = await extensionBundleManager.GetExtensionBundlePath();
+                    var options = ExtensionBundleHelper.GetExtensionBundleOptions();
```
kshyju, line 28: I thought `ExtensionBundleManager.GetExtensionBundlePath` was meant to be the single source of truth for getting the extension bundle path. Is there a reason we are not using it here? Is that method not returning the expected output in some scenarios? Trying to understand the switch to the new logic here.

.vscode/launch.json
```diff
@@ -13,7 +13,7 @@
             },
             "args": "${input:funcArgs} --script-root ${input:scriptRootArg}",
             "cwd": "${workspaceFolder}/src/Cli/func",
-            "console": "internalConsole",
+            "console": "integratedTerminal",
```
kshyju, line 16: Is this change unrelated to flaky test fix?

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
liliankasem: good point, I'll lower it to maybe 2-3 mins?

src/Cli/func/Actions/LocalActions/GetExtensionBundlePath.cs
liliankasem: The problem is that `GetExtensionBundlePath` will also install templates if they don't exist. Given this action is just to return the path itself it doesn't make sense for it to also download the bundles. I believe this is also one of the contributors for the flakeyness happening in the e2e tests. 

Second, `GetExtensionBundlePath` uses the download path that it gets from core tools so this is just shortcuriting all the download bundle stuff that we don't need here.

.vscode/launch.json
liliankasem: No, just an enhacement for the repro in general - didn't feel like it was worth its own pr

aishwaryabh (approved):

Issue #4593, opened 2025-08-11
Clean up `func pack` command and shared args for each stack


aishwaryabh:
Clean up for `func pack` command to include the following arguments that will be shared amongst all stacks

List of args:

- --output 
     - Path of where output zip file should be located after running func pack 

- --no-build 
    - Skip running build for specific language if it is required 

Create internal subcommands for the following worker runtimes as well:

- [ ] .NET
- [ ] Python
- [ ] Powershell
- [ ] Custom handlers
- [ ] Node

aishwaryabh:
Closed by #4600

PR #4600, opened 2025-08-12
Refactor `func pack`

aishwaryabh:

resolves[#4593](https://github.com/Azure/azure-functions-core-tools/issues/4593)

This issue allows users to run `func pack` when they want to create a zip file that they can use the azure CLI to deploy. The general syntax of the command is as follows:

`func pack [path of folder to pack] [language specific options] [--no-build] [--output]`

When the user runs `func --help` the following is displayed:
image

Note that `func pack` will NOT be displayed when the user does `func --help` until we clean up the CLI and add all the validations for each command when running `func pack`. This will be done in a future PR

This PR adds the applicable E2E tests for the different scenarios and for each language worker runtime as well.

### Pull request checklist


src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public async Task RunAsync(PackOptions packOptions)
+        {
+            var functionAppRoot = PackHelpers.ResolveFunctionAppRoot(packOptions.FolderPath);
+            string packingRoot = functionAppRoot;
+
+            if (packOptions.NoBuild)
+            {
+                // For --no-build, treat FolderPath as the build output directory
+                if (string.IsNullOrEmpty(packOptions.FolderPath))
```
liliankasem: Is this saying a path must be provided to use --no-build? Or do we want to default to currect dir?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,120 @@
...
+                await installExtensionAction.RunAsync();
+            }
+
+            await PackHelpers.CreatePackage(packingRoot, outputPath, packOptions.NoBuild, TelemetryCommandEvents);
+        }
+
+        public override Task RunAsync()
+        {
+            // Keep this in case the customer tries to run func pack dotnet, since this subcommand is not meant to be run directly.
+            throw new InvalidOperationException("Invalid command. Please run func pack instead with valid arguments. To see a list of valid arguments, please see func --help.");
```
liliankasem: They should be able to run `func pack --help`

Also do we want to throw here? I know we want folks to just call func pack, but that doesnt mean func pack dotnet should break?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private void ValidateDotNetPublishDirectory(string path)
+        {
+            var requiredFiles = new[] { "host.json" };
+            foreach (var file in requiredFiles)
+            {
+                if (!FileSystemHelpers.FileExists(Path.Combine(path, file)))
+                {
+                    throw new CliException($"Required file {file} not found in build output directory: {path}");
```
liliankasem:
```suggestion
                    throw new CliException($"Required file '{file}' not found in build output directory: {path}");
```

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,120 @@
...
+            var outputPath = Path.Combine(functionAppRoot, "output");
+
+            // Clean the output directory if it exists
+            if (FileSystemHelpers.DirectoryExists(outputPath))
+            {
+                FileSystemHelpers.DeleteDirectorySafe(outputPath);
+            }
+
+            // Run dotnet publish
+            var exe = new Executable("dotnet", $"publish --configuration Release --output \"{outputPath}\"", workingDirectory: functionAppRoot);
```
liliankasem: I think the default for publish is `Release` config so probably don't need this

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,120 @@
...
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Interfaces;
+using Colors.Net;
+using Fclp;
+using static Azure.Functions.Cli.Common.OutputTheme;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack dotnet", ParentCommandName = "pack", ShowInHelp = false, HelpText = ".NET specific arguments")]
```
liliankasem: Can we include better help text?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: For `func pack --help` do we want to add this functionality after? I don't think `func <command> --help` works for any of the commands today so I think we would potentially have to refactor how the help command works today to accommodate for this

I'd initially decided on throwing because we actually run this `RunAsync` method [here](https://github.com/Azure/azure-functions-core-tools/pull/4600/files#diff-d5154e7f9bf52956d6d7205497a03edd625dc00e7f10b3bf4ffcbca81440ec9aR29), which requires `PackOptions` to be passed in. The method is called from `PackAction` [here](https://github.com/Azure/azure-functions-core-tools/pull/4600/files#diff-c652e520ae8eee7ec5871c06c34768b8e8b66c193c449a34a190a9d69ed66ed0R82) from the parent command for obtaining the initial args and letting the subcommand process them.

I had tried passing in `PackAction` into the constructor directly of `DotnetPackSubommand` but `func --help` starts complaining about `PackAction` not being properly injected into the constructor. Any thoughts on how to get around this or if we should keep still throwing?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: ooo we can try default dir. Maybe I can add a log letting the user know we're zipping the current dir?

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,121 @@
...
+                OutputPath = OutputPath,
+                NoBuild = NoBuild,
+                PreserveExecutables = PreserveExecutables
+            };
+
+            // Internally dispatch to runtime-specific subcommand
+            await RunRuntimeSpecificPack(workerRuntime, packOptions);
+        }
+
+        private async Task RunRuntimeSpecificPack(WorkerRuntime runtime, PackOptions packOptions)
```
liliankasem: Any method that uses async/await should end with `Async`

```suggestion
        private async Task RunRuntimeSpecificPackAsync(WorkerRuntime runtime, PackOptions packOptions)
```

liliankasem: Can use switch expression 

```csharp
private async Task RunRuntimeSpecificPack(WorkerRuntime runtime, PackOptions packOptions) =>
    runtime switch
    {
        WorkerRuntime.DotNet => DoDotNetPackAsync(packOptions),
        WorkerRuntime.Node => DoNodePackAsync(packOptions),
        WorkerRuntime.Python => DoPythonPackAsync(packOptions),
        _ => //default behaviour
    };
```

Or maybe `new DotnetPackSubcommandAction(_secretsManager).RunAsync(packOptions)` would work

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (FileSystemHelpers.FileExists(outputPath))
+            {
+                ColoredConsole.WriteLine($"Deleting the old package {outputPath}");
+                try
+                {
+                    FileSystemHelpers.FileDelete(outputPath);
+                }
+                catch (Exception)
+                {
+                    throw new CliException($"Could not delete {outputPath}");
```
liliankasem: Do we want to hard fail if we can't delete? What's the worst that can happen in this scenario?

src/Cli/func/Actions/LocalActions/PackAction/PackOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+
```
liliankasem: remove line

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public bool NoBuild { get; set; }
+
+        public string[] PreserveExecutables { get; set; } = Array.Empty<string>();
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<string>('o', "output")
+                .WithDescription("output path for the packed archive")
```
liliankasem:
```suggestion
                .WithDescription("Specifies the file path where the packed ZIP archive will be created.")
```

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,121 @@
...
+                .Callback(o => OutputPath = o);
+
+            Parser
+                .Setup<bool>("no-build")
+                .WithDescription("Skip running build for specific language if it is required")
```
liliankasem:
```suggestion
                .WithDescription("Do not build the project before packaging.")
```

If they need to provide a path with this arg we should point that out as well.

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,121 @@
...
+                .Callback(n => NoBuild = n);
+
+            Parser
+                .Setup<string>("preserve-executables")
+                .WithDescription("Comma separated list of executables to indicate which bits are to be set as executable in the zip file.")
```
liliankasem:
```suggestion
                .WithDescription("Comma-separated list of executable files to specify which files should be set as executable in the zip archive.")
```

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,121 @@
...
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Interfaces;
+using Colors.Net;
+using Fclp;
+using Microsoft.Azure.WebJobs.Script;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack", HelpText = "Pack function app into a zip that's ready to run.", ShowInHelp = true)]
```
liliankasem: Ready to run or deploy?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
liliankasem: > For func pack --help do we want to add this functionality after? I don't think func <command> --help works for any of the commands today so I think we would potentially have to refactor how the help command works today to accommodate for this

It works today for contexts but not actions? If that makes sense. If it's easy to add, it would be nice to support it. If it's too much work, should be a different PR.

> I'd initially decided on throwing because we actually run this RunAsync method here, which requires PackOptions to be passed in. The method is called from PackAction here from the parent command for obtaining the initial args and letting the subcommand process them.

What happens today if you run a command that doesn't "exist"? Should we maintain the same behaviour?

> I had tried passing in PackAction into the constructor directly of DotnetPackSubommand but func --help starts complaining about PackAction not being properly injected into the constructor. Any thoughts on how to get around this or if we should keep still throwing?

Let me clone and take a look, not sure I am following

aishwaryabh: > It works today for contexts but not actions? If that makes sense. If it's easy to add, it would be nice to support it. If it's too much work, should be a different PR.

Makes sense! I think I might include it in a different PR since it was require refactoring HelpAction

> hat happens today if you run a command that doesn't "exist"? Should we maintain the same behaviour?

This is what happens today:
image
I realized that if the user runs `func pack dotnet` we are also allowing the option to pass in a folder directory where we want to run func pack on, so I think this output makes the most since imo

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: lmk if you can think of a better help text. i didn't make it too detailed for .NET since we don't have any language specific params

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
aishwaryabh: Valid I just added a log warning statement saying we couldn't delete the file

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -211,18 +213,66 @@ private void DisplayActionsHelp(IEnumerable<ActionType> actions)
...
+
+                    ColoredConsole.WriteLine();
+                }
             }
         }
 
+        private void DisplaySubCommandHelp(ActionType subCommand)
+        {
+            // Extract the runtime name from the full command name
+            // E.g., "pack dotnet" -> "Dotnet"
```
swapnil-nagar, line 271: Null check on subCommand.

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -211,18 +213,66 @@ private void DisplayActionsHelp(IEnumerable<ActionType> actions)
...
+            var fullCommandName = subCommand.Names.First();
+            var parts = fullCommandName.Split(' ', StringSplitOptions.RemoveEmptyEntries);
```
swapnil-nagar: subCommand.Names.FirstOrDefault()  would be better?

And what if subCommand.Names.First() is empty or null?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,127 @@
...
+            _secretsManager = secretsManager;
+        }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            // .NET doesn't have any runtime-specific arguments beyond the common ones
+            return base.ParseArgs(args);
+        }
+
+        public async Task RunAsync(PackOptions packOptions)
```
swapnil-nagar, line 30: Null check for packOptions?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            var functionAppRoot = PackHelpers.ResolveFunctionAppRoot(packOptions.FolderPath);
+            string packingRoot = functionAppRoot;
+
+            if (!Directory.Exists(functionAppRoot))
+            {
+                throw new CliException($"Directory not found to pack: {functionAppRoot}");
+            }
+
+            if (packOptions.NoBuild)
```
swapnil-nagar: NIT: Resolving packing root can be its entire new method, for readability.

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,127 @@
...
+using static Azure.Functions.Cli.Common.OutputTheme;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack dotnet", ParentCommandName = "pack", ShowInHelp = false, HelpText = "Arguments specific to .NET apps when running func pack")]
+    internal class DotnetPackSubcommandAction : BaseAction
+    {
+        private readonly ISecretsManager _secretsManager;
+
+        public DotnetPackSubcommandAction(ISecretsManager secretsManager)
```
swapnil-nagar, line 19: Null Checks across the PR

Java Spring has the @NonNull annotation to define parameters as not null, which helps with NPE in the code base and easy to debug. Do we have something similar in .Net? if we do, I will highly encourage the HEAVY use of it.

Java Implementation

```
import org.springframework.lang.NonNull;

public class SomeService{

    public void someService(@NonNull String username) {
      ........
    }
}
```

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
```diff
@@ -0,0 +1,155 @@
...
+using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Interfaces;
+using Colors.Net;
+using Fclp;
+using static Azure.Functions.Cli.Common.OutputTheme;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack node", ParentCommandName = "pack", ShowInHelp = true, HelpText = "Arguments specific to Node.js apps when running func pack")]
+    internal class NodePackSubcommandAction : BaseAction
```
swapnil-nagar: NIT: There are some duplicate code across the languages,

I will suggest to use a abstract class to mode some of the duplicated code.

```
internal abstract class PackSubcommandAction : BaseAction 

internal class NodePackSubcommandAction : PackSubcommandAction

File: src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs [hunk 19]
```diff
@@ -0,0 +1,155 @@
...
+            _secretsManager = secretsManager;
+        }
+
+        public bool SkipInstall { get; set; }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<bool>("skip-install")
+                .WithDescription("Skips running 'npm install' when packing the function app.")
```
swapnil-nagar, line 31: [Question]  We do not have I18N requirement for core tools, correct?

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
```diff
@@ -0,0 +1,155 @@
...
+            try
+            {
+                // Check if package.json has a build script
+                var packageJsonPath = Path.Combine(Environment.CurrentDirectory, "package.json");
+                if (FileSystemHelpers.FileExists(packageJsonPath))
+                {
+                    var packageJsonContent = await FileSystemHelpers.ReadAllTextFromFileAsync(packageJsonPath);
+
+                    // Simple check if build script exists
+                    if (packageJsonContent.Contains("\"build\"") && packageJsonContent.Contains("scripts"))
```
swapnil-nagar, line 118: What is this check at line 122?

src/Cli/func/Actions/LocalActions/PackAction/PythonPackSubcommandAction.cs
```diff
@@ -0,0 +1,72 @@
...
+        public bool BuildNativeDeps { get; set; }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<bool>("build-native-deps")
+                .WithDescription("Builds function app locally using an image that was previously hosted." +
+                " When enabled, core tools launches a docker container with the build env images, builds the function app inside the container," +
+                " and creates a ZIP file with all dependencies restored in .python_packages")
+                .Callback(o => BuildNativeDeps = o);
```
hallvictoria, line 37: slight rewording suggestion:

Builds function app locally using an image that matches the environment used in Azure.
When enabled, Core Tools starts a Docker container, builds the app inside that container, and creates a ZIP file with all dependencies restored in .python_packages.

src/Cli/func/Actions/LocalActions/PackAction/PythonPackSubcommandAction.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            if (!Directory.Exists(functionAppRoot))
+            {
+                throw new CliException($"Directory not found to pack: {functionAppRoot}");
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !BuildNativeDeps)
+            {
+                ColoredConsole.WriteLine(WarningColor("Python function apps is supported only on Linux. Please use the --build-native-deps flag" +
+                    " when building on windows to ensure dependencies are properly restored."));
```
hallvictoria, line 65: nit: "Python function apps are supported only on Linux."

test/Cli/Func.E2ETests/Commands/FuncPack/PythonPackTests.cs
```diff
@@ -116,5 +134,90 @@ public void Pack_PythonFromCache_WorksAsExpected()
...
+
+            var zipFiles = Directory.GetFiles(PythonProjectPath, "*.zip");
+            Assert.True(zipFiles.Length > 0, $"No zip files found in {PythonProjectPath}");
+
+            packResult.Should().ValidateZipContents(
+                zipFiles.First(),
+                new[]
+                {
+                    "host.json",
+                    "requirements.txt"
```
hallvictoria: Should this test still check for `function_app.py`?

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: PackOptions is guaranteed to not be null since we create it [here](https://github.com/Azure/azure-functions-core-tools/pull/4600/files#diff-c652e520ae8eee7ec5871c06c34768b8e8b66c193c449a34a190a9d69ed66ed0R58)

We also already have null checks in place for whatever packOptions is using

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
aishwaryabh: I don't believe we do

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
aishwaryabh: I thought that we have to have "build" defined in package.json to run npm run build. Let me know if that's correct

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: done!

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: added null checks!

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
aishwaryabh: done!

src/Cli/func/Actions/LocalActions/PackAction/PythonPackSubcommandAction.cs
aishwaryabh: done!

test/Cli/Func.E2ETests/Commands/FuncPack/PythonPackTests.cs
aishwaryabh: yup added that

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -211,15 +213,80 @@ private void DisplayActionsHelp(IEnumerable<ActionType> actions)
...
+                    ColoredConsole.WriteLine();
                 }
+            }
+        }
 
-                ColoredConsole.WriteLine();
+        private void DisplaySubCommandHelp(ActionType subCommand)
+        {
+            // Ensure subCommand is valid
+            if (subCommand == null)
```
soninaren: nit :
`(subCommand is null)`

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
swapnil-nagar: That is correct, looks good ignore earlier comment.

hallvictoria (approved):

release_notes.md
```diff
@@ -1,10 +1,9 @@
-# Azure Functions CLI 4.2.2
+# Azure Functions CLI 4.2.3
```
liliankasem: I think new feature = minor

src/Cli/func/Common/SecretsManager.cs
```diff
@@ -36,8 +36,13 @@ public static string AppSettingsFileName
             }
         }
 
-        public IDictionary<string, string> GetSecrets()
+        public IDictionary<string, string> GetSecrets(bool refreshSecrets = false)
         {
+            if (refreshSecrets)
```
liliankasem, line 41: What does refresh secrets have to do with func pack? Does this need to be in this PR?

src/Cli/func/Actions/LocalActions/PackAction/PythonPackSubcommandAction.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public bool BuildNativeDeps { get; set; }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<bool>("build-native-deps")
+                .WithDescription("Builds function app locally using an image that matches the environment used in Azure. " +
+                "When enabled, Core Tools starts a Docker container, builds the app inside that container," +
+                " and creates a ZIP file with all dependencies restored in .python_packages.")
```
liliankasem, line 36: nit: indent desc strings

```suggestion
                  "When enabled, Core Tools starts a Docker container, builds the app inside that container," +
                  " and creates a ZIP file with all dependencies restored in .python_packages.")
```

src/Cli/func/Actions/LocalActions/PackAction/PowershellPackSubcommandAction.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Azure.Functions.Cli.Interfaces;
+using Fclp;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack powershell", ParentCommandName = "pack", ShowInHelp = false, HelpText = "Arguments specific to PowerShell apps when running func pack")]
```
liliankasem, line 10: Is this false because there are no custom pwsh args right now? Can we add a comment that mentions that

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,108 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Helpers;
+using Azure.Functions.Cli.Interfaces;
+using Fclp;
+
+namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
+{
+    [Action(Name = "pack", HelpText = "Pack function app into a zip that's ready to deploy.", ShowInHelp = false)]
```
liliankasem, line 11: I don't think I understood from the PR desc, why are we not showing pack in help?

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            Parser
+                .Setup<string>('o', "output")
+                .WithDescription("Specifies the file path where the packed ZIP archive will be created.")
+                .Callback(o => OutputPath = o);
+
+            Parser
+                .Setup<bool>("no-build")
+                .WithDescription("Do not build the project before packaging. Optionally provide a directory when func pack as the first argument that has the build contents." +
+                "Otherwise, default is the current directory.")
```
liliankasem, line 39: If a line is connected to the one above, you should indent to make that clear

```suggestion
                  "Otherwise, default is the current directory.")
```

src/Cli/func/Actions/LocalActions/PackAction/PackAction.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            // Get the original command line args to pass to subcommands
+            var packOptions = new PackOptions
+            {
+                FolderPath = FolderPath,
+                OutputPath = OutputPath,
+                NoBuild = NoBuild
+            };
+
+            var oldCurrentDirectory = Environment.CurrentDirectory;
```
liliankasem: `originalCurrentDirectory` ?

src/Cli/func/Actions/LocalActions/PackAction/NodePackSubcommandAction.cs
```diff
@@ -0,0 +1,157 @@
...
+            return PackHelpers.CreatePackage(packingRoot, outputPath, options.NoBuild, TelemetryCommandEvents);
+        }
+
+        private async Task RunNodeJsBuildProcess(string functionAppRoot)
+        {
+            // Ensure npm is available
+            EnsureNpmExists();
+
+            // Change to the function app directory for npm operations
+            var previousDirectory = Environment.CurrentDirectory;
```
liliankasem: Let's be consistent across all the commands on what we're calling this

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,126 @@
...
+            PackHelpers.ValidateFunctionAppRoot(functionAppRoot);
+
+            // For --no-build, treat FolderPath as the build output directory
+            if (options.NoBuild)
+            {
+                var packingRoot = functionAppRoot;
+
+                if (string.IsNullOrEmpty(options.FolderPath))
+                {
+                    ColoredConsole.WriteLine(WarningColor("No folder path specified. Using current directory as build output directory."));
```
liliankasem, line 65: For this, can we do more than just defaulting to current? This is probably a good place to check for .props etc.

src/Cli/func/Common/SecretsManager.cs
aishwaryabh: So in the secret manager, the appsettings are lazy loaded. This works for most scenarios, but it doesn't work for `func pack` if the user inputs a directory that they are trying to pack, since the worker runtime is determined by this Init action which is called in ConsoleApp. We could switch directories directly before running this action, but the problem is that we still need to store what the value of the current directory is (in case the user doesn't have an ouput directory where they want the zip file to be located). I also don't think it made sense to have a hardcoded scenario for only if it's func pack, switch directories. Does that make sense?

aishwaryabh: I know that we eventually want to allow all commands to have `func <command> <folder path>` but I thought this was a bit out of the scope of the PR to modify the Init action to account for this for all of the commands (since we only want this to happen for `func pack`), so I decided to have this as a workaround.

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: oo yes that's valid. I can add this scenario

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -261,7 +328,7 @@ private void DisplayPositionalArguments(IEnumerable<CliArgument> arguments)
             longestName += 4; // 4 for coloring and <> characters
             foreach (var argument in arguments)
             {
-                var helpLine = string.Format($"    {{0, {-longestName}}} {{1}}", $"<{argument.Name}>".DarkGray(), argument.Description);
+                var helpLine = string.Format($"{"    "}{{0, {-longestName}}} {{1}}", $"<{argument.Name}>".DarkGray(), argument.Description);
```
soninaren: Can we use standardized indentation through a constant indent size and method that gives an string with set number of empty spaces?

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -311,7 +378,7 @@ private static void DisplayOptions(IEnumerable<ICommandLineOption> options)
                     stringBuilder.Append($" [-{option.ShortName}]");
                 }
 
-                var helpSwitch = string.Format($"    {{0, {-longestName}}} ", stringBuilder.ToString().DarkGray());
+                var helpSwitch = string.Format($"{(addExtraIndent ? "        " : "    ")}{{0, {-longestName}}} ", stringBuilder.ToString().DarkGray());
```
soninaren: same as above, can we use standardized indentation method.

src/Cli/func/Actions/LocalActions/PackAction/CustomPackSubcommandAction.cs
```diff
@@ -0,0 +1,34 @@
...
+            return base.ParseArgs(args);
+        }
+
+        public async Task RunAsync(PackOptions packOptions)
+        {
+            await ExecuteAsync(packOptions);
+        }
+
+        protected override Task<string> GetPackingRootAsync(string functionAppRoot, PackOptions options)
+        {
```
soninaren, line 23: We should get the path for custom handler executable from host.json and add a warning in case the executable is not present in that directory.

release_notes.md
aishwaryabh: Planning on doing minor release bump once we show `func pack` in the help menu!

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: PR is actually getting massive also with this change so I will just defer this to a followup PR

src/Cli/func/Actions/LocalActions/PackAction/CustomPackSubcommandAction.cs
aishwaryabh: this is already done here: https://github.com/Azure/azure-functions-core-tools/blob/c8996cebcb80420ad8e2219cb1674f0c50aba30a/src/Cli/func/Helpers/HostHelpers.cs#L14

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -0,0 +1,126 @@
...
+                }
+
+                return packingRoot;
+            }
+            else
+            {
+                ColoredConsole.WriteLine("Building .NET project...");
+                await RunDotNetPublish(functionAppRoot);
+
+                return Path.Combine(functionAppRoot, "output");
```
soninaren, line 87: Nit: Use a constant for "output"

src/Cli/func/Helpers/ZipHelper.cs
```diff
@@ -40,14 +40,15 @@ public static async Task<Stream> GetAppZipFile(string functionAppRoot, bool buil
...
             {
                 // Remote build for dotnet does not require bin and obj folders. They will be generated during the oryx build
-                return await CreateZip(FileSystemHelpers.GetLocalFiles(functionAppRoot, ignoreParser, false, new string[] { "bin", "obj" }), functionAppRoot, Enumerable.Empty<string>());
+                return await CreateZip(FileSystemHelpers.GetLocalFiles(functionAppRoot, ignoreParser, false, new string[] { "bin", "obj" }), functionAppRoot, Array.Empty<string>());
             }
             else
             {
-                var customHandler = await HostHelpers.GetCustomHandlerExecutable();
+                var customHandler = await HostHelpers.GetCustomHandlerExecutable(functionAppRoot);
                 IEnumerable<string> executables = !string.IsNullOrEmpty(customHandler)
```
soninaren, line 48: We should have a log statement in case we don't find any custom handler executable in host.json

src/Cli/func/Helpers/ZipHelper.cs
```diff
@@ -40,14 +40,15 @@ public static async Task<Stream> GetAppZipFile(string functionAppRoot, bool buil
             else if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Dotnet && buildOption == BuildOption.Remote)
             {
                 // Remote build for dotnet does not require bin and obj folders. They will be generated during the oryx build
-                return await CreateZip(FileSystemHelpers.GetLocalFiles(functionAppRoot, ignoreParser, false, new string[] { "bin", "obj" }), functionAppRoot, Enumerable.Empty<string>());
+                return await CreateZip(FileSystemHelpers.GetLocalFiles(functionAppRoot, ignoreParser, false, new string[] { "bin", "obj" }), functionAppRoot, Array.Empty<string>());
             }
             else
             {
-                var customHandler = await HostHelpers.GetCustomHandlerExecutable();
+                var customHandler = await HostHelpers.GetCustomHandlerExecutable(functionAppRoot);
```
soninaren, line 47: May be nothing actionable here right now. Would be good for a future workitem. We should try to localize logic for each command to their sub action class.

soninaren (approved):
Approved with some minor comments

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
aishwaryabh: will address in a future PR!

src/Cli/func/Helpers/ZipHelper.cs
aishwaryabh: will address in a future PR

src/Cli/func/Helpers/ZipHelper.cs
aishwaryabh: will address in a future PR

PR #4612, opened 2025-08-20
Fix bug with installing .NET templates

### LINKED ISSUE #4608: Templates are not refreshing to the latest core tools version installed
Repro steps:

1. Install version 4.1.1.
2. Execute `func init --worker-runtime dotnet-isolated` command to create a new dni app.
3. Install [4.1.2](https://github.com/Azure/azure-functions-core-tools/releases/tag/4.1.2)
4. Execute `func init --worker-runtime dotnet-isolated --target-framework net10.0` command to create a new net10 dni app.
5. Observe the error saying "net10.0' is not a valid value for --Framework"

>   C:\temp\Apps\ct-test>func init --worker-runtime dotnet-isolated --target-framework net10.0
      Error: Invalid option(s):
      --Framework net10.0
         'net10.0' is not a valid value for --Framework. The possible values are:
            net48    - Target .NET Framework 4.8
            net6.0   - Target .NET 6
            net7.0   - Target .NET 7
            net8.0   - Target .NET 8
            net9.0   - Target .NET 9
      For more information, run:
         dotnet new func -h
      For details on the exit code, refer to https://aka.ms/templating-exit-codes#127
      Error creating project template

See internal thread [here](https://teams.microsoft.com/l/message/19:qNXZXqMm7lYfA6NzDZbSUjcPiwwi6clwOK2p_NK23741@thread.tacv2/1755531754314?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=53ae62b4-734e-484c-888e-f96574d40c7c&parentMessageId=1755531754314&teamName=Azure%20Functions%20Core%20Team&channelName=General&createdTime=1755531754314).

aishwaryabh:

resolves #4608

The issue with installing .NET templates was that we wouldn't install templates at all if we detected that they were already present, but this inhibits the customer from getting new templates. This PR accounts for that use case by reverting back to the original method of installing templates, where we always uninstall the other template, install the template we want, and then uninstall the template we just installed. 

In order to run the E2E tests in parallel, we decided to use custom hives. The templates get installed to whatever directory we want it to install by setting an env variable and adding `--debug:custom-hive <DIRECTORY>` to when we run `dotnet new install`. This prevents the templates from conflicting when running locally and allows us to run the tests in parallel, as the global template cache doesn't get updated.

### Pull request checklist


src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -336,30 +408,41 @@ private static async Task<HashSet<string>> GetInstalledTemplatePackageIds()
...
             var exitCode = await exe.RunAsync(o => output.AppendLine(o), e => output.AppendLine(e));
-
             if (exitCode != 0)
             {
                 throw new CliException("Failed to get list of installed template packages");
             }
 
             var lines = output.ToString()
-                .Split(['\r', '\n'], StringSplitOptions.RemoveEmptyEntries);
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
```
kshyju: Do we need the change here? The previous version should also work.

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -336,30 +408,41 @@ private static async Task<HashSet<string>> GetInstalledTemplatePackageIds()
...
-                if (trimmed.StartsWith(uninstallPrefix, StringComparison.OrdinalIgnoreCase))
+                // Detect package ID line
+                if (trimmed.StartsWith("dotnet new uninstall ", StringComparison.OrdinalIgnoreCase))
                 {
-                    var packageId = trimmed.Substring(uninstallPrefix.Length).Trim();
-                    if (!string.IsNullOrWhiteSpace(packageId))
-                    {
-                        packageIds.Add(packageId);
-                    }
+                    currentPackageId = trimmed.Substring("dotnet new uninstall ".Length).Trim();
```
kshyju: Why the const was removed?

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -295,38 +295,110 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
+        internal static bool AreDotnetTemplatePackagesWithSpecificVersionInstalled(HashSet<string> templates, string templatesPath, string packageIdPrefix)
+        {
+            string[] nupkgFiles = GetNupkgFiles(templatesPath);
+            return AreTemplatesUpdated(templates, packageIdPrefix, nupkgFiles);
+        }
+
+        private static string[] GetNupkgFiles(string templatesPath)
+        {
+            var templatesLocation = Path.Combine(
+                   Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location),
```
kshyju: Replace `Assembly.GetExecutingAssembly().Location` with ` AppContext.BaseDirectory;`

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -295,38 +295,110 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
+
+            if (!FileSystemHelpers.DirectoryExists(templatesLocation))
+            {
+                throw new CliException($"Can't find templates location. Looked under '{templatesLocation}'");
+            }
+
+            return Directory.GetFiles(templatesLocation, "*.nupkg", SearchOption.TopDirectoryOnly);
+        }
+
+        private static bool AreTemplatesUpdated(HashSet<string> templates, string packageIdPrefix, string[] nupkgFiles)
```
kshyju: Can we improve the readability of these param name to indicate one is for the installed (local?) templates, and the other is the template packages from the current version of core tools installation?

kshyju: I suggest `AreLocalTemplatesUpToDate` or something like that to improve readability.

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -295,38 +295,110 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
 
-            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, IsolatedTemplateBasePackId))
+            // Check to see if the isolated templates are already installed with the specific version
+            if (AreDotnetTemplatePackagesWithSpecificVersionInstalled(await _installedTemplatesList.Value, Path.Combine("templates", "net-isolated"), IsolatedTemplateBasePackId))
             {
                 return;
             }
 
+            // First uninstall any existing isolated templates, so .templateengine/packages can be cleaned up
+            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, IsolatedTemplateBasePackId))
```
kshyju: Does this `EnsureIsolatedTemplatesInstalled` method gets executed for every `func init` command execution? I wonder we could cache the result of this method (a boolean variable indicating this method was executed once) and return if we had already done that in that instance of the CLI.

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: changed back

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: added back

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: I added a boolean cache variable for if isolated and webjobs templates are installed. Let me know if it looks ok

aishwaryabh: Actually talked to @soninaren and having this boolean cache wasn't actually doing much since we spawn a new process everytime we run `func init`. We considered having a file that we store locally but it would get complicated to mantain since the customer can delete their `.templateengines` directory locally and then our locally stored file would be incorrect technically since the packages don't exist anymore

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,87 +281,60 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
             {
-                await EnsureWebJobsTemplatesInstalled();
+                await EnsureWebJobsTemplatesInstalled(action);
             }
-
-            await action();
         }
 
-        private static async Task EnsureIsolatedTemplatesInstalled()
+        private static async Task EnsureIsolatedTemplatesInstalled(Func<Task> action)
```
soninaren: Can we rename this to EnsureIsolatedTemplatesInstalledAsync?

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,87 +281,60 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
         {
-            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, WebJobsTemplateBasePackId))
+            try
             {
+                // Uninstall any existing webjobs templates, as they conflict with isolated templates
                 await UninstallWebJobsTemplates();
```
soninaren: The template package name here is a bit misleading. These are actually in-proc templates. Would be good if we can rename this to inproc templates.

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,87 +281,60 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
             {
-                return;
+                await UninstallIsolatedTemplates();
             }
-
-            await FileLockHelper.WithFileLockAsync(TemplatesLockFileName, InstallIsolatedTemplates);
         }
 
-        private static async Task EnsureWebJobsTemplatesInstalled()
+        private static async Task EnsureWebJobsTemplatesInstalled(Func<Task> action)
```
soninaren: EnsureWebJobsTemplatesInstalledAsync

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,87 +281,60 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
-            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, WebJobsTemplateBasePackId))
+            try
             {
+                // Uninstall any existing webjobs templates, as they conflict with isolated templates
                 await UninstallWebJobsTemplates();
-            }
 
-            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, IsolatedTemplateBasePackId))
+                // Install the latest isolated templates
+                await FileLockHelper.WithFileLockAsync(TemplatesLockFileName, InstallIsolatedTemplates);
```
soninaren: I see an issue with the FileLockHelper class here. Could be outside of the context of the PR. But the FileLockHelper is assuming that it is running in windows environment.

soninaren: Also do we need a file lock here? Can we use a simple lock pattern here?

aishwaryabh: Added simple lock pattern.

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,50 +282,138 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
+            bool hasProjectTemplates = false;
+            bool hasItemTemplates = false;
+
+            foreach (var file in templatePackagesFromCoreTools)
+            {
+                var fileName = Path.GetFileNameWithoutExtension(file); // e.g., itemTemplates.4.0.5212
+
+                if (fileName.StartsWith(ItemTemplatesPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    var version = fileName.Substring(ItemTemplatesPrefix.Length);
```
kshyju: Was the original logic (which we've had in place for a long time) handled this doing the version comparison or just uninstalled and installed always?

aishwaryabh: The old logic just uninstalled and installed as always. The issue with the old logic is that the E2E tests which require templates start failing if we reinstate the old logic due to template conflicts. Here is an example of a test run I did when I reverting back to the original logic of uninstalling/installing all the time with the file lock but we still hit the template conflict error:
https://azfunc.visualstudio.com/public/_build/results?buildId=233841&view=logs&j=a84f0933-5bc7-5dba-e8be-4d41f1e90239&t=bd61a0fd-2fc1-52cd-21b3-07d445b45aa9&l=1506

I think if we just keep the new approach of doing the version comparison and only installing the .NET templates if needed, this allows the E2E tests to pass and the logic to function as before to install and refresh the template cache.

soninaren (approved):

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: reverted back to fileLockHelper due to some issues with tests

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: Reverted back to original logic and made func new and func init tests sequential

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,94 +281,67 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
 
-            if (AreDotnetTemplatePackagesInstalled(await _installedTemplatesList.Value, IsolatedTemplateBasePackId))
+                // Install the latest isolated templates
+                await FileLockHelper.WithFileLockAsync(TemplatesLockFileName, InstallIsolatedTemplates);
+                await action();
+            }
+            finally
             {
-                return;
+                await UninstallIsolatedTemplates();
```
kshyju, line 356: Wouldn't this uninstall the isolated templates installed in the `try` block?

aishwaryabh: discussed offline

soninaren:
The tests are being run sequentially. Not sure if we are planning to create additional parallelization in devOps. If yes, then we should create a tracking issue for it.

release_notes.md
```diff
@@ -1,12 +1,10 @@
-# Azure Functions CLI 4.2.1
+# Azure Functions CLI 4.2.2
 
 #### Host Version
 
-- Host Version: 4.1041.200
-- In-Proc Host Version: 4.41.100 (4.841.100,  4.641.100)
+- Host Version: 
+- In-Proc Host Version:
```
liliankasem: Don't reset the host versions since they stay the same

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -283,94 +281,67 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
-                    if (!string.IsNullOrWhiteSpace(packageId))
-                    {
-                        packageIds.Add(packageId);
-                    }
-                }
+                throw new CliException($"Can't find templates location. Looked under '{templatesLocation}'");
             }
 
-            return packageIds;
+            return Directory.GetFiles(templatesLocation, "*.nupkg", SearchOption.TopDirectoryOnly);
```
liliankasem, line 388: Just to check, is this the previous approach of doing things? Or some mix?

aishwaryabh: Yes this is the previous approach! I just liked having this in its own separate method for readability, but the logic is the same

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -13,10 +12,9 @@ namespace Azure.Functions.Cli.Helpers
 {
     public static class DotnetHelpers
     {
-        private const string WebJobsTemplateBasePackId = "Microsoft.Azure.WebJobs";
```
liliankasem, line 16: I think this should still be called the WebJobsTemplates - since that's what they are (beyond just inproc)

aishwaryabh: @soninaren had recommended we call them InProcTemplates since in this specific case we are differentiating isolated and inproc templates for .NET. Thoughts?

liliankasem: that's okay, not a hill i want to die on :D

test/Cli/Func.E2ETests/Traits/TestTraits.cs
```diff
@@ -33,11 +33,5 @@ internal static class TestTraits
         /// This is done since when the dotnet isolated tests are run with dotnet inproc, we run into templating conflict errors.
         /// </summary>
         public const string InProc = "InProc";
-
-        public const string WorkerRuntime = "WorkerRuntime";
-        public const string Dotnet = "Dotnet";
-        public const string DotnetIsolated = "DotnetIsolated";
-        public const string Node = "Node";
-        public const string Powershell = "Powershell";
```
liliankasem, line 41: Why are these test traits getting removed?

test/Cli/Func.UnitTests/HelperTests/DotnetHelpersTests.cs
```diff
@@ -34,49 +34,5 @@ public void GetTemplates_ReturnsExpectedTemplates(WorkerRuntime runtime, int exp
             var templates = DotnetHelpers.GetTemplates(runtime);
             Assert.Equal(expectedCount, templates.Count());
         }
-
-        [Theory]
-        [InlineData("Microsoft.Azure.Functions.Worker")]
-        [InlineData("Microsoft.Azure.WebJobs")]
-        public void AreDotnetTemplatePackagesInstalled_ReturnsTrue_WhenTemplatesExists(string pkgPrefix)
```
liliankasem, line 41: Why are these unit tests removed? Can they be rewritten if needed?

test/Cli/Func.E2ETests/Traits/TestTraits.cs
aishwaryabh: they were leftovers of a previous commit. They're actually not used anywhere!

test/Cli/Func.UnitTests/HelperTests/DotnetHelpersTests.cs
aishwaryabh: we  don't have these methods anymore since we're using the old templating strategy

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -281,115 +303,110 @@ private static async Task TemplateOperationAsync(Func<Task> action, WorkerRuntim
...
             {
-                await EnsureWebJobsTemplatesInstalled();
+                await EnsureInProcTemplatesInstalledAsync(action);
             }
-
-            await action();
         }
 
-        private static async Task EnsureIsolatedTemplatesInstalled()
+        private static async Task EnsureTemplatesInCustomHiveAsync(Func<Task> action, WorkerRuntime workerRuntime)
```
liliankasem, line 324: Let's simplify this method (but keep the comments)

```csharp
private static async Task EnsureTemplatesInCustomHiveAsync(Func<Task> action, WorkerRuntime workerRuntime)
{
    string hivePackagesDir = Path.Combine(GetHivePath(workerRuntime), "packages");

    if (FileSystemHelpers.EnsureDirectoryNotEmpty(hivePackagesDir))
    {
        await action();
        return;
    }

    Func<Task> installTemplates = workerRuntime == WorkerRuntime.DotnetIsolated
        ? InstallIsolatedTemplates
        : InstallInProcTemplates;

    await FileLockHelper.WithFileLockAsync(TemplatesLockFileName, installTemplates);
    await action();
}
```

liliankasem: Also we probably don't need the file lock solution in place anywhere now that we are using the custom hives

test/Cli/Func.UnitTests/HelperTests/DotnetHelpersTests.cs
liliankasem: Can we add unit tests for the old methods?

liliankasem: Honestly E2E tests are sufficient here, but we should consider unit testing this path in future - won't block on it here

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: removed file lock!

test/Cli/Func.UnitTests/HelperTests/DotnetHelpersTests.cs
aishwaryabh: added

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -1,22 +1,84 @@
...
+        // Legacy WinStore aliases (e.g., win8, win81, win10, netcore45, netcore50)
+        private const string LegacyWinStorePattern = @"(?:win(?:8|81|10)|netcore(?:45|451|50)|netcore)";
+
+        /// <summary>
+        /// Regex that matches all valid .NET Target Framework Monikers (TFMs).
+        /// Covers modern TFMs (netX.Y[-osversion]),
+        /// netstandard, netcoreapp, classic .NET Framework, UAP, WP, Silverlight,
+        /// Tizen, NetNano, NetMF, and legacy WinStore aliases.
+        /// </summary>
+        public static readonly Regex TfmRegex = new Regex(
```
liliankasem: I originally thought this was going to be one line of code, any reason it's been split like this? If you want to keep it this way it should be moved into its own file

liliankasem (approved):
approved as my last comment is not blocking but would be nice to clean that up

src/Cli/func/Helpers/DotnetHelpers.cs
aishwaryabh: oh ya copilot suggested it and I liked it b/c it's easier to maintain. Let me split into other file

aishwaryabh: I'm gonna put it in the `TargetFrameworkHelper` since I think it's relevant there

PR #4623, opened 2025-08-26
Implement custom hives for e2e tests

liliankasem:
Implement custom hives for e2e tests

### DIFF
--- a/test/Cli/Func.E2ETests/Commands/FuncNew/DotnetInProcTests.cs
+++ b/test/Cli/Func.E2ETests/Commands/FuncNew/DotnetInProcTests.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System.IO;
 using Azure.Functions.Cli.E2ETests.Traits;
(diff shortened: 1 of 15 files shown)


test/Cli/Func.E2ETests/Commands/FuncNew/CustomRuntimeTests.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Azure.Functions.Cli.E2ETests.Traits;
+using Azure.Functions.Cli.Helpers;
```
aishwaryabh: nit: don't think we need this

test/Cli/Func.E2ETests/Commands/FuncNew/DotnetInProcTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.IO;
 using Azure.Functions.Cli.E2ETests.Traits;
+using Azure.Functions.Cli.Helpers;
```
aishwaryabh: nit: don't think we need this

aishwaryabh: same for the few other test files

PR #4629, opened 2025-08-29
Return HostStartSettings if local.settings.json is not loaded

### LINKED ISSUE #4604: func --help displays exception stacktrace
When I try `func --help` I am seeing an exception stack trace. 

Core Tools Version:      4.1.2

Image

Reproduced on more than one machine.

liliankasem:

resolves #4604

### Pull request checklist

### Additional information

This is an unfortunate overlap with how the help command is designed today, and how the start action's args use customer options.

Unfortunately the way that the help action is designed today makes it so that we need to load an action to get it's arguements/options. So in a situation like this where you run `func` to see the commands, we run the StartAction.ParseArgs -> which inits the secret manager to get host options (if any) -> secret manager looks for local.settings.json -> that doesn't exist outside of a func app dir -> exception. 

This will all be reworked as part of our upcoming refactor, which will seperate these two things and cleanup the help command in general.

Updated `GetHostStartSettings` to catch the exception when local.settings.json does not exist and return a new `HostStartSettings`. If a customer configures these settings, then local.settings.json and/or host.json will exist and the exception won't occur. If a customer does not configure these settings, we use the default.

### DIFF
--- a/src/Cli/func/Common/SecretsManager.cs
+++ b/src/Cli/func/Common/SecretsManager.cs
@@ -110,7 +110,14 @@ namespace Azure.Functions.Cli.Common
 
         public HostStartSettings GetHostStartSettings()
         {
-            return Settings.Host ?? new HostStartSettings();
+            try
+            {
+                return Settings?.Host ?? new HostStartSettings();
+            }
+            catch (CliException)
+            {
+                return new HostStartSettings();
+            }
         }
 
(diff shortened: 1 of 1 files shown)


kshyju (approved):
Synced with Lilian offline. It appears that the current fix is the best available approach for now. Ideally, we want the help command to execute without triggering any additional actions. A future refactor might help us achieve that behavior.

Issue #4633, opened 2025-09-04
Enhance `DotnetPackSubcommandAction` to account for Directory.Build.props


aishwaryabh:
Within `DotnetPackSubcommandAction`, if we detect the presence of a `Directory.build.props` in the function app root, read the `ArtifactsPath` value if it exists and use that value as the build output directory and write appropriate E2E tests for this action

aishwaryabh:
Had a conversation with @jviau and we decided it isn't necessarily a good idea for us to try to infer the publish path. The primary use case for this scenario was to infer the artifacts path when we aren't responsible for running dotnet publish (such as when we have the --no-build flag. There are multiple ways the customer can define their output directory (not just ArtifactsPath) so it's not necessarily worth trying to figure out the scenario for them, as we can never really be 100% accurate and may cause the customer more pain points rather than actually helping them.

If the customer wants a workaround and really wants a specific artifacts path to get zipped they can run the following:
func pack dir_to_artifacts_path --output path_for_zip_file_publish

We can re-open this issue if we see customers wanting to use this feature as well

PR #4643, opened 2025-09-09
Enhance Dotnet Subcommand to account for `ArtifactsPath`

aishwaryabh:

resolves [#4633](https://github.com/Azure/azure-functions-core-tools/issues/4633)

This PR enhances the .NET subcommand by reading the MSBuild property of `ArtifactsPath` and honors the property when publishing the .NET project during the `func pack` operation. That directory is not cleaned up as well, similar to how .NET works today.

### Pull request checklist


src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -55,16 +51,29 @@ protected override async Task<string> GetPackingRootAsync(string functionAppRoot
             // ValidateFunctionApp
             PackHelpers.ValidateFunctionAppRoot(functionAppRoot);
 
+            // Get the artifacts path from MSBuild properties if it exists
+            // If the value exists, we perform the build (if specified by the user) in that directory and do not delete the build output
+            var artifactsPath = DotnetHelpers.TryGetPropertyValueFromMSBuild(functionAppRoot, "ArtifactsPath");
```
soninaren: nit: Can we use a constant here.

src/Cli/func/Actions/LocalActions/PackAction/DotnetPackSubcommandAction.cs
```diff
@@ -82,28 +91,28 @@ protected override async Task<string> GetPackingRootAsync(string functionAppRoot
...
             }
         }
 
         protected override Task PerformCleanupAfterPackingAsync(string packingRoot, string functionAppRoot, PackOptions options)
         {
-            if (!options.NoBuild)
+            if (!options.NoBuild && ArtifactsPathSpecified)
             {
-                // If not no-build, delete packing root after packing
+                // If not no-build and if artifacts path was not specified, delete packing root after packing
```
soninaren, line 95: should the above condition have a `!ArtifactsPathSpecified` or should we change the comment.

aishwaryabh: yes just realized that the name was off so I fixed it!

aishwaryabh:
Had a conversation with @jviau and we decided it isn't necessarily a good idea for us to try to infer the publish path. The primary use case for this scenario was to infer the artifacts path when we aren't responsible for running `dotnet publish` (such as when we have the `--no-build` flag. There are multiple ways the customer can define their output directory (not just `ArtifactsPath`) so it's not necessarily worth trying to figure out the scenario for them, as we can never really be 100% accurate and may cause the customer more pain points rather than actually helping them.

If the customer wants a workaround and really wants a specific artifacts path to get zipped they can run the following:
`func pack dir_to_artifacts_path --output path_for_zip_file_publish`

We can re-open this PR in the future if we get requests for this feature as well

PR #4665, opened 2025-10-02
Reset release notes

castrodd:
Reset release notes after the release of Core Tools v4.3.0

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -2,8 +2,8 @@
 
 #### Host Version
 
-- Host Version: <version>
-- In-Proc Host Version: <version>
+- Host Version: 4.1042.100
+- In-Proc Host Version: 4.41.100 (4.841.100, 4.641.100)
 
 #### Changes
 


release_notes.md
```diff
@@ -1,20 +1,10 @@
-# Azure Functions CLI 4.3.0
+# Azure Functions CLI <version>
 
 #### Host Version
 
-- Host Version: 4.1042.100
-- In-Proc Host Version: 4.41.100 (4.841.100, 4.641.100)
-
+- Host Version: <version>
+- In-Proc Host Version: <version>
```
liliankasem: Let's keep the host version the same until we bump it - we often do releases without a host change and its tedious to have to fill this in every time

liliankasem: I meant to keep these two filled. cli version is also ok

Host Version: 
In-Proc Host Version:

castrodd: Should be up-to-date now

satvu (approved):

PR #4692, opened 2025-10-27
Update durable .NET templates

andystaples:
Adds the new Durable Isolated templates and Durable Entity templates to the list of supported templates.

### DIFF
--- a/eng/ci/templates/official/jobs/linux-deb-build-pack.yml
+++ b/eng/ci/templates/official/jobs/linux-deb-build-pack.yml
@@ -76,6 +76,14 @@ jobs:
       linuxBuildNumber: $(DefaultArtifactVersion)
       consolidatedBuildId: $(ConsolidatedBuildId)
 
+  - task: Bash@3
+    displayName: "Validate DEB architectures"
+    inputs:
+      targetType: 'inline'
+      script: |
+        chmod +x $(Build.SourcesDirectory)/eng/scripts/validate-deb-arch.sh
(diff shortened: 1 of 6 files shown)


andystaples:
One note - we did not create templates named 
DurableFunctionsEnttiyClass-CSharp-1.x
DurableFunctionsEntityFunction-CSharp-1.x

We only created the -2.x variants. 
Just want to expressly check that this is not an issue

src/Cli/func/Helpers/DotnetHelpers.cs
```diff
@@ -207,6 +210,8 @@ internal static IEnumerable<string> GetTemplates(WorkerRuntime workerRuntime)
                 "KafkaTrigger",
                 "KafkaOutput",
                 "DurableFunctionsOrchestration",
+                "DurableFunctionsEntityClass",
+                "DurableFunctionsEntityFunction",
```
liliankasem, line 214: Hey Andy, which template pkg will these be available in? I want to make sure we don't deploy ahead of these templates being available

andystaples: I can't remember exactly which Templates/Bundles version these made it into, but they were merged in Templates back in July and the latest they could possibly have been released was with Bundles 4.25, which was available everywhere long ago. Should not have an issue here, although if there is another way we need to check, happy to assist.

aishwaryabh (approved):

PR #4694, opened 2025-10-27
fix: outputs local.settings.json found even though it isnt

WhiteTomX:

I wondered why I'm asked for the worker runtime, although it told me the `local.settings.json` was found. Turns out it is a new clone and the file is not there. The output is just wrong.

```
local.settings.json found in root directory (C:\Users\WhiteTom\src\api).
Use the up/down arrow keys to select a worker runtime:
dotnet (isolated worker model)
dotnet (in-process model)
Node
Python
Powershell
Custom
```

### Pull request checklist

  * Or should they be added to release notes?
  * I don't think tests are required for that?

### DIFF
--- a/src/Cli/func/Common/SecretsManager.cs
+++ b/src/Cli/func/Common/SecretsManager.cs
@@ -27,7 +27,7 @@ namespace Azure.Functions.Cli.Common
                 });
                 var secretsFilePath = Path.Combine(rootPath, secretsFile);
 
-                ColoredConsole.WriteLine(DarkGray($"'{secretsFile}' found in root directory ({rootPath})."));
+                ColoredConsole.WriteLine(DarkGray($"'{secretsFile}' should be in root directory ({rootPath})."));
                 return secretsFilePath;
             }
         }


WhiteTomX:
I made the easy change to just change the wording. The alternative would be to remove the output there and output if the file was actually loaded or not. Currently we only [output if the file is there, but not a valid json](https://github.com/Azure/azure-functions-core-tools/blob/7c130f44cc5fe03f6f22e7a7afbe7d5476629c76/src/Cli/func/Common/AppSettingsFile.cs#L41).

src/Cli/func/Common/SecretsManager.cs
```diff
@@ -27,7 +27,7 @@ public static string AppSettingsFilePath
                 });
                 var secretsFilePath = Path.Combine(rootPath, secretsFile);
 
-                ColoredConsole.WriteLine(DarkGray($"'{secretsFile}' found in root directory ({rootPath})."));
+                ColoredConsole.WriteLine(DarkGray($"'{secretsFile}' should be in root directory ({rootPath})."));
```
liliankasem, line 30: Thanks for raising this issue!

This isn't the intention of this log here - it should be logging if it _did_ find a local.settings file, and not notifying  were it should be. If this isn't working as intended feel free to file an issue and we can investigate what's actually happening to fix it. You're also welcome to take a stab at this yourself but chaning the verbiage of the log here isn't the approach I would like us to take.

WhiteTomX: Thank you for the clarification. It should be quite straight forward to fix that. I hope I will be able to do so at the weekend.

Issue #2671, opened 2021-07-28
Check for AzureWebJobsStorage fails when using identity-based connection


mattchenderson:
We are currently checking for AzureWebJobsStorage during `func azure functionapp deploy`:

https://github.com/Azure/azure-functions-core-tools/blob/6a070fb40308438c758ba8ab69399845bd698391/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L241-L244

Today this is just looking for an exact-match key, which fails when using an identity-based connection. In that model, there may be one or more keys prefixed with "AzureWebJobsStorage__". For example, the simplest configuration just has "AzureWebJobsStorage__accountName" which just takes the storage account name as a value. The Core Tools should treat these setups as equivalently valid and allow deployment to proceed.

Related to: https://github.com/Azure/azure-cli/issues/19035

apawast:
@mattchenderson can you provide more details about the expected outcome here? design guidance would be helpful here.

balag0:
reopening  since this still needs to be fixed.

mattchenderson:
That closure window is way too aggressive isn't it? This is still an issue. The guidance is to allow prefixing instead of exact name match.

mattchenderson:
The auto-closure without notification to me is particularly irksome.

PR #4695, opened 2025-10-28
Update publish error message

### LINKED ISSUE #2672: Bad error messages on checks for AzureWebJobsStorage during publish
During publish, the Core Tools checks for the presence of AzureWebJobsStorage. This is only strictly required in Linux, but we perform the check on Windows as well. The error message unfortunately assumes Linux though, so the message makes no sense when failing to deploy to a Windows app.

https://github.com/Azure/azure-functions-core-tools/blob/6a070fb40308438c758ba8ab69399845bd698391/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L241-L244

I think it is reasonable to preserve the check over Windows but modify the error message. Related, the conditional here should be revised to capture identity-based connections for AzureWebJobsStorage, which is filed as a separate issue: https://github.com/Azure/azure-functions-core-tools/issues/2671

liliankasem:

resolves #2672, #2671

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -330,15 +330,12 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 }
             }
 
-            if (functionApp.IsLinux &&
-                (functionApp.IsDynamic || functionApp.IsFlex) &&
-                !functionApp.AzureAppSettings.ContainsKey(Constants.AzureWebJobsStorage))
+            if (!functionApp.AzureAppSettings.Keys.Any(k => k.StartsWith("AzureWebJobsStorage", StringComparison.OrdinalIgnoreCase)))
             {
-                var helpUrl = functionApp.IsFlex ?
-                    "https://aka.ms/flex-deployment-settings" :
-                    "https://aka.ms/deployfromurl";
-
-                throw new CliException($"Function App '{FunctionAppName}' is missing the '{Constants.AzureWebJobsStorage}' app setting. Please read the deployment configuration requirements here {helpUrl}");
+                throw new CliException(
+                    $"Function App '{FunctionAppName}' is missing host storage configuration. " +
+                    $"Provide either '{Constants.AzureWebJobsStorage}' (connection string) or identity-based settings " +
+                    $"prefixed with '{Constants.AzureWebJobsStorage}__'. Learn more at https://aka.ms/func-app-settings-storage");
             }
 
             if (functionApp.IsFlex)
@@ -346,11 +343,6 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 if (functionApp.AzureAppSettings.ContainsKey(Constants.WebsiteRunFromPackage))
                 {
                     throw new CliException($"Function Apps on Flex Consumption do not support '{Constants.WebsiteRunFromPackage}'. Please remove the app setting from your Function App.");
-                }
-
-                if (functionApp.AzureAppSettings.ContainsKey(Constants.DeploymentStorageConnectionString))
-                {
-                    throw new CliException($"Function App '{FunctionAppName}' is missing the '{Constants.DeploymentStorageConnectionString}' app setting. Please read the deployment configuration requirements here https://aka.ms/flex-deployment-settings");
                 }
 
                 if (result.ContainsKey(Constants.FunctionsWorkerRuntime))
--- a/src/Cli/func/Common/Constants.cs
+++ b/src/Cli/func/Common/Constants.cs
@@ -49,7 +49,6 @@ namespace Azure.Functions.Cli.Common
         public const string ScmRunFromPackage = "SCM_RUN_FROM_PACKAGE";
         public const string ScmDoBuildDuringDeployment = "SCM_DO_BUILD_DURING_DEPLOYMENT";
         public const string WebsiteRunFromPackage = "WEBSITE_RUN_FROM_PACKAGE";
-        public const string DeploymentStorageConnectionString = "DEPLOYMENT_STORAGE_CONNECTION_STRING";
         public const string WebsiteContentAzureFileConnectionString = "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING";
         public const string WebsiteContentShared = "WEBSITE_CONTENTSHARE";
         public const string TelemetrySentinelFile = "telemetryDefaultOn.sentinel";


src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -330,13 +330,29 @@ private async Task<IDictionary<string, string>> ValidateFunctionAppPublish(Site
...
             }
 
             if (functionApp.IsFlex)
             {
+                if (functionApp.AzureAppSettings.ContainsKey(Constants.WebsiteRunFromPackage))
+                {
+                    throw new CliException($"Function Apps on Flex Consumption do not support '{Constants.WebsiteRunFromPackage}'. Please remove the app setting from your Function App.");
+                }
+
+                if (functionApp.AzureAppSettings.ContainsKey(Constants.DeploymentStorageConnectionString))
```
im-samz: Quick clarification: the setting `DEPLOYMENT_STORAGE_CONNECTION_STRING` is ony required if the property `properties.functionAppConfig.deployment.storage.authentication.type` == `storageaccountconnectionstring` **and** `properties.functionAppConfig.deployment.storage.authentication.storageAccountConnectionStringName` == `DEPLOYMENT_STORAGE_CONNECTION_STRING`. 

Technically, the customer can choose any app setting name, as long as they mark it within `storageAccountConnectionStringName`.

---

Therefore, this validation will be more complex. Kudu.Legion will already perform a similar validation, so this isn't essential. Might be easier to remove the validation?

im-samz: Also, the conditional seems to look for **if** the key exists, but you're trying to check if it doesn't, right?

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -330,13 +330,29 @@ private async Task<IDictionary<string, string>> ValidateFunctionAppPublish(Site
                 }
             }
 
-            if (!functionApp.AzureAppSettings.ContainsKey("AzureWebJobsStorage") && functionApp.IsDynamic && functionApp.IsLinux)
+            if (functionApp.IsLinux &&
+                (functionApp.IsDynamic || functionApp.IsFlex) &&
+                !functionApp.AzureAppSettings.ContainsKey(Constants.AzureWebJobsStorage))
```
im-samz: This conditional is weird. It is checking for `AzureWebJobsStorage`, which isn't related to deployment. If customers are in this state, they're in a broken state...So the first guidance we should give them is to set that setting. Reference doc is here:

https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#azurewebjobsstorage

---

This is the matrix of validations:
| ID | Scenario | SKUs | Action |
| ---|---------|-------|-------|
| 1 | No `AzureWebJobsStorage` | All SKUs | Set the setting via app setting reference doc |
| 2 | No `WEBSITE_RUN_FROM_PACKAGE` | All SKUs except Flex + ACA | Link to aka.ms/deployfromurl |
| 3 | No `WEBSITE_RUN_FROM_PACKAGE` | Linux Consumption v1 | [optional] Link to aka.ms/deployfrom url but make it clear customers need to set it to a URL |
| 4 | `WEBSITE_RUN_FROM_PACKAGE` exists | Flex Consumption | Mention Flex does not support the setting **and** link to aka.ms/flex-deployment-settings |

- Validation ID-1 exists here, but the call to action should be updated;
- Validation IDS-2 + 3 do not exist here; I'm not sure if Kudu validates this
- Validation ID-4 exists here, but we should update the call to action. Kudu validates this too.

im-samz (commented):
Comments on the overall validation strategy + concern regarding the `DEPLOYMENT_STORAGE_CONNECTION_STRING` validation.

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
liliankasem: Yup meant to check "doesn't exist". I think given the complexity of it, I will remove this validation. But it would be nice to update the original AZ message is hard to understand:

`Server Response: {"error":"MissingDeploymentConfigException: functionAppConfig \u003E deployment \u003E storage is missing following configurations: AppSetting name set in authentication \u003E value does not exist. Please add the listed configurations and then redeploy. "}`

im-samz: Noted! This message is emitted from Kudu (not AZ), so I'll create a work item for that component.

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
liliankasem: Discussed offline

liliankasem: @im-samz actually it migth be worth creating a new issue with the above matrix so that we can have a Wi to go investigtae all the core tools deployment validations and update as necessary?

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -330,13 +330,20 @@ private async Task<IDictionary<string, string>> ValidateFunctionAppPublish(Site
...
             {
-                throw new CliException($"Azure Functions Core Tools does not support this deployment path. Please configure the app to deploy from a remote package using the steps here: https://aka.ms/deployfromurl");
+                throw new CliException($"Function App '{FunctionAppName}' is missing the '{Constants.AzureWebJobsStorage}' app setting. Please read the deployment configuration requirements here https://aka.ms/func-app-settings-storage");
```
im-samz: We can leave out `deployment configuration requirements` since not having `AzureWebJobsStorage` is a runtime issue, not a deployment one. 

Maybe something like;

>Function App '{FunctionAppName}' is missing the '{Constants.AzureWebJobsStorage}' app setting. Learn how to set it here https://aka.ms/func-app-settings-storage

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -330,13 +330,20 @@ private async Task<IDictionary<string, string>> ValidateFunctionAppPublish(Site
                 }
             }
 
-            if (!functionApp.AzureAppSettings.ContainsKey("AzureWebJobsStorage") && functionApp.IsDynamic && functionApp.IsLinux)
+            if (functionApp.IsLinux &&
+                (functionApp.IsDynamic || functionApp.IsFlex) &&
```
im-samz: Not having AzureWebJobsStorage is not specific to Linux Consumption and Flex Consumption. If we want full coverage, we can make this validation run for all function apps.

liliankasem: I think this is validated for all SKUs, it's just this weird code path here that is for linux

e.g. 

```                if (!skipAzureWebJobsStorageCheck && string.IsNullOrWhiteSpace(azureWebJobsStorage) &&
                    !ConnectionExists(secrets, storageConnectionKey) && !allNonStorageTriggers)
                {
                    throw new CliException($"Missing value for AzureWebJobsStorage in {SecretsManager.AppSettingsFileName}. " +
                        $"This is required for all triggers other than {string.Join(", ", Constants.TriggersWithoutStorage)}. "
                        + $"You can run 'func azure functionapp fetch-app-settings <functionAppName>', specify a connection string in {SecretsManager.AppSettingsFileName}, or use managed identity to authenticate.");
                }
```

liliankasem: Hmm actually that's in the start path, let me look around publish

satvu (approved):

PR #4710, opened 2025-11-10
Update copilot release instructions

liliankasem:
Update copilot release instructions

### DIFF
--- a/.github/instructions/release.instructions.md
+++ b/.github/instructions/release.instructions.md
@@ -1,7 +1,9 @@
-# Instructions of Azure Functions Core Tools Versioning
+# Instructions of Azure Functions Core Tools Release Preparation
 
 ## Goals
 - Ensure Core Tools version bumps follow [SemVer](https://semver.org/) consistently.
+- Ensure release notes are updated with each version bump.
+- Update the host dependency version when applicable.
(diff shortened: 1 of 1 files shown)


.github/instructions/release.instructions.md
```diff
@@ -24,10 +26,57 @@
...
+1. **Out-of-proc version**:
+    - Checkout the `main` branch
+    - Update `VersionPrefix` in `src/Cli/func/Directory.Version.props`.
+2. **In-proc version** (if requested):
+    - Checkout the `in-proc` branch
+    - Update `VersionPrefix` in `src/Azure.Functions.Cli/Directory.Version.props`.
+
+### Host Version Update Procedure
+
+Updates happen only when there are changes to the host dependency.
```
satvu, line 44: A little confused by this line -is this relaying "[Host version] updates happen only when there are changes to the host dependency."? 

Does changes to the host dependency mean a new host version is released, or does it mean "when the reference to the host dependency is updated, so there was a change in the host dependency".

liliankasem: Yeah I meant `when the reference to the host dependency is updated` i.e. these instructions are used when the agent is asked to update the the host dependency

satvu (approved):

PR #4712, opened 2025-11-10
Update .NET template pkg version & remove -preview from Dockerfile

liliankasem:
Update .NET template pkg version & remove -preview from Dockerfile


mattchenderson (commented):
I believe this is the correct set of changes, so just pending final package version published and validations.

mattchenderson (approved):

fabiocav (approved):

PR #4715, opened 2025-11-11
 Refactor DetermineTargetFramework to use msbuild & support multiple tfms

### LINKED ISSUE #4670: Flaky `DotnetIsolatedInitTests` E2E test
### Version

n/a

### Description

E2E test `Azure.Functions.Cli.E2ETests.Commands.FuncInit.DotnetIsolatedInitTests.Init_DockerOnlyOnExistingProjectWithTargetFramework_GeneratesDockerfile` fails very consistently in the CI pipelines, but only on osx.

Examples: 

https://dev.azure.com/azfunc/public/_build/results?buildId=241533&view=logs&j=a84f0933-5bc7-5dba-e8be-4d41f1e90239&t=bd61a0fd-2fc1-52cd-21b3-07d445b45aa9&s=dd265d6b-20be-5631-1079-c0ccdfa675fd

https://dev.azure.com/azfunc/public/_build/results?buildId=241276&view=logs&j=a84f0933-5bc7-5dba-e8be-4d41f1e90239&t=bd61a0fd-2fc1-52cd-21b3-07d445b45aa9

https://dev.azure.com/azfunc/public/_build/results?buildId=240677&view=logs&j=a84f0933-5bc7-5dba-e8be-4d41f1e90239&t=bd61a0fd-2fc1-52cd-21b3-07d445b45aa9

### Steps to reproduce

n/a

liliankasem:

resolves #4670

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Cli/func/Helpers/DotnetHelpers.cs
+++ b/src/Cli/func/Helpers/DotnetHelpers.cs
@@ -50,6 +50,7 @@ namespace Azure.Functions.Cli.Helpers
             string args =
                 $"msbuild \"{projectFilePath}\" " +
                 "-nologo -v:q -restore:false " +
+                "-getProperty:TargetFrameworks " +
                 "-getProperty:TargetFramework";
 
             var exe = new Executable(
@@ -62,7 +63,7 @@ namespace Azure.Functions.Cli.Helpers
             var stdErr = new StringBuilder();
 
             int exit = await exe.RunAsync(s => stdOut.Append(s), s => stdErr.Append(s));
-            if (exit is not 0)
+            if (exit != 0)
             {
                 throw new CliException(
                     $"Unable to evaluate target framework for '{projectFilePath}'.\nError output:\n{stdErr}");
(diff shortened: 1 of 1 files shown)


mattchenderson (approved):
Approving for CI testing, but not to merge.

jviau:
Probably work for a different PR, if at all. Determining TFM from core tools doesn't work when the function app is multi-targeted. This change is definitely an improvement, but it will have the same multi-target issue.

Is there a framework parameter for `func` cli? Should we add one? Or consider this area not supported and error out when TFM cannot be found?

Using `dotnet run` target does work here, as you can specify the TFM for it.

liliankasem:
> Probably work for a different PR, if at all. Determining TFM from core tools doesn't work when the function app is multi-targeted. This change is definitely an improvement, but it will have the same multi-target issue.

Yeah I think support for multiple TFMs is a little rough at the moment. I can definitely add a check for TargetFramework**s** and prompt the user for which of the TFMs they would like to use from that list.

The determine TFM method comes into play for something like `--docker-only` where you have already initialized the worker with a TFM i.e.

`func init --worker-runtime dotnet-isolated  --target-framework net10.0`

But now you're like "I want docker too!"

`func init --docker-only`

In which case we will a) determine worker from local settings and b) determine tfm via this method

We could also add support for `func init --docker-only --target-framework net10.0` but I don't want to break the scenario where you just do the docker param.

> Is there a framework parameter for func cli? Should we add one? Or consider this area not supported and error out when TFM cannot be found?

Yeah there is already a tfm param for func init 

`--target-framework      Initialize a project with the given target framework moniker. Currently supported only when --worker-runtime set to dotnet-isolated or dotnet. Options are - net10.0, net9.0, net8.0, net7.0, net6.0, net48`

satvu (approved):

PR #4719, opened 2025-11-12
Update CI nightly builds schedule

liliankasem:
Our current schedule runs at ~4pm PST which is still during working hours, updated to run around midnight/11pm PST depending on daylight savings.


fabiocav (commented):
Can we stagger the builds to avoid have them all kicked off at the same time? If you can space them roughly based on their run duration, that would be great.

liliankasem:
> Can we stagger the builds to avoid have them all kicked off at the same time? If you can space them roughly based on their run duration, that would be great.

Great idea!

jviau (approved):

PR #4723, opened 2025-11-13
Update release notes

castrodd:
This PR is removing any changes from the Release Notes that have been published in the latest 4.4.1 release.


liliankasem:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

I will be fixing this soon, it's so annoying

PR #4726, opened 2025-11-13
Update public CI to skip on docs-only changes

liliankasem:
To speed up our PR process, and save on CI agent resources, I would love for us to skip running all of our build and E2E tests when a PR only contains documentation (i.e. .md) file changes. We originally tried to do this with the ADO path exclusion:

```yml
paths:
    include:
    - eng
    - src
    - test
    exclude:
    - docs
    - '**/*.md'
 ```

This did not work as our GitHub ruleset for PRs is that the public CI pipeline MUST run. And with these filters in place, we actually just skipped the CI altogether (which is good and intended), but GitHub would block because the CI needed to run.

I also looked into changing the rulesets, but there is no filter system for GitHub to skip CI for specific files/folder paths; instead it made it so that the CI would run but you can merge before the CI finished/passed (this is bad).

This brings us to this approach, whilst not ideal, it works. We will check to see what the file changes are via a script, and if it's a "docs-only change", we skip CI (but the CI still runs so that satisfys the GitHub ruleset requirement). A couple of things to note:

- This is only for the public CI pipeline, not the official pipeline (no changes there).
- This only applies to Pull Request changes. So if its a manual run or scheduled run, we will always run this pipeline.
- On Pull Requests, we will run the script to identify if its a doc only change or not. This is currently defined as:
  - Any file in docs/
  - Any file in .github/
  - Any file in .vscode/
  - Any file ending in .md
  - The LISENCE and CODEOWNERS files

This definition of docs-only is up for debate. I included .github and .vscode as they both contain files other than "md" type that do NOT have any impact on the product itself - changes in these files do not come into play in CI at all, they are not checked or utilized.

### DIFF
--- a/eng/ci/templates/public/jobs/precheck-docs-only.yml
+++ b/eng/ci/templates/public/jobs/precheck-docs-only.yml
@@ -18,7 +18,3 @@ jobs:
       pwsh: true
       targetType: filePath
       filePath: ${{ parameters.scriptPath }}
-      arguments: >
-        -BuildReason "$(Build.Reason)"
-        -TargetBranch "$(System.PullRequest.TargetBranch)"
-        -SourceBranch "$(System.PullRequest.SourceBranch)"
--- a/eng/scripts/classify-changes.ps1
+++ b/eng/scripts/classify-changes.ps1
@@ -1,28 +1,29 @@
-param(
-    [string]$BuildReason,
-    [string]$TargetBranch,
-    [string]$SourceBranch
-)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: eng/ci/public-build.yml [hunk 1]
```diff
@@ -14,14 +14,8 @@ pr:
...
     - in-proc
-  paths:
-    include:
-    - eng
-    - src
-    - test
-    exclude:
-    - docs
-    - '**/*.md'
+    - featute/*
```
satvu: typo?

Issue #3285, opened 2023-02-24
File 'C:\Program Files\dotnet\dotnet.exe' is not found, 'dotnet' invocation will rely on the PATH environment variable


pguardiario:
Is there a way to disable these warnings in local.settings.json? I'm not sure what dotnet.exe is but I'm pretty sure my js function doesn't need it.

ab958:
any update

K-Vivek-Kumar:
To resolve this warning, you can install the .NET SDK either directly from the website or through Visual Studio (version 2022 or 2017).

Website Link: https://dotnet.microsoft.com/en-us/download

Alternatively, while installing or updating Visual Studio, you can choose the following option:
<img width="941" alt="image" src="https://github.com/Azure/azure-functions-core-tools/assets/102508246/b341f50d-3968-4055-b0c0-5b559183f5d2">

For me, the second step worked, by installing via Visual Studio 17.10.0.

southwolf:
Hi all,

I've already installed .NET SDK8 in a non-standard location (e.g. `C:\software\dotnet\`) and set it into PATH, but function tools still complain about this. Any solutions?

GABRIELNGBTUC:
I am also facing this issue. I use the dotnet SDK installed from the dotnet-install scripts to a non-standard path and the function core tools is refusing to use the dotnet tools from the PATH environment variable. It would be great if the tool only tried to use the default path if no dotnet tool is found in the PATH.

liliankasem:
The dotnet path issue is resolved via #4725 and #4732 - the cli will now check other paths/env vars to see if dotnet is installed.

As for the original issue, this warning 

`File 'C:\Program Files\dotnet\dotnet.exe' is not found, 'dotnet' invocation will rely on the PATH environment variable`

Comes from the [host](https://github.com/Azure/azure-functions-host/blob/90a2950fcfb1f98c9c8e00c5a4f77ed1e2d193a4/src/WebJobs.Script/Workers/Rpc/RpcWorkerDescription.cs#L181), not the cli. But this should only happen if you're trying to run a dotnet or dotnet-isolated function app, and not on other language workers as the check to resolve dotnet path only occurs if:

- The OS is Windows, AND
- The `DefaultExecutablePath` is set to "dotnet" or "dotnet.exe"

[...]

PR #4732, opened 2025-11-17
Use Muxer from .NET SDK for dotnet discovery

### LINKED ISSUE #4725: Azure DevOps: Azure Functions Core Tools Can't Find .NET 10 Installed by UseDotNet@2 Task on Windows Agents
### Version

I was upgrading an Azure Durable Function Application from `.NET 9` to `.NET 10`. Our Azure DevOps pipeline have a job that executes set of integration tests by spinning up the function using `func`. Since we were using `MSSQLLocalDB`, the agent is `Windows`.

After the upgrade, the integration tests was failing to spin up `func` with a frustrating error.
```
You must install or update .NET to run this application.
App: D:\a\1\s\tests\...\bin\Debug\net10.0\FunctionApp.dll
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '10.0.0' (x64)
.NET location: C:\Program Files\dotnet\
The following frameworks were found:
  8.0.6 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  8.0.21 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  9.0.6 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  9.0.10 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
 
Learn more:
https://aka.ms/dotnet/app-launch-failed
To install missing framework, download:
```

### Description

The pipeline used the [UseDotNet@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines&WT.mc_id=DT-MVP-5000570) task to install `.NET 10`.
```yaml
- task: UseDotNet@2
  displayName: Install .NET 10.0.x
  inputs:
    packageType: 'sdk'
    version: '10.0.x'
```
The pipeline debug logs showed [UseDotNet@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-dotnet-v2?view=azure-pipelines&WT.mc_id=DT-MVP-5000570) task was setting `DOTNET_ROOT` and updating `PATH` correctly:
```
##[debug]Absolute path for pathSegments: C:\hostedtoolcache\windows\dotnet\sdk
Successfully installed .NET Core sdk version 10.0.100.
##[debug]Processed: ##vso[task.prependpath]C:\hostedtoolcache\windows/dotnet
##[debug]set DOTNET_ROOT=C:\hostedtoolcache\windows/dotnet
```

And `dotnet --info` confirmed `.NET 10` was installed.
Image

However `func.exe` doesn't seem to recognize it, it kept looking at  `C:\Program Files\dotnet`. 
When starting the worker process, it ignores:
* The `DOTNET_ROOT` environment variable
* The `PATH` environment variable

Since `.NET 10` isn't yet pre-installed on DevOps agents, Azure Functions can't find it.

Shouldn't `func` consider `DOTNET_ROOT` or `PATH` to determine `dotnet`?

### Workaround:

When installing `.NET 10`, override the default installation path which is `$(Agent.ToolsDirectory)/dotnet`  (`C:\hostedtoolcache\windows\dotnet` in `Windows`) to `C:\Program Files\dotnet` where Azure Functions expects to find it.
```
- task: UseDotNet@2
  displayName: Install .NET 10.0.x
  inputs:
    packageType: 'sdk'
    version: '10.0.x'
    installationPath: 'C:\Program Files\dotnet'
```

### Steps to reproduce

- Spin up a .NET 10 function app using `func` in an Azure DevOps Pipeline that runs on `windows-latest`
- Install .NET SDK
```
- task: UseDotNet@2
  displayName: Install .NET 10.0.x
  inputs:
    packageType: 'sdk'
    version: '10.0.x'
```

liliankasem:

resolves #4725, #3285

### Pull request checklist

### Additional information

Replace CommandChecker.CommandExists with DotnetMuxer class copied from the .NET SDK to improve dotnet installation discovery. This provides more comprehensive detection using multiple fallback strategies including DOTNET_HOST_PATH, DOTNET_ROOT, and process path resolution.

Changes:
- Add DotnetMuxer and FileNameSuffixes helper classes from dotnet/sdk
- Update EnsureDotnet() to use DotnetMuxer.MuxerPath with better error handling
- Remove multi-TFM conditionals since CLI only targets net8.0
- Improve error message to include all discoverable environment variables

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -2,9 +2,15 @@
 
 #### Host Version
 
-- Host Version: 4.1044.400
-- In-Proc Host Version: 4.44.100 (4.844.100, 4.644.100)
+- Host Runtime Version: 4.1044.400
+- In-Proc CLI:
+  - CLI Version: 4.3.0
+  - Host Runtime Version: 4.44.100 (includes 4.844.100, 4.644.100)
 
 #### Changes
 
+- Fix .gitignore to allow PowerShell module bin folders (#4574)
+- Refactor to use msbuild for determining .NET target framework & add support multiple TFMs (#4715)
+  - When using `func init --docker-only` on a .NET project with multiple target frameworks, the CLI will now
+   prompt the user to select which target framework to use for the Dockerfile.
 - Enhanced dotnet installation discovery by adopting the same `Muxer` logic used by the .NET SDK itself (#4732)


src/Cli/func/Helpers/DotnetMuxer.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (File.Exists(muxerPathMaybe))
+            {
+                _muxerPath = muxerPathMaybe;
+            }
+        }
+
+        if (_muxerPath is null)
+        {
+            // Best-effort search for muxer.
+            // SDK sets DOTNET_HOST_PATH as absolute path to current dotnet executable
```
jviau, line 41: This comment looks out of place?

src/Cli/func/Helpers/FileNameSuffixes.cs
jviau, line 1: Is this file used anywhere?

test/Cli/Func.UnitTests/HelperTests/DotnetMuxerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            // Arrange & Act
+            var muxer = new DotnetMuxer();
+
+            // Assert
+            Assert.NotNull(muxer.MuxerPath);
+            Assert.False(string.IsNullOrWhiteSpace(muxer.MuxerPath));
+        }
+
+        [Fact]
+        public void MuxerPath_ThrowsInvalidOperationException_WhenNotFound()
```
jviau: This isn't testing anything. I recommend removing this if we can't actually test it.

test/Cli/Func.UnitTests/HelperTests/DotnetMuxerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // If we got here, dotnet was found (which is expected in test environment)
+            Assert.NotNull(muxer.MuxerPath);
+        }
+
+        [Fact]
+        public void MuxerName_IsCorrect()
+        {
+            // Assert
+            Assert.Equal("dotnet", DotnetMuxer.MuxerName);
```
jviau: This test doesn't add much value. Recommend removing it.

test/Cli/Func.UnitTests/HelperTests/DotnetMuxerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                Assert.Equal(string.Empty, DotnetMuxer.ExeSuffix);
+            }
+        }
+
+        [Fact]
+        public void GetDataFromAppDomain_ReturnsNull_ForNonExistentKey()
+        {
+            // Act
+            var result = DotnetMuxer.GetDataFromAppDomain("NON_EXISTENT_KEY_12345");
```
jviau: Nit: not sure this has much value for being separately testable either. You are testing a 1 line proxy to a runtime API. We can assume the runtime API is already tested.

test/Cli/Func.UnitTests/HelperTests/DotnetMuxerTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // Act
+            var path = muxer.MuxerPath;
+
+            // Assert
+            Assert.Contains("dotnet", path, StringComparison.OrdinalIgnoreCase);
+        }
+
+        [Fact]
+        public void MuxerPath_PointsToExecutableFile()
```
jviau: Maybe we can actually invoke the `dotnet` exe with something like `--version` arg to assert this in fact `dotnet` executable that was found.

src/Cli/func/Helpers/DotnetMuxer.cs
liliankasem: Ah I removed some of the TFM switch regions so this is outdated, let me update!

src/Cli/func/Helpers/DotnetMuxer.cs
```diff
@@ -0,0 +1,83 @@
...
+
+public class DotnetMuxer
+{
+    public static readonly string MuxerName = "dotnet";
+
+    private readonly string _muxerPath;
+
+    public static readonly string ExeSuffix = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".exe" : string.Empty;
+
+    public DotnetMuxer()
```
jviau: Having this be a class doesn't seem to offer much here. I understand this is a copy & paste, but I would assume they made it a class to fit their needs (capturing of the lookup result or exception).

The usage in this PR has this class instantiated each time `.MuxerPath` is accessed. This could be a static function in that case.

src/Cli/func/Helpers/FileNameSuffixes.cs
liliankasem: Yeah line 24 in Muxer:

`string muxerPathMaybe = Path.Combine(rootPath, $"{MuxerName}{FileNameSuffixes.CurrentPlatform.Exe}");`

test/Cli/Func.UnitTests/HelperTests/DotnetMuxerTests.cs
liliankasem: This is what happens when you have copilot write tests! I'll remove the ones that don't make sense

src/Cli/func/Helpers/DotnetMuxer.cs
liliankasem: Agree that static makes sense, we can revist this + DI during our refactor work.

src/Cli/func/Helpers/FileNameSuffixes.cs
liliankasem: No longer used - removed from codebase

jviau (approved):

src/Cli/func/Helpers/DotnetMuxer.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        if (muxerPath is null)
+        {
+            throw new InvalidOperationException("Unable to locate dotnet multiplexer");
+        }
+
+        return muxerPath;
+    }
```
jaliyaudagedara, line 60: I am sorry, too much nesting makes it a bit harder to read, early returns would make it more readable.

Something like,
```csharp
public static string GetMuxerPath()
{
    string muxerPath;

    // Most scenarios are running dotnet.dll as the app
    // Root directory with muxer should be two above app base: <root>/sdk/<version>
    string rootPath = Path.GetDirectoryName(Path.GetDirectoryName(AppContext.BaseDirectory.TrimEnd(Path.DirectorySeparatorChar)));
    if (rootPath is not null)
    {
        muxerPath = Path.Combine(rootPath, $"{_muxerName}{_exeSuffix}");
        if (File.Exists(muxerPath))
        {
            return muxerPath;
        }
    }

    // Best-effort search for muxer.
    muxerPath = Environment.ProcessPath;

    // The current process should be dotnet in most normal scenarios except when dotnet.dll is loaded in a custom host like the testhost
    if (muxerPath is not null && Path.GetFileNameWithoutExtension(muxerPath).Equals("dotnet", StringComparison.OrdinalIgnoreCase))
    {
        return muxerPath;
    }

    // SDK sets DOTNET_HOST_PATH as absolute path to current dotnet executable
    muxerPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
    if (muxerPath is not null)
    {
        return muxerPath;
    }

    // Fallback to DOTNET_ROOT which typically holds some dotnet executable
    string root = Environment.GetEnvironmentVariable("DOTNET_ROOT");
    if (root is not null)
    {
        muxerPath = Path.Combine(root, $"dotnet{_exeSuffix}");
        return muxerPath;
    }
   
    throw new InvalidOperationException("Unable to locate dotnet executable");
}
```

Changed the exception message for easy understanding `Unable to locate dotnet executable`.

aishwaryabh (approved):

src/Cli/func/Helpers/DotnetMuxer.cs
liliankasem: Thanks for the suggestion! I'd like to keep this implementation close to the .NET SDK's original structure to make it easier to track upstream changes and ensure we maintain the same behavior. As for the error message, the end user ends up with this error message which should be easy to understand:

`Unable to locate the .NET SDK. Install the .NET SDK from https://aka.ms/dotnet-download or configure PATH | DOTNET_ROOT | DOTNET_HOST_PATH so that 'dotnet' is discoverable`

jaliyaudagedara: Of course, understood!

liliankasem: Actually had a change of heart on this, I want to support more areas where dotnet would be so at this point it does make sense to clean up for readability as we keep adding on: https://github.com/Azure/azure-functions-core-tools/pull/4738

Issue #4296, opened 2025-03-12
Refactor Core tools E2E tests


aishwaryabh:
This work would involve creating a new E2E test project to slowly start migrating the E2E tests from the old way of using the AzCliTester to having a more managed process tracking strategy.

PR #4734, opened 2025-11-17
Clean up repository

liliankasem:
Clean up repository & general housekeeping.

- Update documentation
- Fix version test
- Remove old test project

### DIFF
--- a/test/Cli/Func.UnitTests/ActionsTests/PackAction/PackValidationHelperTests.cs
+++ b/test/Cli/Func.UnitTests/ActionsTests/PackAction/PackValidationHelperTests.cs
@@ -1,0 +1,91 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Azure.Functions.Cli.Actions.LocalActions.PackAction;
+using FluentAssertions;
+using Xunit;
+
+namespace Azure.Functions.Cli.UnitTests.ActionsTests.PackAction
+{
+    public class PackValidationHelperTests : IDisposable
+    {
+        private readonly string _tempDirectory;
+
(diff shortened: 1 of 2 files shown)


test/Azure.Functions.Cli.Tests/Actions/LocalActions/PackAction/PackValidationHelperTests.cs
```diff
@@ -1,63 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See LICENSE in the project root for license information.
```
aishwaryabh, line 2: I think this file is still relevant. Should we move it to the other test project?

liliankasem: Still relvant as in we should be running a test like this? Or as in this is being used today? (I assume the latter) 

I don't want to do any migration in this PR, but anything we should still be testing/need to write tests for we should have a work item to migrate - is there anything missing from this epic that needs to be added? 

https://github.com/Azure/azure-functions-core-tools/issues/4296

aishwaryabh: Yeah I meant that we still use this file today! I think this should technically be a unit test (not sure why it's in the e2e test repo), so maybe we can move it over to `test/Cli/Func.UnitTests`? Does that make sense?

liliankasem: Yup make sense, we do not run this test project anywhere so it is not being used in that way, but I understand what you mean!

aishwaryabh (approved):

PR #4763, opened 2026-01-07
[Design] Ensuring Core Tools can be used to create and run functions offline

liliankasem:
Design proposal for offline support in Core Tools.

One of the major issues is that today we require internet connectivity to download extension bundles from `cdn.functions.azure.com` during `func start`, causing local development to fail in offline. This design proposes making the tool fully functional offline by implementing cache-first fallbacks for all network operations while maintaining optimal online performance.

### DIFF
--- a/docs/design/approved/offline-mode-support.md
+++ b/docs/design/approved/offline-mode-support.md
@@ -270,8 +270,8 @@ When you have network connectivity, you can use `func bundles download` to downl
 - Add clear instructions of how to install bundles
 - These changes should fix the issue for both `func start and func new`
 
-### 3: Validate and shortcircuit other network calls
-- Update `func publish` to shortcircut if offline/no network available
+### 3: Validate and short circuit other network calls
+- Update `func publish` to short circuit if offline/no network available
 - Validate `func pack` works offline
 
 ### 4: Offline Flag (Needs design + review)


docs/design/approved/offline-mode-support.md
```diff
@@ -0,0 +1,291 @@
...
+
+- Skip version check entirely when `--offline` flag is set.
+- Reduce HTTP timeout to 2 seconds (from default 100s) for faster failure on slow networks.
+- Keep existing graceful exception handling.
+
+---
+
+### 5. Global Offline Flag
+
+**Problem**: Have a explicit offline mode to skip network attempts entirely.
```
satvu: nit: an

docs/design/approved/offline-mode-support.md
```diff
@@ -0,0 +1,291 @@
...
+**Proposed Changes**:
+
+- Add `func bundle download` command to explicitly download and cache bundles from host.json configuration.
+- Add `func bundle list` command to display cached bundles with age information. Both commands help users prepare for and troubleshoot offline scenarios.
+
+---
+
+### 4. Version Check Optimization
+
+**Problem**: GitHub version check can timeout and delay startup unnecessarily.
```
satvu: Discussed in chat - GitHub is a typo here?

liliankasem: this refers to the github API call we do to get the latest version of the CLI. will clarify

docs/design/approved/offline-mode-support.md
```diff
@@ -0,0 +1,291 @@
...
+
+### 2: Fallback to bundles download path if network fails
+- Add logic to check the bundle download location for a valid bundles installation if unable to download latest.
+- Add error logs when bundles are not available (unable to download, and not in cache)
+- Add warning logs when cached bundles are in use
+- Add clear instructions of how to install bundles
+- These changes should fix the issue for both `func start and func new`
+
+### 3: Validate and shortcircuit other network calls
+- Update `func publish` to shortcircut if offline/no network available
```
satvu: nit: short circuit

aishwaryabh (approved):

docs/design/approved/offline-mode-support.md
```diff
@@ -0,0 +1,291 @@
...
+↓
+Bundles Exists? → Yes → Use + warn that cached bundles are used
+↓
+No → Error (unable to download bundles and no cache available)
+```
+**Other changes**:
+
+1. We should disable the `EnsureLatest` bundles setting from the host. We should not need this if we
+do the bundles management on the core tools side.
+2. We should not download bundles for stacks that do not require it (e.g. dotnet apps)
```
manvkaur, line 138: dotnet in-proc scripting apps will need bundle.

Another consideration, if user just has http or timer trigger functions in their function app, bundle is not required to run the app. This can only be determined for apps with host based indexing or customer's can remove it themselves.

Only need for bundle in such case would be templates

PR #4764, opened 2026-01-07
Add new `bundles` command to manage extension bundles

### LINKED ISSUE #4769: Implement new command to manage extension bundles
- Implement `func bundle download` commandAdd a comment on  line R262Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
- Implement `func bundle list` command
- Implement `func bundle path` command (this already exists, just need to redirect to this new command path)

liliankasem:

resolves #4769

### Pull request checklist

### Additional information

Add commands to mange bundles installation

### DIFF
--- a/src/Cli/func/Actions/HostActions/Startup.cs
+++ b/src/Cli/func/Actions/HostActions/Startup.cs
@@ -4,6 +4,7 @@
 using Azure.Functions.Cli.Actions.HostActions.WebHost.Security;
 using Azure.Functions.Cli.Diagnostics;
 using Azure.Functions.Cli.ExtensionBundle;
+using Colors.Net;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -16,6 +17,8 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Actions.HostActions
 {
@@ -120,6 +123,11 @@ namespace Azure.Functions.Cli.Actions.HostActions
                         ? extensionBundleOptions.DownloadPath
                         : ExtensionBundleHelper.GetBundleDownloadPath(bundleId);
                     Environment.SetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__downloadPath", downloadPath);
+                }
+                else if (!string.IsNullOrEmpty(extensionBundleOptions.DownloadPath))
+                {
+                    // Both env var and host.json are configured - inform user env var takes precedence
+                    ColoredConsole.WriteLine(WarningColor($"Extension bundle downloadPath is configured in both host.json and environment variable. Using environment variable value: {existingDownloadPath}"));
                 }
 
                 Environment.SetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__ensureLatest", "true");


src/Cli/func/Actions/HostActions/Startup.cs
```diff
@@ -110,7 +110,13 @@ private void SetBundlesEnvironmentVariables()
             var bundleId = ExtensionBundleHelper.GetExtensionBundleOptions(_hostOptions).Id;
             if (!string.IsNullOrEmpty(bundleId))
             {
-                Environment.SetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__downloadPath", ExtensionBundleHelper.GetBundleDownloadPath(bundleId));
+                // Only set the download path if not already set by the user
+                var existingDownloadPath = Environment.GetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__downloadPath");
```
jviau, line 118: Will this be read from host.json also? I ask because that setting looks like a typical host.json setting

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
```diff
@@ -0,0 +1,119 @@
...
+
+                ColoredConsole.WriteLine($"Downloading extension bundle: {options.Id}");
+                ColoredConsole.WriteLine($"Version range: {options.Version}");
+
+                // Check if already downloaded (unless force flag is set)
+                if (!Force)
+                {
+                    try
+                    {
+                        var bundleDetails = await extensionBundleManager.GetExtensionBundleDetails();
```
jviau: nit: it would help to organize some of these large blocks into their own methods. Like a `private bool CheckIfBundleExists(...)`

src/Cli/func/Actions/LocalActions/GetBundlePathAction.cs
```diff
@@ -0,0 +1,53 @@
...
+using Fclp;
+
+namespace Azure.Functions.Cli.Actions.LocalActions
+{
+    [Action(Name = "path", Context = Context.Bundles, HelpText = "Get the path to the downloaded extension bundle.")]
+    internal class GetBundlePathAction : BaseAction
+    {
+        public string Language { get; set; }
+
+        public override ICommandLineParserResult ParseArgs(string[] args)
```
jviau: Does this need to be overriden if you are just calling base?

src/Cli/func/Actions/LocalActions/ListBundleAction.cs
```diff
@@ -0,0 +1,85 @@
...
+            return base.ParseArgs(args);
+        }
+
+        public override async Task RunAsync()
+        {
+            var extensionBundleManager = ExtensionBundleHelper.GetExtensionBundleManager();
+
+            if (!extensionBundleManager.IsExtensionBundleConfigured())
+            {
+                ColoredConsole.WriteLine(WarningColor($"Extension bundle is not configured in host.json."));
```
jviau: two command which expect host.json to have bundles configured in it. Do you think its worth adding a command to help add it to host.json? Like `func bundles add`?

Also do you see any value in being able to list available bundles to download?

liliankasem: I think that's a good idea, let me add that!

As for list, I think this could help offline to see what you have installed without having to browse to that path itself to see what is available

src/Cli/func/Actions/HostActions/Startup.cs
liliankasem: I am pretty sure this is a host.json config - we need to do a better job of updating that schema :D 

But yes you're right, we should check there too

liliankasem:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/Azure/azure-functions-core-tools/pull/4764#pullrequestreview-3753932332)

src/Cli/func/Actions/HostActions/Startup.cs
```diff
@@ -107,10 +107,21 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
...
         {
-            var bundleId = ExtensionBundleHelper.GetExtensionBundleOptions(_hostOptions).Id;
+            var extensionBundleOptions = ExtensionBundleHelper.GetExtensionBundleOptions(_hostOptions);
+            var bundleId = extensionBundleOptions.Id;
             if (!string.IsNullOrEmpty(bundleId))
             {
-                Environment.SetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__downloadPath", ExtensionBundleHelper.GetBundleDownloadPath(bundleId));
+                // Only set the download path if not already set via environment variable
+                var existingDownloadPath = Environment.GetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__downloadPath");
+                if (string.IsNullOrEmpty(existingDownloadPath))
```
aishwaryabh, line 119: so just checking if the user has both env variable and host.json value configured, we fall back to env variable value? Should we log this in case they expect otherwise?

src/Cli/func/Actions/LocalActions/AddBundleAction.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Add the extension bundle configuration based on the selected channel
+            var bundleConfig = await BundleActionHelper.GetBundleConfigForChannel(Channel);
+            var bundleConfigObj = JsonConvert.DeserializeObject<JToken>(bundleConfig);
+            hostJsonObj.Add(Constants.ExtensionBundleConfigPropertyName, bundleConfigObj);
+
+            // Write back to host.json
+            var updatedHostJson = JsonConvert.SerializeObject(hostJsonObj, Formatting.Indented);
+            await FileSystemHelpers.WriteAllTextToFileAsync(hostFilePath, updatedHostJson);
+
+            var channelDisplay = Channel == BundleChannel.GA ? string.Empty : $" ({Channel})";
```
aishwaryabh, line 73: how come we don't want to display the channel for when the bundle channel is GA?

src/Cli/func/Actions/LocalActions/BundleActionHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+            return true;
+        }
+
+        public static void PrintNotConfiguredWarning()
+        {
+            var hostFilePath = Path.Combine(Environment.CurrentDirectory, ScriptConstants.HostMetadataFileName);
+            ColoredConsole.WriteLine(WarningColor($"Extension bundle is not configured in {hostFilePath}."));
+            ColoredConsole.WriteLine("To configure extension bundles, add the following to your host.json:");
+            ColoredConsole.WriteLine(AdditionalInfoColor(@"{
+  ""extensionBundle"": {
```
aishwaryabh, line 116: nit: fix spacing

src/Cli/func/Actions/LocalActions/AddBundleAction.cs
liliankasem: Because GA is the default so `Extension bundle configuration added to ...` makes sense, but if you're not on GA and go preview or experimental we shouls call that out

src/Cli/func/Actions/HostActions/Startup.cs
liliankasem: We do log the location of the bundles so they should be able to tell/troubleshoot if its different

> dotnet run --project src/Cli/func/Azure.Functions.Cli.csproj -- bundles download --script-root ./_pyapp --force
Downloading extension bundle: Microsoft.Azure.Functions.ExtensionBundle
Version range: [4.0.0, 5.0.0)
Clearing existing bundles from /Users/likasem/.azure-functions-core-tools/Functions/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle...
Successfully downloaded extension bundle 4.30.0
Location: /Users/likasem/.azure-functions-core-tools/Functions/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/4.30.0

But that's an interesting edge case, we don't want to override anything the user has set so we can't take what's in host.json and override their env var. But I can add an additional log

src/Cli/func/Actions/LocalActions/BundleActionHelper.cs
liliankasem: intentional for string literals

src/Cli/func/Actions/HostActions/Startup.cs
liliankasem: Priority order:

- Pre-existing environment variable (highest)
- host.json downloadPath
- Default path from GetBundleDownloadPath(bundleId) (lowest)

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
```diff
@@ -0,0 +1,126 @@
...
+                await ExtensionBundleHelper.GetExtensionBundle();
+
+                // Get a fresh manager to verify the download at the correct location
+                var verifyManager = ExtensionBundleHelper.GetExtensionBundleManager();
+                var bundlePath = await verifyManager.GetExtensionBundlePath();
+
+                if (!string.IsNullOrEmpty(bundlePath) && Directory.Exists(bundlePath))
+                {
+                    var details = await extensionBundleManager.GetExtensionBundleDetails();
+                    ColoredConsole.WriteLine(VerboseColor($"Successfully downloaded extension bundle {details.Version}"));
```
aishwaryabh, line 64: nit do we want this to be only logged when verbose is enabled or just logged in general since it has the verbose color? I think logging in general is the right call

liliankasem: VerboseColor is just a color, it does not check if this is --verbose or not; just felt like the right color for this log though

PR #4768, opened 2026-01-09
Reduce noise by making some logs verbose

liliankasem:

n/a

### Pull request checklist

### Additional information

Reduce noise by making some logs verbose

Also fix the mcp template not working for dotnet

### DIFF
--- a/test/Cli/TestFramework/Assertions/CommandResultAssertions.cs
+++ b/test/Cli/TestFramework/Assertions/CommandResultAssertions.cs
@@ -45,29 +45,28 @@ namespace Azure.Functions.Cli.TestFramework.Assertions
 
         public AndConstraint<CommandResultAssertions> StartInProc6Host()
         {
-            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc6 model host."))
+            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc6 model host.") && _commandResult.StdOut.Contains("Selected inproc6 host."))
                 .FailWith($"The command output did not contain expected result for inproc6 host.{Environment.NewLine}");
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> StartInProc8Host()
         {
-            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc8 model host."))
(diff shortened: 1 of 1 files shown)


src/Cli/func/ConsoleApp.cs
```diff
@@ -405,6 +414,46 @@ internal IAction Parse()
...
+                // Global flags (no value)
+                if (lowerArg == "--verbose" || lowerArg == "--version" || lowerArg == "-v")
+                {
+                    continue;
+                }
+
+                // Global options that take values
+                if (lowerArg == "--script-root" || lowerArg == "--prefix")
+                {
+                    skipNext = true; // Skip the next argument (the value)
```
jviau: What if the user doesn't supply the value to the argument? Should probably check in the `if (skipNext)` if the value starts with `-` or not. Or will it already be validated by this point?

liliankasem: UpdateCurrentDirectory already validates that --script-root/--prefix has a valid path value and throws a CliException if the path doesn't exist. This happens before FilterOutGlobalOptions ever runs.

test/Cli/TestFramework/Assertions/CommandResultAssertions.cs
```diff
@@ -45,28 +45,29 @@ public AndConstraint<CommandResultAssertions> NotHaveStdOutContaining(string pat
...
         {
-            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc6 model host.") && _commandResult.StdOut.Contains("Selected inproc6 host."))
+            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc6 model host."))
                 .FailWith($"The command output did not contain expected result for inproc6 host.{Environment.NewLine}");
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> StartInProc8Host()
         {
-            Execute.Assertion.ForCondition(_commandResult.StdOut is not null && _commandResult.StdOut.Contains("Starting child process for inproc8 model host.") && _commandResult.StdOut.Contains("Selected inproc8 host."))
```
aishwaryabh, line 55: how come we're removing the second assert statement?

liliankasem: Let me double check on this, I think it was to do with verbosity but it might be a redundant change

gata-bbs:
FYI, I think this PR removed user log outputs. 
Before this change: `func start` would display Python `logging.info()` messages.
After this change: `func start --verbose` no longer shows Python user logs. I must now use `func start --user-log-level Trace` to see them.

PR #4772, opened 2026-01-12
Fallback to cached bundles if there is no network connection

aishwaryabh:

resolves ##4770

This PR addresses the scenario if we want to fallback to cached bundles specifically for `func start`. For detecting if the system is online, we check by trying to ping an endpoint and cache if the system is offline or not for a specified interval for future calls. We also only download bundles in `func start` if the customer has specified `AzureFunctionsJobHost__extensionBundle__ensureLatest` to be true OR if we detect that the system is offline.

Example of running with no internet connection with no bundles cached:
image

Example of running with no internet connection but with bundles cached already:
image

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/test/Cli/Func.E2ETests/Commands/FuncStart/Core/BaseOfflineBundleTests.cs
+++ b/test/Cli/Func.E2ETests/Commands/FuncStart/Core/BaseOfflineBundleTests.cs
@@ -40,7 +40,7 @@ namespace Azure.Functions.Cli.E2ETests.Commands.FuncStart.Core
                 .WithEnvironmentVariable(Common.Constants.FunctionsWorkerRuntime, language)
                 .Execute(["--offline", "--verbose", "--port", port.ToString()]);
 
-            result.Should().HaveStdOutContaining("Running in offline mode but no cached version of extension bundle");
+            result.Should().HaveStdOutContaining("Running in offline mode. Using cached extension bundle version");
         }
 
         /// <summary>


src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -457,7 +457,21 @@ public override async Task RunAsync()
 
             (var listenUri, var baseUri, var certificate) = await Setup();
 
-            await ExtensionBundleHelper.GetExtensionBundle();
+            // Check if user explicitly disabled EnsureLatest to skip bundle download
+            var ensureLatestEnvVar = Environment.GetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__ensureLatest");
+            var shouldDownloadBundles = !string.IsNullOrEmpty(ensureLatestEnvVar) && ensureLatestEnvVar.Equals("false", StringComparison.OrdinalIgnoreCase);
+
+            // Download bundles if specified by user or if offline mode is detected
```
aishwaryabh: I figured we still want to respect the user's env variable choice here for wanting to download bundles in the host, but we ignore their choice if we detect the system is offline

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -457,7 +457,21 @@ public override async Task RunAsync()
 
             (var listenUri, var baseUri, var certificate) = await Setup();
 
-            await ExtensionBundleHelper.GetExtensionBundle();
+            // Check if user explicitly disabled EnsureLatest to skip bundle download
+            var ensureLatestEnvVar = Environment.GetEnvironmentVariable("AzureFunctionsJobHost__extensionBundle__ensureLatest");
```
jviau: Will this be read from host.json? We are only checking env variable here. The format of the variable itself looks like it is meant for `IConfiguration`, so I think a user would reasonably expect this to also work from host.json.

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -457,7 +457,21 @@ public override async Task RunAsync()
...
+            var shouldDownloadBundles = !string.IsNullOrEmpty(ensureLatestEnvVar) && ensureLatestEnvVar.Equals("false", StringComparison.OrdinalIgnoreCase);
```
jviau: I always find it confusing to figure out boolean logic like this. I recommend either doing a null-coalesce (and empty string coalesce) and then parsing to a boolean. Makes it a bit clearer when your default is when the value is not present.

jviau: Don't use a string comparison to `false`. Use `bool.TryParse`

src/Cli/func/Actions/HostActions/StartHostAction.cs
aishwaryabh: I was under the impression that this wasn't set in host.json, but will check!

jviau: I am asking because the format of the environment variable makes me expect this to work in host.json:

`AzureFunctionsJobHost` prefix -- this is the prefix that `host.json` config is loaded under.
`__` (double underscore separators) -- this is the convention for mapping env variables to IConfiguration sections.

So, everything about this scream "this is interchangeable with other IConfiguration sources, such as host.json". Is this a convention core-tools has been following? Using `AzureFunctionsJobHost` and double underscores as environment variable only settings?

liliankasem: Yes we should check the host.json here too

> Is this a convention core-tools has been following? Using AzureFunctionsJobHost and double underscores as environment variable only settings?

Not that I am aware of; core-tools already reads the host.json in different places to check settings so its not unprecedented to do that

liliankasem: I did just ask copilot to take a look for any instances where we only check env var and looks like there are 2-but,  I think some of this is intentioal. We do want to override customer settings for some specific things like `SequentialRestart`. But log level could probably check the host.json

---

Yes. There are cases where core tools checks for AzureFunctionsJobHost__* env vars without checking host.json:

1. AzureFunctionsJobHost__SequentialRestart in [Program.cs:75]

`EnvironmentHelper.SetEnvironmentVariableAsBoolIfNotExists(Constants.SequentialJobHostRestart);`

Only checks env var, not host.json. Sets to "true" if env var is empty.

2. AzureFunctionsJobHost__logging__logLevel__Function in [StartHostAction.cs:906-910]

```
string functionLogLevelKey = "AzureFunctionsJobHost__logging__logLevel__Function";
if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(functionLogLevelKey)))
{
    Environment.SetEnvironmentVariable(functionLogLevelKey, ...);
}
```

Only checks env var. If someone set logging.logLevel.Function in host.json, this would still set the env var (which then overwrites host.json).

aishwaryabh:
@copilot open a new PR addressing the comments in [this thread](https://github.com/Azure/azure-functions-core-tools/pull/4772#pullrequestreview-3753931948). Can you also rebase to main for the conflicts?

src/Cli/func/Actions/HostActions/StartHostAction.cs
aishwaryabh: added check to also check host.json. The priority order is 1) env variable, 2) host.json, and if both of those aren't populated we just default to EnsureLatest being true

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -64,29 +75,64 @@ public static ExtensionBundleContentProvider GetExtensionBundleContentProvider()
...
-                await RetryHelper.Retry(
-                    func: async () => await extensionBundleManager.GetExtensionBundlePath(httpClient),
-                    retryCount: MaxRetries,
-                    retryDelay: _retryDelay,
-                    displayError: false);
+                if (IsOffline())
+                {
+                    throw new HttpRequestException("System is offline.");
+                }
+
```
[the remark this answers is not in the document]
aishwaryabh: I think my approach is cleaner

src/Cli/func/Azure.Functions.Cli.csproj
```diff
@@ -8,6 +8,10 @@
     <DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>
   </PropertyGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Azure.Functions.Cli.UnitTests" />
```
liliankasem: Curious why this is added here? 

We already have internals visible to set in the `AssemblyInfo` -> `/src/Cli/func/Properties/AssemblyInfo.cs`

src/Cli/func/Helpers/RetryHelper.cs
```diff
@@ -23,6 +23,12 @@ public static async Task Retry(Func<Task> func, int retryCount, TimeSpan retryDe
                 }
                 catch (Exception e)
                 {
+                    // Check if this exception type should not be retried
+                    if (nonRetryableExceptions != null && nonRetryableExceptions.Any(t => t.IsAssignableFrom(e.GetType())))
```
liliankasem: 1. Always default to `is not` - it's a safer option

```suggestion
                    if (nonRetryableExceptions is not null && nonRetryableExceptions.Any(t => t.IsAssignableFrom(e.GetType())))
```

2. Why are we using reflection here? What prompted this change?

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -18,6 +19,7 @@ internal class ExtensionBundleHelper
         private const int MaxRetries = 3;
         private const string ExtensionBundleStaticPropertiesUrl = "https://cdn.functions.azure.com/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/staticProperties.json";
         private const string DefaultExtensionBundleVersionRange = "[4.*, 5.0.0)";
+        private const string AzureFunctionsJobHostExtensionBundleDownloadPath = "AzureFunctionsJobHost__extensionBundle__downloadPath";
```
liliankasem: Let's move this to common/Constants, we have other flags here already i.e. https://github.com/Azure/azure-functions-core-tools/blob/af9404fa5872d8fa5fa2454a7cfeb2ee0cda2832/src/Cli/func/Common/Constants.cs#L69C57-L69C78

There should be a bunch of references to `AzureFunctionsJobHost__extensionBundle__downloadPath` to update

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -79,24 +91,58 @@ public static string GetBundleDownloadPath(string bundleId)
...
+
+            using var httpClient = new HttpClient { Timeout = _httpTimeout };
+
+            // Attempt to get the extension bundle path, which will trigger the download if not already present
+            await RetryHelper.Retry(
+                func: async () => await TryToGetExtensionBundlePath(extensionBundleManager, extensionBundleOptions, httpClient),
+                retryCount: MaxRetries,
+                retryDelay: _retryDelay,
+                displayError: false,
+                nonRetryableExceptions: new[] { typeof(HttpRequestException) });
```
liliankasem: This is too generic, `HttpRequestException` does not always mean we are offline so should not be the indicator that we go into offline mode

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -79,24 +91,58 @@ public static string GetBundleDownloadPath(string bundleId)
...
+                await extensionBundleManager.GetExtensionBundlePath(httpClient);
             }
-            catch (Exception)
+            catch (HttpRequestException)
             {
-                // Don't do anything here.
-                // There will be another attempt by the host to download the Extension Bundle.
-                // If Extension Bundle download fails again in the host then the host will return the appropriate customer facing error.
+                // Mark as offline for future operations
+                MarkAsOffline();
```
liliankasem: Offline mode should be a global state of the CLI, and not just something controlled/is inside of the extension bundles helper.

The design here should be a global setting that if we detect no network connection or a global --offline flag is passed, the global setting for isOffline = true. Other actions can then use this flag to determine how they behave. i.e. the ext bundle helper could check for that global setting and skip download all together and go straight to read from cache.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
liliankasem: Same for ensureLatest

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
liliankasem: i.e. within GlobalCoreToolsSettings:Init: `_isOfflineMode = args.Contains("--offline") || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.FunctionsCoreToolsOffline);`

Can also introduce an env for for setting offline: `public const string FunctionsCoreToolsOffline = "FUNCTIONS_CORE_TOOLS_OFFLINE";`

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -308,5 +354,201 @@ private static int CompareVersions(string v1, string v2)
...
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Detects if the system is currently offline (no network connectivity to CDN).
+        /// Uses caching to avoid excessive network checks.
+        /// </summary>
+        /// <returns>True if offline, false if online.</returns>
+        internal static async Task<bool> IsOfflineAsync()
```
liliankasem: This should all be centralized to a an offline helper, not within ext bundles. i.e.

```csharp
    /// <summary>
    /// Central helper for determining offline/online status across the CLI.
    /// Respects the global --offline flag or FUNCTIONS_CORE_TOOLS_OFFLINE environment variable,
    /// and caches probe results to avoid repeated network calls.
    /// </summary>
    internal static class OfflineHelper
```

This should be responsible for all offline activity and tracking

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: Removed this!

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: done

src/Cli/func/Azure.Functions.Cli.csproj
aishwaryabh: yup removed this was an unnecessary change

src/Cli/func/Helpers/RetryHelper.cs
aishwaryabh: do not need this change anymore! I added it since we don't want to keep retrying if we are truly offline so added another condition

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: missed this earlier but added as global state of CLI!

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -13,11 +13,31 @@ public static class GlobalCoreToolsSettings
         private static WorkerRuntime _currentWorkerRuntime;
         private static bool _isHelpRunning;
         private static bool _isVerbose;
+        private static bool _isExplicitOffline;
```
liliankasem:
```suggestion
        private static bool _isOffline;
```

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// Respects the global --offline flag or FUNCTIONS_CORE_TOOLS_OFFLINE environment variable,
+    /// and caches probe results to avoid repeated network calls.
+    /// </summary>
+    internal static class OfflineHelper
+    {
+        private const string ConnectivityCheckUrl = "https://cdn.functions.azure.com/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/staticProperties.json";
+
+        // Controls how often we re-probe network connectivity
+        private static readonly object _probeLock = new object();
+        private static readonly TimeSpan _probeInterval = TimeSpan.FromSeconds(10);
```
liliankasem: We shouldn't have to probe multiple times in the lifespan of the CLI, the CLI is not a long running process (except for perhaps start and even then we only need to know if offline before we actually start the host process) - the network check should happen once when you run a given command that needs to check for offline mode (such a start and new). Once we're in offline mode it should just be set.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -79,24 +82,56 @@ public static string GetBundleDownloadPath(string bundleId)
...
+        }
+
+        /// <summary>
+        /// Handles the fallback logic when the system is offline or a network failure occurs.
+        /// Attempts to use a cached extension bundle; logs warnings or errors accordingly.
+        /// </summary>
+        private static void HandleOfflineBundleFallback(ExtensionBundleOptions extensionBundleOptions)
+        {
+            // If user didn't specify an extension bundle ID, no need to download extension bundle
+            if (string.IsNullOrEmpty(extensionBundleOptions.Id))
```
liliankasem: Can you walk me through this scenario? What would happen that would lead to a bundles check when there is no bundles defined in host.json?

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -893,6 +905,38 @@ private void PrintMigrationWarningForDotnet6Inproc()
...
         }
 
+        /// <summary>
+        /// Determines whether extension bundles should be downloaded.
+        /// Priority: 1) Environment variable, 2) host.json, 3) Default (download enabled).
+        /// </summary>
+        /// <returns>A tuple indicating whether to download bundles and the source of the setting.</returns>
+        internal (bool ShouldDownload, string Source) ShouldDownloadExtensionBundles()
+        {
+            const string hostJsonConfigKey = "AzureFunctionsJobHost:extensionBundle:ensureLatest";
```
liliankasem: You can do similar to what you did in the bundles config  `Constants.ExtensionBundleEnsureLatest.Replace("__", ":")``

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -79,24 +82,56 @@ public static string GetBundleDownloadPath(string bundleId)
...
-                // Don't do anything here.
-                // There will be another attempt by the host to download the Extension Bundle.
-                // If Extension Bundle download fails again in the host then the host will return the appropriate customer facing error.
+                ColoredConsole.WriteLine(OutputTheme.VerboseColor("Downloading extension bundles..."));
+            }
+
+            using var httpClient = new HttpClient { Timeout = _httpTimeout };
+            var extensionBundleManager = await GetExtensionBundleManagerAsync(extensionBundleOptions);
+
+            await RetryHelper.Retry(
```
liliankasem: There is no cache fallback when bundle download fails after an initial success probe which doesn't meet our requirement of falling back to cahce on network failures.

i.e imagine we run in online mode (no network issues detected, and offline flag not set) but in that split second where we go to download, we do get a network failure, even after multiple retries. At this point we should fallback to cache instead of have an exception

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -308,5 +343,128 @@ private static int CompareVersions(string v1, string v2)
...
+        internal static bool TryGetCachedBundle(string bundleId, string versionRange, out string cachedBundleVersion)
+        {
+            cachedBundleVersion = null;
+
+            if (string.IsNullOrEmpty(bundleId))
+            {
+                return false;
+            }
+
+            // First, check the customer-configured download path (if set via environment variable)
```
liliankasem, line 424: Host json is never checked here - apps with a custom path and no env var will get a “no cached bundle” error even when the bundle is present

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
liliankasem: We should setup a try-catch around every `GetExtensionBundlePath` to reuse HandleOfflineBundleFallback on HttpRequest exceptions - this can be a helper method if the code is the same everywhere. An exception to this might be the specific download bundles action which is meant to download - we could check offline mode here and present a nice message about being unable to download

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -893,6 +905,38 @@ private void PrintMigrationWarningForDotnet6Inproc()
             ColoredConsole.WriteLine(WarningColor($".NET 6 is no longer supported. Please consider migrating to a supported version. For more information, see https://aka.ms/azure-functions/dotnet/net8-in-process. If you intend to target .NET 8 on the in-process model, make sure that '{Constants.InProcDotNet8EnabledSetting}' is set to '1' in {Constants.LocalSettingsJsonFileName}.\n"));
         }
 
+        /// <summary>
+        /// Determines whether extension bundles should be downloaded.
+        /// Priority: 1) Environment variable, 2) host.json, 3) Default (download enabled).
+        /// </summary>
+        /// <returns>A tuple indicating whether to download bundles and the source of the setting.</returns>
+        internal (bool ShouldDownload, string Source) ShouldDownloadExtensionBundles()
```
liliankasem: This all got me thinking about another requirement we had for this work: we want the CLI to manage bundles downloads, not the host. We can do this by making a couple of changes:

1. Force host “ensureLatest” off by default. In Startup.SetBundlesEnvironmentVariables only set AzureFunctionsJobHost__extensionBundle__ensureLatest when it is unset, and set it to "false" (regardless of online/offline). That keeps the host from fetching bundles on its own while still letting a user override via env var if they insist.

2. Stop flipping EnsureLatest back on in the manager options. In ExtensionBundleHelper.GetExtensionBundleManagerAsync, remove the EnsureLatest = !GlobalCoreToolsSettings.IsOfflineMode assignment (or hard‑set it to false unless an explicit env var says otherwise). That ensures the host-side manager respects our “no auto-download” policy and relies on what the CLI has already prepared.

3. Simplify ShouldDownloadExtensionBundles to no longer gate on ensureLatest. If an extension bundle is configured, let the CLI attempt the download/cached fallback every time (the offline logic inside GetExtensionBundle already handles no-network). Remove or ignore the ensureLatest environment/host.json check so we don’t skip our pre-download.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: I was thinking this would apply to .NET function apps, but maybe we can check if the runtime is dotnet and if so we don't even go down this route within `StartHostAction`

src/Cli/func/Actions/HostActions/StartHostAction.cs
aishwaryabh: done!

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: added host.json check!

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: I set up a try catch within `GetExtensionBundlePath`!

src/Cli/func/Helpers/OfflineHelper.cs
aishwaryabh: ah yes that's fair I removed the probing check

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
```diff
@@ -27,7 +28,15 @@ public override ICommandLineParserResult ParseArgs(string[] args)
...
+                throw new CliException(
+                    "Cannot download extension bundles while in offline mode. " +
+                    "Please ensure you have network connectivity and try again.");
+            }
+
+            var (success, extensionBundleManager, options, bundleBasePath) = await BundleActionHelper.TryGetBundleContextAsync();
+            if (!success)
             {
                 return;
             }
```
liliankasem: I prefer the previous syntax for this - can we keep that part as is

`if (!BundleActionHelper.TryGetBundleContext(out var extensionBundleManager, out var options, out var bundleBasePath))`

src/Cli/func/Actions/LocalActions/GetBundlePathAction.cs
```diff
@@ -11,7 +11,8 @@ internal class GetBundlePathAction : BaseAction
     {
         public override async Task RunAsync()
         {
-            if (!BundleActionHelper.TryGetBundleContext(out var extensionBundleManager, out _, out var bundleBasePath))
+            var (success, extensionBundleManager, _, bundleBasePath) = await BundleActionHelper.TryGetBundleContextAsync();
```
liliankasem: Same here, not sure why this is being changed?

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// has not explicitly requested offline mode.
+        /// </summary>
+        /// <returns>True if offline, false if online.</returns>
+        internal static async Task<bool> IsOfflineAsync()
+        {
+            try
+            {
+                using var quickClient = new HttpClient { Timeout = TimeSpan.FromSeconds(1) };
+                using var request = new HttpRequestMessage(HttpMethod.Head, ConnectivityCheckUrl);
+                using var response = await quickClient.SendAsync(request);
```
kshyju: Add `ConfigureAwait(false)` for the async method call since we don't need to capture context for this helper method.

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+                return !response.IsSuccessStatusCode;
+            }
+            catch
+            {
+                return true;
```
kshyju, line 47: Do we need any kind of debug logging here in the catch block?

src/Cli/func/Helpers/OfflineHelper.cs
kshyju: You can use `HttpCompletionOption.ResponseHeadersRead` option (which will avoid the response body check) since all we care about is knowing we get "something back" from the server.

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -52,6 +60,13 @@ public static WorkerRuntime CurrentWorkerRuntimeOrNone
         public static void Init(ISecretsManager secretsManager, string[] args)
         {
             _isVerbose = args.Contains("--verbose");
+            _isOffline = args.Contains("--offline") || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.FunctionsCoreToolsOffline);
+
+            // If offline is not set from args or env var, perform a one-time network check
+            if (!_isOffline)
```
kshyju: Does `Init` method gets called for every `func start` call? If so, are we doing a blocking network check on every call if CLI argument does not contain --offline?

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -555,6 +559,15 @@ private async Task<bool> TryHandleInProcDotNetLaunchAsync()
             return false;
         }
 
+        /// <summary>
+        /// Determines whether extension bundles should be downloaded for the given worker runtime.
+        /// Extension bundles are not needed for dotnet and dotnet-isolated runtimes.
+        /// </summary>
+        internal static bool ShouldDownloadExtensionBundles(WorkerRuntime workerRuntime)
```
liliankasem: I think `RequiresExtensionBundles` makes more sense here; this is misleading as the should download part is actually handled by the get ext bundle method

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,26 +81,119 @@ public static string GetBundleDownloadPath(string bundleId)
...
+            if (TryGetCachedBundle(extensionBundleOptions.Id, versionRange, out string cachedBundleVersion, extensionBundleOptions.DownloadPath))
+            {
+                ColoredConsole.WriteLine(OutputTheme.WarningColor($"Warning: Unable to download extension bundles. Using cached version {cachedBundleVersion}."));
+                ColoredConsole.WriteLine(OutputTheme.WarningColor("When you have network connectivity, you can run `func bundles download` to update."));
+                ColoredConsole.WriteLine();
+            }
+            else
+            {
+                // No cached bundle found
+                throw new HttpRequestException(
```
liliankasem: This is throwing a `HttpRequestException` even though the reason for the exception might not be a HTTP error. We should make this a CliException

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,26 +81,119 @@ public static string GetBundleDownloadPath(string bundleId)
...
+                OfflineHelper.MarkAsOffline();
+                HandleOfflineBundleFallback(extensionBundleOptions);
+                return GetCachedBundlePath(extensionBundleOptions);
+            }
+        }
+
+        /// <summary>
+        /// Returns the path to the best cached bundle version, or null if none is found.
+        /// </summary>
+        private static string GetCachedBundlePath(ExtensionBundleOptions extensionBundleOptions)
```
liliankasem: Both GetCachedBundlePath and HandleOfflineBundleFallback call `TryGetCachedBundle` with nearly identical parameters. In GetExtensionBundle(), both are called sequentially - this should be consolidated so we dont scan the cache dict twice

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,26 +81,119 @@ public static string GetBundleDownloadPath(string bundleId)
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// Determines whether an exception indicates a network connectivity issue
+        /// rather than an HTTP-level error (e.g. 401, 404, 500).
+        /// Walks the inner exception chain looking for common network error types.
+        /// </summary>
+        internal static bool IsNetworkException(Exception ex)
```
liliankasem, line 168: This should cover TaskCanceledException; HttpClient throws TaskCanceledException when a request times out - this should lead to offline fallback but currently won't with how this is implemented

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// </summary>
+        /// <returns>True if offline, false if online.</returns>
+        internal static async Task<bool> IsOfflineAsync()
+        {
+            try
+            {
+                using var quickClient = new HttpClient { Timeout = TimeSpan.FromSeconds(1) };
+                using var request = new HttpRequestMessage(HttpMethod.Head, ConnectivityCheckUrl);
+                using var response = await quickClient.SendAsync(request);
+                return !response.IsSuccessStatusCode;
```
liliankasem: If the CDN returns a 404 or 500, this marks the system as offline, which is incorrect. A server-side error is not the same as no network connectivity. This should only return true on exception (connection failure), not on HTTP error codes.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -308,5 +405,137 @@ private static int CompareVersions(string v1, string v2)
...
+                    {
+                        bundleVersion = latestVersion;
+                        return true;
+                    }
+                }
+            }
+            catch
+            {
+                // If we can't check the directory, return false
+                return false;
```
liliankasem: We should log on exceptions instead of hiding them - makes it more difficult to debug what is happening. We should also avoid generic catches like this

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -52,6 +60,13 @@ public static WorkerRuntime CurrentWorkerRuntimeOrNone
...
+            {
+                _isOffline = OfflineHelper.IsOfflineAsync().GetAwaiter().GetResult();
```
liliankasem: A couple of things here:

1. This performs a synchronous HTTP HEAD request on every CLI invocation (not just func start), even for commands like func --version or func --help; we should lazy init so that we only check offline when a command asks for it instead on every command

2. `.GetAwaiter().GetResult()` is a deadlock risk - I think this is a CLI but this is still fragile and we should avoid anti-patterns. Consider making Init async or using Task.Run(() => ...).GetAwaiter().GetResult() as a safer wrapper.

liliankasem: One suggestion is as follows:

```charp
private static bool _explicitOffline;
private static Lazy<bool> _networkOffline;
public static bool IsOfflineMode => _explicitOffline || _networkOffline.Value;

//init()
    _explicitOffline = args.Contains("--offline") 
        || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.FunctionsCoreToolsOffline);

    // Lazy network probe — only runs on first access of IsOfflineMode when no explicit offline flag was set
    _networkOffline = new Lazy<bool>(() =>
        _explicitOffline
            ? true 
            : Task.Run(() => OfflineHelper.IsOfflineAsync()).GetAwaiter().GetResult();

...

internal static void SetOffline(bool isOffline)
{
    _explicitOffline = isOffline;
}
```

src/Cli/func/Actions/LocalActions/GetBundlePathAction.cs
aishwaryabh: reverted this back!

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
aishwaryabh: added!

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: done

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
aishwaryabh: made the call for one time network check a lazy call

src/Cli/func/Helpers/OfflineHelper.cs
aishwaryabh: added

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -308,5 +393,136 @@ private static int CompareVersions(string v1, string v2)
...
+                    return false;
+                }
+
+                var normalizedVersion = NormalizeVersion(version);
+
+                // Check if version is >= start and < end
+                return CompareVersions(normalizedVersion, range.Value.Start) >= 0 &&
+                       CompareVersions(normalizedVersion, range.Value.End) < 0;
+            }
+            catch
```
liliankasem: We still have a catch-all here - should catch Exception or FormatException

test/Cli/Func.UnitTests/HelperTests/ExtensionBundleHelperTests.cs
```diff
@@ -144,5 +193,666 @@ public void VersionRangesIntersect_DeprecationScenarios_ReturnsExpectedResult(st
...
+                }
+
+                Environment.SetEnvironmentVariable(
+                    Constants.ExtensionBundleDownloadPath,
+                    originalEnvVar);
+            }
+        }
+
+        [Fact]
+        public void TryGetCachedBundle_SimulatesNetworkFailureNoCache_ReturnsTrue()
```
liliankasem: Titled ReturnsTrue but we assert false

`result.Should().BeFalse("no cached bundle should be found");`

release_notes.md
```diff
@@ -23,3 +23,5 @@
   - `func bundles add` - Add extension bundle configuration to host.json with `--channel` flag to select GA (default), Preview, or Experimental bundles
   - Support for custom bundle download paths via `AzureFunctionsJobHost__extensionBundle__downloadPath` environment variable
 - Added `--bundles-channel` option to `func init` command to specify extension bundle channel (GA, Preview, or Experimental) during project initialization
+- Fallback to cached bundles if there is no network connection during `func start` (#4772)
+  - Added global `--offline` variable to run in offline mode
```
```suggestion
- Added global `--offline` variable to run in offline mode (#4772)
```

test/Cli/Func.E2ETests/Commands/FuncStart/Core/BaseOfflineBundleTests.cs
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: Missing test cases for global settings:

1. Init without explicit offline, then access IsOfflineMode — No test calls Init with no offline flag/env var and then reads IsOfflineMode. This is the path that actually triggers the Lazy<bool> network probe. You can't assert on the boolean result (it depends on real connectivity), but you can verify it doesn't throw or deadlock.

2. SetOffline overrides a previously-resolved lazy — No test calls Init (no flags), accesses IsOfflineMode to force the lazy to evaluate, then calls SetOffline(true) and verifies IsOfflineMode flips to true. This validates that SetOffline properly overrides a stale lazy.

3. Init called multiple times resets the lazy — No test verifies that a second Init call replaces the previous Lazy instance, so a prior resolved value doesn't leak across.

liliankasem:
Tested this with a happy and edge cases and its working really well! Only other thing I think we should fix is thr error messaging when offline i.e.

```log
Offline check failed: No route to host (cdn.functions.azure.com:443)

Offline check failed: nodename nor servname provided, or not known (cdn.functions.azure.com:443)
```

Could do:

```csharp
// IsOfflineAsync
            catch (Exception ex)
            {
                ColoredConsole.WriteLine(WarningColor("Unable to resolve network connection, running the CLI in offline mode."));

                {
                    ColoredConsole.WriteLine(VerboseColor($"Details: {ex.Message}"));
                }

                return true;
            }
  }
```

I think we should log a better message here i.e. 

Otherwise lgtm - ready to approve with this and above comments are addressed. Please also make sure to rebase and clean up commits.

src/Cli/func/Actions/HelpAction.cs
```diff
@@ -254,6 +254,7 @@ private static void DisplayGlobalOptions()
             {
                 ("--script-root <PATH>", "Set the root directory of the function app. Changes the working directory to the specified path. Defaults to the current directory."),
                 ("--verbose", "Enable verbose output for detailed logging (not supported by all commands)."),
+                ("--offline", "Run the function app in offline mode (not supported by all commands)."),
```
kshyju: Do we need docs update for this new CLI flag? Are you tracking that?

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -52,6 +63,13 @@ public static WorkerRuntime CurrentWorkerRuntimeOrNone
         public static void Init(ISecretsManager secretsManager, string[] args)
         {
             _isVerbose = args.Contains("--verbose");
+            _explicitOffline = args.Contains("--offline") || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.FunctionsCoreToolsOffline);
+
+            // Lazy network probe — only runs on first access of IsOfflineMode when no explicit offline flag was set
+            _networkOffline = new Lazy<bool>(() =>
+                _explicitOffline
+                    ? true
+                    : Task.Run(() => OfflineHelper.IsOfflineAsync()).GetAwaiter().GetResult());
```
kshyju: Why do we need `Task.Run` here?

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -43,13 +44,16 @@ public static ExtensionBundleOptions GetExtensionBundleOptions(ScriptApplication
             return options;
         }
 
-        public static ExtensionBundleManager GetExtensionBundleManager()
+        public static ExtensionBundleManager GetExtensionBundleManager(ExtensionBundleOptions extensionBundleOption = null)
```
kshyju, line 47: _Copy pasting what review agent flagged for me. Can you validate this is a real issue?_

The new code downloads extension bundles **before** host startup, but the host.json `downloadPath` isn't propagated to the download logic at that point. Combined with `ensureLatest` now being `false`, the host can no longer self-correct, making this a regression for apps using custom `downloadPath` in host.json.

### Repro steps

1. Create a new function app (e.g. Node.js):
   ```bash
   func init MyApp --javascript
   cd MyApp
   func new --name hello --template "HTTP trigger"
   ```
2. Edit `host.json` to add a custom download path:
   ```json
   {
     "version": "2.0",
     "extensionBundle": {
       "id": "Microsoft.Azure.Functions.ExtensionBundle",
       "version": "[4.*, 5.0.0)",
       "downloadPath": "C:\\my-bundles"
     }
   }
   ```
3. Ensure the `AzureFunctionsJobHost__extensionBundle__downloadPath` env var is **not** set.
4. Run `func start`.

**Expected:** Bundles are downloaded to `C:\my-bundles` and the host starts successfully.

**Actual:** Bundles are downloaded to the default path (`~/.azure-functions-core-tools/Functions/ExtensionBundles/...`). The host looks for them in `C:\my-bundles` (from host.json), finds nothing, and fails because `ensureLatest=false` prevents it from downloading them itself.

### Execution order in `StartHostAction.RunAsync()`

```
1. Setup()                      → creates URIs/certs only
2. GetExtensionBundle()         → downloads bundles NOW ← host.json downloadPath not yet in env var
3. BuildWebHost() → Startup     → SetBundlesEnvironmentVariables() sets env var from host.json
```

### Trace through the failure

With `host.json: { extensionBundle: { downloadPath: "C:\\my-bundles" } }`:

1. `GetExtensionBundle()` calls `GetExtensionBundleManager(options)` which **overwrites** `options.DownloadPath` with `GetBundleDownloadPath(bundleId)` ([line 52](https://github.com/Azure/azure-functions-core-tools/pull/4772/files#diff-ExtensionBundleHelper.cs-L52)).
2. `GetBundleDownloadPath` checks the env var `AzureFunctionsJobHost__extensionBundle__downloadPath` — **not set yet** — so it returns the default path `~/.azure-functions-core-tools/Functions/ExtensionBundles/{bundleId}`.
3. Bundle gets downloaded to the **default** path.
4. Later, `Startup.SetBundlesEnvironmentVariables()` reads host.json and sets the env var to `C:\my-bundles` ([line 123](https://github.com/Azure/azure-functions-core-tools/pull/4772/files#diff-Startup.cs-L123)).
5. Host starts, reads env var, looks for bundles in `C:\my-bundles` — **nothing there**.
6. `ensureLatest` is now `false` ([line 140](https://github.com/Azure/azure-functions-core-tools/pull/4772/files#diff-Startup.cs-L140)), so the host **won't download** them itself.
7. **Bundle resolution fails. `func start` is broken.**

### Why this didn't break before

Previously, even though the CLI also downloaded to the default path, `ensureLatest` was `true`. The host would re-download bundles into the correct custom path on its own. That safety net is now gone.

### Impact

This is a regression for any app using `extensionBundle.downloadPath` in host.json (without the env var). It breaks `func start` on every run, not just offline scenarios.

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
```diff
@@ -27,6 +28,13 @@ public override ICommandLineParserResult ParseArgs(string[] args)
 
         public override async Task RunAsync()
         {
+            if (GlobalCoreToolsSettings.IsOfflineMode)
```
kshyju, line 31: Should the download action recheck the connectivity (_or something else in the background periodically check_)? I am wondering if it was a transient network issue, we end up as offline for the rest of the CLI session.

I think it's reasonable for user to experience delay when they execute the "download" action.

src/Cli/func/Actions/HelpAction.cs
kshyju: Do we need docs update for this?

kshyju: Instead of saying "not supported by all commands" . Should we say, "supported only in "X" and "Y" cases). Otherwise it is hard to determine whether the feature is failing or not supported.

src/Cli/func/ExtensionBundle/ExtensionBundleConfigurationBuilder.cs
```diff
@@ -22,8 +23,8 @@ public void Configure(IConfigurationBuilder builder)
             {
                 builder.AddInMemoryCollection(new Dictionary<string, string>
                 {
-                    { "AzureFunctionsJobHost:extensionBundle:downloadPath", ExtensionBundleHelper.GetBundleDownloadPath(bundleId) },
-                    { "AzureFunctionsJobHost:extensionBundle:ensureLatest", "true" }
+                    { Constants.ExtensionBundleDownloadPath.Replace("__", ":"), ExtensionBundleHelper.GetBundleDownloadPath(bundleId) },
+                    { Constants.ExtensionBundleEnsureLatest.Replace("__", ":"), "false" }
```
kshyju: Use `bool.FalseString` instead of "false" string.  (Make sure it works with the casing of what bool.FalseString returns)

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,29 +81,110 @@ public static string GetBundleDownloadPath(string bundleId)
...
+                var extensionBundleManager = GetExtensionBundleManager(extensionBundleOptions);
 
-                // Attempt to get the extension bundle path, which will trigger the download if not already present
-                await RetryHelper.Retry(
+                var bundlePath = await RetryHelper.Retry(
                     func: async () => await extensionBundleManager.GetExtensionBundlePath(httpClient),
                     retryCount: MaxRetries,
-                    retryDelay: _retryDelay,
-                    displayError: false);
+                    retryDelay: _retryDelay);
```
kshyju, line 117: unrelated to your PR. But for 115, you can remove the await

```
func: () => extensionBundleManager.GetExtensionBundlePath(httpClient)
```

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,29 +81,110 @@ public static string GetBundleDownloadPath(string bundleId)
...
+        {
+            while (ex != null)
+            {
+                if (ex is System.Net.Sockets.SocketException)
+                {
+                    return true;
+                }
+
+                // HttpClient throws TaskCanceledException when a request times out
+                if (ex is TaskCanceledException)
```
kshyju: Should we check the base class OperationCanceledException? Can HttpClient throw OperationCanceledException in some cases?

src/Cli/func/Actions/HelpAction.cs
liliankasem: Our CLI documentation is super out of date, this is a bigger concern than just this feature that we need to track

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
liliankasem: This is from my feedback

>.GetAwaiter().GetResult() is a deadlock risk - I know this is a CLI but this is still fragile and we should avoid anti-patterns. Consider making Init async or using Task.Run(() => ...).GetAwaiter().GetResult() as a safer wrapper.

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
liliankasem: The CLI is not long running, each time you run `func` youre starting the process again - so we dont need long running tasks or anything like that

src/Cli/func/Actions/HelpAction.cs
liliankasem: Opened a work item here: https://github.com/Azure/azure-functions-core-tools/issues/4805

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
aishwaryabh: See this comment: https://github.com/Azure/azure-functions-core-tools/pull/4772#discussion_r2829356672

src/Cli/func/Actions/HelpAction.cs
liliankasem: I see how "not supported by all commands" is not helpful - I put it there for verbose because not every command has verbose logs but a lot of them do to the point that this list would get too big. 

For offline mode - give the only thing you can use it for right now is func start, I think its reasonable to say supported by X here

src/Cli/func/Actions/LocalActions/DownloadBundleAction.cs
aishwaryabh: I initially had a periodic check for rechecking the connectivity, but @liliankasem had suggested that we go through `InitAsync` which checks if we are offline at first for every command we individually run. If this was a transient issue, the user can just try again manually after as this is a CLI rather than a service

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: I don't think this is valid; I just tried this out and it works as expected for me since we do read from host.json

src/Cli/func/Actions/HelpAction.cs
aishwaryabh: done

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
kshyju: I don’t think wrapping this in Task.Run changes behavior in the CLI case. There’s no SynchronizationContext, so it’s not helping with preventing any deadlock in this case. It does add an extra thread-pool hop (minor scheduling overhead if you look at it with perf microscope). Probably not a big deal for the CLI use case though.. But the code can be simplified. Best to clarify with the AI Agent though.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,29 +81,110 @@ public static string GetBundleDownloadPath(string bundleId)
...
+        /// <summary>
+        /// Attempts to resolve a cached bundle. Logs a warning if a cached version is found,
+        /// or throws a <see cref="CliException"/> if no cached version is available.
+        /// </summary>
+        private static string GetCachedBundleOrThrow(ExtensionBundleOptions extensionBundleOptions)
+        {
+            string versionRange = extensionBundleOptions.Version?.ToString();
+            if (TryGetCachedBundle(extensionBundleOptions.Id, versionRange, out string cachedVersion, extensionBundleOptions.DownloadPath))
+            {
+                ColoredConsole.WriteLine(OutputTheme.WarningColor($"Warning: Unable to download extension bundles. Using cached version {cachedVersion}."));
```
kshyju: Should we be logging the " Unable to download extension bundles" warning here? If user is explicitly setting the --offline mode, then we don't even try to download. In that case, would this warning message be confusing.? This method (GetCachedBundleOrThrow) is not responsible for downloading.

kshyju: Same question for line 149 exception message.

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
```diff
@@ -77,29 +81,110 @@ public static string GetBundleDownloadPath(string bundleId)
             return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), Constants.UserCoreToolsDirectory, "Functions", ScriptConstants.ExtensionBundleDirectory, bundleId);
         }
 
-        public static async Task GetExtensionBundle()
+        /// <summary>
+        /// Downloads or resolves the extension bundle, with retry logic and offline fallback.
+        /// Returns the resolved bundle path, or null if no bundle is configured.
+        /// On network failure, marks the system as offline and falls back to a cached bundle.
+        /// </summary>
+        public static async Task<string> GetExtensionBundle()
```
kshyju: Should this method be public? Always consider using the least visibility needed.

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -52,6 +63,14 @@ public static WorkerRuntime CurrentWorkerRuntimeOrNone
         public static void Init(ISecretsManager secretsManager, string[] args)
         {
             _isVerbose = args.Contains("--verbose");
+            _explicitOffline = args.Contains("--offline") || EnvironmentHelper.GetEnvironmentVariableAsBool(Constants.FunctionsCoreToolsOffline);
```
kshyju, line 74: Should this be case insensitive check? (Same question of the check in the above line as well).

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    /// <summary>
+    /// Central helper for determining offline/online status across the CLI.
+    /// Respects the global --offline flag or FUNCTIONS_CORE_TOOLS_OFFLINE environment variable.
+    /// The network check happens at most once per CLI invocation via lazy initialization
+    /// in <see cref="GlobalCoreToolsSettings.IsOfflineMode"/>.
+    /// </summary>
+    internal static class OfflineHelper
+    {
+        private const string ConnectivityCheckUrl = "https://cdn.functions.azure.com/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/staticProperties.json";
```
kshyju, line 18: we own/manage this CDN endpoint. rite?  Do we need any additional settings changes on CDN side (like make sure it can handle the new extra traffic) ? Hopefully we don't need as CDN should be able to handles this, but just double checking.

liliankasem: I was also wondering about this, would it be better to base the network check on something else (i.e. microsoft.com)?

kshyju: No. A CDN endpoint we manage/control is the best option. @aishwaryabh Let's make sure the network call is including the "User-Agent" header (I think it does, but double check)

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
aishwaryabh: I think I'm going to leave this as it is for now to match existing patterns, and we can come back to this after if we feel a change is necessary

src/Cli/func/ExtensionBundle/ExtensionBundleHelper.cs
aishwaryabh: added special logging for the case user explicitly set offline status

src/Cli/func/Helpers/OfflineHelper.cs
aishwaryabh: i added the user-agent header

src/Cli/func/Helpers/OfflineHelper.cs
```diff
@@ -28,6 +29,7 @@ internal static async Task<bool> IsOfflineAsync()
             {
                 using var quickClient = new HttpClient { Timeout = TimeSpan.FromSeconds(1) };
                 using var request = new HttpRequestMessage(HttpMethod.Head, ConnectivityCheckUrl);
+                request.Headers.Add("User-Agent", Constants.CliUserAgent);
```
kshyju, line 32: This works. I think a potential optimization we can make in futue is to create a custom DelegatingHandler to add this header so that we don't need this line in multiple places (I noticed we have it in many other files)

liliankasem (approved):

PR #4778, opened 2026-01-21
Update linux E2E test job image to min version

liliankasem:
Our linux E2E tests do not run as the current image does not have enough space. Switching to a min version and installing required tools for our E2E test runs

### DIFF
--- a/eng/ci/templates/jobs/test-e2e-linux.yml
+++ b/eng/ci/templates/jobs/test-e2e-linux.yml
@@ -5,7 +5,7 @@ jobs:
 
   pool:
     name: 1es-pool-azfunc-public
-    image: 1es-ubuntu-22.04
+    image: 1es-ubuntu-24.04-min
     os: linux
 
   strategy:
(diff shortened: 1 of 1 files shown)


eng/ci/templates/jobs/test-e2e-linux.yml
```diff
@@ -27,12 +27,17 @@ jobs:
...
   steps:
-  - pwsh: ./eng/scripts/start-emulators.ps1
-    displayName: 'Start emulators (NoWait)'
-
   - template: /eng/ci/templates/steps/install-tools.yml@self
 
-  - template: /eng/ci/templates/steps/restore-nuget.yml@self
+  - bash: |
+      sudo apt-get update
+      sudo apt-get install -y npm golang
```
jviau, line 34: There are azure pipeline tasks for this which may be more efficient to use:

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/npm-v1?view=azure-pipelines
https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/go-tool-v0?view=azure-pipelines

liliankasem: done!

liliankasem: Using go, but the npm task is for running npm commands, not installing npm itself

liliankasem: OK both of those did not work, so I am going back to the bash script which I know defo worked.

PR #4813, opened 2026-02-25
[in-proc] Update CI linux images to min and install required tools

liliankasem:
Update CI linux images to min and install required tools

### DIFF
--- a/eng/ci/templates/official/jobs/linux-package.yml
+++ b/eng/ci/templates/official/jobs/linux-package.yml
@@ -15,7 +15,7 @@ jobs:
       # Install .NET 8.0 SDK to /usr/share/dotnet so all SDKs are in one location
       curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
       chmod +x dotnet-install.sh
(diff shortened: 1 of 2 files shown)


eng/ci/templates/official/jobs/linux-package.yml
```diff
@@ -4,9 +4,15 @@ jobs:
   timeoutInMinutes: "120"
   pool:
     name: 1es-pool-azfunc
-    image: 1es-ubuntu-22.04
+    image: 1es-ubuntu-24.04-min
     os: linux
   steps:
+  - bash: |
```
jviau, line 10: Would prefer we use the appropriate pipeline tasks for installing these.

liliankasem: iirc from updating main, the tasks didn't work: https://github.com/Azure/azure-functions-core-tools/pull/4778#discussion_r2714472309

Keeping as bash for now so we can unblock this release and can always open a work item to figure this out if pressing.

Francisco-Gamino (approved):

PR #4817, opened 2026-02-26
[in-proc] Update CI linux images to min and install required tools

liliankasem:
[in-proc] Update CI linux images to min and install required tools

### DIFF
--- a/eng/ci/templates/steps/install-linux-tools.yml
+++ b/eng/ci/templates/steps/install-linux-tools.yml
@@ -11,5 +11,5 @@ steps:
     sudo ./dotnet-install.sh --install-dir /usr/share/dotnet --channel 8.0
     sudo chmod -R a+rx /usr/share/dotnet
     echo "##vso[task.prependpath]/usr/share/dotnet"
-    dotnet --list-sdks
+    /usr/share/dotnet/dotnet --version
   displayName: 'Install npm, Go, and .NET'


eng/ci/templates/public/jobs/build-test-public.yml
```diff
@@ -74,9 +74,22 @@ jobs:
...
+  - bash: |
+      sudo apt-get update
+      sudo apt-get install -y npm golang
+      npm --version
+      go version
+      # Install .NET 8.0 SDK to /usr/share/dotnet so all SDKs are in one location
+      curl -sSL https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
+      chmod +x dotnet-install.sh
+      sudo ./dotnet-install.sh --install-dir /usr/share/dotnet --channel 8.0
+      sudo chmod -R a+rwx /usr/share/dotnet
```
fabiocav: What's the reason for the write permission here?

eng/ci/templates/public/jobs/build-test-public.yml
```diff
@@ -74,9 +74,22 @@ jobs:
   timeoutInMinutes: "180"
   pool:
     name: 1es-pool-azfunc-public
-    image: '1es-ubuntu-22.04'
+    image: 1es-ubuntu-24.04-min
     os: 'linux'
   steps:
+  - bash: |
+      sudo apt-get update
```
fabiocav: The steps here are identical to the ones above. Can we reuse this to avoid having to maintain two sets?

eng/ci/templates/public/jobs/build-test-public.yml
```diff
@@ -74,9 +74,22 @@ jobs:
...
+      sudo apt-get install -y npm golang
```
fabiocav: Can we log (or pin) the GO version installed?

aishwaryabh: this is already logged!

eng/ci/templates/public/jobs/build-test-public.yml
aishwaryabh: done

eng/ci/templates/public/jobs/build-test-public.yml
aishwaryabh: let me try removing the write permission to see if it works still but I think it should

PR #4854, opened 2026-04-14
Add Go (Golang) language support to func init via the Native worker runtime

harshivcodes:
This PR adds Go language support to the `func init` command in Azure Functions Core Tools. Go is supported under the Native worker runtime, which hosts language-native binaries directly. When a user runs `func init --worker-runtime native`, the CLI scaffolds a complete Go project with module initialization, dependency resolution, and a starter `main.go`.

Core runtime registration
- Added WorkerRuntime.Native enum value with "go" and "golang" as aliases
- --worker-runtime native, --worker-runtime go, and --worker-runtime golang all resolve to Native
- When go or golang is used, a warning is printed clarifying that Golang runs under the native runtime
- Added Constants.Languages.Golang and wired up default language/programming model mappings
- Native appears in the interactive runtime selection picker

Go project scaffolding (GoHelpers.cs)
- Detects installed Go version (minimum 1.24 required)
- Runs go mod init, go get github.com/azure/azure-functions-golang-worker, and go mod tidy
- --skip-go-mod-tidy flag to skip the tidy step (useful for CI or when the worker module is unreachable)

Static resources
- Embedded main.go template (minimal package main + empty main())
- Reuses existing .funcignore template
- New entry in StaticResources.props and StaticResources.cs

InitAction integration
- InitLanguageSpecificArtifacts handles WorkerRuntime.Native: calls GoHelpers.SetupProject, writes main.go and .funcignore
- Docker flag emits a warning: "Docker support for the native worker runtime is not yet available"
- Language selection wizard surfaces for Native runtime

Testing
- 27 unit tests passing (GoHelperTests + InitActionGoTests)
- 4 E2E test cases defined (NativeGoInitTests)
- Manual validation of 8 scenarios: --worker-runtime native/go/golang, --skip-go-mod-tidy, named subfolder, --force in non-empty dir, --docker warning, and interactive picker

| Test | Scenario                          | Result |
|------|-----------------------------------|--------|
| 1    | `--worker-runtime native`         | All files created; `FUNCTIONS_WORKER_RUNTIME: "native"` set; `go.mod` + `go.sum` present |
| 2    | `--worker-runtime go`             | Same output — alias works correctly |
| 3    | `--worker-runtime golang`         | Same output — alias works correctly |
| 4    | `--skip-go-mod-tidy`              | Shows “Skipped” message; `go.sum` still exists (from `go get`) |
| 5    | Named subfolder `test-go-5`        | Created subfolder and initialized inside it |
| 6    | `--force` in non-empty directory  | `dummy.txt` preserved alongside all new files |
| 7    | `--docker`                        | Docker warning emitted: “Docker support for the native worker runtime is not yet available. Skipping Dockerfile creation.” No Dockerfile created |
| 8    | Interactive (no runtime flag)     | Picker shows **Native** as option 7 in the list |

Output from local testing
[test-results.txt](https://github.com/user-attachments/files/26729550/test-results.txt)

### DIFF
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -141,7 +141,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 .Callback(tf => TargetFramework = tf);
 
             Parser
-                .Setup<string>("language")
+                .Setup<string>('l', "language")
                 .SetDefault(null)
                 .Callback(l => Language = l);
 
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -215,10 +215,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 (ResolvedWorkerRuntime, ResolvedLanguage) = ResolveWorkerRuntimeAndLanguage(WorkerRuntime, Language);
 
-                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
-                {
-                    ColoredConsole.WriteLine(WarningColor("[preview] The native worker runtime is currently in preview. Features and behavior may change in future releases."));
-                }
+                Utilities.WarnIfNativeWorkerRuntime(ResolvedWorkerRuntime);
 
                 // Order here is important: each language may have multiple runtimes, and each unique (language, worker-runtime) pair
                 // may have its own programming model. Thus, we assume that ResolvedLanguage and ResolvedWorkerRuntime are properly set
@@ -249,7 +246,18 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             else
             {
                 bool managedDependenciesOption = ResolveManagedDependencies(ResolvedWorkerRuntime, ManagedDependencies);
-                await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation, SkipGoModTidy);
+
+                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
+                {
+                    await GoHelpers.SetupProject(
+                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_"),
+                        SkipGoModTidy);
+                }
+                else
+                {
+                    await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation);
+                }
+
                 await WriteFiles();
 
                 await WriteHostJson(ResolvedWorkerRuntime, managedDependenciesOption, ExtensionBundle);
@@ -295,7 +303,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 }
                 else
                 {
-                    language = WorkerRuntimeLanguageHelper.GetDefaultTemplateLanguageFromWorker(workerRuntime);
+                    language = LanguageSelectionIfRelevant(workerRuntime);
+                    if (string.IsNullOrEmpty(language))
+                    {
+                        language = WorkerRuntimeLanguageHelper.GetDefaultTemplateLanguageFromWorker(workerRuntime);
+                    }
                 }
             }
             else if (GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone == Helpers.WorkerRuntime.None)
@@ -345,8 +357,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             string language,
             ProgrammingModel programmingModel,
             bool managedDependenciesOption,
-            bool generatePythonDocumentation = true,
-            bool skipGoModTidy = false)
+            bool generatePythonDocumentation = true)
         {
             switch (workerRuntime)
             {
@@ -390,13 +401,6 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                     break;
                 case Helpers.WorkerRuntime.Node:
                     await NodeJSHelpers.SetupProject(programmingModel, language);
-                    break;
-                case Helpers.WorkerRuntime.Native:
-                    await GoHelpers.SetupProject(
-                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_."),
-                        skipGoModTidy);
-                    await FileSystemHelpers.WriteFileIfNotExists("main.go", await StaticResources.MainGo);
-                    await FileSystemHelpers.WriteFileIfNotExists(Constants.FuncIgnoreFile, await StaticResources.FuncIgnore);
                     break;
             }
         }
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -449,7 +449,14 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         private static async Task WriteLocalSettingsJson(WorkerRuntime workerRuntime, ProgrammingModel programmingModel)
         {
             string localSettingsJsonContent = await StaticResources.LocalSettingsJson;
-            localSettingsJsonContent = localSettingsJsonContent.Replace($"{{{Constants.FunctionsWorkerRuntime}}}", WorkerRuntimeLanguageHelper.GetRuntimeMoniker(workerRuntime));
+
+            // The Go worker is registered with the host under the "native" runtime identifier,
+            // so FUNCTIONS_WORKER_RUNTIME in local.settings.json must be "native" even though
+            // the CLI exposes the runtime as "go" everywhere else.
+            string workerRuntimeSetting = workerRuntime == Helpers.WorkerRuntime.Go
+                ? "native"
+                : WorkerRuntimeLanguageHelper.GetRuntimeMoniker(workerRuntime);
+            localSettingsJsonContent = localSettingsJsonContent.Replace($"{{{Constants.FunctionsWorkerRuntime}}}", workerRuntimeSetting);
             localSettingsJsonContent = localSettingsJsonContent.Replace($"{{{Constants.AzureWebJobsStorage}}}", Constants.StorageEmulatorConnectionString);
 
             if (workerRuntime == Helpers.WorkerRuntime.Powershell)
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -215,7 +215,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 (ResolvedWorkerRuntime, ResolvedLanguage) = ResolveWorkerRuntimeAndLanguage(WorkerRuntime, Language);
 
-                Utilities.WarnIfNativeWorkerRuntime(ResolvedWorkerRuntime);
+                Utilities.WarnIfGoWorkerRuntime(ResolvedWorkerRuntime);
 
                 // Order here is important: each language may have multiple runtimes, and each unique (language, worker-runtime) pair
                 // may have its own programming model. Thus, we assume that ResolvedLanguage and ResolvedWorkerRuntime are properly set
@@ -247,7 +247,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 bool managedDependenciesOption = ResolveManagedDependencies(ResolvedWorkerRuntime, ManagedDependencies);
 
-                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
+                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Go)
                 {
                     await GoHelpers.SetupProject(
                         Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_"),
@@ -293,6 +293,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             if (!string.IsNullOrEmpty(workerRuntimeString))
             {
                 workerRuntime = WorkerRuntimeLanguageHelper.NormalizeWorkerRuntime(workerRuntimeString);
+                if (workerRuntime == Helpers.WorkerRuntime.Go)
+                {
+                    return (workerRuntime, string.Empty);
+                }
+
                 if (!string.IsNullOrEmpty(languageString))
                 {
                     language = WorkerRuntimeLanguageHelper.NormalizeLanguage(languageString);
@@ -324,6 +329,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             else
             {
                 workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntime;
+                if (workerRuntime == Helpers.WorkerRuntime.Go)
+                {
+                    return (workerRuntime, string.Empty);
+                }
+
                 language = GlobalCoreToolsSettings.CurrentLanguageOrNull
                     ?? (!string.IsNullOrEmpty(languageString) ? WorkerRuntimeLanguageHelper.NormalizeLanguage(languageString) : null)
                     ?? WorkerRuntimeLanguageHelper.NormalizeLanguage(WorkerRuntimeLanguageHelper.GetRuntimeMoniker(workerRuntime));
@@ -336,8 +346,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
         {
             if (workerRuntime == Helpers.WorkerRuntime.Node
                 || workerRuntime == Helpers.WorkerRuntime.DotnetIsolated
-                || workerRuntime == Helpers.WorkerRuntime.Dotnet
-                || workerRuntime == Helpers.WorkerRuntime.Native)
+                || workerRuntime == Helpers.WorkerRuntime.Dotnet)
             {
                 if (WorkerRuntimeLanguageHelper.WorkerToSupportedLanguages.TryGetValue(workerRuntime, out IEnumerable<string> languages)
                     && languages.Count() != 0)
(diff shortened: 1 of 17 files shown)


src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -276,7 +282,7 @@ private static (WorkerRuntime WorkerRuntime, string WorkerLanguage) ResolveWorke
                 workerRuntime = WorkerRuntimeLanguageHelper.NormalizeWorkerRuntime(workerRuntimeString);
                 language = !string.IsNullOrEmpty(languageString)
                     ? WorkerRuntimeLanguageHelper.NormalizeLanguage(languageString)
-                    : WorkerRuntimeLanguageHelper.NormalizeLanguage(workerRuntimeString);
+                    : WorkerRuntimeLanguageHelper.GetDefaultTemplateLanguageFromWorker(workerRuntime);
```
harshivcodes: this may have side effects for other languages, re-evaluate this chane

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -145,7 +153,16 @@ public static WorkerRuntime NormalizeWorkerRuntime(string workerRuntime)
...
             else if (_normalizeMap.ContainsKey(workerRuntime))
             {
-                return _normalizeMap[workerRuntime];
+                var resolved = _normalizeMap[workerRuntime];
+
+                if (resolved == WorkerRuntime.Native
+                    && !string.Equals(workerRuntime, "native", StringComparison.OrdinalIgnoreCase))
+                {
+                    ColoredConsole.WriteLine(WarningColor(
+                        $"'{workerRuntime}' is not a worker runtime. Golang is supported under the 'native' worker runtime. Resolving to --worker-runtime native."));
```
liliankasem: This is technically correct but may confuse users who intentionally typed `go`. Consider rewording to something like:

> *"'go' maps to the 'native' worker runtime. Resolving to --worker-runtime native."*

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -369,6 +391,13 @@ private static async Task InitLanguageSpecificArtifacts(
                 case Helpers.WorkerRuntime.Node:
                     await NodeJSHelpers.SetupProject(programmingModel, language);
                     break;
+                case Helpers.WorkerRuntime.Native:
+                    await GoHelpers.SetupProject(
+                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_."),
```
liliankasem: Go module names follow different conventions than typical sanitization (they're often URL-like paths, e.g., `github.com/org/repo`). Using the current directory name is reasonable for a default, but the restrictive character set could produce unexpected module names for directories with special characters. Worth verifying edge cases.

src/Cli/func/StaticResources/main.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+func main() {
+}
```
liliankasem: nit: add a new line to end of file, most Go tooling (gofmt, editors, linters) expects a trailing newline

```suggestion
func main() { }

```

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -0,0 +1,130 @@
...
+                    var match = Regex.Match(output, @"go(\d+\.\d+(?:\.\d+)?)");
+                    if (match.Success)
+                    {
+                        return new WorkerLanguageVersionInfo(WorkerRuntime.Native, match.Groups[1].Value, goExe);
+                    }
+                }
+            }
+            catch (Exception)
+            {
+                // Go is not installed or not on PATH
```
liliankasem: Should we log to the user about this? Trying to understand why we would want to swallow an exception here?

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -0,0 +1,130 @@
...
+        public static async Task SetupProject(string moduleName, bool skipGoModTidy)
+        {
+            var goVersion = await GetEnvironmentGoVersion();
+            AssertGoVersion(goVersion);
+
+            // Initialize: Run go mod init
+            var modInitExe = new Executable("go", $"mod init {moduleName}");
+            var modInitExitCode = await modInitExe.RunAsync(
+                l => ColoredConsole.WriteLine(l),
+                e => ColoredConsole.Error.WriteLine(ErrorColor(e)));
```
liliankasem: When I ran "func init --worker-runtime native" I noticed go text is red

```bash
go: creating new go.mod: module go-app

go: downloading github.com/azure/azure-functions-golang-worker v0.0.0-20260414214309-93dd9b8d6637

go: added github.com/azure/azure-functions-golang-worker v0.0.0-20260414214309-93dd9b8d6637

go: warning: "all" matched no packages
```

Given that we are setting the error color here, it appears that the Go stderr output (even informational) renders red. So I am guesing that `go get` and `go mod tidy` write progress to stderr instead of stdout, even though there isnt an error. The stderr handler should use plain output or AdditionalInfoColor for non-failure cases.

Second, we show suppress this information unless the global --verbose flag is passed e.g.

```csharp
 var modInitExitCode = modInitExe exe.RunAsync(
     l =>
     {
         {
             ColoredConsole.WriteLine(VerboseColor(l));
         }
     },
     e =>
     {
         {
             ColoredConsole.WriteLine(VerboseColor(e));
         }
     });
```

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -209,6 +215,11 @@ private async Task InitFunctionAppProject()
             {
                 (ResolvedWorkerRuntime, ResolvedLanguage) = ResolveWorkerRuntimeAndLanguage(WorkerRuntime, Language);
 
+                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
+                {
+                    ColoredConsole.WriteLine(WarningColor("[preview] The native worker runtime is currently in preview. Features and behavior may change in future releases."));
+                }
```
liliankasem: This shouldn't live in the init action, it should be written any time the worker runtime is `native` globally.

The existing pattern is `Utilities.WarnIfPreviewVersion()` in Utilities.cs, it's called from InitAction, StartHostAction, and HelpAction. That's the right place to add something like:

```
     if (GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone == WorkerRuntime.Native)
     {
         ColoredConsole.WriteLine(WarningColor(
             "The native worker runtime is currently in preview. ..."));
     }
```

(I would drop the [preview] pattern)

You can then call `Utilities.WarnIfPreviewVersion()` in new places if needed (i.e. pack action etc. later on)

src/Cli/func/Helpers/GoHelpers.cs
liliankasem: Might want this to be a helper method since it needs to apply to all commands (go mod init, go get, go mod tidy)

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -71,6 +71,8 @@ public InitAction(ITemplatesManager templatesManager, ISecretsManager secretsMan
```
liliankasem, line 71: Running `func init --worker-runtime native` should prompt for language. Right now it just assumes `go` by default.

`func init --worker-runtime native -l go` should be the path to full command run without prompts

src/Cli/func/StaticResources/main.go
```diff
@@ -0,0 +1,4 @@
+package main
```
liliankasem, line 1: Does this need to be created by init? I would expect this to be added by `func new` as an actual http template

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -369,6 +391,13 @@ private static async Task InitLanguageSpecificArtifacts(
                 case Helpers.WorkerRuntime.Node:
                     await NodeJSHelpers.SetupProject(programmingModel, language);
                     break;
+                case Helpers.WorkerRuntime.Native:
+                    await GoHelpers.SetupProject(
+                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_."),
+                        skipGoModTidy);
+                    await FileSystemHelpers.WriteFileIfNotExists("main.go", await StaticResources.MainGo);
+                    await FileSystemHelpers.WriteFileIfNotExists(Constants.FuncIgnoreFile, await StaticResources.FuncIgnore);
```
liliankasem: I would expect all these operations to be part of `GoHelpers.SetupProject`

src/Cli/func/Actions/LocalActions/InitAction/InitNativeSubcommandAction.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Azure.Functions.Cli.Actions.LocalActions
+{
+    [Action(Name = "init native", ParentCommandName = "init", ShowInHelp = true, HelpText = "Options specific to native runtime apps when running func init")]
+    internal class InitNativeSubcommandAction : BaseAction
+    {
+        public override ICommandLineParserResult ParseArgs(string[] args)
+        {
+            Parser
+                .Setup<string>('l', "language")
+                .WithDescription("The language for the function app. Options: golang.")
```
liliankasem: nit: I know not all of the other commands are doing this, but this should ideally not be hardcoded. Can do similar to how we get the tfm for dotnet, but for supported native languages

```csharp
            Parser
                .Setup<string>("target-framework")
                .WithDescription($"The target framework for the .NET project. Options: {string.Join(", ", TargetFrameworkHelper.GetSupportedTargetFrameworks())}.")
                .Callback(_ => { });
```

src/Cli/func/Actions/LocalActions/InitAction/InitNativeSubcommandAction.cs
```diff
@@ -0,0 +1,32 @@
...
+                .Callback(_ => { });
+
+            Parser
+                .Setup<bool>("skip-go-mod-tidy")
+                .WithDescription("Skip running 'go mod tidy' after project creation.")
```
liliankasem: nit 
```suggestion
                .WithDescription("Skip running 'go mod tidy' after Go project creation.")
```

src/Cli/func/Common/Constants.cs
```diff
@@ -113,7 +113,8 @@ internal static partial class Constants
             { WorkerRuntime.Dotnet, new[] { "mcr.microsoft.com/azure-functions/dotnet", "microsoft/azure-functions-dotnet-core2.0", "mcr.microsoft.com/azure-functions/base", "microsoft/azure-functions-base" } },
             { WorkerRuntime.Node, new[] { "mcr.microsoft.com/azure-functions/node", "microsoft/azure-functions-node8" } },
             { WorkerRuntime.Python, new[] { "mcr.microsoft.com/azure-functions/python", "microsoft/azure-functions-python3.6" } },
-            { WorkerRuntime.Powershell, new[] { "mcr.microsoft.com/azure-functions/powershell", "microsoft/azure-functions-powershell" } }
+            { WorkerRuntime.Powershell, new[] { "mcr.microsoft.com/azure-functions/powershell", "microsoft/azure-functions-powershell" } },
+            { WorkerRuntime.Native, new string[] { } }
```
liliankasem: Do we need to add this here? Were you running into any issues without it?

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                ColoredConsole.WriteLine(AdditionalInfoColor("Skipped \"go mod tidy\". You must run \"go mod tidy\" manually."));
+            }
+        }
+
+        private static async Task<WorkerLanguageVersionInfo> GetVersion(string goExe)
+        {
+            try
+            {
+                var exe = new Executable(goExe, "version");
```
liliankasem, line 119: This pattern repeats itself multiple times. A helper like RunGoCommandAsync would clean this up:

```csharp
private static async Task RunGoCommandAsync(string arguments, string errorMessage, bool 
throwOnFailure = true)
 {
     var exe = new Executable("go", arguments);
     var exitCode = await exe.RunAsync(
         l =>
         {
             {
                 ColoredConsole.WriteLine(VerboseColor(l));
             }
         },
         e =>
         {
             {
                 ColoredConsole.WriteLine(VerboseColor(e));
             }
         });
 
     if (exitCode != 0)
     {
         if (throwOnFailure)
         {
             throw new CliException($"{errorMessage} 'go {arguments}' exited with code 
{exitCode}.");
         }
 
         ColoredConsole.WriteLine(WarningColor($"Warning: 'go {arguments}' exited with a 
non-zero code. You may need to run it manually."));
     }
 }
```

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -238,7 +249,7 @@ private async Task InitFunctionAppProject()
             else
             {
                 bool managedDependenciesOption = ResolveManagedDependencies(ResolvedWorkerRuntime, ManagedDependencies);
-                await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation);
+                await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation, SkipGoModTidy);
```
liliankasem: This is a bit of a code smell, and it started with `generatePythonDocumentation` to be frank.

Every runtime-specific flag gets bolted onto this shared method. generatePythonDocumentation was already a smell, and now skipGoModTidy makes it worse. If Rust or another native language lands later, it'll grow again.

I mention this in another comment, but the solution here is to move the file writes into GoHelpers.SetupProject (matching Node's pattern), then the Native case in the switch becomes just:

```csharp
   case Helpers.WorkerRuntime.Native:
       await GoHelpers.SetupProject(
           Utilities.SanitizeLiteral(...), skipGoModTidy);
       break;
```

And `skipGoModTidy` doesn't need to be a parameter of InitLanguageSpecificArtifacts at all, it can stay on the InitAction class and be passed directly. 

I.e. 

 // In InitFunctionAppProject(), before the shared method:
 if (ResolvedWorkerRuntime == WorkerRuntime.Native)
 {
     await GoHelpers.SetupProject(moduleName, SkipGoModTidy);
 }
 else
 {
     await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ...);
 }

or use directly via 

this.SkipGoModTidy

test/Cli/Func.E2ETests/Commands/FuncInit/NativeGoInitTests.cs
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: Missing test cases:

   1. No E2E test for --worker-runtime go alias — there's a unit test for 
  NormalizeWorkerRuntime("go") resolving to Native, but no E2E test verifying func init 
  --worker-runtime go actually works end-to-end and shows the redirect warning.
   2. No E2E test for init in non-empty directory without --force — there's a test with 
  --force, but not one asserting it fails/warns without it (other runtimes test this).
   3. No test for host.json content — Node tests verify host.json contains expected logging 
  config. The Go E2E checks host.json has extensionBundle but doesn't verify logging config.
   4. No test for local.settings.json having FUNCTIONS_WORKER_RUNTIME set to native — it's 
  validated in filesToValidate but only checks the strings exist, not the actual JSON 
  structure. This matches the existing pattern though, so not a regression.
   5. No unit test for GetVersion parsing edge cases — e.g., what happens with unexpected go 
  version output formats, or Go beta versions like go1.25beta1.

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: Here's the flow:
1. GetVersion catches the exception → returns null
2. SetupProject calls AssertGoVersion(null)
3. User sees: "Could not find a Go installation. Go 1.24 or later is required. Please install Go from https://go.dev/dl/"

So the user-facing message is good and actionable. The swallowed exception just hides why it failed (e.g., permission error vs not on PATH). I can log this exception if they are running in verbose mode.

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
harshivcodes: Good point, what appears in the go.mod file could turn into the executable name (unless the customer overrides it). I removed `.`  since that's the only potentially problematic character in the executable name. I think the others are ok.

src/Cli/func/StaticResources/main.go
harshivcodes: Is this behavior I can change later or are we locked into it? If we can change it, I am inclined to leave it here since it's the equivalent of Python's function_app.py — a mandatory entry point, not a function template.

We are likely not going to add func new support for golang during preview - it might come slightly later.

liliankasem: Yeah that's fine with me

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: Done

src/Cli/func/Common/Constants.cs
harshivcodes: removed it.

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: done.

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
harshivcodes: Made this change.

src/Cli/func/Actions/LocalActions/InitAction/InitNativeSubcommandAction.cs
harshivcodes: Updated.

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
harshivcodes: 1. Pulled Native out of the shared InitLanguageSpecificArtifacts method into its own block in InitFunctionAppProject
 2. Removed skipGoModTidy from the InitLanguageSpecificArtifacts signature
 3. GoHelpers.SetupProject is called directly with the flag

test/Cli/Func.E2ETests/Commands/FuncInit/NativeGoInitTests.cs
harshivcodes: Added these tests.

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
harshivcodes: fixed.

liliankasem (approved):

PR #4859, opened 2026-04-20
Add telemetry with OpenTelemetry and Application Insights

liliankasem:
Adds telemetry infrastructure wired into Program.cs

### DIFF
--- a/src/Func.Cli/Program.cs
+++ b/src/Func.Cli/Program.cs
@@ -11,8 +11,8 @@ using Azure.Functions.Cli.Telemetry;
 using Azure.Monitor.OpenTelemetry.Exporter;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
-using OpenTelemetry;
 using OpenTelemetry.Metrics;
+using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 
 // Build the host with DI
@@ -25,30 +25,27 @@ var builder = Host.CreateEmptyApplicationBuilder(new HostApplicationBuilderSetti
 builder.Services.AddSingleton<ITheme, DefaultTheme>();
 builder.Services.AddSingleton<IInteractionService, SpectreInteractionService>();
 
-var host = builder.Build();
-var interaction = host.Services.GetRequiredService<IInteractionService>();
-
 // Wire up the OpenTelemetry SDK only when telemetry is configured (build has
 // a real instrumentation key and the user has not opted out). When it isn't,
 // no listener is subscribed and ActivitySource/Meter calls become no-ops.
-TracerProvider? tracerProvider = null;
-MeterProvider? meterProvider = null;
+// The host owns the providers and flushes/disposes them on shutdown.
 if (CliTelemetry.TryGetConnectionString(out var connectionString))
 {
-    var resourceBuilder = CliTelemetry.CreateResourceBuilder();
-
-    tracerProvider = Sdk.CreateTracerProviderBuilder()
-        .SetResourceBuilder(resourceBuilder)
-        .AddSource(CliTelemetry.SourceName)
-        .AddAzureMonitorTraceExporter(o => o.ConnectionString = connectionString)
-        .Build();
-
-    meterProvider = Sdk.CreateMeterProviderBuilder()
-        .SetResourceBuilder(resourceBuilder)
-        .AddMeter(CliTelemetry.SourceName)
-        .AddAzureMonitorMetricExporter(o => o.ConnectionString = connectionString)
-        .Build();
+    builder.Services
+        .AddOpenTelemetry()
+        .ConfigureResource(r => r
+            .AddService(serviceName: CliTelemetry.SourceName, serviceVersion: CliTelemetry.CliVersion)
+            .AddAttributes(CliTelemetry.GetResourceAttributes()))
+        .WithTracing(t => t
+            .AddSource(CliTelemetry.SourceName)
+            .AddAzureMonitorTraceExporter(o => o.ConnectionString = connectionString))
+        .WithMetrics(m => m
+            .AddMeter(CliTelemetry.SourceName)
+            .AddAzureMonitorMetricExporter(o => o.ConnectionString = connectionString));
 }
+
+var host = builder.Build();
+var interaction = host.Services.GetRequiredService<IInteractionService>();
 
 // Create the command tree
 var rootCommand = Parser.CreateCommand(interaction);
@@ -80,7 +77,7 @@ var commandName = ResolveCommandName(args);
 var stopwatch = Stopwatch.StartNew();
 int exitCode = 0;
 
-using (Activity? activity = ActivitySource.StartCommandActivity(commandName))
+using (Activity? activity = CliTelemetry.Trace.StartCommandActivity(commandName))
 {
     try
     {
@@ -114,7 +111,7 @@ using (Activity? activity = ActivitySource.StartCommandActivity(commandName))
     finally
     {
         stopwatch.Stop();
-        CliTelemetry.Meter.RecordCommand(commandName, exitCode, stopwatch.ElapsedMilliseconds);
+        CliTelemetry.Metric.RecordCommand(commandName, exitCode, stopwatch.ElapsedMilliseconds);
     }
 } // activity disposed (and stopped) here, before flush
 
@@ -125,10 +122,9 @@ try
 }
 finally
 {
-    tracerProvider?.ForceFlush(3000);
-    meterProvider?.ForceFlush(3000);
-    tracerProvider?.Dispose();
-    meterProvider?.Dispose();
+    // Disposing the host stops hosted services and disposes everything in the
+    // DI container — including the OTel TracerProvider/MeterProvider, which
+    // flush pending telemetry on dispose.
     host.Dispose();
 }
 
--- a/src/Func.Cli/Program.cs
+++ b/src/Func.Cli/Program.cs
@@ -33,15 +33,18 @@ var interaction = host.Services.GetRequiredService<IInteractionService>();
 // no listener is subscribed and ActivitySource/Meter calls become no-ops.
 TracerProvider? tracerProvider = null;
 MeterProvider? meterProvider = null;
-var connectionString = CliTelemetry.GetConnectionString();
-if (connectionString is not null)
+if (CliTelemetry.TryGetConnectionString(out var connectionString))
 {
+    var resourceBuilder = CliTelemetry.CreateResourceBuilder();
+
     tracerProvider = Sdk.CreateTracerProviderBuilder()
+        .SetResourceBuilder(resourceBuilder)
         .AddSource(CliTelemetry.SourceName)
         .AddAzureMonitorTraceExporter(o => o.ConnectionString = connectionString)
         .Build();
 
     meterProvider = Sdk.CreateMeterProviderBuilder()
+        .SetResourceBuilder(resourceBuilder)
         .AddMeter(CliTelemetry.SourceName)
         .AddAzureMonitorMetricExporter(o => o.ConnectionString = connectionString)
         .Build();
@@ -75,31 +78,23 @@ var versionCheckTask = VersionChecker.CheckForUpdateAsync(cts.Token);
 // Parse and invoke asynchronously
 var commandName = ResolveCommandName(args);
 var stopwatch = Stopwatch.StartNew();
-int exitCode;
+int exitCode = 0;
 
-using (Activity? activity = CliTelemetry.Source.StartCommandActivity(commandName))
+using (Activity? activity = ActivitySource.StartCommandActivity(commandName))
 {
     try
     {
         var config = new InvocationConfiguration { EnableDefaultExceptionHandler = false };
         exitCode = await rootCommand.Parse(args).InvokeAsync(config, cts.Token);
-
-        stopwatch.Stop();
-        activity?.SetCommandResult(exitCode == 0);
-        CliTelemetry.TrackCommand(commandName, isSuccess: exitCode == 0, durationMs: stopwatch.ElapsedMilliseconds);
     }
     catch (OperationCanceledException)
     {
-        activity?.SetCommandResult(isSuccess: false, "Cancelled");
-        exitCode = 130; // Standard Unix exit code for SIGINT
+        // SIGINT — not a failure to record on the activity.
+        exitCode = 130;
     }
     catch (GracefulException ex)
     {
-        stopwatch.Stop();
-        activity?.SetCommandResult(isSuccess: false, ex.Message);
-        activity?.AddException(ex);
-        CliTelemetry.TrackCommand(commandName, isSuccess: false, durationMs: stopwatch.ElapsedMilliseconds);
-
+        activity?.Fail(ex);
         interaction.WriteError(ex.Message);
 
         if (ex.VerboseMessage is not null)
@@ -107,17 +102,19 @@ using (Activity? activity = CliTelemetry.Source.StartCommandActivity(commandName
             interaction.WriteHint(ex.VerboseMessage);
         }
 
-        exitCode = ex.IsUserError ? 1 : 2;
+        // Standard Unix convention: 1 = general error, 2 = misuse.
+        exitCode = ex.IsUserError ? 2 : 1;
     }
     catch (Exception ex)
     {
-        stopwatch.Stop();
-        activity?.SetCommandResult(isSuccess: false, ex.Message);
-        activity?.AddException(ex);
-        CliTelemetry.TrackCommand(commandName, isSuccess: false, durationMs: stopwatch.ElapsedMilliseconds);
-
+        activity?.Fail(ex);
         interaction.WriteError($"An unexpected error occurred: {ex.Message}");
-        exitCode = 2;
+        exitCode = 1;
+    }
+    finally
+    {
+        stopwatch.Stop();
+        CliTelemetry.Meter.RecordCommand(commandName, exitCode, stopwatch.ElapsedMilliseconds);
     }
 } // activity disposed (and stopped) here, before flush
 
--- a/src/Func.Cli/Program.cs
+++ b/src/Func.Cli/Program.cs
@@ -2,12 +2,18 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.CommandLine;
+using System.Diagnostics;
 using Azure.Functions.Cli;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Console;
 using Azure.Functions.Cli.Console.Theme;
+using Azure.Functions.Cli.Telemetry;
+using Azure.Monitor.OpenTelemetry.Exporter;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using OpenTelemetry;
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Trace;
 
 // Build the host with DI
 var builder = Host.CreateEmptyApplicationBuilder(new HostApplicationBuilderSettings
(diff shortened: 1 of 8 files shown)


src/Func.Cli/Telemetry/TelemetryClient.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    public void Dispose()
+    {
+        _tracerProvider?.Dispose();
+    }
+
+    private static bool CheckEnabled()
+    {
+        var key = Constants.TelemetryInstrumentationKey;
+        if (string.IsNullOrEmpty(key) || key == "00000000-0000-0000-0000-000000000000")
```
satvu: nit: can we make this string value a constant (in this file or elsewhere)

src/Func.Cli/Telemetry/TelemetryClient.cs
```diff
@@ -0,0 +1,155 @@
...
+            foreach (var (key, value) in properties)
+            {
+                activity.SetTag($"command.{key}", value);
+            }
+        }
+
+        activity.SetStatus(isSuccess ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
+
+        // Stop the activity if we started it here
+        if (activity != Activity.Current)
```
satvu: I'm not sure this check is correct - I think `StartActivity()` may set `Activity.Current` to whatever activity it is starting. 

Basing this on `Activity.Start` behavior ([see docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.start?view=net-10.0#remarks)), which is not the same method/type being used in line 70 but I think it's worth checking.

src/Func.Cli/Telemetry/TelemetryClient.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    public bool IsEnabled { get; }
+
+    public AppInsightsTelemetryClient()
+    {
+        IsEnabled = CheckEnabled();
+
+        if (IsEnabled)
+        {
+            var connectionString = $"InstrumentationKey={Constants.TelemetryInstrumentationKey}";
```
satvu, line 40: AI review:

Azure Monitor now prefers full connection strings (InstrumentationKey=...;IngestionEndpoint=...).
Using just the ikey works but may break in sovereign clouds or future SDK versions.

liliankasem: Deferring on this one — the bare InstrumentationKey={key} form is still a valid connection string and is what we've intentionally targeted: Core Tools telemetry has only ever flowed to public-cloud Application Insights, and there's no current ask for sovereign-cloud support. Happy to revisit (switching the build-time constant to a full connection string with explicit IngestionEndpoint) if/when sovereign coverage becomes a requirement.

jviau:
I am curious why an `ITelemetry` abstraction is needed? I would think this is achievable via simple static method helpers that interface directly with `ActivitySource`.

As for no-op, `ActivitySource` already performs a no-op behavior when it has no listeners, and returns `null` when creating the activity. So your 'no-op' can be achieved by simply not wiring up the OTel SDK as an activity source listener.

src/Func.Cli/Telemetry/TelemetryClient.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                activity.SetTag($"command.{key}", value);
+            }
+        }
+
+        activity.SetStatus(isSuccess ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
+
+        if (startedHere)
+        {
+            activity.Stop();
```
jviau: creating and immediately disposing an activity is not its intended use. An activity is meant to exist for the lifetime of the operation it represents. If you are wanting to simply track a commands usage, then you want a _metric_, not a _trace_ (Activity).

src/Func.Cli/Program.cs
```diff
@@ -47,22 +55,43 @@
...
+    {
+        commandActivity = aiClient.StartCommandActivity(commandName);
+    }
+
     var config = new InvocationConfiguration { EnableDefaultExceptionHandler = false };
     var exitCode = await rootCommand.Parse(args).InvokeAsync(config, cts.Token);
 
+    stopwatch.Stop();
+    telemetry.TrackCommand(commandName, isSuccess: exitCode == 0, durationMs: stopwatch.ElapsedMilliseconds);
+    commandActivity?.Stop();
```
jviau: use the disposal pattern for activities, it will call `Stop` for you and lets you ensure it is stopped on exception as well.

src/Func.Cli/Telemetry/AppInsightsTelemetryClient.cs
```diff
@@ -0,0 +1,167 @@
...
+        if (activity is not null)
+        {
+            activity.SetTag("command.name", commandName);
+            AddCommonTags(activity);
+        }
+
+        return activity;
+    }
+
+    public void TrackCommand(string commandName, bool isSuccess, long durationMs, IDictionary<string, string>? properties = null)
```
jviau: I feel like this should just be an extension method for `Activity`. The part where it will create/start an activity if it does not exit doesn't seem important. The caller should be the one controlling if the activity exists or not.

src/Func.Cli/Program.cs
```diff
@@ -47,22 +55,43 @@
...
 
 // Parse and invoke asynchronously
+var commandName = ResolveCommandName(args);
+Activity? commandActivity = null;
+var stopwatch = Stopwatch.StartNew();
+
 try
 {
+    // Start telemetry span wrapping the entire command execution
+    if (telemetry is AppInsightsTelemetryClient aiClient)
```
jviau: needing to cast the contract kind of defeats the purpose of having an abstraction layer here.

jviau (changes requested):
My recommendation is:

1. Remove the `ITelemetry` abstraction. I believe the built in .NET APIs are sufficient for our needs with no abstraction.
2. Add extension methods or helper methods to work with the .NET APIs.
3. Let the default no-listener behavior be the 'no-op' behavior
   - the 'no-op' selection becomes decided at startup time, conditionally set up the OTel SDK. If not set up, the .NET APIs will be effectively a no-op.
4. Make sure to use `Dispose` on activities.

The general end API structure should look like:

``` CSharp
// startup
if (weHaveAppInsightsConnection)
{
  // set up OTel SDK.
}

using Activity? activity = ActivitySource.StartCommandActivity(...); // can use new extensibility model to add a static extension to `ActivitySource`. Or make your own type.

// parse command.

activity?.SetCommandTags(...);
```

src/Func.Cli/Telemetry/CliTelemetry.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// key has not been injected via the build-time attribute.
+    /// </summary>
+    private const string EmptyInstrumentationKey = "00000000-0000-0000-0000-000000000000";
+
+    public static readonly ActivitySource Source = new(SourceName, ActivityExtensions.CliVersion);
+
+    public static readonly Meter Meter = new(SourceName, ActivityExtensions.CliVersion);
+
+    private static readonly Counter<long> _commandCounter =
+        Meter.CreateCounter<long>("func.cli.command.count", unit: "{command}", description: "Number of CLI commands invoked.");
```
jviau: Should we use `azure.functions.cli` to keep the prefix consistent with our other OTel metrics, like in the host?

src/Func.Cli/Telemetry/CliTelemetry.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        return $"InstrumentationKey={key}";
+    }
+
+    /// <summary>
+    /// Records a command-execution metric (count + duration histogram).
+    /// No-op when no <see cref="MeterListener"/> is subscribed.
+    /// </summary>
+    public static void TrackCommand(string commandName, bool isSuccess, long durationMs, IDictionary<string, string>? properties = null)
```
jviau: OTel has semantic conventions for CLI programs. Are we following them?

https://opentelemetry.io/docs/specs/semconv/cli/cli-spans/

It is _experimental_ (not stable)

src/Func.Cli/Telemetry/CliTelemetry.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        var tags = new TagList
+        {
+            { "command.name", commandName },
+            { "command.success", isSuccess },
+            { "cli.version", ActivityExtensions.CliVersion },
```
jviau: CLI version should be part of the _Resource_, not set per span. Also I think `version` is an existing OTel attribute we should use here. I know there is `service.version`, but this isn't a _service_ per-say. But I think we still use it, because Azure Monitor exporter probably translates that to a well-known telemetry dimension (and not a custom dimension). 

https://opentelemetry.io/docs/languages/dotnet/resources/#adding-resources-in-code

src/Func.Cli/Telemetry/ActivityExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    internal static string CliVersion => _cliVersion;
+
+    /// <summary>
+    /// Starts an <see cref="Activity"/> that represents the execution of a
+    /// CLI command. Returns <c>null</c> when no listener is subscribed.
+    /// </summary>
+    public static Activity? StartCommandActivity(this ActivitySource source, string commandName)
+    {
+        var activity = source.StartActivity($"func {commandName}", ActivityKind.Client);
```
jviau: OTel CLI spec says spans for CLI apps should be `INTERNAL`

```suggestion
        var activity = source.StartActivity($"func {commandName}", ActivityKind.Internal);
```

src/Func.Cli/Program.cs
```diff
@@ -22,6 +28,25 @@
...
+// Wire up the OpenTelemetry SDK only when telemetry is configured (build has
+// a real instrumentation key and the user has not opted out). When it isn't,
+// no listener is subscribed and ActivitySource/Meter calls become no-ops.
+TracerProvider? tracerProvider = null;
+MeterProvider? meterProvider = null;
+var connectionString = CliTelemetry.GetConnectionString();
+if (connectionString is not null)
+{
+    tracerProvider = Sdk.CreateTracerProviderBuilder()
+        .AddSource(CliTelemetry.SourceName)
```
jviau: We can use a resource builder to set common attributes that all spans and metrics will have.

See https://opentelemetry.io/docs/languages/dotnet/resources/#adding-resources-in-code

jviau (changes requested):

src/Func.Cli/Telemetry/ActivityExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        activity.SetTag("os.type", RuntimeInformation.OSDescription);
+        activity.SetTag("os.architecture", RuntimeInformation.OSArchitecture.ToString());
+        activity.SetTag("runtime.framework", RuntimeInformation.FrameworkDescription);
+        return activity;
+    }
+
+    /// <summary>
+    /// Marks the activity with success or failure status.
+    /// </summary>
+    public static Activity SetCommandResult(this Activity activity, bool isSuccess, string? errorDescription = null)
```
jviau: This doesn't seem to offer much value over having callers just call `SetStatus` directly. I believe success is also assumed if not explicitly set, so this really only needs to be done on failure.

A helper that may be valuable is something like `Activity.Fail(exception)` which will track and exception and set an error status.

src/Func.Cli/Program.cs
```diff
@@ -22,6 +28,25 @@
 var host = builder.Build();
 var interaction = host.Services.GetRequiredService<IInteractionService>();
 
+// Wire up the OpenTelemetry SDK only when telemetry is configured (build has
+// a real instrumentation key and the user has not opted out). When it isn't,
+// no listener is subscribed and ActivitySource/Meter calls become no-ops.
+TracerProvider? tracerProvider = null;
+MeterProvider? meterProvider = null;
+var connectionString = CliTelemetry.GetConnectionString();
+if (connectionString is not null)
```
jviau:
```suggestion
if (CliTelemetry.GetConnectionString() is { } connectionString)
```

or refactor for

``` CSharp
if (CliTelemetry.TryGetConnectionString(out string connectionString))
```

src/Func.Cli/Telemetry/CliTelemetry.cs
jviau: Having just CliTelemetry.TrackCommand is not entirely clear what kind of telemetry it is. Can we have a separate static class like Metrics.TrackCommand?

Or even, Metrics.Command.Record (or using static extensions, `Meter.Command.Record`) -- this a pattern I use here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckMetrics.cs

I got the pattern from https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Telemetry -- they have source gen for metrics that emits this pattern (but I didn't actually use the source gen).

Another option is an extension off Meter, so you can have CliTelemetry.Meter.RecordCommand, which is more easily identifiable that this is a metric.

src/Func.Cli/Program.cs
```diff
@@ -48,41 +73,101 @@
...
+    try
+    {
+        var config = new InvocationConfiguration { EnableDefaultExceptionHandler = false };
+        exitCode = await rootCommand.Parse(args).InvokeAsync(config, cts.Token);
 
-    // Print version update notice if available (bounded wait)
-    await PrintVersionNotice(interaction, versionCheckTask);
+        stopwatch.Stop();
+        activity?.SetCommandResult(exitCode == 0);
+        CliTelemetry.TrackCommand(commandName, isSuccess: exitCode == 0, durationMs: stopwatch.ElapsedMilliseconds);
```
jviau: Is it possible to avoid duplicating all these telemetry calls? Can we move it all to after the catch blocks?

src/Func.Cli/Program.cs
```diff
@@ -48,41 +73,101 @@
...
-    interaction.WriteError(ex.Message);
+        interaction.WriteError(ex.Message);
 
-    if (ex.VerboseMessage is not null)
+        if (ex.VerboseMessage is not null)
+        {
+            interaction.WriteHint(ex.VerboseMessage);
+        }
+
+        exitCode = ex.IsUserError ? 1 : 2;
```
jviau: I believe the standard is the inverse: `1` is 'general error', `2' is a misuse.

Does System.CommandLine not have default exit codes already for validation errors vs runtime exceptions?

jviau (changes requested):

src/Func.Cli/Telemetry/ActivityExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Azure.Functions.Cli.Telemetry;
+
+/// <summary>
+/// Helpers for starting and tagging <see cref="Activity"/> instances used by
+/// the CLI. Caller owns the activity's lifetime; these helpers only attach
+/// tags. When no listener is subscribed to the underlying
+/// <see cref="ActivitySource"/>, <see cref="ActivitySource.StartActivity(string, ActivityKind)"/>
+/// returns <c>null</c> and the tagging helpers are skipped.
+/// </summary>
+public static class ActivityExtensions
```
jviau, line 21: nit: you can use the new extension syntax, and even add static extensions! https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#extension-members

So you _could_ have something like this:

``` CSharp
public static class ActivityExtensions
{
    public static readonly ActivitySource Source = new("Azure.Functions.Cli", ActivityExtensions.CliVersion);

    extension(ActivitySource)
    {
        public static Activity? StartCommandActivity(string commandName)
        {
            var activity = Source.StartActivity($"func {commandName}", ActivityKind.Client); 
            return activity?.SetCommandTags(commandName);
        }
    }
}
```

With this extension method, no separate static holder needed. Usage is:

``` CSharp
using Activity? activity = ActivitySource.StartCommandActivity("my command name");
```

src/Func.Cli/Program.cs
liliankasem: It should! Let me look into that

src/Func.Cli/Telemetry/CliTelemetry.cs
liliankasem: We are not, but we can!

src/Func.Cli/Program.cs
liliankasem: Okay so,  System.CommandLine does have defaults but it sets exit code `1 ` for both parse errors and for any unhandled exception out of a handler. Instead, we want to have a GracefulException to surface as exit code `2` so scripts and CI can tell misuse apart from a real bug. The catch also lets us print a clean one-line message instead of System.CommandLine's default stack-trace dump for those cases.
 
We could instead register a custom exception handler on the System.CommandLine pipeline that maps GracefulException to exit code 2?

jviau: Personally, I would deprioritize customizing exit codes and let System.CommandLine's default exit of 1 remain. We can consider customizing them in a separate PR, maybe even later.

src/Func.Cli/Telemetry/ActivityExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+/// </remarks>
+public static class ActivityExtensions
+{
+    extension(ActivitySource)
+    {
+        /// <summary>
+        /// Starts an <see cref="Activity"/> that represents the execution of
+        /// a CLI command. Returns <c>null</c> when no listener is subscribed.
+        /// </summary>
+        public static Activity? StartCommandActivity(string commandName)
```
jviau: Thank you for implementing my feedback and apologies to ask changes once more. I would prefer a consistent pattern between activity and meter. So either:

All static extension:
``` CSharp
ActivitySource.StartCommandActivity(...);
Meter.RecordCommand(...); // or Meter.Command.Record(...)
```

note: the `Meter.Command.Record` is more type-verbose, but I like it because it reduces the overloads as you type `.Record`. But totally up to you.

or all off `CliTelemetry`

``` CSharp
// note I named the properties 'Trace' and 'Metric',
// but their types will be 'ActivitySource' and 'Meter' respectively.
// This is to align with what the types represent:
//    ActivitySource/Activity == traces
//    Meter == metrics.
CliTelemetry.Trace.StartCommandActivity(...);
CliTelemetru.Metric.RecordCommand(...);
```

Whichever you use is up to you.

src/Func.Cli/Program.cs
```diff
@@ -22,6 +28,28 @@
...
+// Wire up the OpenTelemetry SDK only when telemetry is configured (build has
+// a real instrumentation key and the user has not opted out). When it isn't,
+// no listener is subscribed and ActivitySource/Meter calls become no-ops.
+TracerProvider? tracerProvider = null;
+MeterProvider? meterProvider = null;
+if (CliTelemetry.TryGetConnectionString(out var connectionString))
+{
+    var resourceBuilder = CliTelemetry.CreateResourceBuilder();
+
+    tracerProvider = Sdk.CreateTracerProviderBuilder()
```
jviau: if we have an IHost with DI, we can use that for OTel construction instead of doing it statically here. This way it will all be flushed and disposed on host disposal.

https://www.nuget.org/packages/OpenTelemetry.Extensions.Hosting

Unless we don't plan on using the host much long term? What is that `IHost` all fulfilling?

src/Func.Cli/Telemetry/ActivityExtensions.cs
liliankasem: No worries! I appreciate all the input, I want to make sure we do this right the first time around :)

liliankasem: Going to go with CliTelemetry pattern

src/Func.Cli/Program.cs
liliankasem: That makes sense to me, happy to make this change if we keep IHost.

IHost right now isn't really serving a purpose yet(?) - I just scaffolded us starting from that point in anticipation for hosted services/Iconfig use etc. We can just as easily stripe this out for now and if we need it later, we can add it?

But if we switch off IHost, I think that  should be another PR

src/Func.Cli/Program.cs
liliankasem: That works, but we since this is already in vnext base, we should drop this in another PR to keep this scoped to telem. Esp if we are talking about dropping IHost - that can be one refactor PR, and this OTel pr can go on top, undoing any changes to the code here.

liliankasem: Had an agent draft up what this looks like: https://github.com/Azure/azure-functions-core-tools/compare/liliankasem/cli-simplify-startup?expand=1

jviau (approved):

src/Func.Cli/Program.cs
jviau: Yeah removing `IHost` in a later PR sounds good to me.

PR #4875, opened 2026-04-22
Add Go language support to func init

harshivcodes:
Changes

 - func init support for Go: new InitGoSubcommandAction and GoHelpers to scaffold a Go project (including main.go, go.mod, and local.settings.json) behind the Native 
worker runtime.
 - Language inference: added TryNormalizeLanguage in WorkerRuntimeLanguageHelper to preserve language inference for existing runtimes while accommodating Go.
 - Static resources: bundled a main.go sample template with the Functions Go SDK.
 - Tests: unit tests (GoHelperTests, InitActionGoTests) and E2E tests (GoInitTests) covering scaffolding, settings, and language normalization.

### DIFF
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -141,7 +141,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 .Callback(tf => TargetFramework = tf);
 
             Parser
-                .Setup<string>("language")
+                .Setup<string>('l', "language")
                 .SetDefault(null)
                 .Callback(l => Language = l);
 
--- a/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
+++ b/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
@@ -215,10 +215,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             {
                 (ResolvedWorkerRuntime, ResolvedLanguage) = ResolveWorkerRuntimeAndLanguage(WorkerRuntime, Language);
 
-                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
-                {
-                    ColoredConsole.WriteLine(WarningColor("[preview] The native worker runtime is currently in preview. Features and behavior may change in future releases."));
-                }
+                Utilities.WarnIfNativeWorkerRuntime(ResolvedWorkerRuntime);
 
                 // Order here is important: each language may have multiple runtimes, and each unique (language, worker-runtime) pair
                 // may have its own programming model. Thus, we assume that ResolvedLanguage and ResolvedWorkerRuntime are properly set
@@ -249,7 +246,18 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             else
             {
                 bool managedDependenciesOption = ResolveManagedDependencies(ResolvedWorkerRuntime, ManagedDependencies);
-                await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation, SkipGoModTidy);
+
+                if (ResolvedWorkerRuntime == Helpers.WorkerRuntime.Native)
+                {
+                    await GoHelpers.SetupProject(
+                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_"),
+                        SkipGoModTidy);
+                }
+                else
+                {
+                    await InitLanguageSpecificArtifacts(ResolvedWorkerRuntime, ResolvedLanguage, ResolvedProgrammingModel, managedDependenciesOption, GeneratePythonDocumentation);
+                }
+
                 await WriteFiles();
 
                 await WriteHostJson(ResolvedWorkerRuntime, managedDependenciesOption, ExtensionBundle);
@@ -295,7 +303,11 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                 }
                 else
                 {
-                    language = WorkerRuntimeLanguageHelper.GetDefaultTemplateLanguageFromWorker(workerRuntime);
+                    language = LanguageSelectionIfRelevant(workerRuntime);
+                    if (string.IsNullOrEmpty(language))
+                    {
+                        language = WorkerRuntimeLanguageHelper.GetDefaultTemplateLanguageFromWorker(workerRuntime);
+                    }
                 }
             }
             else if (GlobalCoreToolsSettings.CurrentWorkerRuntimeOrNone == Helpers.WorkerRuntime.None)
@@ -345,8 +357,7 @@ namespace Azure.Functions.Cli.Actions.LocalActions
             string language,
             ProgrammingModel programmingModel,
             bool managedDependenciesOption,
-            bool generatePythonDocumentation = true,
-            bool skipGoModTidy = false)
+            bool generatePythonDocumentation = true)
         {
             switch (workerRuntime)
             {
@@ -390,13 +401,6 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                     break;
                 case Helpers.WorkerRuntime.Node:
                     await NodeJSHelpers.SetupProject(programmingModel, language);
-                    break;
-                case Helpers.WorkerRuntime.Native:
-                    await GoHelpers.SetupProject(
-                        Utilities.SanitizeLiteral(Path.GetFileName(Environment.CurrentDirectory), allowed: "-_."),
-                        skipGoModTidy);
-                    await FileSystemHelpers.WriteFileIfNotExists("main.go", await StaticResources.MainGo);
-                    await FileSystemHelpers.WriteFileIfNotExists(Constants.FuncIgnoreFile, await StaticResources.FuncIgnore);
                     break;
             }
         }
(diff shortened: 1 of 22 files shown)


test/Cli/Func.E2ETests/Commands/FuncInit/GoInitTests.cs
```diff
@@ -0,0 +1,111 @@
...
+using Azure.Functions.Cli.TestFramework.Assertions;
+using Azure.Functions.Cli.TestFramework.Commands;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Azure.Functions.Cli.E2ETests.Commands.FuncInit
+{
+    [Trait(WorkerRuntimeTraits.WorkerRuntime, WorkerRuntimeTraits.Go)]
+    public class GoInitTests(ITestOutputHelper log) : BaseE2ETests(log)
```
liliankasem, line 15: AI review 

**E2E never compiles the scaffold**

test/Cli/Func.E2ETests/Commands/FuncInit/GoInitTests.cs

The E2E suite asserts files exist and that main.go contains "package
main". It never invokes go build or go vet on the scaffolded
project. So the moment any of the following silently breaks, tests
stay green and users hit a broken first-run:

 - azure-functions-golang-worker SDK API changes (the template 
imports sdk.WithMethods(...), sdk.WithAuth(...), worker.Start(app))
 - The go get …@latest resolves to a major version with breaking 
changes
 - A typo lands in StaticResources/main.go

Fix: add one E2E test that runs go build ./... after init and
asserts exit code 0, gated on go being on PATH (the unit tests
already use SkipIfGoNonExistFact — reuse that).

Also worth pinning the worker dep — go get …@latest in 
GoHelpers.SetupProject will resolve to whatever's newest at install
time, which means func init is non-reproducible. Consider @v0.x.y or
whatever the worker SDK currently recommends.

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                await RunGoCommandAsync("mod tidy", null, throwOnFailure: false);
+            }
+            else
+            {
+                ColoredConsole.WriteLine(AdditionalInfoColor("Skipped \"go mod tidy\". You must run \"go mod tidy\" manually."));
+            }
+        }
+
+        private static async Task RunGoCommandAsync(string arguments, string errorMessage, bool throwOnFailure = true)
```
liliankasem, line 84: **RunGoCommandAsync swallows stderr on throwOnFailure: false**

GoHelpers.cs:107-118

When mod tidy fails (used at SetupProject with throwOnFailure: false), the code prints a generic warning but throws away the captured stderr. Users are told "you may need to run it manually" without seeing why it failed. Append the stderrOutput to the warning when verbose.

harshivcodes: made changes to bubble this up.

test/Cli/Func.E2ETests/Commands/FuncInit/GoInitTests.cs
harshivcodes: I verified that the scaffold code is also built successfully. Pinned the sdk to the latest version we have out.

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -38,7 +40,8 @@ public static class WorkerRuntimeLanguageHelper
             { WorkerRuntime.Python, new[] { "py" } },
             { WorkerRuntime.Java, new string[] { } },
             { WorkerRuntime.Powershell, new[] { "pwsh" } },
-            { WorkerRuntime.Custom, new string[] { } }
+            { WorkerRuntime.Custom, new string[] { } },
+            { WorkerRuntime.Go, new[] { "golang" } }
```
liliankasem, line 44:
```suggestion
            { WorkerRuntime.Go, new[] { "go", "golang" } }
```

liliankasem (approved):

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -14,6 +14,12 @@ public static class GoHelpers
         private const int MinimumGoMajorVersion = 1;
         private const int MinimumGoMinorVersion = 24;
 
+        // Pinned to a known-good release so that `func init` is reproducible and
+        // breaking SDK changes are caught by a Core Tools PR, not user first-runs.
+        // Bump this alongside any corresponding updates to StaticResources/main.go.
+        private const string GoWorkerModule = "github.com/azure/azure-functions-golang-worker";
+        private const string GoWorkerModuleVersion = "v0.2.0-preview";
```
liliankasem: This looks like a maintenance tax to me. There's nothing enforcing this.  A couple of cleaner options that would be more consistent with how the rest of the repo handles worker versions:

#### Option A: Pin in the scaffolded template, drop the explicit go get @version.

The SDK version really belongs in StaticResources/main.go (and/or a templated go.mod). If we ship those with the version pinned and let go mod tidy resolve it, we don't need the explicit go get github.com/.../worker@vX.Y.Z step at all. This is the most idiomatic Go approach. Contributors bumping the SDK only touch template files, never C# logic, and it matches what a Go dev would expect to look at.

#### Option B: Promote to MSBuild properties in eng/build/Packages.props.

If we want CI enforcement that the pinned version is current (the way validate-worker-versions.ps1 covers the other workers), define it as MSBuild properties:

```xml
<GoWorkerModule>github.com/Azure/azure-functions-golang-worker</GoWorkerModule>
<GoWorkerModuleVersion>v0.2.0-preview</GoWorkerModuleVersion>
```

Then surface to C# via a generated constants file or AssemblyMetadata attribute. validate-worker-versions.ps1 can read/update it the same way it handles the NuGet-based workers.

Ideally A on its own, or A + B layered together. The current setup means a future SDK bump touches C# code, the template, and (eventually) a validation script. That's three places that can drift.

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -14,6 +14,12 @@ public static class GoHelpers
         private const int MinimumGoMajorVersion = 1;
         private const int MinimumGoMinorVersion = 24;
 
+        // Pinned to a known-good release so that `func init` is reproducible and
+        // breaking SDK changes are caught by a Core Tools PR, not user first-runs.
+        // Bump this alongside any corresponding updates to StaticResources/main.go.
+        private const string GoWorkerModule = "github.com/azure/azure-functions-golang-worker";
```
liliankasem: nit: Go module paths are case-sensitive on the proxy. The canonical repo is Azure/azure-functions-golang-worker (capital A). Lowercase azure/... works through the Go module proxy's case-insensitive redirect today, but it's safer to match the canonical casing `github.com/Azure/azure-functions-golang-worker`

```suggestion
        private const string GoWorkerModule = "github.com/Azure/azure-functions-golang-worker";
```

src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs
```diff
@@ -490,6 +539,11 @@ private static async Task WriteDockerfile(WorkerRuntime workerRuntime, string la
             {
                 await FileSystemHelpers.WriteFileIfNotExists("Dockerfile", await StaticResources.DockerfileCustom);
             }
+            else if (workerRuntime == Helpers.WorkerRuntime.Go)
+            {
+                ColoredConsole.WriteLine(WarningColor("Docker support for the Go worker runtime is not yet available. Skipping Dockerfile creation."));
```
liliankasem: If a user explicitly passes --docker we should probably fail loudly ("Docker support for Go is not yet available") rather than warn-and-continue, otherwise they'll think it worked. The current behavior is fine when --docker is implicit, but it conflates "asked for it" vs "got it by default."

`throw new CliException("Docker support for the Go worker runtime is not yet available.");`

liliankasem (changes requested):
A couple of smaller follow-ups on how Go is wired into InitAction — most of the per-runtime branching here matches what Python/Node/etc. already do, so no concern there. Two spots stood out though:

#### 1. FUNCTIONS_WORKER_RUNTIME mapping.

GetRuntimeMoniker(WorkerRuntime.Go) returns "go", but the value we actually need to write into local.settings.json is "native". Right now that mapping lives inline in WriteLocalSettingsJson:

```C#
string workerRuntimeSetting = workerRuntime == WorkerRuntime.Go
    ? "native"
    : WorkerRuntimeLanguageHelper.GetRuntimeMoniker(workerRuntime);
```

  Could we move that into WorkerRuntimeLanguageHelper (e.g. a GetFunctionsWorkerRuntimeSetting(workerRuntime) helper)? Every other runtime round-trips between the moniker and the host-facing setting, so anyone else who needs the host value will reach for GetRuntimeMoniker and silently get the wrong string. Centralizing it keeps the divergence in one place.

#### 2. Dispatch in InitFunctionAppProject.

Every other runtime flows through InitLanguageSpecificArtifacts (which has its own internal switch). Go is the first to be handled with a sibling branch at the call site:

```C#
if (ResolvedWorkerRuntime == WorkerRuntime.Go)
{
    await GoHelpers.SetupProject(...);
}
else
{
    await InitLanguageSpecificArtifacts(...);
}
```

  Could we add a case WorkerRuntime.Go: inside InitLanguageSpecificArtifacts that calls GoHelpers.SetupProject instead? Keeps the dispatch in one place and matches the existing shape. Small thing but it's the difference between "Go fits the pattern" and "Go is the exception that splits dispatch in two."

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: This will have to be fixed in the sdk first, I am gonna leave it here like this until we release the fix for the sdk.

liliankasem (approved):

aishwaryabh (changes requested):
I don't see this test running in the pipeline. You would need to add the following to all of the `test-e2e-*.yml`

      - name: Go
        languageWorker: 'Go'

aishwaryabh (approved):
approving to unblock PR. Pipeline changes should be addressed in a future PR

PR #4876, opened 2026-04-22
Simplify CLI startup

liliankasem:
Two related cleanups to Program.cs that don't change observable behavior beyond exit codes:

1. Replace Host.CreateEmptyApplicationBuilder with a bare ServiceCollection. We only register two singletons (ITheme, IInteractionService) and resolve one — no hosted services, configuration pipeline, or logging pipeline. Using IHost was overkill to start with.
  
    - This PR drops the Microsoft.Extensions.Hosting package reference. DI itself (Microsoft.Extensions.DependencyInjection) stays.
    - If we later need IHostedService, IConfiguration, or ILogger pipelines the host is ~10 lines to bring back.

2. Stop customizing exit codes. System.CommandLine returns 1 for parse errors and unhandled handler exceptions; we now return 1 for GracefulException too (whether IsUserError or not). The catch block is preserved purely to suppress the stack-trace dump and print a clean one-line message via IInteractionService.

    - A future PR can introduce a richer exit-code policy (e.g. 2 for user misuse vs 1 for tool bugs, plus a custom System.CommandLine exception handler) on its own merits.

### DIFF
--- a/src/Func.Cli/Program.cs
+++ b/src/Func.Cli/Program.cs
@@ -14,9 +14,6 @@ using OpenTelemetry;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
 
-// DI container — kept deliberately minimal. We only need to resolve a few
-// singletons; we don't run hosted services, configuration, or logging
-// pipelines, so a bare ServiceProvider is enough.
 var services = new ServiceCollection();
 services.AddSingleton<ITheme, DefaultTheme>();
 services.AddSingleton<IInteractionService, SpectreInteractionService>();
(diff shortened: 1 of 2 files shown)


src/Func.Cli/Program.cs
```diff
@@ -10,43 +10,39 @@
...
-using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 
-// Build the host with DI
-var builder = Host.CreateEmptyApplicationBuilder(new HostApplicationBuilderSettings
-{
-    Args = args,
-    DisableDefaults = true
-});
+var services = new ServiceCollection();
```
jviau: Do we even need a service collection at this point? Can we just manually construct the `SpectreInteractionService`, seems simple enough.

jviau (approved):

satvu (approved):

src/Func.Cli/Program.cs
liliankasem: fair :D

PR #4882, opened 2026-04-25
Pivot CLI to HostApplicationBuilder with hosted OpenTelemetry

liliankasem:
Switches Program.cs to Host.CreateEmptyApplicationBuilder and wires the OpenTelemetry SDK through the hosting integration.

## What changes

- Program.cs: build a host, register OTel via AddOpenTelemetry().WithTracing/.WithMetrics (only when a connection string is present), StartAsync before invoking the command, StopAsync(3s) after to flush exporters.
- CliTelemetry.cs: new ConfigureResource(ResourceBuilder) helper used by the hosted pipeline; standalone CreateResourceBuilder() retained for tests.
- Func.Cli.csproj: add OpenTelemetry.Extensions.Hosting and Microsoft.Extensions.Hosting package refs (versions already centralized).

## Why

Sets the stage for DI-based workload extension points (next PR) by giving the CLI a host. Telemetry lifetime is now managed by the host instead of manual ForceFlush + Dispose of providers in Program.cs.

## Verification

- dotnet build clean
- dotnet test — 53/53 pass
- Smoke: func --version and func --help work.

Stacked PR — base: vnext. Next PR (workloads/abstractions) will build on this.

### DIFF
--- a/src/Func.Cli/Program.cs
+++ b/src/Func.Cli/Program.cs
@@ -9,34 +9,41 @@ using Azure.Functions.Cli.Console;
 using Azure.Functions.Cli.Console.Theme;
 using Azure.Functions.Cli.Telemetry;
 using Azure.Monitor.OpenTelemetry.Exporter;
-using OpenTelemetry;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
 
 ITheme theme = new DefaultTheme();
 IInteractionService interaction = new SpectreInteractionService(theme);
 
-// Wire up the OpenTelemetry SDK only when telemetry is configured (build has
-// a real instrumentation key and the user has not opted out). When it isn't,
-// no listener is subscribed and ActivitySource/Meter calls become no-ops.
-TracerProvider? tracerProvider = null;
-MeterProvider? meterProvider = null;
+// Build the host: a CLI process gets the empty variant so we skip the default
+// configuration and logging providers (they're not needed). The host owns
+// shared lifetimes — currently just the OpenTelemetry pipeline, which is
+// registered as hosted services so flush + shutdown happen via host disposal.
+var hostBuilder = Host.CreateEmptyApplicationBuilder(null);
+hostBuilder.Services.AddSingleton(interaction);
+
+// Wire OpenTelemetry only when a real instrumentation key is baked in and
+// the user hasn't opted out. When it isn't, no listener is subscribed and
+// ActivitySource / Meter calls become no-ops.
 if (CliTelemetry.TryGetConnectionString(out var connectionString))
 {
-    var resourceBuilder = CliTelemetry.CreateResourceBuilder();
-
-    tracerProvider = Sdk.CreateTracerProviderBuilder()
-        .SetResourceBuilder(resourceBuilder)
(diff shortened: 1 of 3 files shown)


jviau:
Do we really need to be using a host? We don't use it for the actual System.CommandLine implementation. It seems like a lot of overhead just to get a tiny bit of DI setup (which isn't even really used?)

What aspects of the host are you after? Is there a way to accomplish what you are after with a smaller layer?

src/Func.Cli/Program.cs
```diff
@@ -103,22 +110,15 @@
...
-    tracerProvider?.Dispose();
-    meterProvider?.Dispose();
-}
+// Print version update notice if available (bounded wait)
+await PrintVersionNotice(interaction, versionCheckTask);
+
+// Shut down the host: stops hosted services and disposes the OTel providers,
+// which flushes pending telemetry to Azure Monitor. Bounded so a slow
+// exporter can't hang the CLI on exit.
+await host.StopAsync(TimeSpan.FromSeconds(3));
```
fabiocav: The comment here is a bit misleading.

The work performed by the ` TelemetryHostedService` to flush is all int he dispose path. Stop is essentially a no-op.

Disposal is already handled here, and that's what is going to trigger the flush. If we want to handle this and give the telemetry provider a bounded flush window, we can explicitly call flush here, waiting on that process to finish, up to the window duration.

liliankasem:
> Do we really need to be using a host? We don't use it for the actual System.CommandLine implementation. It seems like a lot of overhead just to get a tiny bit of DI setup (which isn't even really used?)
> 
> What aspects of the host are you after? Is there a way to accomplish what you are after with a smaller layer?

The host is here in anticipation of workloads (landing in the next PRs) - just trying to keep the PRs small. The DI graph is for IWorkload instances + their contributed services (IProjectInitializer, Command registrations, etc.)

https://github.com/Azure/azure-functions-core-tools/pull/4883

src/Func.Cli/Program.cs
```diff
@@ -103,22 +110,17 @@
...
-    meterProvider?.Dispose();
-}
+// Print version update notice if available (bounded wait)
+await PrintVersionNotice(interaction, versionCheckTask);
+
+// Give OTel a bounded window to push pending telemetry before we exit.
+// The actual flush happens in provider disposal (triggered by `using var host`),
+// but ForceFlush lets us cap the wait so a slow exporter can't hang the CLI.
+const int flushTimeoutMs = 3000;
+host.Services.GetService<TracerProvider>()?.ForceFlush(flushTimeoutMs);
```
jviau: I am not sure this is needed. 

Line 42: `using var host = hostBuilder.Build();` -- this will automatically dispose the container, triggering disposal of all services, leading to these being flushed automatically.

And then on CTRL+C, I believe the host builder will also have hooked up disposal there.

src/Func.Cli/Program.cs
jviau: I believe Otel SDK disposal already has a bounded flush. Will adding our own bounded flush add any value?

5 second shutdown window here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/ad7412c319b2c2c6ba5ce3d461d30e9b4e17f8a5/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L369

jviau (approved):

fabiocav (approved):

PR #4883, opened 2026-04-25
Add workload abstractions and DI host wiring

liliankasem:
Introduces the workload extension contract and wires it into the CLI host. The runtime loader and `func workload install` / `uninstall` commands land in a follow-up PR.

## Abstractions (`Func.Cli.Abstractions/Workloads`)

- **`IWorkload`** — entry point. Identity (`PackageId`, `PackageVersion`, `DisplayName`, `Description`) plus `Configure(FunctionsCliBuilder)`.
- **`FunctionsCliBuilder`** — abstract class that exposes the host's `IServiceCollection` to workloads. Abstract (rather than an interface) so we can grow the surface with new properties or virtual members without breaking workloads.
- **`IProjectInitializer`** — `func init` extension for a stack: `Stack`, `SupportedLanguages`, `CanHandle`, `GetInitOptions`, `InitializeAsync(InitContext, ParseResult, …)`.
- **`WorkloadContext` / `InitContext`** — records carrying inputs to providers.

## Host wiring (`Func.Cli/Hosting`)

- **`DefaultFunctionsCliBuilder`** — internal `FunctionsCliBuilder` impl backed by the host's `IServiceCollection`.
- **`BuiltInCommands.AddBuiltInCommands(IServiceCollection)`** — registers every built-in top-level command as a `Command` service so built-ins flow through the same DI path as workload-contributed commands.
- **`WorkloadRegistration.RegisterWorkloads`** — invoked by `Program.cs` during host construction. The runtime loader will land here next; for now it registers an empty `IReadOnlyList<WorkloadInfo>` so DI consumers get a deterministic empty result.
- **`WorkloadInfo`** (CLI-internal) — render-ready manifest entry consumed by `func workload list`.

## Built-in command changes

- **`Parser.CreateCommand(IServiceProvider)`** pulls every `Command` from DI, validates that names are unique (throws on collision), then constructs `HelpCommand` last with a back-reference to the constructed root.
- **`InitCommand`** takes `IEnumerable<IProjectInitializer>` via DI, attaches each initializer's `GetInitOptions()` to the command, and selects an initializer by `--stack` / single-installed / interactive prompt. Throws at startup if two initializers register for the same stack. With zero registrations it falls back to the new `WorkloadHints` empty-state message.
- **`func workload`** parent command + **`func workload list`** (consumes `IReadOnlyList<WorkloadInfo>`).

Workloads contribute new top-level commands by registering a `Command` instance in DI (`builder.Services.AddSingleton<Command>(...)`) — no separate provider interface needed.

## Empty-state behavior

Until the loader lands:
- `func workload list` prints `No workloads installed.`
- `func init` prints `No stacks installed.` and a hint, exit 1
- `func new`, `func pack`, `func start` are unchanged

## Testing

`Parser.CreateCommand` only exposes the production `IServiceProvider` overload. Tests use `TestParser.CreateRoot(interaction)` (in the test project) which builds the same DI shape `Program.cs` builds at runtime.

## Verification

- `dotnet build` clean, `dotnet test` — 53/53 pass
- `func --help`, `func init --help`, `func workload list`, `func init <tmp>` smoke-tested

## Stack

Stacked on **#4882** (host pivot). PR target is `liliankasem/workloads/host-pivot`


src/Func.Cli.Abstractions/Workloads/IFunctionsCliBuilder.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// Bootstrap surface passed to <see cref="IFunctionsCliStartup.Configure"/>.
+/// Workloads register their services — project initializers, command
+/// providers, and any other supporting types — through <see cref="Services"/>.
+///
+/// Modeled after WebJobs' <c>IWebJobsBuilder</c>: the builder exposes the DI
+/// container so workloads can use any standard .NET DI extension method, and
+/// the host picks up the registered services after all startups have run.
+/// </summary>
+public interface IFunctionsCliBuilder
```
jviau: Consider using an abstract type so that we can properties and such without breaking changes later.

src/Func.Cli/Hosting/FunctionsCliBuilder.cs
```diff
@@ -0,0 +1,21 @@
...
+using Azure.Functions.Cli.Workloads;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Azure.Functions.Cli.Hosting;
+
+/// <summary>
+/// Default <see cref="IFunctionsCliBuilder"/> implementation. Internal so
+/// workloads only see the abstraction.
+/// </summary>
+internal sealed class FunctionsCliBuilder : IFunctionsCliBuilder
```
jviau: nit: use primary ctors:

``` csharp
internal sealed class FunctionsCliBuilder(IServiceCollection services) : IFunctionsCliBuilder
{
    public IServiceCollection Services { get; } = services;
}
```

Also consider argument validation.

src/Func.Cli/Parser.cs
```diff
@@ -4,53 +4,77 @@
...
+/// <see cref="Command"/> in DI for entirely new subcommands.
 /// </summary>
 public static class Parser
 {
     /// <summary>
-    /// Creates and configures the root CLI command with all subcommands registered.
+    /// Convenience overload for tests/scenarios that don't need workloads —
+    /// builds a minimal service provider with just the supplied interaction service.
     /// </summary>
     public static FuncRootCommand CreateCommand(IInteractionService interaction)
```
jviau: Does this need to be in the production assembly if it is for tests? Or can we move it to a test helper?

src/Func.Cli/Parser.cs
```diff
@@ -4,53 +4,77 @@
...
+    public static FuncRootCommand CreateCommand(IServiceProvider services)
+    {
+        var interaction = services.GetRequiredService<IInteractionService>();
         var rootCommand = new FuncRootCommand();
 
-        // Create built-in commands
         var helpCommand = new HelpCommand(interaction, rootCommand);
         var versionCommand = new VersionCommand(interaction);
-        var initCommand = new InitCommand(interaction);
+        var initCommand = ActivatorUtilities.CreateInstance<InitCommand>(services);
```
jviau: Can we leverage the DI container for all of our built-in commands as well? Or is there some specific reason not to?

src/Func.Cli.Abstractions/Workloads/InstalledWorkload.cs
```diff
@@ -0,0 +1,24 @@
...
+/// discovery layer; consumed by <c>func workload list</c> and other
+/// commands that need to describe what's available.
+/// </summary>
+/// <param name="PackageId">NuGet package id (e.g. <c>"Azure.Functions.Cli.Workload.Dotnet"</c>).</param>
+/// <param name="DisplayName">Human-readable name.</param>
+/// <param name="Description">One-line description.</param>
+/// <param name="Aliases">User-facing tokens accepted by <c>-s</c> (e.g. <c>["dotnet", "dotnet-isolated"]</c>).</param>
+/// <param name="Version">Installed package version.</param>
+/// <param name="Type">Workload category (stack / tool / extension).</param>
+public sealed record InstalledWorkload(
```
fabiocav: "Installed" here implies a state distinction. 

Can you confirm this will carry the _reference to the workload_ we hydrate from the workloads configuration file? If so:
- Let's rename this type to something like `WorloadInfo`
- Do we need this type to be public? This would be used by the CLI only.

src/Func.Cli/Parser.cs
```diff
@@ -4,53 +4,77 @@
...
 
-        // Replace built-in help rendering with Spectre on all commands
-        ReplaceHelpAction(rootCommand, helpCommand);
+        // Workloads contribute new top-level subcommands by registering
+        // Command instances in DI (e.g. builder.Services.AddSingleton<Command>(new MyCommand())).
+        // Built-ins above are constructed inline rather than registered, so we
+        // never double-add them here.
+        foreach (var command in services.GetServices<Command>())
+        {
+            rootCommand.Subcommands.Add(command);
```
fabiocav, line 87: Do we need to validate the commands here (e.g., ensure we're not hiding built-ins, avoiding dupes, etc.)

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -34,34 +36,101 @@ public class InitCommand : BaseCommand
...
         Options.Add(LanguageOption);
         Options.Add(ForceOption);
+
+        // Workload-contributed options are attached after built-ins so they
+        // appear as a clearly-grouped block in --help output.
+        foreach (var initializer in _initializers)
+        {
+            foreach (var option in initializer.GetInitOptions())
+            {
+                Options.Add(option);
```
fabiocav, line 67: We need some validation on this as well to handle duplications.

src/Func.Cli.Abstractions/Workloads/InstalledWorkload.cs
```diff
@@ -0,0 +1,24 @@
...
+    string PackageId,
+    string DisplayName,
+    string Description,
+    IReadOnlyList<string> Aliases,
+    string Version,
```
fabiocav: We have `PackageVersion` in `IWorkload`. Should we ensure these are consistent?

src/Func.Cli/Commands/WorkloadHints.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    public static void WriteNoMatchingWorkload(
+        IInteractionService interaction,
+        IReadOnlyList<string> installedStacks,
+        string actionDescription,
+        string? requestedStack = null)
+    {
+        if (installedStacks.Count == 0)
+        {
+            interaction.WriteError($"No language workloads installed.");
```
fabiocav: We should align the verbiage here with "stacks" as well for consistency.

src/Func.Cli/Commands/WorkloadHints.cs
```diff
@@ -0,0 +1,52 @@
...
+            return;
+        }
+
+        if (requestedStack is not null)
+        {
+            interaction.WriteError($"No installed workload supports stack '{requestedStack}'.");
+        }
+        else
+        {
+            interaction.WriteError("No stack specified and the installed workloads couldn't be auto-selected.");
```
fabiocav: nit: soften this message here to something like: 

>  Multiple stacks installed; pass --stack <name> to choose.
    [list of available stacks]

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -2,20 +2,22 @@
...
+/// delegated to an <see cref="IProjectInitializer"/> contributed by a workload.
+/// Each registered initializer may also contribute additional options to this
+/// command (e.g. dotnet's <c>--target-framework</c>).
 /// </summary>
 public class InitCommand : BaseCommand
 {
     public static readonly Option<string?> StackOption = new("--stack", "-s")
     {
-        Description = "The stack (language / runtime) for the project"
+        Description = "The stack for the project (e.g. dotnet, node, python)"
```
fabiocav: For this message, we should either have a generic text (e.g., "The stack to use. Run `func workload list` to see what's installed."), or dynamically enumerate the available stacks. This would also be a good opportunity to identify that no stacks are present and provide guidance if needed.

src/Func.Cli.Abstractions/Workloads/IWorkload.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// truth for what version of the workload is installed.
+    /// </summary>
+    public string PackageVersion { get; }
+
+    /// <summary>
+    /// Workload category (stack / tool / extension). Surfaces in
+    /// <c>func workload list</c> and lets the CLI group or filter workloads
+    /// by kind.
+    /// </summary>
+    public WorkloadType Type { get; }
```
fabiocav: I'm not sure we should be exposing a `WorkloadType` here at this point. Workloads can ultimately install anything, including a mix of stack, tools and other functionality.  It's just a "thing" that registers services and augments the CLI behavior. 

As workloads become more complex, the notion of "type" won't necessarily align with what they provide, and the categorization will become inaccurate and confusing. 

At this point, my recommendation is:

- Remove `WorkloadType`
- Drop the `Type` property
- If we feel that is needed, we can populate what the workload provides (a "provides" colum) with the list of commands, templates, etc. 

If we need additional categorization later, we can introduce something to support.

fabiocav (commented):
High level review.

src/Func.Cli/Parser.cs
liliankasem: No reason not to!

liliankasem:
@jviau + @fabiocav - initial feedback addressed if you have time for a second look!

src/Func.Cli.Abstractions/Workloads/IProjectInitializer.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Options this initializer contributes to the <c>func init</c> command.
+    /// Attached during command construction and visible in <c>--help</c>.
+    /// </summary>
+    public IReadOnlyList<Option> GetInitOptions();
+
+    /// <summary>Scaffolds a new project at <see cref="WorkloadContext.ProjectPath"/>.</summary>
+    public Task InitializeAsync(
+        InitContext context,
+        ParseResult parseResult,
```
liliankasem, line 34: Just a thought on this, this leaks System.CommandLine through the workload SDK. Every workload now has a transitive dependency on the System.CommandLine version in Func.Cli, so a future bump of System.CommandLine could break workloads at the binary level.

Do we want to consider having our own wrapper or something like `WorkloadOption`?

jviau: I get the concern, and any bump to System.CommandLine would be problematic. But on the other hand, will creating our own abstraction layer be a rabbit hole? We will need to wrap _a lot_ of `System.CommandLine` public API. Commands, arguments, parameters, validation, parsers, dotnet suggest, etc.

I think the cost/benefit of abstracting it away is not worth it to me. Let's hope they don't major version rev that library with breaking changes any time soon.

liliankasem: Yeah - I would hate to have a full Sys Cmdline abstraction that we have to maintain. But they do have a v2.0.0 in beta at the moment (though it takes them a long time to GA like they did for 1.0.0 - it could be another 3 years out :D )

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -34,35 +37,122 @@ internal class InitCommand : BaseCommand
     };
```
fabiocav, line 37: Is there a reason these are static? While it's unlikely we'll hit issues with the actual product, this does introduce some fragility that may lead to test reliability issues and/or surprises.

src/Func.Cli/Commands/InitCommand.cs
fabiocav: Did you mean to update? Defer? Should we add a comment here if deferring?

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -66,6 +69,23 @@ public InitCommand(
...
+        var dupes = _initializers
+            .GroupBy(i => i.Stack, StringComparer.OrdinalIgnoreCase)
+            .Where(g => g.Count() > 1)
+            .Select(g => g.Key)
+            .ToList();
+        if (dupes.Count > 0)
+        {
+            throw new InvalidOperationException(
+                $"Multiple workloads register an IProjectInitializer for the same stack: {string.Join(", ", dupes)}. " +
+                "Each stack must be owned by exactly one workload.");
```
jviau: Will this message be exposed to the end user? This message sounds like it is more for the workload authors and less for the end-user.

src/Func.Cli/Hosting/BuiltInCommands.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        // VersionCommand is also resolved by Parser to wire `func` (no args)
+        // to detailed-version output, so register the concrete type and
+        // surface it as a top-level Command.
+        services.AddSingleton<VersionCommand>();
+        services.AddSingleton<Command>(sp => sp.GetRequiredService<VersionCommand>());
+
+        services.AddSingleton<Command>(sp => ActivatorUtilities.CreateInstance<InitCommand>(sp));
```
jviau:
``` suggestion
        services.AddSingleton<Command, InitCommand>();
```

This overload will activate the type with the service provider for you.

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -34,35 +37,122 @@ internal class InitCommand : BaseCommand
...
+    {
+        if (_initializers.Count == 0)
+        {
+            return null;
+        }
+
+        if (!string.IsNullOrEmpty(stack))
+        {
+            return _initializers.FirstOrDefault(i =>
+                string.Equals(i.Stack, stack, StringComparison.OrdinalIgnoreCase));
```
fabiocav, line 134: Should we also use the aliases here?

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -108,7 +128,8 @@ protected override async Task<int> ExecuteAsync(ParseResult parseResult, Cancell
 
         if (!string.IsNullOrEmpty(stack))
         {
-            return _initializers.FirstOrDefault(i => i.CanHandle(stack));
+            return _initializers.FirstOrDefault(i =>
```
jviau, line 133: note for the future: do we want some kind of stack alias system? So that customers can use `javascript`, `js`, `node`, etc and they all resolve the same stack?

src/Func.Cli/Hosting/DefaultFunctionsCliBuilder.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Extensions.DependencyInjection;
+using AbstractionsBuilder = Azure.Functions.Cli.Workloads.FunctionsCliBuilder;
```
jviau: Is this alias needed?

src/Func.Cli/Parser.cs
```diff
@@ -4,71 +4,91 @@
...
-        // Command instances in DI (e.g. builder.Services.AddSingleton<Command>(new MyCommand())).
-        // Built-ins above are constructed inline rather than registered, so we
-        // never double-add them here.
-        foreach (var command in services.GetServices<Command>())
+        // First pass: built-ins. Their names form the reserved set, plus
+        // help (added above). Built-ins are trusted by construction — any
+        // collision among them is a CLI bug, not a workload problem, so we
+        // throw rather than skip.
+        var reservedNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { helpCommand.Name };
+        foreach (var command in allCommands.OfType<IBuiltInCommand>().Cast<Command>())
```
jviau: Can also do a `Where`

``` CSharp
allCommands.Where(x => x is IBuiltInCommand)
```

src/Func.Cli/Parser.cs
```diff
@@ -4,71 +4,91 @@
...
+                interaction.WriteWarning(
+                    $"A workload tried to register top-level command '{command.Name}', which is reserved by the CLI. This command was not loaded.");
+                continue;
+            }
+
+            if (workloadNameCounts[command.Name] > 1)
+            {
+                interaction.WriteWarning(
+                    $"Multiple workloads registered top-level command '{command.Name}'. All conflicting registrations were skipped. " +
+                    "Run `func workload uninstall <name>` to keep one.");
```
jviau, line 83: `<name>` isn't clear here. Is there a way we can track down what workload added what command? This might warrant some kind of custom abstraction the workload uses which is what ultimately provides the commands (and so we know what workload produced what commands). Many other options for how we can force workloads to give us Command + Ownership together.

src/Func.Cli/Commands/InitCommand.cs
liliankasem: Yeah I have this in mind already and that's where `WorkloadInfo.Aliases ` will come into play - just not wired up yet, should be in a future pr!

src/Func.Cli/Parser.cs
liliankasem: Yeah that's a good call out. I was thinking to address this in a future pr, where the plan is to replace the raw IEnumerable<Command> registration with a typed contribution (WorkloadCommandContribution { IWorkload Owner; Command Command }) so ownership is a registration-time invariant, and the warning can name the offending workload(s). But open to other ideas for this?

src/Func.Cli/Commands/InitCommand.cs
liliankasem: No reason, just how it started out - I will move them all to non-static in a follow up pr!

src/Func.Cli/Commands/InitCommand.cs
liliankasem: Sorry yes deferring for future fix/convo - added a comment to come back to this

src/Func.Cli/Parser.cs
jviau: That way is what I originally thought as well. Another is checking for an attribute on the `Command`, but that might be less intuitive.

PR #4884, opened 2026-04-27
#4751 - Fix: resolve storage account lookup failure in func fetch-connection-string

### LINKED ISSUE #4751: func azure storage fetch-connection-string fails with “Cannot find storage account” although account exists and Azure CLI retrieves the connection string successfully
### Version

Azure Functions Core Tools: 4.5.0 (install via apt install)
Azure CLI: 2.80.0
OS: Linux

### Description

The command:

`func azure storage fetch-connection-string <storageAccountName>`

consistently fails with:

`Cannot find storage account with name <name>`

For any new storage account, even though the storage account exists in the current subscription, Azure CLI lists it, Azure CLI can successfully retrieve the connection string, and the Core Tools environment has been fully reinstalled and all caches cleared.

This started happening immediately after temporarily disabling another subscription in the tenant.
Re-enabling the subscription did not resolve the issue. The command "func azure storage fetch-connection-string" works fine for storage accounts that existed prior to disabling and re-enabling the other subscription in the tenant.

UPDATE: After reducing the number of storage accounts in the tenant, "Cannot find storage account" happens intermittently, so running the same command multiple times ends up finding the account and fetching the settings.

This problem is reproducible every time in Azure Cloud Shell and GitHub Codespaces / VS Code Dev Containers.
 

**Environment**

Azure Functions Core Tools: 4.5.0 (install via apt install)
Azure CLI: 2.80.0
OS: Linux

**What Works**

Azure CLI sees the account:

`az storage account list -g resource_group -o table`

Azure CLI retrieves its connection string:

```
az storage account show-connection-string \
  --name name \
  --resource-group resource_group \
  --query connectionString -o tsv
```

### Steps to reproduce

In an Azure tenant with multiple subscriptions disable one of them. Then, in an active subscription, create a storage account and run:

`func azure storage fetch-connection-string <storageAccountName>`

umangsriv:

Replaces /subscriptions/{id}/resources lookup (eventually consistent) with provider-specific storage API. Fixes intermittent “Cannot find storage account” errors.

resolves #4751

### Pull request checklist

### Additional information

Added unit tests with MockHttpMessageHandler and validated against real Azure.

### DIFF
--- a/src/Cli/func/Arm/ArmUriTemplates.cs
+++ b/src/Cli/func/Arm/ArmUriTemplates.cs
@@ -17,12 +17,9 @@ namespace Azure.Functions.Cli.Arm
         public const string BasicAuthCheckApiVersion = "2022-03-01";
         public const string FunctionsStacksApiVersion = "2020-10-01";
         public const string FlexFunctionsStacksApiVersion = "2020-10-01";
-        public const string StorageApiVersion = "2018-02-01";
         public const string ArgUri = "providers/Microsoft.ResourceGraph/resources";
 
         public static readonly ArmUriTemplate SubscriptionResourceByNameAndType = new ArmUriTemplate($"resources?$filter=(SubscriptionId eq '{{subscriptionId}}' and name eq '{{resourceName}}' and resourceType eq '{{resourceType}}')", ArmApiVersion);
-        public static readonly ArmUriTemplate StorageAccountsList = new ArmUriTemplate("subscriptions/{subscriptionId}/providers/Microsoft.Storage/storageAccounts", StorageApiVersion);
-        public static readonly ArmUriTemplate StorageAccountByResourceGroup = new ArmUriTemplate("subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Storage/storageAccounts/{storageAccountName}", StorageApiVersion);
     }
 
     public class ArmUriTemplate
--- a/src/Cli/func/Arm/ArmUriTemplates.cs
+++ b/src/Cli/func/Arm/ArmUriTemplates.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Text.RegularExpressions;
@@ -17,10 +17,12 @@ namespace Azure.Functions.Cli.Arm
         public const string BasicAuthCheckApiVersion = "2022-03-01";
         public const string FunctionsStacksApiVersion = "2020-10-01";
         public const string FlexFunctionsStacksApiVersion = "2020-10-01";
-
+        public const string StorageApiVersion = "2018-02-01";
         public const string ArgUri = "providers/Microsoft.ResourceGraph/resources";
 
         public static readonly ArmUriTemplate SubscriptionResourceByNameAndType = new ArmUriTemplate($"resources?$filter=(SubscriptionId eq '{{subscriptionId}}' and name eq '{{resourceName}}' and resourceType eq '{{resourceType}}')", ArmApiVersion);
+        public static readonly ArmUriTemplate StorageAccountsList = new ArmUriTemplate("subscriptions/{subscriptionId}/providers/Microsoft.Storage/storageAccounts", StorageApiVersion);
+        public static readonly ArmUriTemplate StorageAccountByResourceGroup = new ArmUriTemplate("subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Storage/storageAccounts/{storageAccountName}", StorageApiVersion);
     }
 
     public class ArmUriTemplate
--- a/src/Cli/func/Arm/ArmUriTemplates.cs
+++ b/src/Cli/func/Arm/ArmUriTemplates.cs
@@ -18,8 +18,6 @@ namespace Azure.Functions.Cli.Arm
         public const string FunctionsStacksApiVersion = "2020-10-01";
         public const string FlexFunctionsStacksApiVersion = "2020-10-01";
         public const string ArgUri = "providers/Microsoft.ResourceGraph/resources";
-
-        public static readonly ArmUriTemplate SubscriptionResourceByNameAndType = new ArmUriTemplate($"resources?$filter=(SubscriptionId eq '{{subscriptionId}}' and name eq '{{resourceName}}' and resourceType eq '{{resourceType}}')", ArmApiVersion);
     }
 
     public class ArmUriTemplate
--- a/src/Cli/func/Helpers/AzureHelper.cs
+++ b/src/Cli/func/Helpers/AzureHelper.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Azure.Functions.Cli.Arm;
@@ -17,6 +17,8 @@ namespace Azure.Functions.Cli.Helpers
     public static class AzureHelper
     {
         private static readonly string _storageApiVersion = "2018-02-01";
+
+        internal static Func<Func<Task<object>>, Task<object>> ArmCallOverrideAsync { get; set; }
 
         internal static async Task<Site> GetFunctionApp(string name, string accessToken, string managementURL, string slot = null, string defaultSubscription = null, IEnumerable<ArmSubscription> allSubs = null, Func<Site, string, string, Task<Site>> loadFunction = null)
         {
@@ -288,20 +290,63 @@ namespace Azure.Functions.Cli.Helpers
             var subscriptions = await GetSubscriptions(accessToken, managementURL);
             foreach (var subscription in subscriptions)
             {
-                var storageAccount =
-                    await ArmHttpAsync<ArmArrayWrapper<ArmGenericResource>>(
-                        HttpMethod.Get,
-                        ArmUriTemplates.SubscriptionResourceByNameAndType.Bind(managementURL, new
-                        {
-                            subscriptionId = subscription.SubscriptionId,
-                            resourceName = storageAccountName,
-                            resourceType = "Microsoft.Storage/storageAccounts"
-                        }),
-                        accessToken);
-
-                if (storageAccount.Value.Any())
+                try
                 {
-                    return await GetStorageAccount(storageAccount.Value.First(), accessToken, managementURL);
+                    // List all storage accounts in the subscription
+                    var listUrl = new Uri(
+                        $"{managementURL}/subscriptions/{subscription.SubscriptionId}" +
+                        $"/providers/Microsoft.Storage/storageAccounts" +
+                        $"?api-version={_storageApiVersion}");
+
+                    var accounts =
+                        await ExecuteArmCallAsync(() => ArmHttpAsync<ArmArrayWrapper<ArmGenericResource>>(
+                            HttpMethod.Get,
+                            listUrl,
+                            accessToken));
+
+                    if (accounts?.Value == null || !accounts.Value.Any())
+                    {
+                        continue;
+                    }
+
+                    // Filter in-memory by storage account name
+                    var match = accounts?.Value?
+                        .FirstOrDefault(a => string.Equals(
+                            a.Name,
+                            storageAccountName,
+                            StringComparison.OrdinalIgnoreCase));
+
+                    if (match == null)
+                    {
+                        continue;
+                    }
+
+                    var resourceGroup = ExtractResourceGroupFromId(match.Id);
+
+                    // Fetch storage account using RG-specific endpoint
+                    var getUrl = new Uri(
+                        $"{managementURL}/subscriptions/{subscription.SubscriptionId}" +
+                        $"/resourceGroups/{resourceGroup}" +
+                        $"/providers/Microsoft.Storage/storageAccounts/{storageAccountName}" +
+                        $"?api-version={_storageApiVersion}");
+
+                    var storageAccount =
+                        await ArmHttpAsync<ArmWrapper<ArmGenericResource>>(
+                            HttpMethod.Get,
+                            getUrl,
+                            accessToken);
+
+                    if (storageAccount != null)
+                    {
+                        return await GetStorageAccount(storageAccount, accessToken, managementURL);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (StaticSettings.IsDebug)
+                    {
+                        ColoredConsole.Error.WriteLine(ErrorColor(ex.ToString()));
+                    }
                 }
             }
 
@@ -758,6 +803,32 @@ namespace Azure.Functions.Cli.Helpers
             {
                 return null;
             }
+        }
+
+        private static string ExtractResourceGroupFromId(string resourceId)
+        {
+            var parts = resourceId.Split('/', StringSplitOptions.RemoveEmptyEntries);
+
+            for (int i = 0; i < parts.Length; i++)
+            {
+                if (parts[i].Equals("resourceGroups", StringComparison.OrdinalIgnoreCase))
+                {
+                    return parts[i + 1];
+                }
+            }
+
+            throw new InvalidOperationException("Resource group not found in resource ID");
+        }
+
+        private static async Task<T> ExecuteArmCallAsync<T>(Func<Task<T>> actualCall)
+        {
+            if (ArmCallOverrideAsync != null)
+            {
+                var result = await ArmCallOverrideAsync(() => actualCall().ContinueWith(t => (object)t.Result));
+                return (T)result;
+            }
+
+            return await actualCall();
         }
     }
 }
--- a/src/Cli/func/Helpers/AzureHelper.cs
+++ b/src/Cli/func/Helpers/AzureHelper.cs
@@ -286,89 +286,34 @@ namespace Azure.Functions.Cli.Helpers
         public static async Task<StorageAccount> GetStorageAccount(string storageAccountName, string accessToken, string managementURL)
         {
             var subscriptions = await GetSubscriptions(accessToken, managementURL);
-            Exception lastException = null;
+            var subIds = subscriptions?.Select(s => s.SubscriptionId).ToList();
 
-            foreach (var subscription in subscriptions)
+            if (subIds == null || subIds.Count == 0)
             {
-                try
-                {
-                    // Build URL using ArmUriTemplates
-                    var listUrl = ArmUriTemplates.StorageAccountsList.Bind(managementURL, new
-                    {
-                        subscriptionId = subscription.SubscriptionId
-                    });
-
-                    // Fetch all pages
-                    var allAccounts = new List<ArmWrapper<ArmGenericResource>>();
-
-                    var response = await ArmHttpAsync<ArmArrayWrapper<ArmGenericResource>>(
-                        HttpMethod.Get,
-                        listUrl,
-                        accessToken);
-
-                    while (response != null)
-                    {
-                        if (response.Value != null)
-                        {
-                            allAccounts.AddRange(response.Value);
-                        }
-
-                        if (string.IsNullOrEmpty(response.NextLink))
-                        {
-                            break;
-                        }
-
-                        response = await ArmHttpAsync<ArmArrayWrapper<ArmGenericResource>>(
-                        HttpMethod.Get,
-                        new Uri(response.NextLink),
-                        accessToken);
-                    }
-
-                    var match = allAccounts.FirstOrDefault(a =>
-                        string.Equals(a.Name, storageAccountName, StringComparison.OrdinalIgnoreCase));
-
-                    if (match == null)
-                    {
-                        continue;
-                    }
-
-                    var resourceGroup = ExtractResourceGroupFromId(match.Id);
-
-                    var getUrl = ArmUriTemplates.StorageAccountByResourceGroup.Bind(managementURL, new
-                    {
-                        subscriptionId = subscription.SubscriptionId,
-                        resourceGroup = resourceGroup,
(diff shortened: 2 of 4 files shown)


src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -288,20 +290,63 @@ public static async Task<StorageAccount> GetStorageAccount(string storageAccount
...
-                        {
-                            subscriptionId = subscription.SubscriptionId,
-                            resourceName = storageAccountName,
-                            resourceType = "Microsoft.Storage/storageAccounts"
-                        }),
-                        accessToken);
+                try
+                {
+                    // List all storage accounts in the subscription
+                    var listUrl = new Uri(
```
liliankasem: The list endpoint returns nextLink past ~100 results. The new code only inspects the first page, so subscriptions with many storage accounts can miss the target and reintroduce the exact "Cannot find storage account" failure mode this PR is fixing. GetSubscriptions (~line 387) already follows NextLink in a loop — mirror that.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -288,20 +290,63 @@ public static async Task<StorageAccount> GetStorageAccount(string storageAccount
...
+                            HttpMethod.Get,
+                            getUrl,
+                            accessToken);
+
+                    if (storageAccount != null)
+                    {
+                        return await GetStorageAccount(storageAccount, accessToken, managementURL);
+                    }
+                }
+                catch (Exception ex)
```
liliankasem: Auth failures (401/403), throttling (429), transient 5xx all silently get logged-in-debug-only and the loop continues, producing a misleading ArmResourceNotFoundException at the end. Narrow the catch, or rethrow on the last subscription so users see the real cause.

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
```diff
@@ -39,5 +41,101 @@ public async Task GetFunctionKeyTest(string key, string value)
...
+                // Mock subscriptions
+                mockHttp.When(HttpMethod.Get, subscriptionsUrl).Respond("application/json", @"{'value': [{ 'subscriptionId': 'sub', 'displayName': 'Test Sub' }]}");
+
+                // Mock empty storage accounts list
+                mockHttp.When(HttpMethod.Get, listUrl).Respond("application/json", @"{ 'value': [] }");
+
+                ArmClient.SetTestHandler(mockHttp);
+
+                // Act
+                Func<Task> act = async () =>
```
liliankasem: Func<Task> act is dead code, act is assigned but never invoked. Assert.ThrowsAsync is then called with a separate invocation of GetStorageAccount. Either drop act or use await act.Should().ThrowAsync<ArmResourceNotFoundException>().

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
```diff
@@ -39,5 +41,101 @@ public async Task GetFunctionKeyTest(string key, string value)
...
+                // Assert
+                await Assert.ThrowsAsync<ArmResourceNotFoundException>(() => AzureHelper.GetStorageAccount(storageName, accessToken, managementUrl));
+            }
+            finally
+            {
+                ArmClient.SetTestHandler(null);
+            }
+        }
     }
 }
```
liliankasem, line 173: ArmCallOverrideAsync + ExecuteArmCallAsync (:21, :823)

Not used by the new tests (they use ArmClient.SetTestHandler). Also asymmetric, only wraps the list call, not the GET. A mutable internal static is risky for cross-test contamination. Remove it.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -759,5 +804,31 @@ public static async Task<FlexFunctionsStacks> GetFlexFunctionsStacks(string acce
                 return null;
             }
         }
+
+        private static string ExtractResourceGroupFromId(string resourceId)
+        {
+            var parts = resourceId.Split('/', StringSplitOptions.RemoveEmptyEntries);
```
liliankasem: parts[i + 1] throws if resourceGroups is the last segment.

 Add i + 1 < parts.Length and fall through to the existing InvalidOperationException.

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Azure.Functions.Cli.Arm;
+using Azure.Functions.Cli.Arm.Models;
 using Azure.Functions.Cli.Helpers;
 using FluentAssertions;
+using Moq;
```
liliankasem: Moq, and likely Azure.Functions.Cli.Arm.Models, are unused usings

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -288,20 +290,63 @@ public static async Task<StorageAccount> GetStorageAccount(string storageAccount
...
+                            listUrl,
+                            accessToken));
+
+                    if (accounts?.Value == null || !accounts.Value.Any())
+                    {
+                        continue;
+                    }
+
+                    // Filter in-memory by storage account name
+                    var match = accounts?.Value?
```
liliankasem: The redundant accounts?.Value? null-conditional on line 313 is unreachable (the guard on line 307 already returned).

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -288,20 +290,63 @@ public static async Task<StorageAccount> GetStorageAccount(string storageAccount
...
-                            subscriptionId = subscription.SubscriptionId,
-                            resourceName = storageAccountName,
-                            resourceType = "Microsoft.Storage/storageAccounts"
-                        }),
-                        accessToken);
+                try
+                {
+                    // List all storage accounts in the subscription
+                    var listUrl = new Uri(
+                        $"{managementURL}/subscriptions/{subscription.SubscriptionId}" +
```
liliankasem: Raw string interpolation for ARM URLs bypasses the existing ArmUriTemplates.Bind pattern used elsewhere in the file which is a minor inconsistency.

liliankasem (changes requested):

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Added pagination support by following nextLink (same pattern as GetSubscriptions) to ensure all storage account are evaluated.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Updated exception handling to preserve and rethrow the actual error instead of "not found".

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
umangsriv: Removed dead code.

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
umangsriv: Removed unused code.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Added check (i + 1 < parts.Length) to avoid potential index error.

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
umangsriv: Removed unused usings.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Fixed code.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Updated URL construction to use ArmUriTemplates.Bind for consistency with existing pattern.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,25 +286,88 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
+                    if (StaticSettings.IsDebug)
+                    {
+                        ColoredConsole.Error.WriteLine(ErrorColor(ex.ToString()));
+                    }
                 }
             }
 
+            if (lastException != null)
+            {
+                throw lastException;
```
liliankasem: Only rethrow lastException when no subscription completed listing successfully. You can track a bool i.e. `anySubscriptionSucceeded` and otherwise fall through to ArmResourceNotFoundException.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,25 +286,88 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
                         accessToken);
+                    }
+
+                    var match = allAccounts.FirstOrDefault(a =>
+                        string.Equals(a.Name, storageAccountName, StringComparison.OrdinalIgnoreCase));
+
+                    if (match == null)
+                    {
+                        continue;
+                    }
```
liliankasem: The private GetStorageAccount(ArmWrapper<ArmGenericResource>, ...) overload only reads armWrapper.Id and armWrapper.Name to call listKeys. The match from the list already has both, so you can pass match directly and drop:

the StorageAccountByResourceGroup ARM template
the ExtractResourceGroupFromId helper
the extra HTTP round-trip per call

liliankasem (changes requested):
I'd push back on the approach. This fans out across every subscription, pages through all storage accounts, then does a redundant GET before listKeys — and the lastException rethrow regresses the error UX when a sub returns 403 (common when the caller lacks Storage RBAC somewhere).

The repo already has an Azure Resource Graph helper: ArmUriTemplates.ArgUri + AzureHelper.GetResourceIDFromArg. One call resolves the storage account id across all accessible subs, server-side filtered by name, no pagination, no per-sub 403s:

```csharp
var query = $"Resources | where type =~ 'microsoft.storage/storageaccounts' "
          + $"and name =~ '{storageAccountName}' | project id | limit 1";
var resourceId = await GetResourceIDFromArg(subIds, query, accessToken, managementURL);
return await GetStorageAccount(
    new ArmWrapper<ArmGenericResource> { Id = resourceId, Name = storageAccountName },
    accessToken, managementURL);
```

Caveat: ARG has a few seconds of ingestion delay. If we need strict strong consistency for brand-new accounts, use ARG only to resolve subscriptionId + resourceGroup, then call the provider GET.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Apply ARG implementation

umangsriv:
Hi @liliankasem ,
Could you please review this PR when you get a chance? I’d really appreciate your feedback.

src/Cli/func/Arm/ArmUriTemplates.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem: Undo stages to this file?

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,34 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
-                    return await GetStorageAccount(storageAccount.Value.First(), accessToken, managementURL);
-                }
+            // Resolve storage account via Azure Resource Graph (ARG)
+            var query =
+                $"Resources | where type =~ 'microsoft.storage/storageaccounts' " +
+                $"and name =~ '{storageAccountName}' | project id | limit 1";
+
+            var resourceId = await GetResourceIDFromArg(subIds, query, accessToken, managementURL);
+
+            if (string.IsNullOrWhiteSpace(resourceId))
```
liliankasem: This branch is unreachable. GetResourceIDFromArg already throws CliException("Error finding the Azure Resource information.") when ARG returns no rows (line 119), so we never get here.

one solution is to wrap the GetResourceIDFromArg call in try/catch (CliException) and rethrow ArmResourceNotFoundException($"Cannot find storage account with name {storageAccountName}")

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,34 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
+            }
 
-                if (storageAccount.Value.Any())
-                {
-                    return await GetStorageAccount(storageAccount.Value.First(), accessToken, managementURL);
-                }
+            // Resolve storage account via Azure Resource Graph (ARG)
+            var query =
+                $"Resources | where type =~ 'microsoft.storage/storageaccounts' " +
+                $"and name =~ '{storageAccountName}' | project id | limit 1";
```
liliankasem: `storageAccountName` is interpolated into the KQL query without escaping. if input validation ever loosens or this helper gets reused, a single quote would break the query. Worth either validating up front or escaping single quotes.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,34 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
+                throw new InvalidOperationException("No subscriptions found for the current account.");
+            }
 
-                if (storageAccount.Value.Any())
-                {
-                    return await GetStorageAccount(storageAccount.Value.First(), accessToken, managementURL);
-                }
+            // Resolve storage account via Azure Resource Graph (ARG)
+            var query =
+                $"Resources | where type =~ 'microsoft.storage/storageaccounts' " +
```
liliankasem: Nit: this `Resources |` prefix is inconsistent with the existing TryGetFunctionAppFromArg query (line 57) which omits the table name. Both work since Resources is the default table, but pick one style and stick with it across the file - always better to be consistent.

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,34 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
         public static async Task<StorageAccount> GetStorageAccount(string storageAccountName, string accessToken, string managementURL)
         {
             var subscriptions = await GetSubscriptions(accessToken, managementURL);
-            foreach (var subscription in subscriptions)
+            var subIds = subscriptions?.Select(s => s.SubscriptionId).ToList();
+
+            if (subIds == null || subIds.Count == 0)
```
liliankasem:
```suggestion
            if (subIds.Count == 0)
```

GetSubscriptions always returns a non-null List<>

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: The UTF-8 BOM was stripped from this file - can you revert this?

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Azure.Functions.Cli.Arm;
```
liliankasem, line 4: Consider adding a multi-subscription test, since collapsing N per-subscription HTTP calls into a single ARG query is the core motivation of this PR. The current happy-path test only exercises one subscription, which the old code path also handled.

src/Cli/func/Arm/ArmUriTemplates.cs
```diff
@@ -17,7 +17,6 @@ internal static class ArmUriTemplates
         public const string BasicAuthCheckApiVersion = "2022-03-01";
         public const string FunctionsStacksApiVersion = "2020-10-01";
         public const string FlexFunctionsStacksApiVersion = "2020-10-01";
-
         public const string ArgUri = "providers/Microsoft.ResourceGraph/resources";
 
         public static readonly ArmUriTemplate SubscriptionResourceByNameAndType = new ArmUriTemplate($"resources?$filter=(SubscriptionId eq '{{subscriptionId}}' and name eq '{{resourceName}}' and resourceType eq '{{resourceType}}')", ArmApiVersion);
```
liliankasem: This is no longer used - please remove to clean up

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Updated the implementation to catch CliException.

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Fixed query

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Fixed query

src/Cli/func/Helpers/AzureHelper.cs
umangsriv: Removed unnessary logic

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
umangsriv: Test added

src/Cli/func/Arm/ArmUriTemplates.cs
umangsriv: Removed

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,43 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
         public static async Task<StorageAccount> GetStorageAccount(string storageAccountName, string accessToken, string managementURL)
         {
             var subscriptions = await GetSubscriptions(accessToken, managementURL);
-            foreach (var subscription in subscriptions)
+            var subIds = subscriptions?.Select(s => s.SubscriptionId).ToList();
```
liliankasem:
```suggestion
            var subIds = subscriptions.Select(s => s.SubscriptionId).ToList();
```

src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -286,26 +286,43 @@ await ArmHttpAsync<ArmWrapper<ArmWebsitePublishingCredentials>>(
...
+            try
+            {
+                resourceId = await GetResourceIDFromArg(
+                    subIds,
+                    query,
+                    accessToken,
+                    managementURL);
+            }
+            catch (CliException ex) when (
+               ex.Message == "Error finding the Azure Resource information.")
```
liliankasem: String-matching internal exception messages is fragile: any typo fix, message tweak, or future localization in GetResourceIDFromArg silently bypasses this catch and the user sees the generic CliException again, regressing exactly what the previous round fixed.

In GetResourceIDFromArg

```
return argResponse.Data?.Rows?.FirstOrDefault()?.FirstOrDefault()
    ?? throw new ArmResourceNotFoundException("Error finding the Azure Resource information.");
```

Then here you can do

```
string resourceId;
try
{
    resourceId = await GetResourceIDFromArg(subIds, query, accessToken, managementURL);
}
catch (ArmResourceNotFoundException)
{
    throw new ArmResourceNotFoundException(
        $"Cannot find storage account with name {storageAccountName}");
}

```

test/Cli/Func.UnitTests/HelperTests/AzureHelperTests.cs
```diff
@@ -39,5 +40,134 @@ public async Task GetFunctionKeyTest(string key, string value)
...
+                const string accessToken = "token";
+
+                var subscriptionsUrl = $"{managementUrl}/subscriptions?api-version=2018-09-01";
+                var argUrl = $"{managementUrl}/providers/Microsoft.ResourceGraph/resources?api-version=2019-04-01";
+                var resourceId = "/subscriptions/sub2/resourceGroups/rg/providers/Microsoft.Storage/storageAccounts/mystorage";
+                var keysUrl = $"{managementUrl}/subscriptions/sub2/resourceGroups/rg/providers/Microsoft.Storage/storageAccounts/{storageName}/listKeys?api-version=2018-02-01";
+                var mockHttp = new MockHttpMessageHandler();
+
+                // Mock multiple subscriptions
+                mockHttp.When(HttpMethod.Get, subscriptionsUrl).Respond("application/json", @"{'value': [{ 'subscriptionId': 'sub1', 'displayName': 'Sub One' }, { 'subscriptionId': 'sub2', 'displayName': 'Sub Two' },{ 'subscriptionId': 'sub2', 'displayName': 'Sub Three' }]}");
```
liliankasem:
```suggestion
                mockHttp.When(HttpMethod.Get, subscriptionsUrl).Respond("application/json", @"{'value': [{ 'subscriptionId': 'sub1', 'displayName': 'Sub One' }, { 'subscriptionId': 'sub2', 'displayName': 'Sub Two' },{ 'subscriptionId': 'sub3', 'displayName': 'Sub Three' }]}");
```

liliankasem (commented):
Almost there

umangsriv:
> Almost there

Haha 😄 , thanks for the detailed reviews and patience throughout this, sorry for stretching this PR a bit longer than expected but it's definitely a good learning for me, have addressed the latest comments, hopefully this will be the final 🤞.

liliankasem (approved):

liliankasem (approved):
Rebae off latest main and add a release note for this fix:

`- Fix `func azure storage fetch-connection-string` failing with "Cannot find storage account" due to ARM eventual consistency (#4884)`

liliankasem (approved):

PR #4889, opened 2026-04-28
Refactor [path] handling: typed WorkingDirectory threaded through commands and workloads

liliankasem:
### Problem
`BaseCommand.ApplyPath` validated the optional `[path]` argument by mutating the process working directory via `Directory.SetCurrentDirectory`. Commands and workloads then read it back with `Directory.GetCurrentDirectory()` — an implicit hand-off invisible to method signatures, plus a global side effect that forced tests to share a serial xUnit collection (`WorkingDirectoryTests`) to avoid cwd races.

The shared `[path]` argument was also stringly-typed (`Argument<string?>`), pushing path validation onto every caller.

### Approach
- New **`WorkingDirectory`** record (in `Func.Cli.Abstractions`) wrapping `(DirectoryInfo Info, bool WasExplicit, string? OriginalPath)` with `Exists` and `CreateIfNotExists()` helpers. `WasExplicit` lets callers distinguish a user-supplied path from a defaulted cwd; `OriginalPath` preserves the typed string for user-facing error messages.
- New **`PathArgument : Argument<WorkingDirectory>`** — strongly-typed binding so commands receive a real domain object from System.CommandLine. Keeps the dash-prefix rejection in `CustomParser` (so `func start --bogus` errors out as an unrecognized option instead of being treated as a path) and defaults to `WorkingDirectory.FromCwd()`. Created per command instance via `BaseCommand.AddPathArgument()`, matching the instance-options pattern from #4895.
- **`BaseCommand.ApplyPath` is gone.** Commands now read `parseResult.GetValue(PathArgument!)` and thread the `WorkingDirectory` through their own logic. The CLI never calls `Directory.SetCurrentDirectory`.
  - `InitCommand` calls `workingDirectory.CreateIfNotExists()` before dispatching to a workload.
  - `StartCommand` checks `workingDirectory.Exists` and throws `GracefulException` (echoing `OriginalPath`) if missing.
  - `NewCommand` is still a stub; the wiring will land alongside template support.
- **Workloads receive `WorkingDirectory` on `WorkloadContext`** instead of a bare `ProjectPath` string. Workload authors get the same typed handle the CLI uses, plus the `WasExplicit` signal for behaviour that depends on whether the user typed a path.
- **Tests:** `BaseCommandTests` (which depended on cwd mutation) replaced by `PathArgumentTests` (parser behaviour) and `WorkingDirectoryTests` (record behaviour). Added an `InitCommand` test that asserts the `[path]` directory is created before workload dispatch. Deleted the `WorkingDirectoryTests` collection definition; `StartCommandTests` no longer opts into it.
- **Docs:** `cli-architecture.md`, `adding-a-command.md`, and `building-a-workload.md` updated to describe the new `[path]` flow and the CLI's "directory exists before dispatch" guarantee, while keeping #4895's instance-options pattern. `AGENTS.md` adds an XML-doc style rule that the new types follow.


satvu (approved):

src/Func.Cli/Commands/PackCommand.cs
```diff
@@ -39,9 +39,9 @@ public PackCommand(IInteractionService interaction)
 
     protected override Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
     {
-        ApplyPath(parseResult);
+        var location = ResolveProjectLocation(parseResult);
```
aishwaryabh: i like this approach a lot :)

aishwaryabh (approved):

src/Func.Cli/Commands/ProjectArgument.cs
```diff
@@ -0,0 +1,47 @@
...
+/// The shared <c>[path]</c> argument used by every command that operates on
+/// a project directory (<c>func init</c>, <c>func new</c>, <c>func pack</c>,
+/// <c>func start</c>). Strongly typed as <see cref="DirectoryInfo"/> so the
+/// binding layer hands callers a real directory handle instead of a string.
+///
+/// Resolution to an absolute path / existence checks / create-if-missing
+/// behavior live in <see cref="BaseCommand.ResolveProjectLocation"/> — this
+/// type is just the parser-level binding.
+/// </summary>
+internal sealed class ProjectArgument : Argument<DirectoryInfo?>
```
liliankasem: Keeping this here for now but we can move this to abstractions in future PRs when/if it makes sense to. If that is the case, we should promote InitContext.ProjectPath from string to DirectoryInfo (or ProjectLocation) for consistency.

src/Func.Cli/Commands/BaseCommand.cs
```diff
@@ -59,34 +47,35 @@ protected void AddPathArgument()
...
         {
-            return;
+            return new ProjectLocation(new DirectoryInfo(Directory.GetCurrentDirectory()), WasExplicit: false);
         }
 
-        var fullPath = Path.GetFullPath(path);
-        if (!Directory.Exists(fullPath))
+        if (!directory.Exists)
         {
             if (createIfNotExists)
```
jviau: I would move the `.Exists` and create-if-not-exists logic to be on `ProjectLocation` itself, so the consumer can decide when and if to call that. Also have `ProjectArgument` be `Argument<ProjectLocation>`. This way all that logic stays with `ProjectLocation`. The usage becomes:

``` csharp
var location = parseResult.GetValue(projectArgument);
// can check if exists:
if (location.Exists)
{
}

location.CreateIfNotExists(); // create it if doesn't exist.
```

src/Func.Cli/Commands/InitCommand.cs
```diff
@@ -105,7 +105,7 @@ protected override async Task<int> ExecuteAsync(ParseResult parseResult, Cancell
         }
 
         var context = new InitContext(
-            ProjectPath: Directory.GetCurrentDirectory(),
+            ProjectPath: location.Path,
```
jviau: I would move away from stringly-typed when appropriate. With a concrete type here, callers won't need to wonder "does this exist on disk yet?" "is it a relative path, fully qualified?" etc.

src/Func.Cli/Commands/InitCommand.cs
jviau, line 122: Side ask: is there a constants file we can use for any non-ASCII character? Anything not typically on a keyboard that is.

src/Func.Cli/Commands/NewCommand.cs
```diff
@@ -45,7 +45,7 @@ public NewCommand(IInteractionService interaction)
 
     protected override Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
     {
-        ApplyPath(parseResult, createIfNotExists: true);
+        _ = ResolveProjectLocation(parseResult, createIfNotExists: true);
```
jviau: Do we want to bother creating the directory if there are no workloads?

src/Func.Cli/Commands/ProjectLocation.cs
```diff
@@ -0,0 +1,19 @@
...
+/// a project directory explicitly instead of reading <see cref="Directory.GetCurrentDirectory"/>
+/// after a hidden cwd mutation. Always points at an existing directory by
+/// the time the command sees it.
+/// </summary>
+/// <param name="Directory">The project directory.</param>
+/// <param name="WasExplicit">True when the user passed <c>[path]</c> on the command line; false when defaulted to the process cwd.</param>
+internal sealed record ProjectLocation(DirectoryInfo Directory, bool WasExplicit)
+{
+    /// <summary>Convenience accessor for <see cref="DirectoryInfo.FullName"/>.</summary>
+    public string Path => Directory.FullName;
```
jviau: Uncertain if this is needed. Accessing `Directory` doesn't seem like an issue to me. Also for some stacks the project can also be considered a _file_ (dotnet isolated). So is `ProjectLocation.Path` the csproj or the directory for dotnet isolated? Leaving it as `.Directory` removes that ambiguity.

src/Func.Cli/Commands/InitCommand.cs
liliankasem: I think this aligns with my comment here? https://github.com/Azure/azure-functions-core-tools/pull/4889#discussion_r3163418556

But you're suggesting we just do this from the get-go?

src/Func.Cli/Commands/InitCommand.cs
liliankasem: If we continue using this, then yes I think we will need a constatns file or similar. But lets revisit this when we flesh out the actual theme we want for the new cli - going to skip in this pr

src/Func.Cli/Commands/InitCommand.cs
liliankasem: Updated InitContext to use project location type

docs/adding-a-command.md
```diff
@@ -92,24 +94,24 @@ internal class DeployCommand : BaseCommand
...
+- **`PathArgument`**: After calling `AddPathArgument()`, read the resolved value with `parseResult.GetValue(PathArgument)`. Returns a `WorkingDirectory` (a `DirectoryInfo` plus a `WasExplicit` flag) and never mutates the process cwd. Call `workingDirectory.CreateIfNotExists()` if your command should auto-create the directory (e.g. `init`, `new`); otherwise check `workingDirectory.Exists` and throw `GracefulException` yourself.
 - **Never use `Console.Write*`**: Always go through `IInteractionService`
 - **Throw `GracefulException`** for user-facing errors (not `Exception`)
 
 ### Options vs Arguments
 
 ```csharp
 // Options: named, optional by default
-public Option<string> NameOption { get; } = new("--name", "-n")
+public static readonly Option<string> NameOption = new("--name", "-n")
```
jviau: Docs are going back to static. Is that intentional? I _think_ it is okay? I think options and arguments are reusable between commands, they dont hold parser state.

But I also see the `BaseCommand` has `ProjectArgument` as an instance member still. Any chance we can make it consistent?

jviau (approved):

docs/adding-a-command.md
liliankasem: It shouldn't be going back to static! Let me fix that

PR #4890, opened 2026-04-28
Rename Stacks to KnownInstallableWorkloads

liliankasem:
### Problem
`Common/Stacks.cs` was named like a runtime registry of stacks (`Stacks.LanguageMap`), but its actual job is much narrower: it's a hard-coded catalog used **only** to power install hints when no workload is registered. Once a workload loader exists, the runtime list of stacks comes from DI'd `IProjectInitializer` contributions — this static would just go away. The misleading name kept inviting new readers to treat it as the source of truth for what stacks the CLI supports.

It also conflated data (the catalog) with behaviour (`WriteWorkloadInstallHints` writing through `IInteractionService`).

### Approach
- Move `src/Func.Cli/Common/Stacks.cs` → `src/Func.Cli/Workloads/KnownInstallableWorkloads.cs`. Same data, intent-revealing name + doc-comment that flags it as install-hint-only and slated for removal once the loader manifest lands.
- Split rendering into `KnownInstallableWorkloadsRenderer` next to the catalog so they evolve together.
- Update callers (`WorkloadHints`, `NewCommand`).

### Caveat
This PR doesn't delete the static — the loader hasn't landed, so install hints still need a fallback. The rename plus the doc-comment is the deepening; full deletion is a follow-up after the loader PR.

Based off `liliankasem/workloads/abstractions` (PR #4883). Will retarget to `vnext` once that merges.


src/Func.Cli/Workloads/KnownInstallableWorkloads.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Frozen;
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Static catalog of workloads the CLI knows how to suggest installing when no
+/// suitable workload is registered at runtime. This is **not** the registry of
+/// what is currently installed — that comes from DI'd workload contributions.
```
satvu, line 11: nit: switch from acronym "DI'd" to explicit words

satvu (approved):

aishwaryabh (approved):

PR #4892, opened 2026-04-28
Add func start support for Go

harshivcodes:
Implements func start for the Go worker runtime. Builds the user's Go project and launches the Functions host with the Go worker registered under the native runtime identifier

**Build & Launch Flow**
- src/Cli/func/Helpers/GoHelpers.cs:
  - BuildProject() runs go build -o app[.exe] . and surfaces stderr / non-zero exits as CliException
  - AssertBinaryExists() validates the compiled binary for the --no-build path with an actionable error
 - src/Cli/func/Actions/HostActions/StartHostAction.cs: PreRunConditions adds a Go branch — version check, then BuildProject() or AssertBinaryExists() based on --no-build
 - src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs: maps FUNCTIONS_WORKER_RUNTIME=native → WorkerRuntime.Go so the host's runtime identifier resolves correctly inside
 the CLI

**How it works**
 1. func start resolves runtime via FUNCTIONS_WORKER_RUNTIME=native (written by func init)
 2. CLI verifies Go ≥
  1.24, then builds app (or app.exe on Windows)
 3. Host loads workers/native/worker.config.json and launches the binary

### DIFF
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -733,6 +733,7 @@ namespace Azure.Functions.Cli.Actions.HostActions
 
         private async Task PreRunConditions()
         {
+            ResolveNativeWorkerRuntime();
             EnsureWorkerRuntimeIsSet();
 
             if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Python)
@@ -893,6 +894,37 @@ namespace Azure.Functions.Cli.Actions.HostActions
                 ? await SecurityHelpers.GetOrCreateCertificate(CertPath, CertPassword)
                 : null;
             return (new Uri($"{protocol}://0.0.0.0:{Port}"), new Uri($"{protocol}://localhost:{Port}"), cert);
+        }
+
+        internal void ResolveNativeWorkerRuntime()
+        {
+            // 'native' is the host's runtime identifier for any worker registered with
+            // language="native" (today: Go via workers/native/worker.config.json).
+            // Resolve it here by inspecting project markers; add new arms as additional native
+            // workers (e.g. Rust) are onboarded.
+            var setting = Environment.GetEnvironmentVariable(Constants.FunctionsWorkerRuntime)
+                          ?? _secretsManager.GetSecrets().FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
+
+            if (string.IsNullOrWhiteSpace(setting)
+                || !string.Equals(setting, "native", StringComparison.OrdinalIgnoreCase))
+            {
+                return;
+            }
+
+            if (FileSystemHelpers.FileExists(Path.Combine(Environment.CurrentDirectory, "go.mod")))
+            {
+                if (VerboseLogging == true)
+                {
+                    ColoredConsole.WriteLine(VerboseColor("Resolving native worker runtime to 'go' (go.mod found)."));
+                }
+
+                GlobalCoreToolsSettings.CurrentWorkerRuntime = WorkerRuntime.Go;
+                return;
+            }
+
+            throw new CliException(
+                $"Could not determine the worker language for the project at '{Environment.CurrentDirectory}'. " +
+                "Run 'func init' to initialize a supported function app in this directory.");
         }
 
         private void EnsureWorkerRuntimeIsSet()
--- a/src/Cli/func/Actions/HostActions/StartHostAction.cs
+++ b/src/Cli/func/Actions/HostActions/StartHostAction.cs
@@ -750,6 +750,20 @@ namespace Azure.Functions.Cli.Actions.HostActions
             {
                 throw new CliException("Dotnet is required for PowerShell Functions. Please install dotnet (.NET Core SDK) for your system from https://www.microsoft.com/net/download");
             }
+            else if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Go)
+            {
+                var goVersion = await GoHelpers.GetEnvironmentGoVersion();
+                GoHelpers.AssertGoVersion(goVersion);
+
+                if (NoBuild)
+                {
+                    GoHelpers.AssertBinaryExists();
+                }
+                else
+                {
+                    await GoHelpers.BuildProject();
+                }
+            }
 
             if (!NetworkHelpers.IsPortAvailable(Port))
             {
--- a/src/Cli/func/Helpers/GoHelpers.cs
+++ b/src/Cli/func/Helpers/GoHelpers.cs
@@ -9,11 +9,11 @@ using static Azure.Functions.Cli.Common.OutputTheme;
 
 namespace Azure.Functions.Cli.Helpers
 {
-    public static class GoHelpers
+    internal static class GoHelpers
     {
         private const int MinimumGoMajorVersion = 1;
         private const int MinimumGoMinorVersion = 24;
-        public const string GoBinaryName = "app";
+        private const string GoBinaryName = "app";
 
         public static async Task<WorkerLanguageVersionInfo> GetEnvironmentGoVersion()
         {
@@ -95,19 +95,13 @@ namespace Azure.Functions.Cli.Helpers
             ColoredConsole.WriteLine($"Building Go worker binary '{outputName}'...");
 
             var exe = new Executable("go", $"build -o \"{outputName}\" .", workingDirectory: workingDirectory);
-            var stderr = new StringBuilder();
             var exitCode = await exe.RunAsync(
                 l => ColoredConsole.WriteLine(l),
-                e =>
-                {
-                    stderr.AppendLine(e);
-                    ColoredConsole.Error.WriteLine(ErrorColor(e));
-                });
+                e => ColoredConsole.Error.WriteLine(ErrorColor(e)));
 
             if (exitCode != 0)
             {
-                var detail = stderr.Length == 0 ? string.Empty : $" {stderr.ToString().Trim()}";
-                throw new CliException($"Go build failed with exit code {exitCode}.{detail}");
+                throw new CliException($"Go build failed with exit code {exitCode}. See output above for details.");
             }
         }
 
--- a/src/Cli/func/Helpers/GoHelpers.cs
+++ b/src/Cli/func/Helpers/GoHelpers.cs
@@ -13,6 +13,7 @@ namespace Azure.Functions.Cli.Helpers
     {
         private const int MinimumGoMajorVersion = 1;
         private const int MinimumGoMinorVersion = 24;
+        public const string GoBinaryName = "app";
 
         public static async Task<WorkerLanguageVersionInfo> GetEnvironmentGoVersion()
         {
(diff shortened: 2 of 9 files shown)


src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -41,7 +41,7 @@ public static class WorkerRuntimeLanguageHelper
             { WorkerRuntime.Java, new string[] { } },
             { WorkerRuntime.Powershell, new[] { "pwsh" } },
             { WorkerRuntime.Custom, new string[] { } },
-            { WorkerRuntime.Go, new[] { "golang" } }
+            { WorkerRuntime.Go, new[] { "golang", "native" } }
```
liliankasem: `native` is not Go-specific. If the host ever onboards another native worker (Rust, C++, etc.), this mapping silently misroutes them to the Go build path. Consider gating this e.g., only treat native as Go when a go.mod is present in the project, or revisit whether native should be a sibling of Go rather than an alias.

src/Cli/func/StaticResources/nativeWorkerConfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "description": {
+        "language": "native",
+        "defaultExecutablePath": "app",
+        "workerIndexing": "true"
```
liliankasem, line 5: Not 100% sure, but isn't this supposed to be a bool?

```suggestion
        "workerIndexing": true
```

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -81,6 +82,54 @@ public static async Task SetupProject(string moduleName, bool skipGoModTidy)
...
+                e =>
+                {
+                    stderr.AppendLine(e);
+                    ColoredConsole.Error.WriteLine(ErrorColor(e));
+                });
+
+            if (exitCode != 0)
+            {
+                var detail = stderr.Length == 0 ? string.Empty : $" {stderr.ToString().Trim()}";
+                throw new CliException($"Go build failed with exit code {exitCode}.{detail}");
```
liliankasem: Each stderr line is already written in red via ColoredConsole.Error.WriteLine(ErrorColor(e)), then the full buffer is concatenated into the CliException message. Users see the same compiler output twice (once inline, once in the thrown error). Either drop the inline write and let the exception carry it, or trim/summarize what's appended to the message.

src/Cli/func/Helpers/GoHelpers.cs
liliankasem: Forgot to comment in previous pr, but any new class should default to internal and then if it needs to be public for any reason, we can open it

```suggestion
    internal static class GoHelpers
```

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -13,6 +13,7 @@ public static class GoHelpers
     {
         private const int MinimumGoMajorVersion = 1;
         private const int MinimumGoMinorVersion = 24;
+        public const string GoBinaryName = "app";
```
liliankasem: `app` is a common name; if a user already has ./app in their function project for an unrelated reason, it gets clobbered on every func start. Consider something more specific like func-go-app (matching the worker.config.json accordingly).

Also why does this need to be public?

```suggestion
        private const string GoBinaryName = "app";
```

test/Cli/Func.UnitTests/HelperTests/GoHelperTests.cs
```diff
@@ -88,6 +88,99 @@ public void AssertGoVersion_ExtraVersionParts_DoesNotThrow(string version)
...
+                try
+                {
+                    await GoHelpers.BuildProject(dir);
+                }
+                catch (CliException ex)
+                {
+                    caught = ex;
+                }
+
+                caught.Should().NotBeNull("BuildProject should throw when go build fails");
```
liliankasem: This pattern:

```
CliException caught = null;
try { await GoHelpers.BuildProject(dir); }
catch (CliException ex) { caught = ex; }
caught.Should().NotBeNull(...);
```

is inconsistent with the surrounding FluentAssertions style. Replace with:

```
await FluentActions.Invoking(() => GoHelpers.BuildProject(dir))
    .Should().ThrowAsync<CliException>()
    .Where(e => e.Message.Contains("Go build failed"));
```

src/Cli/func/StaticResources/nativeWorkerConfig.json
harshivcodes: checked this for the golang and pythong workers and they are both strings fsr

https://github.com/Azure/azure-functions-golang-worker/blob/main/worker.config.json
https://github.com/Azure/azure-functions-python-worker/blob/62edfe776deff8cc27a0a68af83ebd8e8421bab0/workers/python/prodV4/worker.config.json

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: I originally made it public so the unit tests could reference the constant, but that can be hard coded - i made this private.

harshivcodes: also, for preview we're likely gonna go with the name app, since the image changes are checked in already.

for GA we're likely gonna let customers pick their executable name, if not, I'll surface this feedback to the team.

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
harshivcodes: Good catch! I removed native from the alias list. Instead I am looking for the language marker file if this ever resolves to native.

src/Cli/func/StaticResources/nativeWorkerConfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "description": {
+        "language": "native",
+        "defaultExecutablePath": "app",
```
aishwaryabh, line 4: should the scaffolded funcignore exclude app?

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -750,6 +750,20 @@ private async Task PreRunConditions()
             {
                 throw new CliException("Dotnet is required for PowerShell Functions. Please install dotnet (.NET Core SDK) for your system from https://www.microsoft.com/net/download");
             }
+            else if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Go)
```
aishwaryabh, line 754: we should also add e2e tests for the func start flow as well. Doesn't have to be this PR but this should be flagged

src/Cli/func/StaticResources/nativeWorkerConfig.json
aishwaryabh: not sure if it's needed just wondering

test/Cli/Func.UnitTests/HelperTests/GoHelperTests.cs
```diff
@@ -88,6 +88,91 @@ public void AssertGoVersion_ExtraVersionParts_DoesNotThrow(string version)
...
+        {
+            var dir = Path.Combine(Path.GetTempPath(), "func-go-build-fail-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(dir);
+            try
+            {
+                File.WriteAllText(Path.Combine(dir, "go.mod"), "module example.com/test\n\ngo 1.24\n");
+                File.WriteAllText(Path.Combine(dir, "main.go"), "package main\n\nthis is not valid go\n");
+
+                Func<Task> act = () => GoHelpers.BuildProject(dir);
+                act.Should().Throw<CliException>("BuildProject should throw when go build fails")
```
liliankasem: Should be ThrowAsync

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -196,6 +196,29 @@ public static WorkerRuntime GetCurrentWorkerRuntimeLanguage(ISecretsManager secr
             var setting = Environment.GetEnvironmentVariable(Constants.FunctionsWorkerRuntime)
                           ?? secretsManager.GetSecrets(refreshSecrets)?.FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
 
+            // 'native' is the host's runtime identifier for any worker registered with
+            // language="native" (today: Go via workers/native/worker.config.json). It is not a
+            // language moniker and is intentionally not in _normalizeMap. Resolve it to a concrete
+            // worker runtime by inspecting project markers; add new arms here as additional native
+            // workers (e.g. Rust) are onboarded.
+            if (!string.IsNullOrWhiteSpace(setting)
```
liliankasem: No test coverage for this new path

liliankasem: Can we revert the native resolution block in WorkerRuntimeLanguageHelper.cs (lines 196-218) and move it into StartHostAction.PreRunConditions instead? Two reasons:

1. The CliException thrown here gets silently swallowed by GlobalCoreToolsSettings.cs:128-136, so func start users never see the actionable error.
2. native isn't Go-specific so keeping the "native → Go if go.mod exists" assumption inside the Go-specific code path keeps the shared helper clean.

In PreRunConditions, gate the Go pre-flight on CurrentWorkerRuntime == Go OR (env var is native AND go.mod exists). If native with no go.mod, throw the "Run func init" error there. Proper multi-language native modeling can be a follow-up PR.

liliankasem (commented):
We don't need to handle this in this pr, but as a follow up - I have some thoughts about this native -> language resolution/story.

We should refactor how the CLI handles FUNCTIONS_WORKER_RUNTIME=native. native is the host's identifier for any worker registered with language="native" so it isn't really a worker runtime in the same sense as python or node. It's more of a category.

Options to explore:

1. Sidecar setting in local.settings.json (e.g. FUNCTIONS_WORKER_NATIVE_LANGUAGE=go) written by func init and read by every CLI action that branches on language (when FUNCTIONS_WORKER_RUNTIME is native).

2. Subcommand-style language dispatch for start, similar to how init already uses InitNodeSubcommandAction / InitDotnetSubcommandAction to differentiate languages within a runtime.

3. Marker-file detection as a fallback only (current PR's go.mod check) but driven by an explicit setting first.

The goal being that we keep WorkerRuntimeLanguageHelper a pure mapping (no project sniffing), keep native language-agnostic in the helper, and surface a single source of truth that downstream actions (start, pack, publish) can branch on.

Might be worth a conversation about this. But also we don't need to overinvst here with v5 in the works at the moment.

src/Cli/func/Actions/HostActions/StartHostAction.cs
harshivcodes: Flagging for follow-up — func start for Go runs go build, which fetches modules from the proxy unless they're already cached or vendored. That's the same outbound-network dependency that made Init_WithGo_ScaffoldedProjectCompiles flaky on restricted CI agents. Before adding a func start E2E, we'll need to either vendor deps in the test fixture or pre-warm GOMODCACHE on the agents so the test doesn't hit the proxy at runtime. Tracking separately.

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
harshivcodes: Added 3 new tests after refactoring:

 - ResolveNativeWorkerRuntime_NativeWithGoMod_ResolvesToGo — native + go.mod → resolves to Go
 - ResolveNativeWorkerRuntime_NativeWithoutGoMod_Throws — native without go.mod → throws CliException ("Run 'func init'")
 - ResolveNativeWorkerRuntime_NonNativeRuntime_NoOp — non-native env → no-op

harshivcodes: Fixed

aishwaryabh (approved):

src/Cli/func/StaticResources/nativeWorkerConfig.json
harshivcodes: app can't be ignored since that's the artifact that's deployed to the function app.

liliankasem (approved):
This is fine for now, but before we release the go support preview - we need to have a conversation about go/native support in v5 and make sure the experience in v4 aligns with our future plans to avoid supporting 2 different experiences for native

PR #4893, opened 2026-04-28
Add workload storage repository

### LINKED ISSUE #4896: Workload engine: storage layer (paths + manifest read/write)
## What to build

The pure data layer for the workload engine: the on-disk paths layout, the manifest POCOs, and atomic read/write of `~/.azure-functions/workloads.json`. No commands, no DI wiring — just the substrate every other workload-engine module sits on.

## Acceptance criteria

- [x] `WorkloadPaths` exposes the workload root, install dir, and manifest path; honors `FUNC_CLI_HOME` env var override
- [x] `GlobalManifest`, `GlobalManifestEntry`, and `EntryPointSpec` POCOs round-trip through JSON
- [x] `GlobalManifestStore.Read` returns an empty manifest when the file is missing
- [x] `GlobalManifestStore.Write` is atomic (write-temp + rename) so a crash never produces partial JSON
- [x] Unit tests cover paths layout, env override, roundtrip, and missing-file read

## Status

Implemented in PR #4893 (draft).

## Blocked by

None - can start immediately.

liliankasem:
Closes #4896

First in the workload install/loader stack. Pure data layer — no commands or loader wired yet.

## Storage layer

- **`WorkloadPathsOptions`** — pure config data (`Home` only) with `[Required, MinLength(1)]`. Bound from the `Workloads` configuration section; settable from the `FUNC_CLI_Workloads__Home` environment variable for test isolation.
- **`IWorkloadPaths` / `WorkloadPaths`** — computes `WorkloadsRoot`, `GlobalManifestPath`, and per-package install directories from `WorkloadPathsOptions.Home`. Separated from the options class so consumers inject path computation without touching configuration.
- **`IGlobalManifestStore` / `GlobalManifestStore`** — repository over the global manifest. API: `GetWorkloadsAsync`, `SaveWorkloadAsync` (upsert by `PackageId`, case-insensitive), `RemoveWorkloadAsync`. The on-disk `GlobalManifest` aggregate is not exposed — callers can't race on a shared mutable list.
- **`GlobalManifestEntry` / `EntryPointSpec`** — JSON-serializable rows. Source-generated via `WorkloadJsonContext` (AOT/trim-friendly).
- **Atomic write** — serialize to `<guid>.json.tmp` in the same directory, then `File.Move(..., overwrite: true)`. A failed write leaves the previous manifest byte-identical and removes the temp file.

## Hosting

- **`WorkloadStorageRegistration.AddWorkloadStorage(IServiceCollection, IConfiguration)`** — binds `WorkloadPathsOptions` from the `Workloads` config section with `ValidateDataAnnotations().ValidateOnStart()` and registers `IWorkloadPaths` + `IGlobalManifestStore` as singletons.
- **`Program.cs`** — wires environment-variable configuration (`FUNC_CLI_` prefix) and calls `AddWorkloadStorage` after `AddBuiltInCommands`.

## Configuration

The `FUNC_CLI_` prefix is stripped and `__` maps to section nesting:

| Setting              | Env var                       | Default                  |
|----------------------|-------------------------------|--------------------------|
| `Workloads:Home`     | `FUNC_CLI_Workloads__Home`    | `~/.azure-functions`     |

## Excluded (next PRs)

- Loader (`WorkloadLoader`, `WorkloadLoadContext`, `WorkloadProbe`)
- Install commands (`WorkloadInstaller`, `func workload install/uninstall/list`)
- `NuspecReader` (install-time only)

## Tests

13 new, 63/63 total. No env-var manipulation — every test injects `Options.Create(new WorkloadPathsOptions { Home = tempDir })`. Includes a failure-path test for the atomic-write guarantee (asserts the previous manifest bytes are preserved and no temp file leaks when serialization throws).


src/Func.Cli/Workloads/Storage/GlobalManifestStore.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// <summary>Writes the global manifest, creating the directory if needed.</summary>
+    public static void Write(GlobalManifest manifest)
+    {
+        var path = WorkloadPaths.GlobalManifestPath;
+        Directory.CreateDirectory(Path.GetDirectoryName(path)!);
+
+        using var stream = File.Create(path);
+        JsonSerializer.Serialize(stream, manifest, _options);
+    }
```
fabiocav, line 202: Let's ensure this finishes writing before persisting (i.e., write to a temp file like `<guid>.json.tmp`) then rename the file to the final name once completed (use a File.Move). Perform the write on the same folder to ensure atomicity.

src/Func.Cli/Workloads/Storage/WorkloadPaths.cs
```diff
@@ -0,0 +1,37 @@
...
+
+/// <summary>
+/// Filesystem layout for installed workloads. Honors the <c>FUNC_CLI_HOME</c>
+/// environment variable so tests can redirect installs to a temp folder.
+/// </summary>
+internal static class WorkloadPaths
+{
+    public const string GlobalManifestFileName = "workloads.json";
+
+    /// <summary>Root for everything the func CLI persists for the user (<c>~/.azure-functions</c>).</summary>
```
liliankasem: In v4, this is `~/.azure-functions-core-tools/`, core tools is dropped for v5. Open to other names as well i.e `~/.func/` or `~/.func-cli/` etc.

src/Func.Cli/Workloads/Storage/GlobalManifest.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Azure.Functions.Cli.Workloads.Storage;
+
+/// <summary>
+/// Aggregated manifest at <c>~/.azure-functions/workloads.json</c>. Lists
+/// every workload <c>func workload install</c> has installed. Source of
+/// truth for the loader and <c>func workload list</c>.
+/// </summary>
+internal sealed class GlobalManifest
+{
+    [JsonPropertyName("workloads")]
```
jviau: Instead of individual attributes to accomplish camel casing, how about configuring the serializer to use camel casing? That way it will ensure consistency.

Or maybe if you use JSON serializer source gen, is there a way to tell it to generate with camel case serialization?

src/Func.Cli/Workloads/Storage/WorkloadJsonContext.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Azure.Functions.Cli.Workloads.Storage;
+
+/// <summary>
+/// System.Text.Json source-generated serialization context for the workload
+/// storage shapes. Build-time generation keeps the JSON path reflection-free
+/// (AOT/trim-friendly) and centralizes naming policy so individual properties
+/// don't carry <c>[JsonPropertyName]</c> attributes.
+/// </summary>
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
```
liliankasem, line 14: @jviau is this what you had in mind?

src/Func.Cli/Workloads/Storage/GlobalManifest.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public required string DisplayName { get; init; }
+
+    public string Description { get; init; } = string.Empty;
+
+    public required string Version { get; init; }
+
+    public IReadOnlyList<string> Aliases { get; init; } = Array.Empty<string>();
+
+    /// <summary>Absolute path to the install directory (contains the assembly + workload.json).</summary>
```
jviau: nit: newlines for summary tags. Also if you have xml doc here, do you want to add it to the rest of the file?
```suggestion
    /// <summary>
    /// Absolute path to the install directory (contains the assembly + workload.json).
    /// </summary>
```

src/Func.Cli/Workloads/Storage/GlobalManifestStore.cs
```diff
@@ -0,0 +1,82 @@
...
+        var tempPath = Path.Combine(directory, $"{Guid.NewGuid():N}.json.tmp");
+
+        try
+        {
+            using (var stream = File.Create(tempPath))
+            {
+                JsonSerializer.Serialize(stream, manifest, WorkloadJsonContext.Default.GlobalManifest);
+            }
+
+            File.Move(tempPath, path, overwrite: true);
```
jviau, line 166: Can you add a comment explaining the intermediary temp file. (Avoiding partial write failures corrupting the original file)

src/Func.Cli/Workloads/Storage/WorkloadJsonContext.cs
jviau: Exacty!

src/Func.Cli/Workloads/Storage/WorkloadPaths.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads.Storage;
+
+/// <summary>
+/// Filesystem layout for installed workloads. Honors the <c>FUNC_CLI_HOME</c>
+/// environment variable so tests can redirect installs to a temp folder.
+/// </summary>
+internal static class WorkloadPaths
```
jviau: I wonder if we will want this to be a DI'd service so that it can be easily testable. Otherwise no test which uses this can be ran in parallel as it will need an env variable.

Probably want to use `IOptions` for this.

src/Func.Cli/Workloads/Storage/GlobalManifest.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>Path to the assembly relative to the install directory (e.g. <c>lib/net10.0/Foo.dll</c>).</summary>
+    public required string Assembly { get; init; }
+
+    /// <summary>
+    /// Fully-qualified type name implementing <see cref="IWorkload"/>. Stored as a string so the manifest
+    /// stays loadable even when the assembly isn't on the runtime probe path (e.g. listing workloads
+    /// without loading them).
+    /// </summary>
+    public required string Type { get; init; }
```
jviau, line 81: Do we need this to be explicitly supplied? I wonder if with only the assembly we can load it and just scan for all public implementations of `IWorkload`. It will be less error prone if we can avoid needing a string type ref.

Or we can define an attribute in abstractions that these assemblies use:

``` CSHarp
[assembly: ExportCliWorkload<MyWorkload>()]
```

src/Func.Cli/Hosting/WorkloadStorageRegistration.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public static IServiceCollection AddWorkloadStorage(
+        this IServiceCollection services,
+        IConfiguration configuration)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+        ArgumentNullException.ThrowIfNull(configuration);
+
+        services.AddOptions<WorkloadPathsOptions>()
+            .Bind(configuration.GetSection(ConfigurationSection))
```
jviau: You can avoid needing `IConfiguration` here if you use `BindConfiguration(<Secion>)` -- it will resolve `IConfiguration` from the service provider and then bind it.

src/Func.Cli/Workloads/Storage/GlobalManifestStore.cs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public async Task SaveWorkloadAsync(
+        GlobalManifestEntry entry,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(entry);
+
+        var manifest = await ReadManifestAsync(cancellationToken).ConfigureAwait(false);
+        var existing = manifest.Workloads.FindIndex(
```
jviau: This code is assuming workloads are unique by `PackageId` in the JSON.

Should it be a dictionary / json object with the PackageId as the key instead then? This will guarantee uniqueness instead of an array.

src/Func.Cli/Workloads/Storage/WorkloadPaths.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Azure.Functions.Cli.Workloads.Storage;
+
+/// <summary>
+/// Default <see cref="IWorkloadPaths"/> implementation. Reads
+/// <see cref="WorkloadPathsOptions.Home"/> once at construction time
+/// (CLI config doesn't change mid-run) and computes everything else from it.
+/// </summary>
+internal sealed class WorkloadPaths(IOptions<WorkloadPathsOptions> options) : IWorkloadPaths
```
jviau: Not a fan of the service-wrapping-options approach. All of these values are options in themselves, they just happen to be computed from another option.

I would add all of this to `WorkloadPathsOptions`. You can keep only `Home` as settable, everything else is computed. If you want to keep the interface `IWorkloadPaths` you can, have the options object implement it. Then register it to DI by resolving the options value.


``` csharp
internal sealed class WorkloadPathsOptions : IWorkloadPaths
{
}

services.AddSingleton<IWorkloadPaths>(sp => sp.GetRequiredService<IOptions<WorkloadPathsOptions>>().Value);
```

src/Func.Cli/Workloads/Storage/GlobalManifestStore.cs
liliankasem: I think a basic dictioanry would limit our side-by-side versioning, unless we use a composite key like `packageId@version`, or we do a wrapper i.e. `Dictionary<string, Dictionary<string, GlobalManifestEntry>>`

e.g.

```json
{
  "workloads": {
    "Azure.Functions.Cli.Workload.Dotnet": {
      "1.0.0": { "displayName": "...", "installPath": "...", "entryPoint": {...}, ... },
      "2.0.0": { "displayName": "...", "installPath": "...", "entryPoint": {...}, ... }
    },
    "Azure.Functions.Cli.Workload.Node": {
      "1.0.0": { ... }
    }
  }
}
```

I have a preferance for keeping this a list for simplicity, but the above json looks good to me too - added benefit: it would make the repository API for "list versions" / "remove all versions of X" easier too

src/Func.Cli/Workloads/Storage/GlobalManifest.cs
liliankasem: I think this conversation is for the next pr (loader) - the [assembly: ExportCliWorkload<T>] attribute approach makes sense, but the manifest still needs to persist the resolved type so listing/loading don't have to open assemblies. We can revist this in the workload loader conversaton

jviau (approved):

src/Func.Cli/Workloads/Storage/GlobalManifestStore.cs
jviau: I see this often done with `Id/Version` in JSON, so `Azure.Functions.Cli.Workload.Dotnet/1.0.0` and  `Azure.Functions.Cli.Workload.Dotnet/2.0.0` , but the nested version also looks good to me!

My main complaint is: if a list is supposed to enforce a uniqueness on some key, you want a dictionary not a list.

PR #4895, opened 2026-04-28
Refactor command options/arguments to instance members

liliankasem:
Converts static `Option`/`Argument` fields and helper methods on commands to instance members so each command owns its own argument graph and DI-resolved state.

- `Option`/`Argument` fields on `FuncRootCommand`, `StartCommand`, `InitCommand`, `NewCommand`, `PackCommand`, `HelpCommand` → instance properties
- `BaseCommand.PathArgument` / `ApplyPath` → instance (argument created in `AddPathArgument`)
- `VersionCommand.GetVersion` / `GetInformationalVersion` → instance; `HelpCommand` now takes `VersionCommand` via ctor
- `HelpCommand.FormatOptionName` → instance
- `Parser.cs` and `VersionCommandTests.cs` updated for new shapes
- Static utility classes (`ProjectDetector`, `WorkloadHints`) left as-is

All 53 tests pass.

### DIFF
--- a/docs/adding-a-command.md
+++ b/docs/adding-a-command.md
@@ -31,21 +31,21 @@ using Azure.Functions.Cli.Console;
 
 namespace Azure.Functions.Cli.Commands;
 
-public class DeployCommand : BaseCommand
+internal class DeployCommand : BaseCommand
 {
-    private readonly IInteractionService _interaction;
-
-    // Define options as static readonly fields
-    public static readonly Option<string> AppNameOption = new("--app-name", "-a")
+    // Define options as instance properties
+    public Option<string> AppNameOption { get; } = new("--app-name", "-a")
     {
         Description = "The Azure Functions app name to deploy to",
         Required = true
     };
 
-    public static readonly Option<bool> DryRunOption = new("--dry-run")
+    public Option<bool> DryRunOption { get; } = new("--dry-run")
     {
         Description = "Preview the deployment without making changes"
     };
+
+    private readonly IInteractionService _interaction;
 
     public DeployCommand(IInteractionService interaction)
         : base("deploy", "Deploy the function app to Azure")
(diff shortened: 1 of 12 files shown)


fabiocav (commented):
We need to update the docs on how to "add a command" to align with the changes, right?

liliankasem:
> We need to update the docs on how to "add a command" to align with the changes, right?

Good call, let me add an agent instruction to update docs with each pr

docs/adding-a-command.md
```diff
@@ -31,22 +31,22 @@ using Azure.Functions.Cli.Console;
...
         Required = true
     };
 
-    public static readonly Option<bool> DryRunOption = new("--dry-run")
+    public Option<bool> DryRunOption { get; } = new("--dry-run")
     {
         Description = "Preview the deployment without making changes"
     };
 
+    private readonly IInteractionService _interaction;
```
jviau, line 48: I know this is a doc, but this field should be above the properties.

src/Func.Cli/Commands/BaseCommand.cs
```diff
@@ -18,23 +18,10 @@ namespace Azure.Functions.Cli.Commands;
...
-            var token = result.Tokens.Count > 0 ? result.Tokens[0].Value : null;
-            if (token is not null && token.StartsWith('-'))
-            {
-                result.AddError($"Unrecognized option '{token}'.");
-                return null;
-            }
-            return token;
-        }
-    };
+    protected Argument<string?>? PathArgument { get; private set; }
```
jviau, line 24: Is this only a directory, or can it be a file also? Can we use a stronger type with validation as necessary? `Argument<DirectoryInfo>` if it is a directory. I believe System.CommandLine might have some built in validation helpers for directories also.

jviau: I recommend this overall, move away from stringly-typed when possible and use more concrete types. You can even derive your own arguments if you want, like `class ProjectArgument : Argument<DirectoryInfo>` or something.

liliankasem: I'm going to fold this change into the existing PR touching project path: https://github.com/Azure/azure-functions-core-tools/pull/4889

Then in a follow up, I'll review all the other args/options to see what could benefit from this kind of change as well

docs/adding-a-command.md
liliankasem: @copilot can you move this field about props in this doc

PR #4908, opened 2026-04-30
Add func setup design proposal

liliankasem:
Purpose of this design is to figure out the onboarding story for the new CLI now that a lot of base functionality from v4 is coming from workloads. Given that workloads are essential to the use of the new CLI - we need to make it easy for users to onboard in the CLI, whether that is setting up a new local environment, or CI environment.

### DIFF
--- a/proposed/func-setup-design.md
+++ b/proposed/func-setup-design.md
@@ -1,0 +1,91 @@
+# `func setup` & Onboarding Design (Draft)
+
+This document sketches the proposed `func setup` command and the broader onboarding story for the Azure Functions CLI (`func`). It is a working draft — open questions are called out inline.
+
+## Goals
+
+- Give new users a single command to go from "I installed `func`" to "I can run a function".
+- Let users pick the workloads they care about up front (language stacks like Node / Python / .NET, plus extensions like Durable Functions, Azure-integration tooling, etc.), instead of discovering workloads ad-hoc.
+- Support a non-interactive mode for CI / scripted environments.
+- Centralize first-run concerns: telemetry opt-in, CLI config, baseline workload install.
+
+## Command Surface (proposed)
+
+```
+func setup [--profile <name>] [--non-interactive] [--yes]
+```
+
+| Option              | Description                                                                 |
+| ------------------- | --------------------------------------------------------------------------- |
+| `--profile <name>`  | Apply a named profile (e.g. `node`, `python`, `dotnet`) without prompting.  |
+| `--non-interactive` | Never prompt; fail if a required answer is missing. Intended for CI.        |
+| `--yes`, `-y`       | Accept defaults for any prompt that would otherwise block.                  |
+
+### Interactive flow (default)
+
+1. Welcome + telemetry opt-in.
+2. Pick workloads to install — language stacks (Node, Python, .NET, Java, …) and/or feature workloads (e.g. Durable Functions, Azure-integration tooling).
+3. Confirm and install the corresponding workloads (or workload meta-packages — see below).
+4. Write `.func/config.json` (or update the user-level config) with the chosen profile.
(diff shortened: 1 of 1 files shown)


proposed/func-setup-design.md
```diff
@@ -0,0 +1,161 @@
...
+| ------------------- | ------------------------------------------------------------------------------------------------------------------- |
+| `--workloads <list>`| Comma-separated workload IDs to ensure are installed (e.g. `node,python,durable`).                                  |
+| `--profiles <list>` | Comma-separated profile names whose dependencies should be pre-installed (see [cli-profiles.md](./cli-profiles.md)).|
+| `--non-interactive` | Never prompt. Fail if a required answer is missing or a step would otherwise block.                                 |
+| `--yes`, `-y`       | Accept defaults for any prompt that would otherwise block. Implies non-interactive.                                 |
+| `--check`           | Report what would change. Make no mutations. Exits non-zero if anything is missing.                                 |
+
+### 3.1 Interactive flow (default)
+
+1. If running inside a project directory, infer suggested workloads from project files (read-only).
```
satvu: This is running `func setup` only with no options added? What project files are being read here?

Will this be reusing the "find missing workloads" flow described [here](https://github.com/Azure/azure-functions-core-tools/pull/4923/changes#r3190018800)?

proposed/func-setup-design.md
```diff
@@ -0,0 +1,161 @@
...
+2. Present a grouped picker (language stacks, feature workloads).
```
satvu, line 49: Do we have a default set of workloads that we'll always feature/hard-code into func cli? Otherwise, how is this list created for the customer to pick from?

proposed/func-setup-design.md
liliankasem: Yes to all three:

1. **Trigger.** Step 1 fires on bare `func setup` inside a project directory. Outside a project, there is nothing to infer from, so the picker just shows the catalog with no pre-selection.
2. **Signals.** Same set as the thread you linked: `local.settings.json` `FUNCTIONS_WORKER_RUNTIME`, `host.json`, and project markers (`*.csproj`, `requirements.txt`, `package.json`, `pom.xml`, etc.).
3. **Reuse.** Yes, this should be the **same** detection module as the workloads-spec "find missing workloads" flow (PR #4923, thread r3190018800). One detection module, two callers (setup and the workloads missing-flow). Avoids two slightly-different detectors drifting over time.

I will update §3.1 and §9.6 to cite the shared detection module and link back to PR #4923.

proposed/func-setup-design.md
liliankasem: Two lists, owned by different layers:

1. **Full catalog (everything pickable).** Comes from `func workload search` against the configured NuGet feed (workload-spec §4.1). `setup` does not maintain its own list; it presents the catalog dynamically. So "how is this list created" is: *it isn't*, the workload subsystem produces it.

2. **CLI-curated stacks + recommendations (what gets featured / pre-selected).** This *is* a small hard-coded list inside the CLI. It already has to exist for the workloads-spec "known stacks" install-hint flow (§4.2 line 282). `setup` reuses the same list. Members are the well-known Functions worker runtimes: `node`, `python`, `dotnet-isolated`, `java`, `powershell`, `custom`. A "Python recommendation" is whatever bundle the CLI ships for Python (initially: just the `python` workload; can grow over time).

So the picker has two zones: **Featured / recommended** (CLI-curated, hard-coded) on top, then **Browse all** (catalog-driven, dynamic).

This avoids two failure modes:
- All hard-coded → CLI ships stale lists, cannot surface third-party workloads.
- All catalog → no opinion, no quick path for new users.

The hard-coded list is small and is *already required* by the workloads spec, so this is not new surface, just reuse.

Adding a §3.3 "Catalog and recommendations" section to make this explicit in the proposal.

PR #4912, opened 2026-04-30
Updating command registration models

fabiocav:
This pull request updates the documentation to reflect significant architectural changes in how commands are contributed and registered in the CLI, especially around extensibility for workloads. The changes clarify the distinction between built-in and workload-contributed commands, introduce new abstractions (`FuncCliCommand`, `FuncCommand`), and update all guides and diagrams to match the new model. The documentation now consistently instructs contributors to use the new registration patterns and base classes.

**Key documentation and architectural updates:**

### Built-in Commands and Registration Model

- All built-in commands now extend `FuncCliCommand` instead of `BaseCommand`, and should be registered in `BuiltInCommands.cs` as `FuncCliCommand` instances implementing `IBuiltInCommand`. Registration and checklist steps have been updated throughout the docs to reflect this. [[1]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9L12-R12) [[2]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L12-R23) [[3]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L34-R34) [[4]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L118-R143) [[5]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L218-R217) [[6]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L229-R228) [[7]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L247-R246)

### Workload-Contributed Commands

- Workloads now contribute commands by calling `builder.RegisterCommand(...)` in `IWorkload.Configure`, not by registering raw `Command` instances in DI. The documentation introduces the new `FuncCommand` abstraction for parser-independent command definitions, and explains the new registration overloads and options. [[1]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L48-R49) [[2]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L59-R60) [[3]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L118-R119) [[4]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L195-R246) [[5]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L216-R270) [[6]](diffhunk://#diff-b0dcd1381bda45e1202f8452362a84c05b6c9c876ba98d85a278b9f57cd80448L275-R324) [[7]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L20-R34) [[8]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L223-R245)

### Command Tree Composition and Collision Handling

- The CLI now resolves all `FuncCliCommand` instances from DI, partitions them into built-in and workload-contributed (via `ExternalCommand` wrappers), and handles name collisions by skipping conflicting workload commands and emitting warnings. This is reflected in updated architecture diagrams and explanations. [[1]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L20-R34) [[2]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L63-R75) [[3]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L223-R245)

### Abstractions and Directory Structure

- Documentation now describes new abstractions (`FuncCommand`, `FuncCommandOption`, `FuncCommandInvocationContext`, etc.) and updates the repo structure to include the new `Workloads/` directory for `ExternalCommand` adapters. [[1]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L208-R219) [[2]](diffhunk://#diff-bdd4f9ee6075a30ad6c61495a199188afcb9d9827b37ca7ec09749d02262c17bL42-R50)

### Consistency and Clarity in Command Authoring

- All code samples and guides now use `FuncCliCommand` (for built-ins) and `FuncCommand` (for workloads), and clarify the use of descriptor instances for option/argument parsing. Checklist and quickstart sections have been updated for clarity and accuracy. [[1]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L12-R23) [[2]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L34-R34) [[3]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L118-R143) [[4]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L218-R217) [[5]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L229-R228) [[6]](diffhunk://#diff-6c92b9c800ca63d85253b2dd9b0a117e2ab08aa3a89ab1c1bb4867aa87d62f00L247-R246) [[7]](diffhunk://#diff-2a79134c003d258441e742e0ad07b3ccfc1fce3b55b50af234305a6fdb6ca583L112-R118)

These updates ensure that both internal and external contributors follow the new extensibility model and that the documentation accurately reflects the current CLI architecture.

### Pull request checklist

### DIFF
--- a/src/Func.Cli.Abstractions/Workloads/FuncCommandInvocationContext.cs
+++ b/src/Func.Cli.Abstractions/Workloads/FuncCommandInvocationContext.cs
@@ -1,0 +1,37 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Surface a <see cref="FuncCommand"/> uses at execution time to read parsed option
+/// and argument values. The host supplies a concrete subclass that bridges to the
+/// underlying parser; workload authors only see this abstract type.
+///
+/// Lookups use reference identity on the descriptor — pass the same
+/// <see cref="FuncCommandOption{T}"/> / <see cref="FuncCommandArgument{T}"/> instance
+/// that the command declared in <see cref="FuncCommand.Options"/> /
+/// <see cref="FuncCommand.Arguments"/>. Two distinct descriptors with identical
+/// fields are treated as different keys.
+/// </summary>
+public abstract class FuncCommandInvocationContext
+{
+    /// <summary>Reads the parsed value of a typed option.</summary>
+    /// <returns>
+    /// The parsed value when supplied on the command line, the option's
+    /// <see cref="FuncCommandOption{T}.DefaultValue"/> when not supplied and a default
+    /// was declared, or <c>default(T)</c> otherwise.
+    /// </returns>
+    /// <exception cref="ArgumentException">The option was not declared on the
+    /// command being executed.</exception>
+    public abstract T? GetValue<T>(FuncCommandOption<T> option);
+
+    /// <summary>Reads the parsed value of a typed positional argument.</summary>
+    /// <returns>
+    /// The parsed value when supplied on the command line, or <c>default(T)</c>
+    /// otherwise (including when the argument is optional and was omitted).
+    /// </returns>
+    /// <exception cref="ArgumentException">The argument was not declared on the
+    /// command being executed.</exception>
+    public abstract T? GetValue<T>(FuncCommandArgument<T> argument);
+}
--- a/src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
+++ b/src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
@@ -34,7 +34,7 @@ public abstract class FuncCommandOption
     public string Description { get; }
 
     /// <summary>Type-erased view of the value type carried by this option (used by the parser adapter to construct a typed parser option).</summary>
-    internal abstract Type ValueType { get; }
+    public abstract Type ValueType { get; }
 }
 
 /// <summary>
@@ -76,5 +76,5 @@ public sealed class FuncCommandOption<T> : FuncCommandOption
     /// </summary>
     public T DefaultValue { get; }
 
-    internal override Type ValueType => typeof(T);
+    public override Type ValueType => typeof(T);
 }
--- a/src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
+++ b/src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
@@ -1,0 +1,80 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Parser-independent description of an option accepted by a <see cref="FuncCommand"/>.
+/// Use <see cref="FuncCommandOption{T}"/> to declare options of a specific value type;
+/// instances are kept by reference (identity) and used as the lookup key when reading
+/// parsed values from <see cref="FuncCommandInvocationContext"/>.
+/// </summary>
+public abstract class FuncCommandOption
+{
+    /// <param name="name">Long form (e.g. <c>"--name"</c>). Required.</param>
+    /// <param name="shortName">Optional short form (e.g. <c>"-n"</c>). Pass <c>null</c> for none.</param>
+    /// <param name="description">Help text shown in <c>--help</c> output.</param>
+    private protected FuncCommandOption(string name, string? shortName, string description)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
+        ArgumentNullException.ThrowIfNull(description);
+
+        Name = name;
+        ShortName = shortName;
+        Description = description;
+    }
+
+    /// <summary>Long form (e.g. <c>"--name"</c>).</summary>
+    public string Name { get; }
+
+    /// <summary>Short form (e.g. <c>"-n"</c>), or <c>null</c> if the option has only a long form.</summary>
+    public string? ShortName { get; }
+
+    /// <summary>Help text shown in <c>--help</c> output.</summary>
+    public string Description { get; }
+
+    /// <summary>Type-erased view of the value type carried by this option (used by the parser adapter to construct a typed parser option).</summary>
+    internal abstract Type ValueType { get; }
+}
+
+/// <summary>
+/// A typed <see cref="FuncCommandOption"/>. Construct with one of the two
+/// constructors: omit <c>defaultValue</c> for a non-defaulting option, or pass
+/// it to provide an explicit default. The two-constructor shape distinguishes
+/// "no default" from "default is <c>default(T)</c>" — for value types like
+/// <see langword="int"/>, the unset case is <c>HasDefaultValue == false</c>,
+/// not <c>DefaultValue == 0</c>.
+/// </summary>
+/// <typeparam name="T">Value type read by <see cref="FuncCommandInvocationContext.GetValue{T}(FuncCommandOption{T})"/>.</typeparam>
+public sealed class FuncCommandOption<T> : FuncCommandOption
+{
+    /// <summary>Declares an option without a default value.</summary>
+    public FuncCommandOption(string name, string? shortName, string description)
+        : base(name, shortName, description)
+    {
+        HasDefaultValue = false;
+        DefaultValue = default!;
+    }
+
+    /// <summary>Declares an option with an explicit default value.</summary>
+    public FuncCommandOption(string name, string? shortName, string description, T defaultValue)
+        : base(name, shortName, description)
+    {
+        HasDefaultValue = true;
+        DefaultValue = defaultValue;
+    }
+
+    /// <summary>
+    /// <c>true</c> if a default value was supplied at construction. When <c>false</c>,
+    /// the value of <see cref="DefaultValue"/> is unspecified and should not be read.
+    /// </summary>
+    public bool HasDefaultValue { get; }
+
+    /// <summary>
+    /// The default value supplied at construction. Read only when
+    /// <see cref="HasDefaultValue"/> is <c>true</c>.
+    /// </summary>
+    public T DefaultValue { get; }
+
+    internal override Type ValueType => typeof(T);
+}
--- a/src/Func.Cli.Abstractions/Workloads/FunctionsCliBuilder.cs
+++ b/src/Func.Cli.Abstractions/Workloads/FunctionsCliBuilder.cs
@@ -8,7 +8,8 @@ namespace Azure.Functions.Cli.Workloads;
 /// <summary>
 /// Bootstrap surface passed to <see cref="IWorkload.Configure"/>.
 /// Workloads register their services — project initializers, commands, and
-/// any other supporting types — through <see cref="Services"/>.
+/// any other supporting types — through <see cref="Services"/> and the
+/// <see cref="RegisterCommand(FuncCommand)"/> overloads.
 ///
 /// Modeled after WebJobs' <c>IWebJobsBuilder</c>: the builder exposes the DI
 /// container so workloads can use any standard .NET DI extension method, and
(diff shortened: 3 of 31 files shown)


liliankasem (approved):

src/Func.Cli/Hosting/DefaultFunctionsCliBuilder.cs
```diff
@@ -9,8 +10,69 @@ namespace Azure.Functions.Cli.Hosting;
...
+    protected override void OnRegisterCommand(FuncCommand command)
+    {
+        var workload = RequireWorkload();
+        Services.AddSingleton<FuncCliCommand>(_ => new ExternalCommand(workload, command));
+    }
+
+    protected override void OnRegisterCommand<TCommand>()
+    {
+        var workload = RequireWorkload();
+        Services.AddSingleton<TCommand>();
```
jviau, line 55: I don't think we should add this to DI ourselves. Let the caller decide if/how it is added. We should be calling `ActivatorUtilities.GetServiceOrCreateInstance<TCommand>(sp)`

src/Func.Cli.Abstractions/Workloads/FunctionsCliBuilder.cs
```diff
@@ -22,4 +23,59 @@ public abstract class FunctionsCliBuilder
...
+
+    /// <summary>Hook implemented by the host to wire a <see cref="FuncCommand"/> instance.</summary>
+    protected abstract void OnRegisterCommand(FuncCommand command);
+
+    /// <summary>Hook implemented by the host to wire a <see cref="FuncCommand"/> by type.</summary>
+    protected abstract void OnRegisterCommand<TCommand>()
+        where TCommand : FuncCommand;
+
+    /// <summary>Hook implemented by the host to wire a <see cref="FuncCommand"/> via a factory.</summary>
+    protected abstract void OnRegisterCommand(Func<IServiceProvider, FuncCommand> factory);
```
jviau, line 79: Is it worth having 3 different overloads of this? Or can we get away with just this version? The `RegistryCommand` overloads can all call this:

``` CSharp
public void RegisterCommand(FuncCommand command)
{
    ArgumentNullException.ThrowIfNull(command);
    OnRegisterCommand(_ => command);
}

public void RegisterCommand<TCommand>()
    where TCommand : FuncCommand
{
    if (typeof(TCommand).IsAbstract)
    {
        throw new ArgumentException(
            $"Cannot register abstract command type '{typeof(TCommand)}'. Pass a concrete FuncCommand subclass.",
            nameof(TCommand));
    }

    OnRegisterCommand(s => ActivatorUtilities.GetServiceOrCreateInstance<TCommand>(sp));
}
```

src/Func.Cli.Abstractions/Workloads/FunctionsCliBuilder.cs
```diff
@@ -22,4 +23,59 @@ public abstract class FunctionsCliBuilder
...
     /// <summary>The DI service collection workloads contribute to.</summary>
     public abstract IServiceCollection Services { get; }
+
+    /// <summary>
+    /// Registers a top-level <see cref="FuncCommand"/> instance. The same
+    /// instance is used for every invocation; do not depend on per-invocation
+    /// state being held on the command itself.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="command"/> is <c>null</c>.</exception>
+    public void RegisterCommand(FuncCommand command)
```
jviau, line 33: How about a `public void RegisterCommand(Type command)` overload?

src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Parser-independent description of an option accepted by a <see cref="FuncCommand"/>.
+/// Use <see cref="FuncCommandOption{T}"/> to declare options of a specific value type;
+/// instances are kept by reference (identity) and used as the lookup key when reading
+/// parsed values from <see cref="FuncCommandInvocationContext"/>.
+/// </summary>
+public abstract class FuncCommandOption
```
jviau, line 12: I feel this abstraction is dropping a significant amount of value for workloads compared to System.CommandLine.

Compared with the API surface of `Option`, this is limiting: https://learn.microsoft.com/en-us/dotnet/api/system.commandline.option?view=net-10.0-pp

Help name, arity, aliases (I know you have shortname, but it isn't the same level of aliases), recursive, validators, completion sources.

I still question if having our own abstraction layer is worth the cost of this degraded workload implementation experience.

src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        HasDefaultValue = true;
+        DefaultValue = defaultValue;
+    }
+
+    /// <summary>
+    /// <c>true</c> if a default value was supplied at construction. When <c>false</c>,
+    /// the value of <see cref="DefaultValue"/> is unspecified and should not be read.
+    /// </summary>
+    public bool HasDefaultValue { get; }
```
jviau, line 71: `HasDefaultValue` is on `Command` is SCL, not `Command<T>`  -- why the difference?

src/Func.Cli.Abstractions/Workloads/FuncCommandInvocationContext.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>Reads the parsed value of a typed option.</summary>
+    /// <returns>
+    /// The parsed value when supplied on the command line, the option's
+    /// <see cref="FuncCommandOption{T}.DefaultValue"/> when not supplied and a default
+    /// was declared, or <c>default(T)</c> otherwise.
+    /// </returns>
+    /// <exception cref="ArgumentException">The option was not declared on the
+    /// command being executed.</exception>
+    public abstract T? GetValue<T>(FuncCommandOption<T> option);
```
jviau, line 27: This layer is forcing usage of the generic types. What about non-generic versions? Or types derived from the non-generic version.

I believe it is possible in SCL via `parseResult.GetResult(<non-generic-option>).GetValueOrDefault<T>();`

src/Func.Cli/Workloads/ExternalCommand.cs
```diff
@@ -0,0 +1,264 @@
...
+            Subcommands.Add(new ExternalCommand(Workload, subcommand));
+        }
+    }
+
+    private InvalidOperationException BuildSourceError(string message)
+        => new($"Workload '{Workload.PackageId}' {message}");
+
+    private static Option<T> CreateTypedOption<T>(FuncCommandOption descriptor)
+    {
+        var typed = (FuncCommandOption<T>)descriptor;
```
jviau, line 185: `FuncCommand` abstraction does not mandate using the generic types. This cast will fail if a workload opts to derive their own type from `FuncCommandOption` (which is totally valid to do per `FuncCommand`). Same thing with arguments.

src/Func.Cli/Workloads/ExternalCommand.cs
```diff
@@ -0,0 +1,264 @@
...
+            {
+                throw BuildSourceError(
+                    $"command '{Source.Name}' declared subcommand '{subcommand.Name}' more than once.");
+            }
+
+            Subcommands.Add(new ExternalCommand(Workload, subcommand));
+        }
+    }
+
+    private InvalidOperationException BuildSourceError(string message)
```
jviau, line 180: Should we have a derived exception type? `WorkloadOperationException`

jviau:
I am still uncertain abstracting away SCL is the right move. This is a significant reduction in API flexibility for workloads compared to using SCL directly. This might force workloads to implement these gaps themselves, which might lead to an inconsistent experience. Take validation for example, workloads will now have to perform validation during execute, and they will print errors all sorts of different ways. With SCL, they could use validation directly and give a hopefully more consistent experience.

I like the stronger association between workload and registered commands though.

src/Func.Cli/Hosting/DefaultFunctionsCliBuilder.cs
fabiocav: I debated on this and ultimately left with the registration. We use `GetServiceOrCreateInstance` in other flows, including WebJobs and Worker, so I'll likely align here as well.

src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
fabiocav: It's a tradeoff. We spent a bunch of time evaluating the pros and cons. Will add more details in response to the comment for the PR.

fabiocav:
> I am still uncertain abstracting away SCL is the right move. This is a significant reduction in API flexibility for workloads compared to using SCL directly. This might force workloads to implement these gaps themselves, which might lead to an inconsistent experience. Take validation for example, workloads will now have to perform validation during execute, and they will print errors all sorts of different ways. With SCL, they could use validation directly and give a hopefully more consistent experience.
> 
> I like the stronger association between workload and registered commands though.

 I agree that this abstraction is less expressive than SCL, but this is an intentional tradeoff rather than an accidental gap.
 
The workload-facing API is effectively an installed extension contract. If we expose SCL types there, SCL becomes part of the workload public binary contract, and future SCL changes can invalidate already-installed workloads. Keeping SCL internal gives Core Tools room to absorb parser changes in the adapter layer without forcing coordinated updates across Core Tools, workload packages, and customer installations. No plans at this point, but it also gives us flexibility to change the command line library down the road if needed.
 
I also think the recovery paths are very different. If we start with our own smaller abstraction and later find that it is too limiting, the recovery is additive: we can expand the abstraction, add specific capabilities as workload needs appear, or even introduce a separate advanced registration API that accepts SCL types. If we expose SCL now and later regret it, the recovery is much harder. Existing workloads may already have compiled against those types and customers may already have them installed, so changing or removing that dependency would invalidate existing installations.
 
I’d prefer to keep the public workload command model intentionally basic and stable for now. That still supports the command patterns we expect most workloads to need, and we can grow the surface area deliberately based on real workload requirements rather than mirroring all of SCL up front.
 
For commands that genuinely need richer SCL integration, we still have the option to implement those commands as built-in Core Tools commands (the commands only, not the actual workloads). Most workloads are maintained by our team, so when a command is tightly coupled to the CLI/parser experience, baking that command into the CLI is a reasonable escape valve without making SCL part of the public workload contract.

Worth mentioning... this is the current state. We'll have a preview and time to validate some assumptions, work with the team responsible for SCL and reevaluate all this. 

For the remaining comments, I'll follow up with a PR.

src/Func.Cli/Workloads/ExternalCommand.cs
fabiocav: Those types have `private protected` constructors.

src/Func.Cli.Abstractions/Workloads/FuncCommandOption.cs
fabiocav: There's no real need for that in the base, as workloads can't derive from the base type due to the closed-hierarchy and all usage is from the generic type.

We can move, but that would be cosmetic :)

src/Func.Cli.Abstractions/Workloads/FuncCommandInvocationContext.cs
fabiocav: With the closed hierarchy, this is not currently possible with our abstractions.

PR #4913, opened 2026-05-01
CLI Profiles proposal

fabiocav:
Proposal for profile support in the CLI

### Pull request checklist

### DIFF
--- a/proposed/cli-profiles.md
+++ b/proposed/cli-profiles.md
@@ -1,0 +1,723 @@
+# Design Document: Profile-Based Local Development for Azure Functions CLI
+
+**Status:** Draft
+
+## 1. Problem Statement
+
+Azure Functions supports multiple deployment SKUs — Flex Consumption, Linux Premium, Windows Consumption, Windows Dedicated, and Linux Consumption. Each SKU runs its own version of the Functions runtime, and our desire is to support different deployment cadences.
+
+Today, Azure Functions Core Tools bundles a **single host version**. To prevent "works locally, breaks in cloud" failures when the developer's code depends on host behaviors, APIs, or extension bundle features not yet deployed to their target SKU, Core Tools follows a strict deployment sequence to ensure it never gets ahead of the host versions deployed to the cloud.
+
+### Impact
+
+- **Silent correctness failures**: Code passes local testing but fails in production on an older host version.
+- **Extension bundle drift**: Locally-resolved bundle versions may exceed what the target environment supports, causing runtime errors after deployment.
+- **Side-by-side comparison impossible**: Developers cannot run two environment configurations simultaneously to validate migration or compare behavior.
+
+---
+
+## 2. Key Concept: Profiles
+
+Rather than coupling directly to SKU names, this design introduces **profiles** as the primary abstraction. A profile is a **constraint set** that defines the local development environment: host version bounds, extension bundle bounds, and feature toggles.
+
+Built-in profiles are named after SKUs (e.g., `flex`, `windows-consumption`), but the SKU is metadata _within_ the profile, not its identity. This decoupling enables:
+
+- **Custom profiles**: Teams can define profiles that pin host versions, use preview channels, or enforce organizational constraints — without waiting for a registry update.
+- **Dynamic resolution**: Instead of a static SKU → exact-version lookup, the CLI resolves the best available host version within a profile's version range from installed workloads.
+- **Cross-flow reuse**: The same profile informs `func start` (which host to launch), `func pack` (validate compatibility), and potentially `func deploy` (warn on mismatches).
+- **Composability**: Custom profiles can extend built-in profiles, overriding only what differs.
+
+---
+
+## 3. Goals
+
+1. **Environment alignment**: `func start --profile <name>` launches a host version matching the constraints of the target environment, so local behavior matches cloud behavior at the host-version level.
+2. **Side-by-side execution**: Developers can run two different profiles simultaneously on different ports for comparison testing.
+3. **Extension bundle constraining**: The extension bundle version resolved locally is constrained by the intersection of the project's `host.json` range and the profile's bundle range.
+4. **Automatic resolution**: When a project declares its target profile(s), `func start` uses the correct host version without manual intervention.
+5. **Custom profiles**: Developers and teams can define custom profiles that extend built-in profiles to pin versions, test preview channels, or enforce organizational standards.
+6. **Integrate with Workloads**: Profile resolution builds on the Workloads model (§4.6 of the Workloads Spec) where the host runtime is a separately-versioned workload.
+7. **Offline-capable**: Profile resolution degrades gracefully when network is unavailable, using cached or bundled profile data.
+
+## 4. Non-Goals and Scope Boundary
+
+This design targets **profile-level host version alignment**, not full cloud parity.
+
+### In scope
+- Matching the host runtime version within a profile's constraints
+- Constraining extension bundle versions to what the profile supports
+- Side-by-side host version execution
+- Built-in profile registry distribution and caching
+- Custom profile definition at project and user levels
+- Feature toggles (boolean constraints on host behavior)
+- Cross-flow profile usage (start, pack)
+
+### Out of scope
+- **Region/stamp/ring-level parity**: Cloud rollouts are staged (UD 0 → UD N). Built-in profiles model a SKU as a single version target, not per-region state.
+- **OS-level behavior differences**: A developer on macOS using the `windows-consumption` profile will get the correct host version but not Windows-specific OS behaviors.
+- **SKU-specific scale/networking/billing**: Consumption cold-start behavior, VNET integration, instance limits, etc. are not emulated.
+
+> NOTE: Planned future iterations of CLI will support running on target environment containers, more closely aligning with the cloud environments.
+
+---
+
+## 5. Profile Schema
+
+### 5.1 Built-in Profile Registry
+
+The registry uses a `$schema` URI for self-description and versioning, following the pattern established by Azure Resource Manager templates:
+
+```json
+{
+  "$schema": "https://aka.ms/func-profiles/v1/schema.json",
+  "generatedAt": "2026-04-20T00:00:00Z",
+  "profiles": {
+    "flex": {
+      "sku": "flex-consumption",
+      "status": "GA",
+      "host": {
+        "version": "[4.1048.0, 4.1049.0)"
+      },
+      "extensionBundle": {
+        "version": "[4.0.0, 4.99.1)"
+      },
+      "features": {
+        "proxies": false
+      },
+      "supportedRuntimes": ["node", "python", "java", "powershell", "dotnet-isolated", "custom"],
+      "notes": "Newest host — Flex gets bits first"
+    }
+  }
+}
+```
+
+The `$schema` URI encodes the version (`v1`). When the schema evolves in a breaking way, a new URI is published (`v2`). The CLI checks the URI against its supported set — an unrecognized schema triggers: `"Profile registry requires a newer CLI. Run: func upgrade"`
+
+Unknown fields within a supported schema version are ignored (forward-compatible additive changes do not require a new schema version).
+
+### 5.2 Custom Profile (User or Project Defined)
+
+Custom profiles also use `$schema` for editor support and validation:
+
+```json
+{
+  "$schema": "https://aka.ms/func-custom-profiles/v1/schema.json",
+  "staging": {
+    "extends": "flex",
+    "host": {
+      "version": "[4.1046.100]"
+    }
+  },
+  "locked-prod": {
+    "extends": "windows-consumption",
+    "extensionBundle": {
+      "version": "[4.0.0, 4.22.1)"
+    }
+  }
+}
+```
+
+### 5.3 Field Reference
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `sku` | string | no | Azure SKU identifier. Metadata only — not used for resolution. |
+| `status` | string | no | `GA`, `preview`, `deprecated`. CLI warns on deprecated profiles. |
+| `deprecationUrl` | string | no | URL to deprecation documentation with migration guidance. Used when `status` is `deprecated`. |
+| `extends` | string | no | Name of a parent profile to inherit from. Single inheritance only. |
+| `host.version` | version range | yes (or inherited) | NuGet-style version range for the host runtime. |
+| `extensionBundle.version` | version range | no | Version range constraining extension bundles (applied on top of `host.json`). |
+| `features` | object | no | Boolean feature toggles. See §7. |
+| `supportedRuntimes` | string[] | no | Worker runtimes this profile supports. Enables early validation. |
+| `notes` | string | no | Human-readable description. |
+
+### 5.4 Version Range Expressions
+
+All version constraints use NuGet-style version range syntax:
+
+| Intent | Expression |
+|--------|-----------|
+| Latest within SKU bounds | `[4.1045.0, 4.1049.0)` |
+| Pinned to exact version | `[4.1046.100]` |
+| Any 4.x | `[4.0.0, 5.0.0)` |
+| Preview channel | `[4.1048.100-preview, 4.1049.0)` |
+
+A single field replaces what would otherwise be separate `maxVersion`, `pinnedVersion`, and `channel` properties. The resolution rule is: **highest installed (or available) workload version satisfying the range**.
+
+---
+
+## 6. Profile Sources and Layering
+
+Profiles are resolved from four sources, in precedence order:
+
+```
+1. CLI flag           --profile <name>              (highest priority)
+2. Project config     .func/config.json             (team-shared)
+3. User profiles      ~/.azure-functions/profiles.json
+4. Built-in profiles  Registry (remote → cache → bundled fallback)
+```
+
+### 6.1 Built-in Profiles (Registry)
+
+Built-in profiles are distributed via the **profile registry** — a lightweight JSON artifact published to a well-known CDN endpoint, separate from host workload packages.
+
+**Distribution:**
+
+```
+1. Remote fetch    → https://aka.ms/func-profiles (canonical URL)
+                     https://aka.ms/func-profiles-sha256 (detached checksum)
+2. Local cache     → ~/.azure-functions/profiles/registry.json (1-hour TTL)
+3. Bundled fallback → Shipped with the CLI package (updated each CLI release)
+```
+
+**Publishing:**
+
+The registry is maintained in the same CDN-backed storage that serves extension bundles today. Updates are made by the existing SKU deployment pipelines:
+
+1. SKU deployment pipeline deploys a new host version
+2. Same pipeline updates `registry.json` in CDN-backed storage
+3. CDN serves the updated registry at the well-known URL
+4. **Publish ordering is inherent**: the pipeline that publishes the host binary is the same pipeline that updates the registry, so the host package is always available before the registry references it.
+
+**Integrity:**
+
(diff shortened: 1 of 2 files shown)


proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+- Matching the host runtime version within a profile's constraints
+- Constraining extension bundle versions to what the profile supports
+- Side-by-side host version execution
+- Built-in profile registry distribution and caching
+- Custom profile definition at project and user levels
+- Feature toggles (boolean constraints on host behavior)
+- Cross-flow profile usage (start, pack)
+
+### Out of scope
+- **Region/stamp/ring-level parity**: Cloud rollouts are staged (UD 0 → UD N). Built-in profiles model a SKU as a single version target, not per-region state.
```
Frulfump, line 56: Ideally the host itself would look at the `.func/config.json` to determine if it supports what's there (but then the host would need to be provided with which profile it is you intended to run on the host.)

Frulfump (commented):
Just one thought

proposed/cli-profiles.md
fabiocav: I agree the host should remain the authority on what it supports, but it shouldn't consume `.func/config.json` directly since that file is intended to capture CLI/profile selection rather than app runtime requirements, and profiles are a CLI concern. 
 
For feature enablement, the host already provides feature flags that environments can use to control behavior, and the CLI can translate the selected profile into those host settings when needed.
 
The host also exposes supported capabilities, which the CLI can use for early validation. The missing piece today is a contract that lets the app/project declare its runtime requirements. This design helps pave the way for those requirements to be expressed explicitly, which would also enable earlier validation.

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+
+---
+
+## 12. Profile Listing
+
+### 12.1 Commands
+
+```bash
+func profile list                    # List all available profiles (built-in + user + project)
+func profile list --built-in         # Built-in profiles only
```
liliankasem: Instead of a built-in flag, do we want to consider a source flag e.g. `func profile list --source user|built-in|project` etc.

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+
+When looking up a profile by name, the CLI searches in order:
+
+```
+1. Project-level (.func/profiles.json)
+2. User-level (~/.azure-functions/profiles.json)
+3. Built-in (registry)
+```
+
+First match wins. A project-level profile can shadow a built-in profile name — this is intentional, allowing a project to override the built-in `flex` with a pinned version.
```
liliankasem, line 258: Agent brought up this concern, what are your thoughts? This might also be solved with something as simple a log to say the source of the profile


> Shadowing built-in profile names is a footgun. Two teammates running `func profile show flex` can see different things depending on what's in their `.func/`, with no signal that the name is shadowed. Tooling/docs that mention "the flex profile" become ambiguous.

> Either forbid shadowing (require unique custom names; error links to "extend instead") or require an explicit opt-in field (`shadows: "flex"`) so it's grep-able in a repo and listable in `func profile list`.

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+
+### 6.4 Project Configuration
+
+The project declares which profiles it targets in `.func/config.json`:
+
+```json
+{
+  "$schema": "https://aka.ms/func-config/v1/schema.json",
+  "profiles": ["flex", "staging"],
+  "defaultProfile": "flex"
```
liliankasem, line 234: Should we consider letting defaultProfile point at a profile that isn't in profiles[] e.g. a personal my-preview profile as my local-dev default while the project's deployment targets stay clean?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+      "host": {
+        "version": "[4.1048.0, 4.1049.0)"
+      },
+      "extensionBundle": {
+        "version": "[4.0.0, 4.99.1)"
+      },
+      "features": {
+        "proxies": false
+      },
+      "supportedRuntimes": ["node", "python", "java", "powershell", "dotnet-isolated", "custom"],
```
liliankasem, line 87: What happens if a profile lists `node` as supported but the user hasn't installed the node workload? auto-prompt install, or hard error?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+```
+
+### 15.3 Deprecation Warnings
+
+```
+Warning: Profile "linux-consumption" is deprecated (retirement: 2028-09-30).
+See: https://aka.ms/func-linux-consumption-deprecation
+```
+
+---
```
liliankasem, line 650: Should we land telemetry conventions alongside this design (e.g. `profile.resolved`, `profile.registry.fetched`, `profile.bundle.intersected`) using the existing `CliTelemetry`/`ActivityExtensions`?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+   - Unrecognized `$schema` URI → error: `"Profile registry requires a newer CLI. Run: func upgrade"`
+   - Applies to: registry (`func-profiles`), custom profiles (`func-custom-profiles`), and project config (`func-config`).
+
+6. **~~Feature catalog~~** *(resolved)*
+   - Features are **host configuration directives**, not validation checks. The CLI disables features in the host process at launch time, so local behavior matches the target environment. Absent features default to enabled; `false` actively disables.
+   - v1 set: `proxies`, `inProcess`. New features added by the CLI team as SKU differences are identified.
+   - No warn/block mode — the host simply runs with the target environment's capabilities.
+
+7. **~~Project config file name and location~~** *(resolved)*
+   - `.func/` directory in the project root. Contains `config.json` (project configuration) and `profiles.json` (custom profile definitions). Groups all CLI config together, avoids overloading `host.json` (which is deployed) or `func.json` (which has legacy v1 meaning). Follows the pattern of `.cargo/`, `.nuget/`.
```
liliankasem, line 700: We use `.func/` in project root and `.azure-functions/` in user profile - do we want these folder names to match at all? Or is the different naming easier to understand/differentiate?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+   - **Resolved**: No default profile. When no profile is specified (no `--profile` flag, no `.func/config.json`), the CLI uses the latest installed host with no constraints (backward-compatible) and emits a one-time warning suggesting the developer configure a target profile using the appropriate CLI command. The warning is suppressed once a `.func/config.json` exists (the project has made an explicit choice). This avoids locking in a default SKU that becomes a breaking change to update later.
+
+3. **~~One host workload ID or one per major version?~~**
+   - **Resolved**: One workload ID (`host`) with semver versions. The profile's version range constrains which versions are valid. Major version boundaries are handled naturally by version range expressions (e.g., `[4.0.0, 5.0.0)`).
+
+4. **~~Pre-caching for CI/offline:~~**
+   - **Resolved**: `func profile install <name>` resolves the profile and installs all required dependencies (host workload + extension bundle). `func profile install` (no argument) installs dependencies for all profiles declared in `.func/config.json`. This is the recommended CI step for ensuring offline readiness.
```
liliankasem, line 685: The agent is picking up on this user experience:

>L682 implies `func profile install` orchestrates over `func workload install`, but L548/L549 still surface `func workload install host --version X` as the user-facing recovery command. Should we pick one as the supported entry point and stop showing the other?

I don't think there is anything wrong with exposing both of these entry points to users, but it would be nice to be more explict in the preferred approach we want users to follow, or signpost a specifc way? If profile install is supposed to install missing/required deps, then why are we telling users to install manually?

proposed/cli-profiles.md
fabiocav: I like that!

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+  }
+}
+```
+
+### 5.3 Field Reference
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `sku` | string | no | Azure SKU identifier. Metadata only — not used for resolution. |
+| `status` | string | no | `GA`, `preview`, `deprecated`. CLI warns on deprecated profiles. |
```
jviau: Is `GA` a well known enough term? Would something like `stable` or `latest` be better?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,723 @@
...
+### 6.1 Built-in Profiles (Registry)
+
+Built-in profiles are distributed via the **profile registry** — a lightweight JSON artifact published to a well-known CDN endpoint, separate from host workload packages.
+
+**Distribution:**
+
+```
+1. Remote fetch    → https://aka.ms/func-profiles (canonical URL)
+                     https://aka.ms/func-profiles-sha256 (detached checksum)
+2. Local cache     → ~/.azure-functions/profiles/registry.json (1-hour TTL)
```
jviau, line 169: Can you clarify the 1-hour TTL and the order precedence here. Is the cache only used when remote fetch is not available? If so, what is the purpose of the TTL? Or is the cache preferred, only performed remote fetch once cache expires?

proposed/cli-profiles.md
fabiocav: I don't think so. This needs to align with code that is committed into the repo. If I have a profile defined in my local dev environment, and you don't, adding a reference to that profile here would break your experience (that profile wouldn't be available). 

It's simple enough to pass `--profile myprofile`

proposed/cli-profiles.md
fabiocav: Independent from profiles, the prompt would come when attempting to run that workload that has a dependency on a stack workload that isn't available locally.

With no profiles defined, if we run `func run` for a node app and the node workload isn't installed, we need to provide guidance on how to install that (or prompt for an install at that point). We'd just leverage the same functionality here.

proposed/cli-profiles.md
fabiocav: Probably, but I propose we merge this first and iterate on updates for capture telemetry. Thoughts?

proposed/cli-profiles.md
fabiocav: That's a great question.

Although aligning would be good, I do like `.func` for the local project, but feel that wouldn't be clear for the user profile folder. Thoughts?

liliankasem: Yeah I was thinking the same, ./func being global might be too vague - let's keep them split

proposed/cli-profiles.md
fabiocav: I don't think there's anything wrong with that either. Profiles are also an optional feature, so having the ability to install features independently would be needed.

If profiles are defined, `func profile install` ensures you have all the required dependencies without having to install them individually, but the ability to install workloads individually is also needed for scenarios where profiles are not used or you're installing a workload that isn't defined in a profile.

proposed/cli-profiles.md
liliankasem: I think we should prompt for install - can discuss as we flesh out the rest

proposed/cli-profiles.md
liliankasem: Yeah fair it won't break my experience - but I probably wouldn't want to check in my local dev profile either

proposed/cli-profiles.md
liliankasem: Yeah agree, I think I was just saying we should be explicit in saying profiles will install all required workloads (which we do this part of the spec ``func profile install <name>` resolves the profile and installs all required dependencies`) - but I thought I had seen a contradiction somewhere in the spec where it also says we will tell users to install themselves but I can't seem to find this. Can ignore and resolve :)

proposed/cli-profiles.md
fabiocav: Yeah, I think with the source information, this becomes a less of a problem, but we can have strict enforcement if needed. Let me think more, but I don't think we need to block on this now.

liliankasem (approved):

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
...
+
+```bash
+func profile list                                # List all available profiles (built-in + user + project)
+func profile list --source built-in              # Filter to built-in profiles only
+func profile list --source user                  # Filter to user-level profiles only
+func profile list --source project               # Filter to project-level profiles only
+func profile list --source built-in,project      # Combine sources (comma-separated)
+func profile list --json                         # Machine-readable output
+func profile show flex                           # Detailed view of a specific profile (resolved)
+func profile show staging --raw                  # Show without inheritance resolution
```
kshyju, line 507: Don't we need a command to remove a profile?

Also, we may need a background/auto cleanup mechanism to remove old/unused host versions. Otherwise, this feature may accumulate different host versions and consume customer's disk space.

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
...
+- Custom profile definition at project and user levels
+- Feature toggles (boolean constraints on host behavior)
+- Cross-flow profile usage (start, pack)
+
+### Out of scope
+- **Region/stamp/ring-level parity**: Cloud rollouts are staged (UD 0 → UD N). Built-in profiles model a SKU as a single version target, not per-region state.
+- **OS-level behavior differences**: A developer on macOS using the `windows-consumption` profile will get the correct host version but not Windows-specific OS behaviors.
+- **SKU-specific scale/networking/billing**: Consumption cold-start behavior, VNET integration, instance limits, etc. are not emulated.
+
+> NOTE: Planned future iterations of CLI will support running on target environment containers, more closely aligning with the cloud environments.
```
kshyju, line 60: Shouldn't this be in P0/P1 list of this feature(profiles) instead of future iterations?, running host on the specific OS (same as SKU)

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
+# Design Document: Profile-Based Local Development for Azure Functions CLI
+
+**Status:** Draft
+
+## 1. Problem Statement
+
+Azure Functions supports multiple deployment SKUs — Flex Consumption, Linux Premium, Windows Consumption, Windows Dedicated, and Linux Consumption. Each SKU runs its own version of the Functions runtime, and our desire is to support different deployment cadences.
```
mathewc, line 7: "Each SKU runs its own version" - because of different rollout schedules right, not because different SKUs are pinned to a different version.

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
...
+---
+
+## 3. Goals
+
+1. **Environment alignment**: `func start --profile <name>` launches a host version matching the constraints of the target environment, so local behavior matches cloud behavior at the host-version level.
+2. **Side-by-side execution**: Developers can run two different profiles simultaneously on different ports for comparison testing.
+3. **Extension bundle constraining**: The extension bundle version resolved locally is constrained by the intersection of the project's `host.json` range and the profile's bundle range.
+4. **Automatic resolution**: When a project declares its target profile(s), `func start` uses the correct host version without manual intervention.
+5. **Custom profiles**: Developers and teams can define custom profiles that extend built-in profiles to pin versions, test preview channels, or enforce organizational standards.
+6. **Integrate with Workloads**: Profile resolution builds on the Workloads model (§4.6 of the Workloads Spec) where the host runtime is a separately-versioned workload.
```
mathewc, line 39: Can you add a link to this spec, or is it not public?

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
...
+- Side-by-side host version execution
+- Built-in profile registry distribution and caching
+- Custom profile definition at project and user levels
+- Feature toggles (boolean constraints on host behavior)
+- Cross-flow profile usage (start, pack)
+
+### Out of scope
+- **Region/stamp/ring-level parity**: Cloud rollouts are staged (UD 0 → UD N). Built-in profiles model a SKU as a single version target, not per-region state.
+- **OS-level behavior differences**: A developer on macOS using the `windows-consumption` profile will get the correct host version but not Windows-specific OS behaviors.
+- **SKU-specific scale/networking/billing**: Consumption cold-start behavior, VNET integration, instance limits, etc. are not emulated.
```
mathewc, line 58: Scale/Networking/Billing are never concerns when running locally, right? Wondering if you need to call this out explicitly, since it should be clear to users when running locally that they're running an emulation

proposed/cli-profiles.md
```diff
@@ -0,0 +1,726 @@
...
+- **Extension bundle drift**: Locally-resolved bundle versions may exceed what the target environment supports, causing runtime errors after deployment.
+- **Side-by-side comparison impossible**: Developers cannot run two environment configurations simultaneously to validate migration or compare behavior.
+
+---
+
+## 2. Key Concept: Profiles
+
+Rather than coupling directly to SKU names, this design introduces **profiles** as the primary abstraction. A profile is a **constraint set** that defines the local development environment: host version bounds, extension bundle bounds, and feature toggles.
+
+Built-in profiles are named after SKUs (e.g., `flex`, `windows-consumption`), but the SKU is metadata _within_ the profile, not its identity. This decoupling enables:
```
satvu, line 23: So we're updating each built-in profile as soon as a SKU completes rollout across all regions for a host version? Is that owned by deployment team or core-tools?

And for versions w/ a bug leading to a rollback or hotfix (post rollout completion), what do we do with the version in the profile in that gap between regression -> hotfix release?

proposed/cli-profiles.md
fabiocav: Some of this was pulled from the SKU docs. I think stable would be a better terminology.

proposed/cli-profiles.md
fabiocav: The time here is coming straight from what was captured in the SKU support doc and we can change this around (the 1 hour window here feels a bit aggressive). The behavior is that our cache would be used without performing a fetch if the cache hasn't gone stale.

proposed/cli-profiles.md
fabiocav: > Don't we need a command to remove a profile?

Are you talking about the option to remove a custom profile? The current commands don't expose any authoring experiences. The initial experience to define, remove, modify profiles will be based on documentation. We'll also have our built-in profiles on GitHub for reference (in addition to `show` dumping the details).

> Also, we may need a background/auto cleanup mechanism to remove old/unused host versions. Otherwise, this feature may accumulate different host versions and consume customer's disk space.

I'm not expecting to tie this to profiles, but to workloads instead, as ultimately, this pruning would be required for all workloads.

proposed/cli-profiles.md
fabiocav: Not for the current scope. The goal of this initial scope is to match the runtime expectations of the current Core Tools experience, while providing the ability to install the **Functions** dependencies and the flexibility to run them side-by-side.

Virtualization, running on containers or emulating other managed environment behaviors are not within scope for the initial iteration.

proposed/cli-profiles.md
fabiocav: Correct. Some of the future plans will also lead to more fragmentation.

proposed/cli-profiles.md
fabiocav: The spec is in PR right now here: https://github.com/Azure/azure-functions-core-tools/pull/4923

proposed/cli-profiles.md
fabiocav: Questions came up about networking and scaling behaviors, so I decided to add this here explicitly. I can remove if causing confusion.

proposed/cli-profiles.md
fabiocav: > So we're updating each built-in profile as soon as a SKU completes rollout across all regions for a host version? Is that owned by deployment team or core-tools?

Deployment pipelines would update this manifest.

> And for versions w/ a bug leading to a rollback or hotfix (post rollout completion), what do we do with the version in the profile in that gap between regression -> hotfix release?

The typical flow is a hotfix deployment (roll forward), and the profile would be updated with that version. If a version is indeed rolled back, the profile would be updated with the previous host version as its cap.

PR #4915, opened 2026-05-01
Add workload loader

### LINKED ISSUE #4897: Workload engine: loader (manifest → loaded IWorkload instances)
## What to build

A pure function that takes a `GlobalManifest` and returns hydrated `IWorkload` instances ready to participate in DI. Each workload loads into its own `AssemblyLoadContext` for dependency isolation, with host-shared assemblies (Abstractions, BCL) delegated back to the default context to preserve type identity across the boundary.

## Acceptance criteria

- [x] `WorkloadLoader.LoadInstalledAsync(GlobalManifest, CT)` returns `IReadOnlyList<LoadedWorkload>`
- [x] Each loaded workload runs in its own `AssemblyLoadContext`
- [x] Host-shared assemblies resolve through the default context (so `(IWorkload)instance` cast works)
- [x] Missing assembly throws `GracefulException` naming the workload
- [x] Missing entry-point type throws `GracefulException` naming the workload
- [x] Type that does not implement `IWorkload` throws `GracefulException` naming the workload
- [x] Multiple workloads load with isolated dependency graphs
- [x] Tests use a real fixture workload assembly (not Roslyn or mocks)

## Status

Implemented in PR #4894 (draft).

## Blocked by

- #4896

liliankasem:
Resolves #4897

## Summary

Adds the workload loader: turns persisted manifest entries (from #4893, now in `vnext`) into runnable `IWorkload` instances, each loaded into its own collectible `AssemblyLoadContext` so dependency versions don't collide across workloads.

The discovery/scanner side (`[assembly: ExportCliWorkload<T>]`) is intentionally deferred to a follow-up install-pipeline PR.

## What this PR adds

- **`IWorkloadLoader` / `WorkloadLoader`** — pure: takes `IReadOnlyList<InstalledWorkload>`, returns `IReadOnlyList<LoadedWorkload>`. No I/O on the manifest, no DI surface yet — the composition root reads from `IGlobalManifestStore` and hands entries here.
- **`WorkloadLoadContext`** — per-workload **collectible** `AssemblyLoadContext`. Resolves probe-path dependencies (`AssemblyDependencyResolver`) and **delegates a small allow-list of host assemblies** (`Azure.Functions.Cli.Abstractions`, `Microsoft.Extensions.*`, `System.*`, `Microsoft.Win32.*`, `netstandard`, `mscorlib`) to `Default` so type identity for `IWorkload` and DI primitives is preserved across the boundary.
- **`LoadedWorkload`** — pairs the activated `IWorkload` instance with its `WorkloadInfo`.
- **Test fixture project** (`Func.Cli.Tests.Fixtures.Workload`) — real on-disk assembly the loader tests load through the ALC; not referenced by the test project at compile time.

## Design decisions

- **Sync, pure API.** `LoadInstalled` is synchronous and accepts an in-memory list rather than the store. Reflection/`Activator.CreateInstance` is CPU-bound, and decoupling from the store keeps loader tests free of filesystem fixtures.
- **Collectible ALCs.** `isCollectible: true` enables future unload/hot-reload (e.g. `func workload uninstall` while the host is alive). Cost is small; the option can't be added later without a breaking change to ALC identity.
- **Explicit shared-assembly allow-list** instead of `AssemblyLoadContext.Default.Assemblies.Any(...)`. The latter accidentally shares any assembly the host happens to have loaded; the allow-list makes the contract explicit and reviewable.
- **All-or-nothing loading.** A failure on any entry throws `GracefulException` — no partially-loaded workload set. Errors are prefixed `[<package-id>]` and include `InstallPath` per spec §6.3 / §7.

## Tests

- `LoadInstalled_ReturnsEmpty_WhenNoEntries`
- `LoadInstalled_HydratesEntry_FromAssemblyOnDisk`
- `LoadInstalled_Throws_WhenAssemblyFileMissing` / `WhenTypeNotFoundInAssembly` / `WhenTypeDoesNotImplementIWorkload`
- `LoadInstalled_LoadsEachWorkloadInOwnAssemblyLoadContext`
- `LoadInstalled_LoadContextIsCollectible`
- `LoadInstalled_DelegatesAbstractionsAssemblyToDefaultContext` — pins the type-identity contract; if the ALC ever loaded its own copy of `Azure.Functions.Cli.Abstractions`, the cast in `WorkloadLoader` would fail.

## Out of scope (follow-up PR)

- `[assembly: ExportCliWorkload<T>]` attribute + scanner
- `func workload install / uninstall / list` commands
- Composition root wiring (DI registration, command parser bridge)

## Spec
[`docs/workload-spec.md`](https://github.com/Azure/azure-functions-core-tools/blob/liliankasem/docs/workloads/docs/workload-spec.md)

### DIFF
--- a/src/Func/Workloads/Loading/WorkloadLoadContext.cs
+++ b/src/Func/Workloads/Loading/WorkloadLoadContext.cs
@@ -1,0 +1,87 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace Azure.Functions.Cli.Workloads.Loading;
+
+/// <summary>
+/// Per-workload <see cref="AssemblyLoadContext"/>. Each installed workload
+/// gets its own so dependency versions don't collide across workloads.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Collectible so a future <c>func workload unload</c> / hot-reload can
+/// release the context. Probe path is scoped to the workload's install
+/// directory only — workloads do not see each other's dependencies.
+/// </para>
+/// <para>
+/// Assemblies on <see cref="_sharedAssemblyPrefixes"/> are delegated back to
+/// the default context so type identity is preserved across the host /
+/// workload boundary. Without this, a cast to <see cref="IWorkload"/> would
+/// fail because the host and the workload would each load their own copy
+/// of <c>Azure.Functions.Cli.Abstractions</c> and see two different
+/// <c>IWorkload</c> types. The list is intentionally explicit (rather than
+/// "whatever the default context has loaded") so the boundary is
+/// deterministic regardless of host warm-up state.
+/// </para>
+/// </remarks>
+internal sealed class WorkloadLoadContext : AssemblyLoadContext
+{
+    /// <summary>
+    /// Assembly-name prefixes that must resolve from the default context to
+    /// preserve type identity across the host / workload boundary. Order
+    /// doesn't matter; matching is case-sensitive (CLR convention).
+    /// </summary>
+    private static readonly string[] _sharedAssemblyPrefixes =
+    [
+        "Azure.Functions.Cli.Abstractions",
+        "Microsoft.Extensions.",
+        "System.",
+        "Microsoft.Win32.",
+        "netstandard",
+        "mscorlib",
+    ];
+
+    private readonly AssemblyDependencyResolver _resolver;
+
+    public WorkloadLoadContext(string assemblyPath)
+        : base(name: Path.GetFileNameWithoutExtension(assemblyPath), isCollectible: true)
+    {
+        _resolver = new AssemblyDependencyResolver(assemblyPath);
+    }
+
+    protected override Assembly? Load(AssemblyName assemblyName)
+    {
+        if (IsShared(assemblyName))
+        {
+            // Returning null delegates to the default context, which is what
+            // we want for shared host types so identity matches across the
+            // boundary.
+            return null;
+        }
+
+        var resolved = _resolver.ResolveAssemblyToPath(assemblyName);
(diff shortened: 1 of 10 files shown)


liliankasem:
Before we go too deep into this implementation, we should decide about AOT. This approach is definitly not with that in mind. But we can at this stage pivot.

fabiocav:
Small comment here before review: 

>Collectible ALCs. isCollectible: true enables future unload/hot-reload (e.g. func workload uninstall while the host is alive). Cost is small; the option can't be added later without a breaking change to ALC identity.

This wouldn't have an impact on the host hot reload. This would reload the workload, but the host will be running in a separate process, and collecting the workload ALC wouldn't impact the host context.

src/Func/Workloads/Loading/WorkloadLoadContext.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// doesn't matter; matching is case-sensitive (CLR convention).
+    /// </summary>
+    private static readonly string[] _sharedAssemblyPrefixes =
+    [
+        "Azure.Functions.Cli.Abstractions",
+        "Microsoft.Extensions.",
+        "System.",
+        "Microsoft.Win32.",
+        "netstandard",
+        "mscorlib",
```
fabiocav, line 44: This is a bit broad and what's in TPA won't be required here.

src/Func/Workloads/Loading/WorkloadLoadContext.cs
```diff
@@ -0,0 +1,87 @@
...
+    ];
+
+    private readonly AssemblyDependencyResolver _resolver;
+
+    public WorkloadLoadContext(string assemblyPath)
+        : base(name: Path.GetFileNameWithoutExtension(assemblyPath), isCollectible: true)
```
fabiocav, line 50: We should use the package ID for the name here as different workloads could, in theory, have the same assembly name.

PR #4916, opened 2026-05-01
Add workload entry-point discovery (attribute + scanner)

liliankasem:
## Summary

Adds the install-time discovery primitives that turn a workload package's assemblies into a persistable `EntryPointSpec`, so the load path never has to crack open assemblies just to find the workload entry point.

This is a pure primitive — no DI wiring, no NuGet acquisition, no commands. Those land in follow-up PRs that consume `IWorkloadEntryPointScanner`.

## What this PR adds

- **`ExportCliWorkloadAttribute<T> where T : IWorkload`** (in `Func.Cli.Abstractions`) — assembly-level, single-use, non-inherited. Public contract; workload package authors put `[assembly: ExportCliWorkload<MyWorkload>]` somewhere in their package.
- **`IWorkloadEntryPointScanner` / `WorkloadEntryPointScanner`** — given an install directory, walks top-level `*.dll`s inside a **collectible** `AssemblyLoadContext`, inspects assembly-level custom attributes, and returns the matching `EntryPointSpec(Assembly, Type)`. Errors:
  - directory missing
  - **zero** assemblies declare the attribute
  - **multiple** assemblies declare the attribute (lists offenders)
- **Test fixture projects** — `Func.Cli.Tests.Fixtures.Workload` and `Func.Cli.Tests.Fixtures.Workload.Second`. Real on-disk assemblies the scanner reads via reflection from a temp install dir; not referenced by the test project at compile time.

## Design decisions

- **Collectible scan context, unloaded after each scan.** Install/uninstall cycles must not pin assemblies for the lifetime of the host. The scan context is unloaded inside `Scan(...)`; a regression test forces GC and asserts the ALC was released.
- **Type resolution falls through to `Default`.** `attr.AttributeType.GetGenericTypeDefinition() == typeof(ExportCliWorkloadAttribute<>)` works by reference equality because `Func.Cli.Abstractions` is loaded in the default context.
- **`BadImageFormatException` / `FileLoadException` are silently skipped.** Native binaries, resource-only DLLs, locked files, etc. shouldn't bring down a scan — only the absence of a valid `[assembly: ExportCliWorkload<T>]` is fatal. Covered by a regression test.
- **Generic constraint `where T : IWorkload`** enforces the workload contract at the call site rather than at install time. A package that points the attribute at a non-workload type fails to compile.
- **Errors throw `GracefulException`** with `IsUserError = true` and the install path included, matching the pattern from the loader.

## Tests

- `ExportCliWorkloadAttribute_TargetsAssembly_DisallowsMultiple_AndIsNotInherited`
- `ExportCliWorkloadAttribute_GenericParameter_IsConstrainedToIWorkload`
- `Scan_ReturnsEntryPoint_PointingAtAssemblyAndType`
- `Scan_Throws_WhenNoAssemblyDeclaresAttribute`
- `Scan_Throws_WhenMultipleAssembliesDeclareAttribute`
- `Scan_Throws_WhenInstallDirectoryDoesNotExist`
- `Scan_DoesNotPinAssembly_WhenScanCompletes`
- `Scan_SkipsFiles_ThatAreNotManagedAssemblies`

110/110 total tests pass.

## Out of scope (follow-up PRs)

- NuGet acquisition / `WorkloadInstaller`
- `NuspecReader` (version + tags → aliases)
- `func workload install / uninstall` commands
- Wiring `IWorkloadEntryPointScanner` into the install pipeline
- Composition root / DI registration

## Spec
[`docs/workload-spec.md`](https://github.com/Azure/azure-functions-core-tools/blob/liliankasem/docs/workloads/docs/workload-spec.md) §6.1 (entry-point discovery)

### DIFF
--- a/src/Abstractions/Workloads/ExportCliWorkloadAttribute.cs
+++ b/src/Abstractions/Workloads/ExportCliWorkloadAttribute.cs
@@ -1,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Marks the assembly as a Functions CLI workload and identifies the
+/// <see cref="IWorkload"/> implementation the CLI should activate. Authors
+/// add a single <c>[assembly: ExportCliWorkload&lt;MyWorkload&gt;]</c>
+/// declaration anywhere in their package; the install pipeline scans the
+/// package's lib assemblies for this attribute and persists the result into
+/// the global manifest so subsequent CLI invocations don't have to crack
+/// open the assembly again.
+/// </summary>
+/// <typeparam name="T">
+/// The <see cref="IWorkload"/> implementation to activate. Constrained at
+/// compile time so authors can't accidentally point the CLI at a non-workload
+/// type — surfaces the requirement at the call site rather than at install
+/// time.
+/// </typeparam>
+/// <remarks>
+/// <para>
+/// Targets only the assembly (not types or methods) because workload
+/// activation is a per-package concern, not a per-type concern. A single
+/// package can ship at most one workload entry point — the install scanner
+/// rejects packages that declare more than one occurrence of this attribute.
+/// </para>
+/// <para>
+/// <see cref="AttributeUsageAttribute.Inherited"/> is <c>false</c>: the
+/// scanner walks each assembly explicitly and inheritance is meaningless on
+/// an assembly target.
+/// </para>
+/// </remarks>
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+public sealed class ExportCliWorkloadAttribute<T> : Attribute
+    where T : IWorkload;
--- a/src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
+++ b/src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
@@ -1,0 +1,114 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.Loader;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Workloads.Storage;
+
+namespace Azure.Functions.Cli.Workloads.Discovery;
+
+/// <summary>
+/// Default <see cref="IWorkloadEntryPointScanner"/>. Walks the install
+/// directory's top-level <c>*.dll</c> files inside a collectible
+/// <see cref="AssemblyLoadContext"/> and inspects assembly-level custom
+/// attributes for <see cref="ExportCliWorkloadAttribute{T}"/>.
+/// </summary>
+/// <remarks>
+/// The scan context is collectible and unloaded after the scan completes so
+/// repeat install/uninstall cycles don't pin assemblies for the lifetime of
+/// the host. Type resolution falls through to the default load context,
+/// which is what gives us reference equality for
+/// <see cref="ExportCliWorkloadAttribute{T}"/>.
+/// </remarks>
+internal sealed class WorkloadEntryPointScanner : IWorkloadEntryPointScanner
+{
+    /// <inheritdoc />
+    public EntryPointSpec Scan(string installDirectory)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(installDirectory);
+
+        if (!Directory.Exists(installDirectory))
+        {
+            throw new GracefulException(
+                $"Could not scan workload entry point: directory '{installDirectory}' does not exist.",
+                isUserError: true);
+        }
+
+        var matches = new List<EntryPointSpec>();
+        var ctx = new AssemblyLoadContext("workload-scan", isCollectible: true);
+        try
+        {
+            foreach (var assemblyPath in Directory.EnumerateFiles(installDirectory, "*.dll", SearchOption.TopDirectoryOnly))
+            {
+                var match = TryReadEntryPoint(ctx, assemblyPath, installDirectory);
+                if (match is not null)
+                {
+                    matches.Add(match);
+                }
+            }
+        }
+        finally
+        {
+            ctx.Unload();
+        }
+
+        return matches.Count switch
+        {
+            0 => throw new GracefulException(
+                $"No workload entry point found in '{installDirectory}'. " +
+                $"A workload package must declare exactly one [assembly: ExportCliWorkload<T>].",
+                isUserError: true),
+            1 => matches[0],
+            _ => throw new GracefulException(
+                $"Multiple workload entry points found in '{installDirectory}': " +
+                $"{string.Join(", ", matches.Select(m => m.Assembly))}. " +
+                $"A workload package must declare exactly one [assembly: ExportCliWorkload<T>].",
+                isUserError: true),
+        };
+    }
+
+    private static EntryPointSpec? TryReadEntryPoint(
+        AssemblyLoadContext ctx,
+        string assemblyPath,
+        string installDirectory)
+    {
+        Assembly assembly;
+        try
+        {
+            assembly = ctx.LoadFromAssemblyPath(assemblyPath);
+        }
+        catch (BadImageFormatException)
+        {
+            // Native binaries, resource-only DLLs, or otherwise non-managed
+            // files in the install directory aren't workload candidates.
+            return null;
+        }
+        catch (FileLoadException)
+        {
+            return null;
+        }
+
+        foreach (var attr in assembly.GetCustomAttributesData())
+        {
+            if (!attr.AttributeType.IsGenericType)
+            {
+                continue;
+            }
+
+            if (attr.AttributeType.GetGenericTypeDefinition() != typeof(ExportCliWorkloadAttribute<>))
+            {
+                continue;
+            }
+
+            var workloadType = attr.AttributeType.GetGenericArguments()[0];
+            return new EntryPointSpec
+            {
+                Assembly = Path.GetRelativePath(installDirectory, assemblyPath),
+                Type = workloadType.FullName!,
+            };
+        }
+
+        return null;
+    }
+}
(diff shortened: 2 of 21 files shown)


src/Abstractions/Workloads/ExportCliWorkloadAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+/// rejects packages that declare more than one occurrence of this attribute.
+/// </para>
+/// <para>
+/// <see cref="AttributeUsageAttribute.Inherited"/> is <c>false</c>: the
+/// scanner walks each assembly explicitly and inheritance is meaningless on
+/// an assembly target.
+/// </para>
+/// </remarks>
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+public sealed class ExportCliWorkloadAttribute<T> : Attribute
```
fabiocav: naming: `ExportCliWorkloadAttribute` -> `CliWorkloadAttribute`

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+            // Native binaries, resource-only DLLs, or otherwise non-managed
+            // files in the install directory aren't workload candidates.
+            return null;
+        }
+        catch (FileLoadException)
+        {
+            return null;
+        }
+
+        foreach (var attr in assembly.GetCustomAttributesData())
```
fabiocav: The attribute type here can throw `FileNotFoundException`, `TypeLoadException`, `FileLoadException` if a transitive dep referenced by the attribute type or its generic argument isn't on disk or in the host context, which will lead to an uncaught exception and will prevent further workload discovery.

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        if (!Directory.Exists(installDirectory))
+        {
+            throw new GracefulException(
+                $"Could not scan workload entry point: directory '{installDirectory}' does not exist.",
+                isUserError: true);
+        }
+
+        var matches = new List<EntryPointSpec>();
+        var ctx = new AssemblyLoadContext("workload-scan", isCollectible: true);
```
fabiocav: With this model, probing will rely on all types required during probing to be available in the loaded assembly or in shared assemblies coming from the default context and TPA. If the type used for registration is either defined in a separate assembly or, for any reason, uses an attribute, interface of base type from a separate assembly, probing would fail.

It's reasonable to expect those cases to be unsupported, but we need to make sure we're correctly handling and providing guidance if those cases are found.

fabiocav (changes requested):
Some comments here, but I think we'd benefit from using the `MetadataLoadContext` here  (https://learn.microsoft.com/en-us/dotnet/standard/assembly/inspect-contents-using-metadataloadcontext) as all you need is the ability to inspect the assembly metadata. Combine that with the `PathAssemblyResolver` and that will address most of the issues mentioned and simplify things a bit.

When switching to that model, validate that your generic argument is part of the assembly you've loaded and fail if that isn't the case (making the requirement strict and validated).

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                matches.Add(match);
+            }
+        }
+
+        return matches.Count switch
+        {
+            0 => throw new GracefulException(
+                $"No workload entry point found in '{installDirectory}'. " +
+                $"A workload package must declare exactly one [assembly: CliWorkload<T>].",
```
jviau: Is this message going to be shown to the user? If so, I think it has too much implementation details in it. An average user is not going to know about assembly attributes needed by workloads.

There are several exceptions here that are all `GracefulException` with `isUserError: true`, but they stem from workload authoring mistakes. Do we need a separate class of exceptions for workload authoring issues? Printing the authoring mistakes to the end user isn't very valuable. We probably need some `--verbose` or `--diagnostics` switch which will capture a trace with the error details so they can submit to us for debugging issues like this.

test/Func.Tests.Fixtures.Workload.CrossAssembly/Func.Tests.Fixtures.Workload.CrossAssembly.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <AssemblyName>$(TopLevelNamespace).Tests.Fixtures.Workload.CrossAssembly</AssemblyName>
```
jviau: Let's avoid using `Func` as the start of the project name. `Func` and `Func.Tests` is the only exception for that.

If we rename the csproj we can rely on the default assembly name logic. Something like `Workload.Tests.Fixtures.CrossAssembly`. Then the namespace and assembly becomes `Azure.Functions.Cli.Workload.Tests.Fixtures.CrossAssembly`.

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
```diff
@@ -0,0 +1,156 @@
...
+        if (!Directory.Exists(installDirectory))
+        {
+            throw new GracefulException(
+                $"Could not scan workload entry point: directory '{installDirectory}' does not exist.",
+                isUserError: true);
+        }
+
+        var installAssemblies = Directory.GetFiles(installDirectory, "*.dll", SearchOption.TopDirectoryOnly);
+        var resolverPaths = new List<string>(installAssemblies);
+        resolverPaths.AddRange(Directory.GetFiles(RuntimeEnvironment.GetRuntimeDirectory(), "*.dll"));
```
jviau, line 42: How does `RuntimeEnvironment.GetRuntimeDirectory()` behave when we are self-contained single file?

jviau (changes requested):
I think the assembly resolver needs to be tested with single-file self-contained.

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
liliankasem: Good call out, I think we do need to figure out our error handling story.

Perhaps for now we dumb this down to a user message and add these details in `verbose`? I don't know how a workload authoring error would work - how do we determine a workload author is running the CLI? (Probably some env var or we tell them to depend on --verbose?)

liliankasem: Another suggestion here is to keep the scanner internal so it can throw an exception for 

No [assembly: CliWorkload<T>] found in '/Users/foo/...'. 

Then the install command (when it lands) catches the exception from the scanner and rethrows a new GracefulException with proper user message + VerboseMessage = ex.Message.

Something like:

```text
ERROR: Could not install workload 'Azure.Functions.Cli.Workload.Node':
       package contents are malformed.
HINT:  Try a different version: `func workload install Azure.Functions.Cli.Workload.Node --version <x>`,
       or report the issue to the package author.
       Run with --verbose for technical details.
```

test/Func.Tests.Fixtures.Workload.CrossAssembly/Func.Tests.Fixtures.Workload.CrossAssembly.csproj
liliankasem: Will fix CrossAssembly here and have a follow up PR to fix the rest of the Tests + update agents.md

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
liliankasem: Good question - let me test this out, will make required changes to ensure this works with self-contained single file

liliankasem: Yup fails 

`src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs(48,27): warning IL3000: 'System.Reflection.Assembly.Location.get' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.`

And when I do a mock test run of loading -> `Scan failed: ArgumentException: The path '' is not valid. (Parameter 'assemblyPaths')`

Any ideas on how we should address this? I am getting this suggestion from an agent and I would prefer your opinion on this

##### Suggestion 

Drop MetadataLoadContext entirely. Use the lower-level PEReader + MetadataReader which read directly from a Stream. No resolver, no framework-DLL lookup, no Assembly.Location dependency. Works identically in framework-dependent, self-contained, and single-file.

```C#
using var stream = File.OpenRead(assemblyPath);
using var pe = new PEReader(stream);
var reader = pe.GetMetadataReader();
foreach (var handle in reader.GetAssemblyDefinition().GetCustomAttributes())
{
    var attr = reader.GetCustomAttribute(handle);
    // resolve attr.Constructor parent → check if it's CliWorkloadAttribute`1
    // from Azure.Functions.Cli.Abstractions, extract the generic arg type name
}
```

jviau: 😐 I am definitely not an expert in this space. Maybe @fabiocav knows?

fabiocav: I'll sync with you on this @liliankasem.

src/Func/Workloads/Discovery/WorkloadEntryPointScanner.cs
jviau: When I mean "workload author error", I mean the error originates from the authoring of a workload. IE, it is not a user error, it is not a CLI error, it is a workload issue (it was 'authored' incorrectly).

But what you suggested is good. A general error pointing to a workload issue. Then a hint telling them to run with `--verbose` for more details (which can be shared with the authors). I wouldn't suggest installing a different version, that is guidance I would leave to the workload author to recommend (when the user reports the issue).

Also as for:

> I don't know how a workload authoring error would work - how do we determine a workload author is running the CLI? (Probably some env var or we tell them to depend on --verbose?)

Tell them to use `--verbose` would be the answer

PR #4921, opened 2026-05-01
Add $schema field to workload manifest

liliankasem:
Stamps the on-disk workload manifest with a versioned **JSON Schema URL** rather than an opaque integer version field.

```json
{
  "$schema": "https://aka.ms/func/workloads/v1/schema.json",
  "workloads": { ... }
}
```

## Why `$schema` URL instead of `schemaVersion: 1`

Matches the convention used by `tsconfig.json`, `azure-pipelines.yml`, and `dotnet/global.json`:

- Editors (VS Code) and external JSON Schema validators can fetch the schema document at the URL and validate the manifest without our CLI in the loop.
- Breaking changes ship under a new versioned URL (`/v1/`, `/v2/`, ...) so older and newer CLIs coexist on disk without one silently corrupting the other.
- The artifact is something tooling, package authors, and CI can consume; an integer is meaningful only to our loader code.

## URL shape

The URL uses a slash-namespaced path (`aka.ms/func/<config>/<version>/schema.json`) so sibling CLI configs can share the namespace:

- `https://aka.ms/func/workloads/v1/schema.json` (this PR)
- `https://aka.ms/func/profiles/v1/schema.json` (planned)
- `https://aka.ms/func/config/v1/schema.json` (planned)

Per Fabio's guidance, version lives in exactly one place (the URL path). No parallel `version` field in the document, so there's no drift surface.

## Behavior

- The store writes `$schema` first in the document so it's the first thing on disk after `{`.
- A manifest whose `$schema` URL the CLI doesn't recognize is rejected with a `GracefulException` that lists the supported schemas and suggests checking for typos or updating the CLI. No partial parse.
- A manifest with no `$schema` field (legacy, written before this PR) is treated as v1 and re-emitted with the current URL on the next write.

## Tests

New `WorkloadManifestSchemaTests`:
- `$schema` URL constant matches the v1 URL.
- Save writes `$schema` ahead of `workloads`.
- Load accepts the known URL.
- Load accepts a manifest with no `$schema` field as legacy v1.
- Load rejects an unknown URL with a graceful, actionable error that lists supported schemas.
- Round-trip preserves `$schema`.

## Follow-up

The schema document itself and the `aka.ms` redirect are tracked in #4936 (one issue covering all v5 CLI schemas).

Addresses review feedback: "is there a reason we wouldn't use an actual versioned schema instead?"


src/Func/Workloads/Storage/GlobalManifestStore.cs
```diff
@@ -106,6 +106,15 @@ private async Task<GlobalManifest> ReadManifestAsync(CancellationToken cancellat
                 cancellationToken).ConfigureAwait(false)
                 ?? new GlobalManifest();
 
+            if (!WorkloadManifestSchema.IsSupported(deserialized.Schema))
+            {
+                throw new GracefulException(
+                    $"The workload manifest at '{path}' declares schema '{deserialized.Schema}', "
+                    + $"but this version of the Functions CLI only understands '{WorkloadManifestSchema.CurrentSchema}'. "
+                    + "Please update the CLI to the latest version.",
```
jviau: I think this message could use some refining. What if it's a typo on the schema? No amount of updating will resolve that.

I think if we just have an error of declared schema vs supported schemas. If you want to talk about updating, I think we need to be more upfront that it might not help.


```
The schema '{schema}' declared by manifest '{path}' is not supported.
Supported schemas are:
  - {list_supported_schemas}

Check for spelling or try updating the CLI to the latest version.
```

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public const string CurrentSchema = "https://aka.ms/func-workloads/v1/schema.json";
+
+    /// <summary>
+    /// Returns <c>true</c> if this CLI understands the supplied schema URL.
+    /// <c>null</c> or empty is treated as legacy v1 so manifests written
+    /// before this field was introduced still load.
+    /// </summary>
+    public static bool IsSupported(string? schema)
+        => string.IsNullOrEmpty(schema)
```
jviau: Is an exact URI comparison how schema validation is really done? Huh, I would expect it to be more robust, but not sure exactly _how_.

test/Func.Tests/Workloads/Storage/WorkloadManifestSchemaTests.cs
```diff
@@ -0,0 +1,130 @@
...
+
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Workloads.Storage;
+using Xunit;
+
+namespace Azure.Functions.Cli.Tests.Workloads.Storage;
+
+public class WorkloadManifestSchemaTests : IDisposable
+{
+    private const string V1Schema = "https://aka.ms/func-workloads/v1/schema.json";
```
jviau: Can you not reuse `WorkloadManifestSchema.CurrentSchema`?

Or if you are worried about pinning the schema then update `WorkloadManifestSchema` to be:

``` CSharp
public const string V1Schema = "https://aka.ms/func-workloads/v1/schema.json";

public const string CurrentSchema = V1Schema;
```

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
liliankasem: Honestly, I don't actually know how schema validation is done with the URI so I just looked into it and it turns out almost nobody does runtime validation against it (other than ARM):

- dotnet CLI / global.json: [docs](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json) explicitly say $schema is "reserved for tools that consume JSON schemas" (i.e. editors). The CLI never reads the field.
    - `"$schema": "https://json.schemastore.org/global.json",`

- tsconfig.json, omnisharp.json, azure-pipelines.yml, NuGet configs: same pattern. $schema is purely an editor IntelliSense hint, never compared or enforced at runtime.

So the exact-URI check here is non-standard. The convention seems to be  to leave $schema alone and use a dedicated version / manifestVersion field if you actually need runtime version negotiation.

But @fabiocav was leaning towards version in schema URI so maybe he has some thoughts on how we want to do validation here?

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
```diff
@@ -0,0 +1,50 @@
...
+/// doesn't recognize is rejected at load time with an actionable error.
+/// </remarks>
+internal static class WorkloadManifestSchema
+{
+    /// <summary>
+    /// v1 manifest schema URL. Pinned so tests and external tooling can
+    /// reference a stable identifier that does not move when
+    /// <see cref="CurrentSchema"/> is bumped.
+    /// </summary>
+    public const string V1Schema = "https://aka.ms/func-workloads/v1/schema.json";
```
liliankasem: I think the base URL will be `https://json.schemastore.org/func-workloads-v1.json` or -`1.0` - we can decide on this

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
fabiocav: The principle I'd anchor on is single source of truth for the version. Two patterns satisfy it:

   1. Versioned $schema URI. Version lives in the URL; loader closed-set-matches.
   2. Stable $schema URI + version field in the data, with the schema discriminating via $ref to per-version files

I don't have a strong preference here. There are tools and services using both approaches. Aspire has a versioned schema (though, I think this aligns with @liliankasem's findings and it's only used by tools for editing/completion) , host already sets a precedent with a stable one.

Both avoid drift (only one version source per document). The pattern I want us to avoid is "versioned URI and a parallel  schemaVersion field," which reintroduces the drift surface ($schema: …/v1/… + schemaVersion: 2.0 — which wins?).

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
liliankasem: Do we want to use that or an aka link?

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
liliankasem: I am going to continue with versioned $schema URI - will create work items to track creating those schemas in schema store and to track aka link creation

src/Func/Workloads/Storage/WorkloadManifestSchema.cs
jviau: using a link we ultimately control sounds more reliable.

PR #4923, opened 2026-05-01
Add func workload design proposal

liliankasem:
Proposed spec for the v5 workload extension model. Lives under `proposed/` so it can be iterated on without affecting `vnext`.

## Highlights

- **DI extension model.** Workloads register services (e.g. `IProjectInitializer`, `ITemplateProvider`, `IPackProvider`) into a hosted DI container. Built-in commands consume those services via `IEnumerable<T>`. No per-workload manifest file: workloads are pure NuGet packages with `[assembly: ExportCliWorkload<T>]`.
- **Single global manifest** at `~/.azure-functions/workloads.json` records what's installed; the loader is a single JSON read + N assembly loads, no filesystem walking.
- **`$schema` URL** identifies the manifest format (`https://aka.ms/func-workloads/v1/schema.json`). Editors and validators consume it directly; breaking changes ship under a new versioned URL.
- **`packageType: FuncCliWorkload`** is the discriminator on NuGet, mirroring how `dotnet` finds tools (`packageType=dotnettool`).
- **`func new` (no args)** lists templates from every installed workload that targets the active stack (resolved via `local.settings.json` or `.func/config.json`).
- **`--stack`** replaces `--worker-runtime` as the user-facing language selector.

## Updates from earlier rounds

- Removed per-workload manifest file (DI replaces it).
- Switched manifest version field from `schemaVersion: <int>` to `$schema: <url>`.
- Documented `packageType` filter convention.
- Renamed `--worker-runtime` to `--stack` throughout.
- Clarified `func new` template scoping by active stack.

## Base branch

Targets `docs` so the proposal can be iterated on independently of code branches. Once accepted, content moves into the main spec under `docs/`.

### DIFF
--- a/proposed/workload-spec.md
+++ b/proposed/workload-spec.md
@@ -1,0 +1,364 @@
+# Workload Spec
+
+> **Status:** Draft. This spec defines the contract between the Func CLI
+> and workload packages. v1 uses an **in-process** model: workloads are
+> .NET assemblies loaded into the `func` process inside isolated
+> collectible `AssemblyLoadContext`s.
+
+## 1. Goals
+
+- Allow the v5 `func` CLI to be extended with language- and feature-
+  specific functionality without rebuilding or shipping a new CLI.
+- Keep the CLI binary small and stable; defer language-specific work to
+  workloads acquired on demand.
+- Provide a uniform UX across all languages: a user runs the same
+  `func init`, `func new`, `func pack` regardless of which workload backs
+  the project.
+- Allow the Functions team and external authors to ship workloads on their
+  own cadence, independent of the Func CLI's release schedule.
+- Make workload acquisition, update, and removal a first-class CLI
+  experience (`func workload …`).
+
+## 2. Non-goals (v1)
+
+- **Sandboxing.** Workloads run with the same OS privileges as `func`. We
+  rely on package provenance (signed NuGets) for trust, not runtime
+  containment.
+- **Cross-workload coordination.** Workloads do not call each other.
+- **Authoring telemetry framework.** Workloads may emit logs/diagnostics;
+  shared telemetry primitives are out of scope for v1.
+- **GUI / IDE integration.** This spec covers the CLI only. IDE
+  integrations consume `func` as a black box.
+- **Workload-supplied UI prompts.** v1 keeps interactive prompts on the
+  CLI side; the workload returns data, the CLI renders.
+
+## 3. Definitions
+
+| Term                  | Meaning                                                                 |
+|-----------------------|-------------------------------------------------------------------------|
+| **Func CLI**          | The `func` binary itself. Owns command parsing, UX, workload lifecycle. Distinct from the Functions host runtime, which is itself delivered as a workload (see §4.6). |
+| **Host runtime**      | `Microsoft.Azure.WebJobs.Script.WebHost` — the process that actually executes Functions. Acquired and versioned as a workload, launched by `func start`. |
+| **Workload**          | An installable extension contributing init/new/pack support and/or commands. |
+| **Workload package**  | The distributable unit. NuGet in v1; format may evolve.                 |
+| **Global manifest**   | The single CLI-owned JSON file at `~/.azure-functions/workloads.json` recording which workload versions are installed and how to load them. |
+| **Worker runtime**    | The Functions runtime language identifier (`dotnet`, `node`, `python`, `java`, `powershell`, `custom`). |
+| **Contribution point** | An interface in `Func.Cli.Abstractions` (e.g. `IProjectInitializer`) that a workload registers to participate in a built-in command. |
+| **Project marker**    | A glob/file pattern (`*.csproj`, `package.json`, …) that hints which workload owns a given directory. |
+
+## 4. User-facing requirements
+
+### 4.1 `func workload` — package management
+
+| Command                                | Behavior                                                          |
+|----------------------------------------|-------------------------------------------------------------------|
+| `func workload list`                   | List installed workloads with id, version, capabilities, install path. |
+| `func workload search [<query>]`       | Search the configured workload catalog (NuGet feed). Returns id, latest version, description. |
+| `func workload install <id> [--version <v>] [--from <path>]` | Acquire and install a workload. `--from` installs from a local path/feed for development. |
+| `func workload uninstall <id> [--all-versions]` | Remove an installed workload. By default removes only the active version. |
+| `func workload update [<id>] [--all]` | Update one or all workloads to the latest compatible version.     |
+
+**Required behaviors:**
+
+- All commands must be **non-interactive** when given complete arguments and
+  must exit with a non-zero code on failure.
+- `install`, `uninstall`, `update` must be **idempotent**: re-running with
+  the same effective state must succeed without error.
+- A failed install must leave no partial state on disk (atomic move into
+  the install root, or rollback).
+- Concurrent invocations must not corrupt the workload store. (Cross-
+  process locking on the install root.)
+
+### 4.2 `func init` — scaffold a project
+
+- If no workload is installed for the requested runtime, the Func CLI **must**
+  print an actionable hint with the exact command to install one (e.g.
+  `func workload install python`) and exit with a clear error.
+- If a workload is installed, the Func CLI delegates project scaffolding to it.
+- If `--worker-runtime` is omitted, the Func CLI **may** prompt or auto-select
+  based on installed workloads. Behavior must be consistent across
+  workloads.
+
+### 4.3 `func new` — create a function from a template
+
+- Templates come **only** from installed workloads. The Func CLI has no
+  built-in templates.
+- `func new` (no args) lists templates from every installed workload.
+- `func new --template <name> --name <fn>` materializes the template
+  identified by name. If multiple workloads expose the same template name,
+  the Func CLI disambiguates by `--worker-runtime` or via an error listing the
+  conflicts.
+
+### 4.4 `func pack` — build & package for deployment
+
+- The Func CLI inspects the current directory and asks each candidate workload
+  (filtered by project markers, then by `project.detect`) whether it owns
+  the directory.
+- Exactly one workload must claim the directory. Zero → error with install
+  hint. More than one → error listing the contenders, suggest
+  `--worker-runtime` to disambiguate.
+- Pack output paths and naming are workload-defined; the Func CLI surfaces the
+  resulting artifact path back to the user.
+
+### 4.5 `func start` and other built-in commands
+
+- `func start` launches the Functions **host runtime**. The Func CLI **must
+  not** embed the host runtime in its own binary; the runtime is acquired
+  and updated as a workload (see §4.6) so it can ship and version
+  independently of the CLI.
+- In v1, `func start` resolves the host-runtime workload, ensures it is
+  installed (offering to install if missing), and delegates process
+  startup to it. Argument forwarding, port selection, and log streaming
+  remain CLI responsibilities.
+- Workloads **may** register additional command trees (e.g., `func durable
+  …`). The Func CLI must surface these in `--help` output alongside built-in
+  commands and label the source workload in verbose help.
+
+### 4.6 Host runtime as a workload
+
+- The Functions host runtime is distributed as a first-class workload
+  (e.g. `Azure.Functions.Cli.Workload.Host`). It is **not** bundled
+  with the CLI binary.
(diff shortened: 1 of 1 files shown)


proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+## 4. User-facing requirements
+
+### 4.1 `func workload` — package management
+
+| Command                                | Behavior                                                          |
+|----------------------------------------|-------------------------------------------------------------------|
+| `func workload list`                   | List installed workloads with id, version, capabilities, install path. |
+| `func workload search [<query>]`       | Search the configured workload catalog (NuGet feed). Returns id, latest version, description. |
+| `func workload install <id> [--version <v>] [--from <path>]` | Acquire and install a workload. `--from` installs from a local path/feed for development. |
```
jviau: I have seen `--source` more (instead of `--from`) for declaring where a package comes from.

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+| `func workload uninstall <id> [--all-versions]` | Remove an installed workload. By default removes only the active version. |
```
jviau: Do we need a `--version` option also specify an exact version to remove.

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+| **Contribution point** | An interface in `Func.Cli.Abstractions` (e.g. `IProjectInitializer`) that a workload registers to participate in a built-in command. |
+| **Project marker**    | A glob/file pattern (`*.csproj`, `package.json`, …) that hints which workload owns a given directory. |
+
+## 4. User-facing requirements
+
+### 4.1 `func workload` — package management
+
+| Command                                | Behavior                                                          |
+|----------------------------------------|-------------------------------------------------------------------|
+| `func workload list`                   | List installed workloads with id, version, capabilities, install path. |
```
jviau: A command to check if a specific workload (with optional version) is installed would be useful for desired-state scripts.

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+- A failed install must leave no partial state on disk (atomic move into
+  the install root, or rollback).
+- Concurrent invocations must not corrupt the workload store. (Cross-
+  process locking on the install root.)
+
+### 4.2 `func init` — scaffold a project
+
+- If no workload is installed for the requested runtime, the Func CLI **must**
+  print an actionable hint with the exact command to install one (e.g.
+  `func workload install python`) and exit with a clear error.
```
jviau, line 256: How is the CLI going to know what workload the user is missing?

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+  the CLI itself be smaller and AOT-friendly.
+
+## 5. Workload contract — what a workload provides
+
+A workload is a NuGet package containing a .NET assembly that implements
+`IWorkload`. The assembly declares its entry point with a single
+assembly-level attribute:
+
+```csharp
+[assembly: ExportCliWorkload<MyWorkload>]
```
jviau:
```suggestion
[assembly: CliWorkload<MyWorkload>]
```

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+Workload packages do **not** ship a per-package manifest file. The
+metadata the CLI needs is split between two sources:
+
+1. **NuGet metadata (`.nuspec`)** — owned by the package author, captured
+   at install time:
+    - `id` → package id (NuGet rules apply, case-insensitive).
+    - `version` → semver, the workload version.
+    - `title` → display name shown in `func workload list`.
+    - `description` → one-line summary shown in `func workload list`.
+    - `tags` → space-separated short aliases (e.g. `dotnet csharp`)
```
jviau: We need a spec of the exact tags we will use for determining the alias, we can't use all the tags. Would it be a tag like `alias:python`?

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+those services via `IEnumerable<T>` collections and dispatches based on
+which workload contributed which service. There is **no static
+capability list** — what a workload can do is whatever it has registered.
+
+### 5.1 Contribution points (DI)
+
+A workload **may** register any subset of these services. Each service
+the workload registers means it participates in the corresponding command:
+
+| Service registered                | The workload can…                                          | Used by                          |
```
jviau: There are not many details on these interfaces. `building-a-workload.md` is a bit stale also. I have a feeling the contract(s) are going to need to be revisited, particularly the relationship between `IProjectDetector` and all the other contracts. The CLI will need some mapping for many of the commands to find what workload owns it (init, pack, new, start).

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+|-----------------------------------|-----------------------------------------------------------|----------------------------------|
+| `IProjectInitializer`             | Scaffold a new project for a given worker runtime.        | `func init`                      |
+| `IProjectDetector`                | Decide whether a directory is "its" project.              | `func pack` directory routing    |
```
jviau: I think more than pack will need `IProjectDetector`.

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+- Workloads use **semver**. `func workload update` uses the same major
+  version by default; major bumps require an explicit opt-in.
+- The CLI declares a **minimum supported global-manifest schema version**
+  (see §10) and rejects manifests it cannot parse.
+
+## 6. Lifecycle
+
+### 6.1 Install
+
+1. Resolve `<id>` to a package via the catalog (NuGet by default),
```
jviau: Install will need a tag alias-based flow as well, so that a user can run `func workload install python` and we will find `Azure.Functions.Cli.Workload.Python` and install it. Same with my other comment, we would have a tag alias convention (`alias:python`) and install it. If multiple matched that, then we will need an explicit package ID.

proposed/workload-spec.md
jviau: Install will also need a `--exact/-e` option if we add tag-alias support. `--exact` will not perform any alias matching, and install based purely on package id.

IE: package is `Azure.Functions.Cli.Workload.Python`, with tag `alias:python`.

`func workload install python` will succeed.
`func workload install -e python` will fail, must use `func workload install -e Azure.Functions.Cli.Workload.Python`.

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+- Errors from a workload **must** surface to the user with the workload id
+  prefixed (e.g. `[python] error: ...`).
+- The Func CLI **must** distinguish between workload-protocol errors (bug, ask
+  user to file an issue) and user-facing errors (display verbatim).
+
+### 6.4 Update
+
+- `func workload update <id>` resolves a newer version, installs it
+  side-by-side, then atomically swaps the "active" version pointer.
+- Old versions remain on disk until `--prune` is passed, allowing rollback.
```
jviau: This is the first time `--prune` is mentioned. Should the doc be declaring all options for the relevant commands upfront?

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+- Other installed workloads must be unaffected.
+
+### 6.6 Background staleness check
+
+- The Func CLI **may** periodically check the catalog for newer versions of
+  installed workloads and surface a one-line hint (`A new version of
+  python (1.2.3) is available. Run 'func workload update python' to
+  upgrade.`).
+- Frequency: at most once per 24h. Network failures must be silent.
+- Disabled by `FUNCTIONS_DISABLE_WORKLOAD_UPDATE_CHECK=1`.
```
jviau: Use the common env prefix.

`FUNC_CLI_` I think? Or what is it?

proposed/workload-spec.md
```diff
@@ -0,0 +1,390 @@
...
+  run `func workload update`".
+
+## 11. Telemetry expectations
+
+- The Func CLI emits anonymous telemetry per command invocation including:
+  `command`, `workload-id` (when one was selected), `outcome`
+  (success / user-error / workload-error / cli-error), `duration`.
+- Workloads **may** emit their own telemetry but **must not** propagate
+  user-identifying data through the Func CLI.
+- Telemetry is opt-out via the existing `FUNCTIONS_CORE_TOOLS_TELEMETRY_OPTOUT`
```
jviau: `FUNC_CLI`_ prefix as well here. If `FUNCTIONS_CORE_TOOLS_TELEMETRY_OPTOUT` is from the existing core-tools then lets support both the legacy and new value for back-compat.

jviau:
As I review more and more error handling, I think we need a spec for a diagnostics/verbose mode. Where we will collect errors/traces into a shareable log file users can attach to github issues.

proposed/workload-spec.md
liliankasem: That's a good question - I am not sure that this is going to be possible for every command, but I think we can do some basic project detection to say "looks like this might be a python function app, install the python workload". I just want us to be able to provide clear signposting for the basic commands to help onboard folks to the new cli

proposed/workload-spec.md
liliankasem: Yeah it's `FUNC_CLI_` at the moment - will update to use that

proposed/workload-spec.md
liliankasem: Yeah let me make sure it's documented with all the proposed options

proposed/workload-spec.md
liliankasem: Good call out, we should document all planned tags here. For alias, I think `alias:python` makes sense. But I recall we discussed having a `func-workload` tag as well for easier discoverability of workloads in search

proposed/workload-spec.md
liliankasem: Do we want this to be a new command, or can we just make this a property of list?

Option A: New dedicated subcommand

`func workload check <id> [--version <v>]`

> Exit 0 if the workload (and optional version) is installed; non-zero otherwise. Prints nothing by default; `--verbose` prints resolved version.

Option B: Flag on list

`func workload list <id> [--version <v>] --exists   # exit code only`

proposed/workload-spec.md
liliankasem: We can use:

- An explicit --stack <name> flag, if provided.
- FUNCTIONS_WORKER_RUNTIME from local.settings.json, if present.
- Lightweight project-marker detection (e.g. presence of *.csproj, requirements.txt, package.json) to make a best-effort guess.

I am not sure about other workloads, but I at least want to ensure the base important ones are signposted.

Another idea I have here is that we can have some list or legacy map of previous commands from v4 that are now workloads, so if someone tries to run `func azure` or `func kubernetes` we should say "this is now a workload, please install _"

liliankasem:
> As I review more and more error handling, I think we need a spec for a diagnostics/verbose mode. Where we will collect errors/traces into a shareable log file users can attach to github issues.

Agree - would you be interested in throwing a quick spec for that together?

proposed/workload-spec.md
```diff
@@ -0,0 +1,574 @@
...
+1. **NuGet metadata (`.nuspec`)** — owned by the package author, captured
+   at install time:
+    - `id` → package id (NuGet rules apply, case-insensitive).
+    - `version` → semver, the workload version.
+    - `title` → display name shown in `func workload list`.
+    - `description` → one-line summary shown in `func workload list`.
+    - `tags` → NuGet tags. The CLI gives meaning to two reserved tag
+      conventions; all other tags are treated as free-form metadata
+      for search ranking only:
+        - `func-workload` → **required** for discoverability. The
```
liliankasem: search on type instead

proposed/workload-spec.md
```diff
@@ -0,0 +1,663 @@
...
+- `--all`
+  - Updates every installed workload. Mutually exclusive with `<package>`.
+- `--major`
+  - Allows crossing a major-version boundary. Default: same major
+    version only, to protect against breaking changes.
+- `--prune`
+  - Deletes superseded versions from disk after a successful swap.
+    Default: keep prior versions to allow rollback (see §6.4).
+- `--source <path>`
+  - Resolves updates from an alternate feed instead of the configured
```
satvu, line 208: What are the alternate feeds? Are they owned by third parties (are we allowing third-party integration here)? 

Follow up if so: Would 3rd party feeds need a warning as a not-officially supported feed?

fabiocav:
Merging this in the current state. We'll continue to iterate, but it will be helpful to have all documentation checked in so we can start reviewing everything together to identify discrepancies.

PR #4925, opened 2026-05-01
Read installed workloads from manifest in func workload list

### LINKED ISSUE #4900: Workload engine: `func workload list` reads from real manifest
## What to build

The existing `func workload list` command currently returns a hard-coded set of `KnownInstallableWorkloads`. Rewire it to read the real `GlobalManifest` so it reflects what the user has actually installed.

## Acceptance criteria

- [ ] `func workload list` reads from `GlobalManifestStore`
- [ ] Output shows package id and version for each installed workload
- [ ] Empty manifest produces a clear "no workloads installed" message
- [ ] Existing `KnownInstallableWorkloads` list either removed or repurposed (e.g. for a `--available` discovery view — decide during implementation)
- [ ] Unit tests cover empty, single, and multi-entry manifests

## Blocked by

- #4897

liliankasem:
Resolves #4900

## Summary

Wires `func workload list` to actually read the global workload manifest. Until now the command resolved an `IReadOnlyList<WorkloadInfo>` registered as `Array.Empty<WorkloadInfo>()`, so it always printed "No workloads installed." regardless of what was on disk.

## What changes

- **`WorkloadListCommand`** now depends on `IGlobalManifestStore` (already in DI from `AddWorkloadStorage`) and projects each `InstalledWorkload` entry into a table row at execute time. No assembly loads, no `AssemblyLoadContext` creation: listing works entirely from the manifest JSON.
- **`WorkloadRegistration.RegisterWorkloads`** drops the empty `IReadOnlyList<WorkloadInfo>` stub and is now a no-op until the loader lands ([#4915](https://github.com/Azure/azure-functions-core-tools/pull/4915)).
- **`Func.csproj`** exposes internals to `DynamicProxyGenAssembly2` so NSubstitute (Castle DynamicProxy) can mock the internal `IGlobalManifestStore` from tests.
- **`TestParser`** registers a stub `IGlobalManifestStore` in place of the old empty-list registration so existing parser tests still build.

## Drive-by

Replaced the em-dash placeholder for empty aliases with `-`.

## Tests

`WorkloadListCommandTests` (4 cases): empty manifest, single entry, missing aliases, multiple entries. Full suite: 88/88 passing.

## Out of scope

Install / uninstall commands and `NuspecReader` land in the next slice. The manifest writers are already in `IGlobalManifestStore`; the next PR adds the install pipeline that populates them.

### DIFF
--- a/src/Func/Commands/Workload/WorkloadListCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadListCommand.cs
@@ -3,39 +3,43 @@
 
 using System.CommandLine;
 using Azure.Functions.Cli.Console;
-using Azure.Functions.Cli.Workloads;
+using Azure.Functions.Cli.Workloads.Storage;
 
 namespace Azure.Functions.Cli.Commands.Workload;
 
 /// <summary>
-/// Lists workloads recorded in the global manifest. Source of truth is
-/// <see cref="WorkloadInfo"/>, populated by the install / discovery layer
-/// from each package's <c>workload.json</c>.
+/// Lists workloads recorded in the global manifest. Reads directly from
+/// <see cref="IGlobalManifestStore"/> so listing works without invoking the
+/// workload loader (no assembly loads, no ALC creation).
 /// </summary>
-internal sealed class WorkloadListCommand(IInteractionService interaction, IReadOnlyList<WorkloadInfo> workloads)
+internal sealed class WorkloadListCommand(IInteractionService interaction, IGlobalManifestStore store)
     : FuncCliCommand("list", "List installed workloads.")
 {
-    private readonly IInteractionService _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
-    private readonly IReadOnlyList<WorkloadInfo> _workloads = workloads ?? throw new ArgumentNullException(nameof(workloads));
+    private const string AliasesPlaceholder = "-";
 
-    protected override Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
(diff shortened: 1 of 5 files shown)


src/Func/Commands/Workload/WorkloadListCommand.cs
```diff
@@ -3,39 +3,43 @@
...
     : FuncCliCommand("list", "List installed workloads.")
 {
+    private const string AliasesPlaceholder = "-";
+
     private readonly IInteractionService _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
-    private readonly IReadOnlyList<WorkloadInfo> _workloads = workloads ?? throw new ArgumentNullException(nameof(workloads));
+    private readonly IGlobalManifestStore _store = store ?? throw new ArgumentNullException(nameof(store));
 
-    protected override Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
```
jviau, line 23: Is `int` return even needed? I feel like we either exit with success or throw an exception. Are there any cases we would use a non-0 return?

jviau (approved):

src/Func/Commands/Workload/WorkloadListCommand.cs
liliankasem: The signature comes from the FuncCliCommand base class (protected virtual Task<int> ExecuteAsync), which all commands override, so we can't drop the int here without changing the base. You're right that this command's flow only ever returns 0 or throws; non-zero exit codes are produced by GracefulException handling in Program.cs rather than by command return values.

PR #4926, opened 2026-05-01
Add func workload install/uninstall commands

liliankasem:
Adds `func workload install` and `func workload uninstall`. Read-side (`func workload list`) already shipped on `vnext` (#4925, #4937). Stacked on the workload abstraction refactor in #4939.

## `func workload install <id>`

```
func workload install <id> [options]

Arguments:
  <id>   Workload to install. Currently must be a path to a .nupkg on disk.
```

- Nupkg-on-disk only for now. Resolving an id against a NuGet feed is a follow-up.
- Validates the package's nuspec carries the `FuncCliWorkload` package type.
- Extracts to `~/.azure-functions/workloads/<id>/<version>/` and records a registry entry.
- Aliases come from nuspec tags prefixed `alias:` (e.g. `alias:dotnet alias:dotnet-isolated`).

## `func workload uninstall <id> [-v <version>] [-a]`

```
func workload uninstall <id> [options]

Arguments:
  <id>   ID or alias of the workload to uninstall.

Options:
  --version, -v       Specific version to uninstall. Omit when only one version is installed.
  --all-versions, -a  Uninstall every installed version of the workload.
```

- Resolves `<id>` against `PackageId` first, then aliases (case-insensitive). Ambiguous aliases prompt for an id.
- With multiple installed versions and no `--version` / `--all-versions`, fails and lists the installed versions.
- `--version` and `--all-versions` are mutually exclusive.
- Removes the `<id>/<version>/` directory and the registry entry.

## Path shape

`assemblyPath` has two shapes that the installer translates between:

- **`workload.json` (package-root-relative)** - what the author writes, e.g. `tools/any/Foo.dll`.
- **`workloads.json` (workloads-root-relative)** - what the installer persists, e.g. `<id>/<version>/tools/any/Foo.dll`.

Loader resolves entries with a single concat against `WorkloadsRoot`. No magic `tools/any` injection.

## Tests

188 pass. New: `WorkloadInstallCommandTests`, `WorkloadUninstallCommandTests`, `WorkloadInstallerTests`.

## Docs

`docs/building-a-workload.md`, `docs/cli-architecture.md`, `docs/repo-structure.md` updated.

### DIFF
--- a/.github/instructions/dotnet.instructions.md
+++ b/.github/instructions/dotnet.instructions.md
@@ -112,6 +112,18 @@ fire:
 - No unused `using` directives.
 - Don't run `dotnet format` across unrelated files; no drive-by reformatting.
 - Don't disable analyzers to silence warnings, fix the underlying issue.
+## Code comments
+
+Comment to explain *why*, not *what*. XML doc summaries
+are one or two sentences; reach for `<remarks>` only when there's a single
+non-obvious clarification, and avoid stacking `<para>` blocks. Don't cite
+spec sections or document URLs from inside code (`§6.2`,
+`workload-package-layout §5.4`). Don't justify naming choices, list
+convention origins ("matches tsconfig.json…"), or narrate the alternative
+you didn't pick. Skip comments that restate the next line (DI registrations,
+obvious defaults, "Empty when no X" on a collection). Lead with the
+operative verb. Keep pointers to follow-up issues, cross-platform quirks,
+and rationale that isn't visible from the surrounding code.
 
 ## Code comments
 
--- a/.github/instructions/dotnet.instructions.md
+++ b/.github/instructions/dotnet.instructions.md
@@ -112,6 +112,7 @@ fire:
 - No unused `using` directives.
 - Don't run `dotnet format` across unrelated files; no drive-by reformatting.
 - Don't disable analyzers to silence warnings, fix the underlying issue.
+
 ## Code comments
 
 Comment to explain *why*, not *what*. XML doc summaries
--- a/.github/instructions/dotnet.instructions.md
+++ b/.github/instructions/dotnet.instructions.md
@@ -71,13 +71,42 @@ the lightest option:
 
 ## Error handling
 
-- Throw `GracefulException` (`Abstractions/Common/`) for **expected,
-  user-facing** errors. `Program.Main` catches it, prints the message, and
-  returns a non-zero exit code without a stack trace.
+Throw the **most specific** exception type at the failure site. Services and
+helpers shouldn't shape exceptions around CLI presentation; commands are the
+boundary that decides what's user-facing.
+
+- Framework / domain exceptions in services / helpers for known failure modes:
+  - `FileNotFoundException` / `DirectoryNotFoundException` for missing paths.
+  - `InvalidOperationException` for "operation isn't valid in current state".
+  - Domain exceptions like `InvalidWorkloadException` for subsystem-specific
+    validation failures.
+  - `ArgumentNullException` / `ArgumentException` for programmer errors on
+    method contracts.
+- **Commands wrap.** A command knows which exceptions from its dependencies
+  are user errors. Catch those at the command and wrap in
+  `GracefulException(message, isUserError: true)` (preserve the inner
+  exception). Anything you don't catch is a runtime bug and surfaces with a
+  stack trace.
+  - Keep the `try` **narrow**, ideally one method call. A wide `try` makes
+    a generic catch (e.g. `FileNotFoundException`) ambiguous about which
+    call site produced it. A one-call `try` plus the `<exception>` tags
+    on that method's xmldoc fully define what each catch arm means.
+  - Catching framework types (`FileNotFoundException`,
+    `InvalidOperationException`, ...) is fine **only when** the `try` is
+    narrow and the called method's xmldoc documents that type as a
+    user-facing failure mode. Otherwise promote the failure to a domain
+    exception so the catch is unambiguous.
+- `GracefulException` (`Abstractions/Common/`) is the CLI-facing wrapper.
+  `Program.Main` only catches `GracefulException`. Do **not** add broad
+  framework types to the top-level catch list, that hides bugs as user
+  errors.
 - Don't call `Environment.Exit` from inside a command.
-- Use specific framework exceptions (`ArgumentNullException`,
-  `InvalidOperationException`, ...) for programmer errors. Don't swallow
-  exceptions silently.
+- Default to rethrowing with context or converting to a more specific
+  domain exception. Swallowing is allowed only for genuine best-effort
+  cleanup paths (e.g. removing a leftover temp directory after another
+  failure), and the `catch` must include a comment explaining why losing
+  the exception is acceptable. Never `catch (Exception)` in business
+  logic.
 
 ## Logging vs user output
 
--- a/src/Abstractions/Workloads/EntryPointSpec.cs
+++ b/src/Abstractions/Workloads/EntryPointSpec.cs
@@ -4,24 +4,19 @@
 namespace Azure.Functions.Cli.Workloads;
 
 /// <summary>
-/// Identifies the type that implements <see cref="Workload"/>. Shared between
-/// the per-workload <see cref="WorkloadMetadata"/> (in the package root) and
-/// the CLI's global workload registry, so a workload's entry-point is
-/// described the same way at author time and at install time.
+/// Identifies the type that implements <see cref="Workload"/>.
 /// </summary>
 public sealed class EntryPointSpec
 {
     /// <summary>
-    /// Path to the assembly relative to the package's content root
-    /// (<c>tools/any/</c> by convention), e.g. <c>Foo.dll</c>.
+    /// Relative path to the workload assembly. In <c>workload.json</c> it is
+    /// relative to the package root; in the global <c>workloads.json</c> it
+    /// is relative to the workloads root. Must not contain <c>..</c> segments.
     /// </summary>
     public required string AssemblyPath { get; init; }
 
     /// <summary>
-    /// Fully-qualified type name implementing <see cref="Workload"/>. Stored
-    /// as a string so the metadata stays loadable even when the assembly
-    /// isn't on the runtime probe path (e.g. listing workloads without
-    /// loading them).
+    /// Fully-qualified type name implementing <see cref="Workload"/>.
     /// </summary>
     public required string Type { get; init; }
 }
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -1,0 +1,97 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.CommandLine;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Console;
+using Azure.Functions.Cli.Workloads;
+using Azure.Functions.Cli.Workloads.Discovery;
+using Azure.Functions.Cli.Workloads.Install;
+using Azure.Functions.Cli.Workloads.Storage;
+
+namespace Azure.Functions.Cli.Commands.Workload;
+
+/// <summary>
+/// <c>func workload install &lt;package&gt;</c>. Installs a workload from a
+/// <c>.nupkg</c> on disk. Feed-based acquisition is not yet supported.
+/// </summary>
+internal sealed class WorkloadInstallCommand : FuncCliCommand
+{
+    private const string NupkgExtension = ".nupkg";
+
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+
+    public Argument<string> WorkloadArgument { get; } = new("id")
+    {
+        Description = "Workload to install. Currently must be a path to a .nupkg on disk.",
+    };
+
+    public Option<bool> ForceOption { get; } = new("--force", "-f")
+    {
+        Description = "Overwrite an existing install of the same id and version.",
+    };
+
+    public WorkloadInstallCommand(IInteractionService interaction, IWorkloadInstaller installer)
+        : base("install", "Install a workload.")
+    {
+        _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
+        _installer = installer ?? throw new ArgumentNullException(nameof(installer));
+
+        WorkloadArgument.Validators.Add(result =>
+        {
+            string? value = result.GetValue(WorkloadArgument);
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                result.AddError("A workload id is required.");
+                return;
+            }
+
+            if (!value.EndsWith(NupkgExtension, StringComparison.OrdinalIgnoreCase))
+            {
+                result.AddError(
+                    $"Resolving '{value}' against a NuGet feed is not yet supported. " +
+                    "Pass a path to a .nupkg on disk.");
+            }
+        });
+
+        Arguments.Add(WorkloadArgument);
+        Options.Add(ForceOption);
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string workload = parseResult.GetValue(WorkloadArgument)!;
+        bool force = parseResult.GetValue(ForceOption);
+
+        WorkloadInstallResult result;
+        try
+        {
+            result = await _installer.InstallFromPackageAsync(workload, force, cancellationToken);
+        }
+        catch (Exception ex) when (ex is FileNotFoundException or InvalidWorkloadException or InvalidOperationException)
+        {
+            throw new GracefulException(ex.Message, isUserError: true);
+        }
+
+        _interaction.WriteSuccess(BuildSuccessMessage(result));
+        return 0;
+    }
+
+    private static string BuildSuccessMessage(WorkloadInstallResult result)
+    {
+        WorkloadEntry entry = result.Entry;
+        string verb = result.AlreadyInstalled
+            ? $"Workload '{entry.PackageId}' version '{entry.PackageVersion}' is already installed"
+            : $"Installed workload '{entry.PackageId}' version '{entry.PackageVersion}'";
+
+        return entry.Kind switch
+        {
+            WorkloadKind.Workload when entry.EntryPoint is not null
+                => $"{verb} (entry point: {entry.EntryPoint.Type}).",
+            WorkloadKind.Content
+                => $"{verb} (content at '{entry.Source}').",
+            _ => $"{verb}.",
+        };
+    }
+}
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -64,18 +64,26 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         string workload = parseResult.GetValue(WorkloadArgument)!;
         bool force = parseResult.GetValue(ForceOption);
 
-        WorkloadInstallResult result;
         try
         {
-            result = await _installer.InstallFromPackageAsync(workload, force, cancellationToken);
+            WorkloadInstallResult result = await _installer.InstallFromPackageAsync(workload, force, cancellationToken);
+            _interaction.WriteSuccess(BuildSuccessMessage(result));
+            return 0;
         }
-        catch (Exception ex) when (ex is FileNotFoundException or InvalidWorkloadException or InvalidOperationException)
+        catch (FileNotFoundException ex)
         {
             throw new GracefulException(ex.Message, isUserError: true);
         }
-
-        _interaction.WriteSuccess(BuildSuccessMessage(result));
-        return 0;
+        catch (InvalidWorkloadException ex)
+        {
+            throw new GracefulException(ex.Message, isUserError: true);
+        }
+        catch (InvalidOperationException ex)
+        {
+            throw new GracefulException(
+                $"{ex.Message} Pass --force to repair the install.",
+                isUserError: true);
+        }
     }
 
     private static string BuildSuccessMessage(WorkloadInstallResult result)
--- a/src/Func/Commands/Workload/WorkloadUninstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadUninstallCommand.cs
@@ -1,0 +1,166 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.CommandLine;
+using System.CommandLine.Parsing;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Console;
+using Azure.Functions.Cli.Workloads.Install;
+using Azure.Functions.Cli.Workloads.Storage;
+
+namespace Azure.Functions.Cli.Commands.Workload;
+
+/// <summary>
+/// <c>func workload uninstall &lt;packageId&gt; [--version &lt;v&gt;] [--all-versions] [--exact]</c>.
+/// Removes one or all installed versions of a workload.
+/// </summary>
+internal sealed class WorkloadUninstallCommand : FuncCliCommand
+{
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+    private readonly IWorkloadStore _store;
+
+    public Argument<string> WorkloadArgument { get; } = new("id")
+    {
+        Description = "ID or alias of the workload to uninstall.",
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version", "-v")
+    {
+        Description = "Specific version to uninstall. Omit when only one version is installed.",
+    };
+
+    public Option<bool> AllVersionsOption { get; } = new("--all-versions", "-a")
+    {
+        Description = "Uninstall every installed version of the workload.",
+    };
+
+    public Option<bool> ExactOption { get; } = new("--exact", "-e")
+    {
+        Description = "Match the argument as a literal package id; do not look up aliases.",
+    };
+
+    public WorkloadUninstallCommand(
+        IInteractionService interaction,
+        IWorkloadInstaller installer,
+        IWorkloadStore store)
+        : base("uninstall", "Uninstall a workload.")
+    {
+        _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
+        _installer = installer ?? throw new ArgumentNullException(nameof(installer));
+        _store = store ?? throw new ArgumentNullException(nameof(store));
+
+        WorkloadArgument.Validators.Add(result =>
+        {
+            string? value = result.GetValue(WorkloadArgument);
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                result.AddError("A workload id is required.");
+            }
+        });
+
+        Arguments.Add(WorkloadArgument);
+        Options.Add(VersionOption);
+        Options.Add(AllVersionsOption);
+        Options.Add(ExactOption);
+
+        Validators.Add(result =>
+        {
+            int specified = result.Children
+                .OfType<OptionResult>()
+                .Count(or => or.Option == VersionOption || or.Option == AllVersionsOption);
+
+            if (specified > 1)
+            {
+                result.AddError("--all-versions and --version cannot be combined.");
+            }
+        });
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string identifier = parseResult.GetValue(WorkloadArgument)!;
+        string? version = parseResult.GetValue(VersionOption);
+        bool all = parseResult.GetValue(AllVersionsOption);
+        bool exact = parseResult.GetValue(ExactOption);
+
+        IReadOnlyList<WorkloadEntry> installed = await _store.GetWorkloadsAsync(cancellationToken);
+        List<WorkloadEntry> matches = [.. installed
+            .Where(w => string.Equals(w.PackageId, identifier, StringComparison.OrdinalIgnoreCase)
+                || (!exact && (w.Aliases?.Any(a => string.Equals(a, identifier, StringComparison.OrdinalIgnoreCase)) ?? false)))];
+
+        if (matches.Count == 0)
+        {
+            _interaction.WriteWarning(
+                $"Workload '{identifier}' is not installed; nothing to do.");
+            return 0;
+        }
+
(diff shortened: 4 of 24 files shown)


src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,87 @@
...
+/// the same <see cref="IWorkloadInstaller"/>.
+/// </summary>
+internal sealed class WorkloadInstallCommand : FuncCliCommand
+{
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+
+    public Argument<string> PackageIdArgument { get; } = new("id")
+    {
+        Description = "Package id (or alias) of the workload to install.",
```
liliankasem:
```suggestion
        Description = "Package ID (or alias) of the workload to install.",
```

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version", "-v")
```
liliankasem: If we have -v for verbose globally, we cant have -v shorthand for version here

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+/// </summary>
+internal sealed class WorkloadUninstallCommand : FuncCliCommand
+{
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+    private readonly IGlobalManifestStore _store;
+
+    public Argument<string> PackageIdArgument { get; } = new("packageId")
+    {
+        Description = "Package id of the workload to uninstall.",
```
liliankasem:
```suggestion
        Description = "Package ID of the workload to uninstall.",
```

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version", "-v")
```
liliankasem, line 28: Same comment about -v here

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,87 @@
...
+/// command is the install pipeline's entry point and is wired so that
+/// follow-up only needs to resolve a package id to a directory and call
+/// the same <see cref="IWorkloadInstaller"/>.
+/// </summary>
+internal sealed class WorkloadInstallCommand : FuncCliCommand
+{
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+
+    public Argument<string> PackageIdArgument { get; } = new("id")
```
liliankasem:
```suggestion
    public Argument<string> PackageIdArgument { get; } = new("packageId")
```

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        Description = "Package ID (or alias) of the workload to install.",
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version")
+    {
+        Description = "Specific version to install. Defaults to the latest available.",
+    };
+
+    public Option<DirectoryInfo?> FromOption { get; } = new("--from")
```
jviau: I think `--source` is more popular. Also, add a short name?

Or can we just remove this for now, only adding it when nuget feed resolution is added.

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        var packageId = parseResult.GetValue(PackageIdArgument)
+            ?? throw new GracefulException("packageId is required.", isUserError: true);
+        var version = parseResult.GetValue(VersionOption);
+        var from = parseResult.GetValue(FromOption);
+
+        if (from is null)
```
jviau: Why do we need to specify `from` (or `source`) when installing a package from disk? Could we not just check `packageId` and see if it ends with `.nupkg`, and then assume it is a path on disk?

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+/// command is the install pipeline's entry point and is wired so that
+/// follow-up only needs to resolve a package id to a directory and call
+/// the same <see cref="IWorkloadInstaller"/>.
+/// </summary>
+internal sealed class WorkloadInstallCommand : FuncCliCommand
+{
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+
+    public Argument<string> PackageIdArgument { get; } = new("packageId")
```
jviau: Can we rename to `PackageArgument`? I think this should be flexible to support package alias, nupkg on disk, and package ID on a feed (remote or local).

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        Description = "Package ID (or alias) of the workload to install.",
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version")
```
jviau: Lets remove this for now until nuget feed resolution is added.

src/Func/Workloads/Install/NuspecMetadata.cs
```diff
@@ -0,0 +1,24 @@
...
+/// the install pipeline only persists what <c>func workload list</c> shows
+/// and what the loader uses to activate a workload.
+/// </summary>
+/// <param name="PackageId">The <c>id</c> element. NuGet ids are case-insensitive.</param>
+/// <param name="Version">The <c>version</c> element verbatim. Validation lives at the catalog layer.</param>
+/// <param name="Title">The <c>title</c> element, or empty if absent. Falls back to <see cref="PackageId"/> in the manifest.</param>
+/// <param name="Description">The <c>description</c> element, or empty if absent.</param>
+/// <param name="Aliases">Whitespace-split <c>tags</c>, used as user-facing short names (e.g. <c>"dotnet"</c>).</param>
+/// <param name="PackageTypes">Names from <c>packageTypes/packageType/@name</c>. Contains <c>"FuncCliWorkload"</c> for valid workload packages.</param>
+internal sealed record NuspecMetadata(
```
jviau: Do we want to just use `NuGet.Client` types? We will need that library for reading from feeds anyways.

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/ManifestMetadata.cs#L18

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// </summary>
+    public const string FuncCliWorkloadPackageType = "FuncCliWorkload";
+
+    private readonly IWorkloadPaths _paths = paths ?? throw new ArgumentNullException(nameof(paths));
+    private readonly IGlobalManifestStore _store = store ?? throw new ArgumentNullException(nameof(store));
+    private readonly INuspecReader _nuspecReader = nuspecReader ?? throw new ArgumentNullException(nameof(nuspecReader));
+    private readonly IWorkloadEntryPointScanner _scanner = scanner ?? throw new ArgumentNullException(nameof(scanner));
+
+    /// <inheritdoc />
+    public async Task<InstalledWorkload> InstallFromDirectoryAsync(
```
jviau: Do we want to support installing from a directory? I understand installing from a .nupkg on disk. But from a directory implies you _had_ a nupkg and then unzipped it. Why not skip that unzipped requirement and support a nupkg directly.

jviau (changes requested):
Two primary suggestions:

1. Bring in NuGet library directly, use its types for loading and parsing a manifest.
2. For first iteration of `install`, if you don't want to support feeds then add support for `.nupkg` on disk only. Don't add `--version` or `--from` yet, just `func workload install <package-on-disk>.nupkg`, we will detect "on disk" via the `.nupkg` extension.
    - nuget libraries probably have tooling to work with a `.nupkg` on disk directly, like extracting manifest from it.

src/Func/Commands/Workload/WorkloadInstallCommand.cs
liliankasem: Removed from this pr for now, but agree --source makes more sense

src/Func/Workloads/Install/WorkloadInstaller.cs
liliankasem: Very good point, I will switch to .nupkg and remove from dir

src/Func/Commands/Workload/WorkloadInstallCommand.cs
liliankasem: WorkloadInstallCommand takes a single package argument and detects .nupkg

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+    private readonly IInteractionService _interaction;
+    private readonly IWorkloadInstaller _installer;
+    private readonly IGlobalManifestStore _store;
+
+    public Argument<string> PackageIdArgument { get; } = new("packageId")
+    {
+        Description = "Package ID of the workload to uninstall.",
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version")
```
jviau: Should we add short-names for these options? `-v` for this one. Maybe `-a` for `--all-versions`?

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        using var reader = OpenPackage(nupkgPath);
+        var nuspec = reader.NuspecReader;
+
+        var packageId = nuspec.GetId();
+        var version = nuspec.GetVersion().ToNormalizedString();
+        var title = nuspec.GetTitle();
+        var description = nuspec.GetDescription();
+        var aliases = SplitTags(nuspec.GetTags());
```
jviau: Do we want to treat all tags as aliases? I think this would prevent workloads from using tags for other means. We should probably have a convention. Aliases are `alias:<name>` or `workload_alias:<name>`

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,188 @@
...
+
+    private static PackageArchiveReader OpenPackage(string nupkgPath)
+    {
+        try
+        {
+            return new PackageArchiveReader(File.OpenRead(nupkgPath));
+        }
+        catch (Exception ex) when (ex is InvalidDataException or PackagingException)
+        {
+            throw new GracefulException(
```
jviau: Do we want to include the inner exception?

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,188 @@
...
+    /// Package-type name a workload package must declare in its .nuspec.
+    /// </summary>
+    public const string FuncCliWorkloadPackageType = "FuncCliWorkload";
+
+    private readonly IWorkloadPaths _paths = paths ?? throw new ArgumentNullException(nameof(paths));
+    private readonly IGlobalManifestStore _store = store ?? throw new ArgumentNullException(nameof(store));
+    private readonly IWorkloadEntryPointScanner _scanner = scanner ?? throw new ArgumentNullException(nameof(scanner));
+
+    /// <inheritdoc />
+    public async Task<InstalledWorkload> InstallFromPackageAsync(
```
jviau: We need to settle on the nupkg format. Nupkg can have many folders. Do we have any requirements on what folders, if any, they must put their entry point into?

We synced offline: will require packages have a `workload.json`.

jviau (commented):
I think some of these comments are already discussed in the sync we had,

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
liliankasem: -v is global verbose (we can discuss if we dont want this) - but -a for all makes sense

src/Func/Workloads/Install/WorkloadInstaller.cs
liliankasem: Thanks for catching! Should be on alias:, will make sure to update that

src/Func/Workloads/Install/WorkloadInstaller.cs
liliankasem: Will operate on workload.json at root for now - Fabio is working on the nupkg format

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
liliankasem: Actually let's support -v and maybe just switch to diagnostics | -d instead of verbose for global?

liliankasem: have a refactor pr for this

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Azure.Functions.Cli.Commands.Workload;
+
+/// <summary>
+/// <c>func workload install &lt;package&gt;</c>. Installs a workload from a
+/// <c>.nupkg</c> on disk. Feed-based acquisition (alias or package id
+/// resolved against a NuGet feed) lands in a follow-up; the argument is
+/// shaped to accept either form so the command surface doesn't have to
+/// change when feeds are added.
+/// </summary>
+internal sealed class WorkloadInstallCommand : FuncCliCommand
```
jviau, line 18: Should we add a `--force|-f` switch to allow for overriding existing workloads (if ID + Version matches an existing workload)

src/Abstractions/Workloads/EntryPointSpec.cs
```diff
@@ -12,8 +12,18 @@ namespace Azure.Functions.Cli.Workloads;
...
 {
     /// <summary>
-    /// Path to the assembly relative to the package's content root
-    /// (<c>tools/any/</c> by convention), e.g. <c>Foo.dll</c>.
+    /// Path to the workload assembly. The exact base depends on context:
+    /// <list type="bullet">
+    ///   <item><description>In <c>workload.json</c> (per-package manifest the
+    ///   author writes): relative to the package root, e.g.
+    ///   <c>tools/any/Foo.dll</c>.</description></item>
+    ///   <item><description>In <c>workloads.json</c> (global registry the
```
jviau: Ah I thought we were going to leave it always relative to package root. When resolving from global manifest, we will `Path.Combine(workload.Id, workload.Version, workload.EntryPoint.AssemblyPath)`

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        Arguments.Add(WorkloadArgument);
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string workload = parseResult.GetValue(WorkloadArgument)
+            ?? throw new GracefulException("id is required.", isUserError: true);
+
+        bool isNupkgPath = workload.EndsWith(NupkgExtension, StringComparison.OrdinalIgnoreCase);
```
jviau: Should this be a validator?

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,151 @@
...
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string identifier = parseResult.GetValue(WorkloadArgument)
+            ?? throw new GracefulException("id is required.", isUserError: true);
+        string? version = parseResult.GetValue(VersionOption);
+        bool all = parseResult.GetValue(AllVersionsOption);
+
+        if (all && !string.IsNullOrEmpty(version))
+        {
+            throw new GracefulException(
```
jviau: Should this be a validator? Found a stackoverflow regarding mutually exclusive options.

https://stackoverflow.com/questions/76997020/system-commandline-mutally-exclusive-options

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,239 @@
...
+    /// <inheritdoc />
+    public async Task<WorkloadEntry> InstallFromPackageAsync(
+        string nupkgPath,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(nupkgPath);
+
+        if (!File.Exists(nupkgPath))
+        {
+            throw new GracefulException(
```
jviau: I think we are using `GracefulException` as too much of a catch-all exception. We should still use other exception types when appropriate. 

This should be a `FileNotFoundException` for example.

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,239 @@
...
+        using PackageArchiveReader reader = OpenPackage(nupkgPath);
+        NuspecReader nuspec = reader.NuspecReader;
+
+        string packageId = nuspec.GetId();
+        string version = nuspec.GetVersion().ToNormalizedString();
+        IReadOnlyList<string> aliases = ParseAliases(nuspec.GetTags());
+
+        if (!nuspec.GetPackageTypes().Any(t => string.Equals(t.Name, FuncCliWorkloadPackageType, StringComparison.OrdinalIgnoreCase)))
+        {
+            throw new GracefulException(
```
jviau: `InvalidWorkloadException` seems more appropriate here.

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,239 @@
...
+                $"Package '{packageId}' is not a func CLI workload " +
+                $"(missing package type '{FuncCliWorkloadPackageType}' in its .nuspec).",
+                isUserError: true);
+        }
+
+        string installPath = _paths.GetInstallDirectory(packageId, version);
+        if (Directory.Exists(installPath))
+        {
+            throw new GracefulException(
```
jviau: `InvalidOperationException`?

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,239 @@
...
+        catch (Exception ex) when (ex is InvalidDataException or PackagingException)
+        {
+            throw new GracefulException(
+                $"Failed to read .nupkg at '{nupkgPath}': {ex.Message}",
+                isUserError: true,
+                verboseMessage: ex.ToString());
+        }
+    }
+
+    private static async Task ExtractPackageAsync(
```
jviau, line 170: The nuget APIs don't have something to extract a nupkg?

It technically is just a .zip file. I wonder if you can simply use dotnet `ZipArchive` APIs to extract the steam?

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,239 @@
...
+    /// </summary>
+    private static IReadOnlyList<string> ParseAliases(string? tags)
+    {
+        if (string.IsNullOrWhiteSpace(tags))
+        {
+            return [];
+        }
+
+        List<string> aliases = [];
+        foreach (string tag in tags.Split((char[]?)null, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries))
```
jviau: why `(char[]?)null` over `' '`?

src/Func/Workloads/Loading/WorkloadLoader.cs
```diff
@@ -34,42 +34,40 @@ public IReadOnlyList<WorkloadInfo> Load(IReadOnlyList<WorkloadEntry> entries)
 
     private WorkloadInfo LoadEntry(WorkloadEntry entry)
     {
-        string installPath = _paths.GetInstallDirectory(entry.PackageId, entry.PackageVersion);
-        string contentRoot = Path.GetFullPath(Path.Combine(installPath, "tools", "any"));
-        string assemblyPath = Path.GetFullPath(Path.Combine(contentRoot, entry.EntryPoint.AssemblyPath));
+        string workloadsRoot = Path.GetFullPath(_paths.WorkloadsRoot);
+        string assemblyPath = Path.GetFullPath(Path.Combine(workloadsRoot, entry.EntryPoint.AssemblyPath));
```
jviau: Do we want to go with the multi-combine so we don't have to re-write the entrypoint on install:

``` CSharp
string workloadsRoot = Path.GetFullPath(_paths.WorkloadsRoot);
string workloadRelative = Path.Combine(entry.PackageId, entry.PackageVersion, entry.EntryPoint.AssemblyPath);
string assemblyPath = Path.GetFullPath(Path.Combine(workloadsRoot, workloadRelative));
```

jviau:
Looking out how much `GracefulException` is used throughout all the codebase, do we want to re-evaluate that? Can we switch to using more fitting per-scenario exception types?

src/Abstractions/Workloads/EntryPointSpec.cs
liliankasem: Discussed offline - keeping as is for easier comprehension + avoid having to append the id and version each time

liliankasem:
> Looking out how much `GracefulException` is used throughout all the codebase, do we want to re-evaluate that? Can we switch to using more fitting per-scenario exception types?

Yeah you're right on this - it is overused. Program.cs ultimately will surface this as a GracefulException if needed so individual items should be using the right exception type. I'll update the agent instructions for this too

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,296 @@
...
+        foreach (ZipArchiveEntry entry in archive.Entries)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (ShouldSkipEntry(entry))
+            {
+                continue;
+            }
+
+            string targetPath = Path.Combine(destination, entry.FullName.Replace('/', Path.DirectorySeparatorChar));
```
satvu: AI Review: 

The extraction loop does Path.Combine(destination, packageFile.Replace('/', ...)) without validating the resolved path stays within destination. A crafted .nupkg with entries like ../../evil.dll could write outside the install directory. Add a `Path.GetFullPath` check:

```
 string fullTarget = Path.GetFullPath(targetPath);
 if (!fullTarget.StartsWith(Path.GetFullPath(destination) + Path.DirectorySeparatorChar))
     throw new InvalidWorkloadException("Package contains entry that escapes the install directory.");
```

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -157,39 +148,104 @@ public async Task<bool> UninstallAsync(
...
             await entryStream.CopyToAsync(output, cancellationToken);
         }
     }
 
+    /// <summary>
+    /// Skips OPC packaging metadata, the nuspec itself, signatures, and
+    /// directory entries so the extracted layout matches what the legacy
+    /// NuGet reader produced.
+    /// </summary>
+    private static bool ShouldSkipEntry(ZipArchiveEntry entry)
```
jviau: Didn't realize using ZipArchive would require this much custom logic. Was hoping it will spare you from needing to iterate each entry entirely and to just unzip it to disk. Seeing this, not sure it offers any value over what you had previously.

There might be a more convenient way to extract using Nuget APIs. I do see a `PackageArchiveReader.CopyFilesAsync`, but its not actually async (the delegate doesn't allow for async I/O on each stream).

There is an `EnumeratePackageEntries` which will give you the path + ZipArchiveEntry pair

src/Func/Program.cs
```diff
@@ -118,6 +119,18 @@
 
         exitCode = 1;
     }
+    catch (Exception ex) when (ex is FileNotFoundException
```
jviau: Ah not sure I like this result of us switching away from `GracefulException`. These exceptions might not be exclusively user-failures, they could be runtime bugs. How can we avoid using a single exception for everything, decouple different services from needing to knowing the context of an error, but also avoid accidentally hiding runtime bugs as user issues.

Should it be the responsibility of each individual command to catch exceptions and decide how and if to wrap in a `GracefulException`?

liliankasem: 100% - agent went too far here, each command being the boundary was my intention - fixed!

src/Func/Workloads/Install/WorkloadInstaller.cs
liliankasem: Fixed!

fabiocav (changes requested):
This should block on the package structure/layout spec. I'll be pushing that proposal tomorrow and it will help define a lot of the install requirements.

.github/instructions/dotnet.instructions.md
```diff
@@ -69,13 +69,27 @@ the lightest option:
...
+- `GracefulException` (`Abstractions/Common/`) is the **top-level adapter**,
+  not the default. Use it only when you need CLI-specific context the base
+  type can't carry (e.g. `VerboseMessage`, explicit `IsUserError`). The
+  framework + domain types above are already surfaced as clean user errors
+  by `Program.Main`.
 - Don't call `Environment.Exit` from inside a command.
-- Use specific framework exceptions (`ArgumentNullException`,
-  `InvalidOperationException`, ...) for programmer errors. Don't swallow
-  exceptions silently.
+- Don't swallow exceptions; rethrow with context or convert to a more
```
jviau: Not sure how strictly agents follow this. Do we need to permit deviations from this when appropriate? There are always cases when swallowing an exception is _okay_.

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+        });
+
+        Arguments.Add(WorkloadArgument);
+        Options.Add(ForceOption);
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string workload = parseResult.GetValue(WorkloadArgument)
+            ?? throw new ArgumentException("id is required.");
```
jviau: Is there also a validator asserting this is required?

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,156 @@
...
+            if (specified > 1)
+            {
+                result.AddError("--all-versions and --version cannot be combined.");
+            }
+        });
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string identifier = parseResult.GetValue(WorkloadArgument)
```
jviau: Same here, is there a validator for this?

src/Func/Program.cs
jviau: We are still catching these exceptions, is that intentional?

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        // An alias may resolve to multiple workload IDs. Require the user to disambiguate by ID in that case.
+        string[] distinctPackageIds = matches
+            .Select(m => m.PackageId)
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .ToArray();
+        if (distinctPackageIds.Length > 1)
+        {
+            throw new InvalidOperationException(
```
jviau: Maybe we can derive a `InvalidUserActionException : InvalidOperationException` - this will give us a specific exception to capture in `Program.Main` and convert to a user-facing error. It is better than `InvalidOperationException`, because libraries might throw that exception for authoring (not user) mistakes.

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -0,0 +1,79 @@
...
+        try
+        {
+            installed = await _installer.InstallFromPackageAsync(workload, force, cancellationToken);
+        }
+        catch (Exception ex) when (ex is FileNotFoundException or InvalidWorkloadException or InvalidOperationException)
+        {
+            throw new GracefulException(ex.Message, isUserError: true);
+        }
+
+        _interaction.WriteSuccess(
```
jviau: this can move within the `try` block. No need to declare `WorkloadEntry` variable separately from the install action then.

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        IReadOnlyList<WorkloadEntry> installed = await _store.GetWorkloadsAsync(cancellationToken);
+        List<WorkloadEntry> matches = installed
+            .Where(w => string.Equals(w.PackageId, identifier, StringComparison.OrdinalIgnoreCase)
+                || (w.Aliases?.Any(a => string.Equals(a, identifier, StringComparison.OrdinalIgnoreCase)) ?? false))
+            .ToList();
+
+        if (matches.Count == 0)
+        {
+            throw new GracefulException(
```
jviau: Do we want this to be an error? If we aim for idempotence, calling uninstall on a non-installed package seems "successful" to me - the package is uninstalled at the end. The fact there was nothing to uninstall is besides the point.

src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        string installPath = _paths.GetInstallDirectory(packageId, version);
+        if (Directory.Exists(installPath))
+        {
+            if (!force)
+            {
+                throw new InvalidOperationException(
+                    $"Workload '{packageId}' version '{version}' is already installed at '{installPath}'. " +
+                    "Pass --force to overwrite the existing install.");
```
jviau: Having the `--force` here is leaking the workload concept into this service. It looks like to me this service is meant to be command agnostic. How can we better decouple the error and the specific user-facing suggestion?

src/Func/Commands/Workload/WorkloadUninstallCommand.cs
jviau: stale from before your changes.

.github/instructions/dotnet.instructions.md
fabiocav: I think it's fine to have this defined as a rule and iterate on exceptions. We may want to be explicit about that here.

PR #4929, opened 2026-05-02
Choose unique default names for func new

### LINKED ISSUE #3262: Make sure default trigger name is unique
For `func new`, if users don't pass in a name, the default is trigger name, like "HttpTrigger". If a trigger with this name already exists, users would be asked about overwriting. But if the intention wasn't to overwrite (because the user didn't know about the default name), you'd have to cancel and start again.  It could get annoying if you forget that the default name is duplicated and need to cancel and start again each time.

VSCode currently increments trigger name automatically to ensure there's no duplicate.

Genmin:
Fixes #3262.

## Summary
- when `func new` offers a default function name, choose the first available name instead of defaulting to an existing artifact
- use `HttpTrigger`, `HttpTrigger1`, `HttpTrigger2`, ... style naming for classic folder templates and Node v4 file templates
- add unit coverage for the default-name increment helper

## Validation
- `git diff --check`
- Not run locally: this machine does not have the .NET SDK (`dotnet`, `csc`, and `msbuild` are unavailable).

### DIFF
--- a/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
+++ b/src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
@@ -270,9 +270,12 @@ namespace Azure.Functions.Cli.Actions.LocalActions
                         ConfigureAuthorizationLevel(template);
                     }
 
-                    ColoredConsole.Write($"Function name: [{template.Metadata.DefaultFunctionName}] ");
+                    var defaultFunctionName = GetUniqueDefaultFunctionName(
+                        template.Metadata.DefaultFunctionName,
+                        functionName => FunctionArtifactExists(functionName, FileName, template));
+                    ColoredConsole.Write($"Function name: [{defaultFunctionName}] ");
                     FunctionName = FunctionName ?? Console.ReadLine();
-                    FunctionName = string.IsNullOrEmpty(FunctionName) ? template.Metadata.DefaultFunctionName : FunctionName;
+                    FunctionName = string.IsNullOrEmpty(FunctionName) ? defaultFunctionName : FunctionName;
                     await _templatesManager.Deploy(FunctionName, FileName, template);
                     PerformPostDeployTasks(FunctionName, Language);
                 }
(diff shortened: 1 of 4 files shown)


src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -270,9 +270,12 @@ public override async Task RunAsync()
                         ConfigureAuthorizationLevel(template);
                     }
 
-                    ColoredConsole.Write($"Function name: [{template.Metadata.DefaultFunctionName}] ");
+                    var defaultFunctionName = GetUniqueDefaultFunctionName(
+                        template.Metadata.DefaultFunctionName,
+                        functionName => FunctionArtifactExists(functionName, FileName, template));
```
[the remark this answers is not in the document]
liliankasem: When FileName is set, FunctionArtifactExists ignores the candidate name and only checks the fixed file path. If that file already exists, every candidate returns true and GetUniqueDefaultFunctionName loops forever, hanging func new before it ever prompts. Easiest fix is to skip the uniqueness loop when FileName != null (renaming the function can't avoid a fixed-file collision anyway).

Cleanest fix:

```csharp
var defaultFunctionName = FileName != null
    ? template.Metadata.DefaultFunctionName
    : GetUniqueDefaultFunctionName(
        template.Metadata.DefaultFunctionName,
        functionName => FunctionArtifactExists(functionName, FileName, template));
```

src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -470,6 +473,45 @@ private void ConfigureAuthorizationLevel(Template template)
...
+                    .Where(file => !file.Key.EndsWith(".dat"))
+                    .Select(file => fileName ?? file.Key.Replace("%functionName%", functionName))
+                    .Any(name => FileSystemHelpers.FileExists(
+                        Path.Combine(Environment.CurrentDirectory, "src", "functions", name)));
+            }
+
+            return FileSystemHelpers.DirectoryExists(Path.Combine(Environment.CurrentDirectory, functionName));
+        }
+
+        private static bool IsNodeV4Template(Template template)
```
liliankasem: This exact code/check already exists inlined in TemplatesManager.cs:145 - you should extract this a shared helper instead of duplicating

src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -270,9 +270,12 @@ public override async Task RunAsync()
                         ConfigureAuthorizationLevel(template);
                     }
 
-                    ColoredConsole.Write($"Function name: [{template.Metadata.DefaultFunctionName}] ");
+                    var defaultFunctionName = GetUniqueDefaultFunctionName(
```
liliankasem: Right now this PR only covers the traditional/Node v4 path, .NET (in-proc and isolated) and the Python v2 programming model take their own branches earlier in RunAsync and still silently prompt with a colliding default. Please extend the uniqueness behavior to those two paths as well.

- .NET (in-proc & isolated) branch: lines 144–162 (if (WorkerRuntimeLanguageHelper.IsDotnet(_workerRuntime) && !Csx)), default name is set at line 158 (FunctionName = string.IsNullOrEmpty(FunctionName) ? TemplateName : FunctionName;).

- Python v2 (new programming model) branch: lines 165–234 (else if (IsNewPythonProgrammingModel())), default name is resolved at lines 227–230 (if (string.IsNullOrEmpty(FunctionName)) FunctionName = providedInputs[GetFunctionNameParamId];).


Both bypass the new GetUniqueDefaultFunctionName call

liliankasem (changes requested):

Genmin:
Checked the current branch against the requested changes. The follow-up commit c11254d8 addresses the open review items: Node v4 fixed --file templates skip the unique-name loop, the Node v4 check uses the shared TemplateHelpers.IsNodeV4Template helper, and .NET plus Python v2 paths now derive unique default function names before prompting.\n\nI attempted to rerun the focused unit tests locally with:\n\n- dotnet test test/Cli/Func.UnitTests/Azure.Functions.Cli.UnitTests.csproj --filter FullyQualifiedName~CreateFunctionActionTests --no-restore\n\nbut this environment does not have dotnet installed. The remaining non-code blocker is the Microsoft CLA status.

liliankasem (approved):

PR #4937, opened 2026-05-06
Refactor workload manifest model and add per-package metadata

liliankasem:
- IWorkload (interface) -> Workload (abstract class) in Abstractions.
- Add per-package WorkloadMetadata + shared EntryPointSpec in Abstractions.
- GlobalManifest (nested dict) -> WorkloadRegistry (flat IList<WorkloadEntry>);
  each entry self-describes PackageId, PackageVersion, Aliases, EntryPoint.
- Rename GlobalManifestStore/IGlobalManifestStore -> WorkloadStore/IWorkloadStore
  and IWorkloadPaths.GlobalManifestPath -> WorkloadRegistryPath.
- WorkloadLoader now resolves install dir via IWorkloadPaths.
- WorkloadInfo carries the loaded Workload instance; WorkloadListCommand
  reads display name/description from it.
- Update tests and fixtures accordingly.


src/Abstractions/Workloads/EntryPointSpec.cs
```diff
@@ -0,0 +1,27 @@
...
+/// Identifies the type that implements <see cref="Workload"/>. Shared between
+/// the per-workload <see cref="WorkloadMetadata"/> (in the package root) and
+/// the CLI's global workload registry, so a workload's entry-point is
+/// described the same way at author time and at install time.
+/// </summary>
+public sealed class EntryPointSpec
+{
+    /// <summary>
+    /// Path to the assembly relative to the package root
+    /// (e.g. <c>lib/net10.0/Foo.dll</c>).
```
fabiocav: nit: The comment here should align with what we end up documenting in the package spec, which currently, would be `tools/any`, but the path should be relative to that base, so in this case, simply `Foo.dll`

Non blocking.

src/Abstractions/Workloads/Workload.cs
```diff
@@ -28,17 +31,21 @@ public interface IWorkload
     /// SemVer 2.0 string (e.g. <c>"1.2.3"</c>, <c>"1.2.3-preview.1"</c>);
     /// the CLI does not currently enforce or normalize the format.
     /// </summary>
-    public string PackageVersion { get; }
+    public abstract string Version { get; }
```
fabiocav: Did we want to have some of this implement a default behavior or resolving the version from the assembly?

src/Func/Hosting/WorkloadStorageRegistration.cs
```diff
@@ -35,7 +42,25 @@ public static IServiceCollection AddWorkloadStorage(this IServiceCollection serv
...
+
+        // Materialize the installed-workloads list once, on first resolve.
+        // The store + loader are still registered above so install / uninstall
+        // commands can mutate / reload the registry; consumers that only need
+        // to enumerate already-loaded workloads (list, alias routing, command
+        // contribution) inject this singleton instead. Sync-over-async is
+        // intentional: this runs at most once per process, before any command
+        // executes, and there is no synchronization context to deadlock on
+        // in a CLI.
+        services.AddSingleton<IReadOnlyList<WorkloadInfo>>(sp =>
```
fabiocav: Instead of registering a collection here, we should either introduce a service (e.g., `IWorkloadProvider`) or repurpose one of the services you already have to handle the caching and construction. We should also avoid the sync-over-async logic here, which a service would help with.

It also gives you a natural home for caching, lazy materialization, and CancellationToken propagation, in a way that is idiomatic.

The current factory lambda will also not handle exceptions here as cleanly.

> this runs at most once per process, before any command executes, and there is no synchronization context to deadlock on in a CLI.

We shouldn't make assumptions on how the service (in this case, the collection) will be consumed. A proper service registration here ensures safety.

Although we won't have a default synchronization context in the running product, we can avoid the fragility of sync-over-async (and avoid the anti-pattern), protecting us in case this is ever used in a different context, including under test conditions.

src/Func/Workloads/Loading/WorkloadLoader.cs
```diff
@@ -11,60 +13,58 @@ namespace Azure.Functions.Cli.Workloads.Loading;
...
     }
 
-    private static LoadedWorkload LoadEntry(InstalledWorkload installed)
+    private WorkloadInfo LoadEntry(WorkloadEntry entry)
     {
-        var entry = installed.Entry;
-        var assemblyPath = Path.Combine(entry.InstallPath, entry.EntryPoint.Assembly);
+        string installPath = _paths.GetInstallDirectory(entry.PackageId, entry.PackageVersion);
+        string contentRoot = Path.Combine(installPath, WorkloadMetadataReader.ContentDirectoryName);
+        string assemblyPath = Path.Combine(contentRoot, entry.EntryPoint.AssemblyPath);
```
fabiocav: This is using a path coming from the metadata in a package. Let's make sure we're avoiding path traversal here by constraining the resulting path to the content root (i.e., defend against a package publishing a path like `..\..` or an absolute path)

src/Func/Hosting/WorkloadStorageRegistration.cs
liliankasem: Good point - I will introduce IWorkloadProvider

fabiocav (approved):

PR #4938, opened 2026-05-06
Add FUNC_CLI_TELEMETRY_OPTOUT, drop -v from --verbose, add -v alias for --host-version

liliankasem:
## Changes

1. **Telemetry opt-out env var.** Honor `FUNC_CLI_TELEMETRY_OPTOUT` (matches the v5 `FUNC_CLI_*` convention). The legacy `FUNCTIONS_CORE_TOOLS_TELEMETRY_OPTOUT` from Core Tools v4 is still honored. Opt-out is **true for any value except** `no`, `n`, `0`, `false`, `off` (case-insensitive), so unrecognized values fail safe toward not collecting telemetry.

2. **Drop `-v` short alias from global `--verbose`.** `--verbose` stays; the `-v` alias is removed so subcommands can claim it.

3. **`-v` shortcut on `func start --host-version`.** First subcommand to take advantage of the freed-up `-v`.

## Tests

- `ParserTests` asserts `--verbose` exists and has no `-v` alias.
- `TelemetryTests` covers all "off" sentinels and a sample of truthy values, plus both env var names.
- Full suite: 162 tests pass.

## Docs

- README telemetry section updated to describe the new env var and the opt-out semantics, and call out the legacy back-compat.

### DIFF
--- a/src/Func/Telemetry/CliTelemetry.cs
+++ b/src/Func/Telemetry/CliTelemetry.cs
@@ -82,6 +82,9 @@ internal static class CliTelemetry
             .AddAttributes(GetResourceAttributes());
     }
 
+    private static readonly HashSet<string> _optOutFalseSentinels =
+        new(StringComparer.OrdinalIgnoreCase) { "no", "n", "0", "false", "off" };
+
     /// <summary>
     /// Returns the Azure Monitor connection string when telemetry is
     /// configured (build has a real instrumentation key) and the user has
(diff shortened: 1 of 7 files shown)


src/Func/Telemetry/CliTelemetry.cs
```diff
@@ -109,6 +110,20 @@ public static bool TryGetConnectionString([NotNullWhen(true)] out string? connec
...
+        var value = Environment.GetEnvironmentVariable(envVarName);
+        if (string.IsNullOrEmpty(value))
+        {
+            return false;
+        }
+
+        // Explicit "off" sentinels mean the user has not opted out;
+        // anything else (including unrecognized values) is treated as
+        // opt-out, so we fail safe toward not collecting telemetry.
+        return !_optOutFalseSentinels.Contains(value);
```
jviau, line 124: I know you have `StringComparer.OrdinalIgnoreCase`, but we might want to `Trim()` as well.

```suggestion
        return !_optOutFalseSentinels.Contains(value.Trim());
```

Issue #4901, opened 2026-04-28
Workload engine: command ownership wrapper + named conflict diagnostics


liliankasem:
## What to build

Today, workloads register top-level commands as a raw `IEnumerable<Command>`, so when two workloads register the same command name, the conflict diagnostic in `Parser` cannot name the offending workload (placeholder `<name>` in the warning text).

Replace the raw registration with a typed contribution wrapper — `WorkloadCommandContribution { IWorkload Owner; Command Command }` — so ownership is a registration-time invariant. Update `Parser` to consume the wrapper and produce a diagnostic that names every workload involved in a conflict.

## Acceptance criteria

- [ ] New `WorkloadCommandContribution` type in `Func.Cli.Abstractions` (or appropriate layer)
- [ ] Workload-contributed commands flow through this wrapper, not raw `Command`
- [ ] `Parser` consumes the wrapper and names the offending workload(s) in conflict diagnostics
- [ ] Existing built-in commands and tests keep passing
- [ ] Unit test: two fixture workloads contributing the same command name produces a diagnostic naming both
- [ ] Documentation in `docs/building-a-workload.md` updated to show the new registration shape

## Blocked by

- #4897

## Background

Originated from review feedback on PR #4894 (Jacob): "isn't clear here. Is there a way we can track down what workload added what command?"

liliankasem:
Implemented by #4941 (workload command ownership + DI wiring).

PR #4941, opened 2026-05-06
Wire installed workloads into host DI at startup

### LINKED ISSUE #4902: Workload engine: wire loaded workloads into host DI at startup
## What to build

The loader produces hydrated `IWorkload` instances, but nobody currently calls `instance.Configure(builder)` against the host. Without this wiring, installed workloads cannot register `IProjectInitializer`s, top-level `Command`s, or any other DI services — so `func init` still reports "no workloads installed" even after a successful install.

Wire the loader into `Program.cs` so that during host construction, every loaded workload gets `Configure(FunctionsCliBuilder)` invoked against the real builder.

## Acceptance criteria

- [ ] `Program.cs` (or `WorkloadRegistration.RegisterWorkloads`) loads installed workloads via `WorkloadLoader` and calls `Configure(builder)` on each
- [ ] Loader failures are surfaced as `GracefulException` and abort startup with a clear message
- [ ] Built-in commands that consume workload contributions (e.g. `InitCommand` consuming `IEnumerable<IProjectInitializer>`) see registrations from installed workloads
- [ ] Workload-contributed top-level `Command`s appear in the root command tree
- [ ] Integration test: a fixture workload registers an `IProjectInitializer`; running `func init` with that workload installed exercises the initializer

## Blocked by

- #4897

## Notes

This is the missing link that turns the storage + loader + install plumbing into actual user-visible behavior. Without it, the engine is fully built but inert.

liliankasem:
Resolves #4902.

Replaces the no-op `WorkloadRegistration` stub with the real load+Configure flow so installed workloads' commands and services actually show up in the running CLI.

## Boot pipeline

1. `Program.cs` → `CliHostFactory.CreateHostAsync` (single composition path, also used by tests with a config-override hook for `Workloads:Home`).
2. `WorkloadRegistration.RegisterWorkloadsAsync`:
   - Reads `~/.azure-functions/workloads.json` via `WorkloadStore`.
   - **Skips content-only entries** (workload-spec §6.2 step 3).
   - **Picks highest semver per `packageId`** via `NuGet.Versioning` (workload-spec §6.2 step 2).
   - **Loads each entry one-at-a-time** with try/catch + `IInteractionService.WriteWarning`, so a single bad workload cannot brick the CLI (workload-spec §6.2 "Failure isolation"). The `IWorkloadLoader.Load` contract stays all-or-nothing per call; per-entry isolation lives in this layer.
   - Publishes `IReadOnlyList<WorkloadInfo>` as a singleton.
   - Calls each `Workload.Configure(builder)` with the same per-workload isolation.
   - Records boot duration to telemetry as `func.workload.boot_duration_ms` (histogram tagged with `func.workload.count`).

## Schema changes (foundation for content-only / install)

- `WorkloadEntry.EntryPoint` and `WorkloadMetadata.EntryPoint` are now nullable.
- Both gain `bool ContentOnly { get; init; }`.
- `WorkloadMetadataReader` short-circuits validation when `contentOnly: true`; any declared `entryPoint` is silently ignored, not an error.

JSON shapes the loader now accepts:

**Per-package `workload.json` (normal):**
```json
{
  "entryPoint": {
    "assemblyPath": "Foo.Workload.dll",
    "type": "Foo.Workload.MyWorkload"
  }
}
```

**Per-package `workload.json` (content-only):**
```json
{ "contentOnly": true }
```

**Global `workloads.json`:**
```json
{
  "$schema": "https://aka.ms/func/workloads/v1/schema.json",
  "workloads": [
    { "packageId": "Foo.Workload", "packageVersion": "1.0.0", "aliases": ["foo"],
      "entryPoint": { "assemblyPath": "Foo.Workload.dll", "type": "Foo.Workload.MyWorkload" } },
    { "packageId": "Azure.Functions.Cli.Workload.Host", "packageVersion": "2.0.0", "aliases": ["host"],
      "contentOnly": true }
  ]
}
```

`assemblyPath` is always a bare filename; the loader resolves `<workloadHome>/workloads/<pkgId>/<pkgVer>/tools/any/<assemblyPath>`.

## Cleanup

`IWorkloadProvider`, `WorkloadProvider`, and `WorkloadProviderTests` are removed (~150 LOC). Eager loading at boot replaces the lazy provider abstraction; `WorkloadListCommand` consumes the singleton list directly.

`Program.cs` slimmed from ~140 lines to ~85: theme/interaction → `CliHostFactory` → `Parser.CreateCommand` → Ctrl+C → `RunAsync`.

## Foundation for install/uninstall (next)

Install/uninstall (#4901 follow-ups) get this for free:
- `WorkloadStore.SaveAsync` / `GetWorkloadsAsync` already exist.
- `ContentOnly` field on both `WorkloadEntry` and `WorkloadMetadata` means install can stamp content-only host-runtime packages with no further schema work.
- "Highest semver per packageId at boot" lives in the runtime, so install can blindly write side-by-side rows and uninstall can blindly delete a row, no "active version" pointer needed.
- Per-entry load-failure isolation means `func workload uninstall <bad>` keeps working even if `<bad>` itself can't load.

## Tests

- New `CliHostFactoryTests` (3 integration tests): happy path, Configure-throws warn-and-skip, empty home.
- New fixture project `Workload.Tests.Fixtures.WithCommand` with a command-registering and a Configure-throwing workload.
- `WorkloadStoreTests` and `WorkloadMetadataReaderTests` updated for nullable `EntryPoint`.

161/161 tests pass; `CI=true` build is warning-clean.

### DIFF
--- a/src/Abstractions/Workloads/WorkloadKind.cs
+++ b/src/Abstractions/Workloads/WorkloadKind.cs
@@ -1,0 +1,34 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Discriminator for the three shapes a workload package can take, declared
+/// in <c>workload.json</c>'s <c>kind</c> field
+/// (workload-package-layout §5.4). Determines whether the loader activates
+/// the package, skips it, or treats it as install-time-only.
+/// </summary>
+public enum WorkloadKind
+{
+    /// <summary>
+    /// Default. Carries a runtime payload and a <see cref="Workload"/>
+    /// entry-point class; the loader activates it and calls
+    /// <see cref="Workload.Configure"/>.
+    /// </summary>
+    Workload = 0,
+
+    /// <summary>
+    /// Carries a payload but no entry-point class. The loader skips
+    /// activation; built-in commands resolve the install directory by
+    /// package id (e.g. the Functions host runtime).
+    /// </summary>
+    Content = 1,
+
+    /// <summary>
+    /// Carries no payload of its own. Bundles other workload-or-content
+    /// packages via the <c>.nuspec</c>'s <c>&lt;dependencies&gt;</c>. The
+    /// loader never activates it.
+    /// </summary>
+    Meta = 2,
+}
--- a/src/Abstractions/Workloads/WorkloadMetadata.cs
+++ b/src/Abstractions/Workloads/WorkloadMetadata.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Text.Json.Serialization;
+
 namespace Azure.Functions.Cli.Workloads;
 
 /// <summary>
--- a/src/Abstractions/Workloads/WorkloadMetadata.cs
+++ b/src/Abstractions/Workloads/WorkloadMetadata.cs
@@ -4,21 +4,27 @@
 namespace Azure.Functions.Cli.Workloads;
 
 /// <summary>
-/// Per-workload manifest authored alongside the <see cref="Workload"/>
-/// implementation and shipped at the root of the workload's NuGet package
-/// (<c>workload.json</c>). Required: a package without this file is not a
-/// valid workload.
-///
-/// Read at install time to discover the entry-point assembly and type, which
-/// the CLI then records in its global workload registry. Distinct from the
-/// global registry itself: this file describes a single workload as authored;
-/// the registry tracks every workload installed on the machine.
+/// Contents of a workload's <c>workload.json</c> manifest. Every workload
+/// package must ship one at its root; packages without it are not valid
+/// workloads and are rejected at install time.
 /// </summary>
 public sealed class WorkloadMetadata
 {
     /// <summary>
-    /// Where to find the <see cref="Workload"/> implementation inside the
-    /// package. Path is relative to the package root.
+    /// JSON Schema URL identifying the manifest format.
     /// </summary>
-    public required EntryPointSpec EntryPoint { get; init; }
+    [JsonPropertyName("$schema")]
+    [JsonPropertyOrder(-1)]
+    public required string Schema { get; init; }
+
+    /// <summary>
+    /// Package shape. Defaults to <see cref="WorkloadKind.Workload"/>.
+    /// </summary>
+    public WorkloadKind Kind { get; init; } = WorkloadKind.Workload;
+
+    /// <summary>
+    /// Entry-point assembly and type. Required for
+    /// <see cref="WorkloadKind.Workload"/>; otherwise null.
+    /// </summary>
+    public EntryPointSpec? EntryPoint { get; init; }
 }
--- a/src/Func/Commands/Workload/WorkloadListCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadListCommand.cs
@@ -8,11 +8,7 @@ using Azure.Functions.Cli.Workloads;
 namespace Azure.Functions.Cli.Commands.Workload;
 
 /// <summary>
-/// Lists workloads recorded in the global registry. Reads from the cached
-/// <see cref="WorkloadInfo"/> list materialized by
-/// <see cref="IWorkloadProvider"/> so display name and description come from
-/// each loaded <see cref="Workloads.Workload"/> instance and the file-backed
-/// registry isn't re-read on every invocation.
+/// Lists workloads from <see cref="IWorkloadProvider"/>.
 /// </summary>
 internal sealed class WorkloadListCommand(
     IInteractionService interaction,
@@ -24,14 +20,14 @@ internal sealed class WorkloadListCommand(
     private readonly IInteractionService _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
     private readonly IWorkloadProvider _workloads = workloads ?? throw new ArgumentNullException(nameof(workloads));
 
-    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    protected override Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
     {
-        IReadOnlyList<WorkloadInfo> workloads = await _workloads.GetWorkloadsAsync(cancellationToken);
+        IReadOnlyList<WorkloadInfo> workloads = _workloads.GetWorkloads();
 
         if (workloads.Count == 0)
         {
             _interaction.WriteHint("No workloads installed.");
-            return 0;
+            return Task.FromResult(0);
         }
 
         IEnumerable<string[]> rows = workloads.Select(w => new[]
@@ -44,6 +40,6 @@ internal sealed class WorkloadListCommand(
         });
 
         _interaction.WriteTable(["ID", "Aliases", "Display Name", "Description", "Version"], rows);
-        return 0;
+        return Task.FromResult(0);
     }
 }
--- a/src/Func/Hosting/CliHostFactory.cs
+++ b/src/Func/Hosting/CliHostFactory.cs
@@ -1,0 +1,85 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Console;
+using Azure.Functions.Cli.Telemetry;
+using Azure.Monitor.OpenTelemetry.Exporter;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Azure.Functions.Cli.Hosting;
+
+/// <summary>
+/// Builds the CLI host with all services, configuration, and workloads wired in.
+/// </summary>
+internal static class CliHostFactory
+{
+    /// <summary>
+    /// Creates and starts the CLI host. Registers built-in commands and
+    /// workload storage, loads installed workloads, and invokes each
+    /// workload's <see cref="Workloads.Workload.Configure"/>. The returned
+    /// host is already started so OpenTelemetry listeners are subscribed and
+    /// hosted services are running before any command code executes.
+    /// </summary>
+    /// <param name="interaction">The interaction service used for warnings, errors, and prompts. Registered as a singleton in the host.</param>
+    /// <param name="configureConfiguration">Optional hook for tests to override configuration sources (e.g. point <c>Workloads:Home</c> at a temp directory).</param>
+    /// <param name="cancellationToken">Cancellation propagated to <see cref="IHost.StartAsync"/> and the workload registration step.</param>
+    public static async Task<IHost> CreateHostAsync(
+        IInteractionService interaction,
+        Action<IConfigurationBuilder>? configureConfiguration = null,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(interaction);
+
+        // Empty builder: skip the default config and logging providers a CLI
+        // doesn't need. The host owns shared lifetimes (currently just the
+        // OpenTelemetry pipeline) so flush + shutdown happen via host disposal.
+        HostApplicationBuilder hostBuilder = Host.CreateEmptyApplicationBuilder(null);
+        hostBuilder.Services.AddSingleton(interaction);
+
+        // Only wire OpenTelemetry when a connection string is available and the
+        // user hasn't opted out; otherwise ActivitySource / Meter calls no-op.
+        if (CliTelemetry.TryGetConnectionString(out string? connectionString))
+        {
+            hostBuilder.Services.AddOpenTelemetry()
+                .ConfigureResource(r => CliTelemetry.ConfigureResource(r))
+                .WithTracing(t => t
+                    .AddSource(CliTelemetry.SourceName)
+                    .AddAzureMonitorTraceExporter(o => o.ConnectionString = connectionString))
+                .WithMetrics(m => m
+                    .AddMeter(CliTelemetry.SourceName)
+                    .AddAzureMonitorMetricExporter(o => o.ConnectionString = connectionString));
+        }
+
+        hostBuilder.Services.AddBuiltInCommands();
+
+        // FUNC_CLI_ prefix is stripped and "__" maps to section nesting, so
+        // FUNC_CLI_Workloads__Home binds to WorkloadPathsOptions.Home.
+        hostBuilder.Configuration.AddEnvironmentVariables(prefix: Constants.EnvironmentVariablePrefix);
+        configureConfiguration?.Invoke(hostBuilder.Configuration);
+        hostBuilder.Services.AddWorkloadStorage();
+
+        // Runs before Build() so workloads can still mutate IServiceCollection.
+        // The boot duration is recorded after StartAsync below — OTel listeners
+        // aren't subscribed until the host starts, so emitting earlier drops it.
+        WorkloadRegistrationResult registration = await WorkloadRegistration.RegisterWorkloadsAsync(
+            hostBuilder.Services,
+            hostBuilder.Configuration,
+            interaction,
+            cancellationToken);
+
+        IHost host = hostBuilder.Build();
+
+        // Start before any command code emits telemetry so OTel listeners are
+        // subscribed.
+        await host.StartAsync(cancellationToken);
+
+        CliTelemetry.Metric.RecordWorkloadBoot(
+            registration.WorkloadCount,
+            registration.ElapsedMilliseconds);
+
+        return host;
+    }
+}
--- a/src/Func/Hosting/WorkloadRegistration.cs
+++ b/src/Func/Hosting/WorkloadRegistration.cs
@@ -1,30 +1,150 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Diagnostics;
+using Azure.Functions.Cli.Console;
+using Azure.Functions.Cli.Telemetry;
 using Azure.Functions.Cli.Workloads;
+using Azure.Functions.Cli.Workloads.Loading;
+using Azure.Functions.Cli.Workloads.Storage;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using NuGet.Versioning;
 
 namespace Azure.Functions.Cli.Hosting;
 
 /// <summary>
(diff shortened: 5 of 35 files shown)


satvu (approved):

src/Abstractions/Workloads/WorkloadMetadata.cs
```diff
@@ -18,7 +18,19 @@ public sealed class WorkloadMetadata
...
+    public EntryPointSpec? EntryPoint { get; init; }
+
+    /// <summary>
+    /// True for content-only workloads (e.g. the host runtime): the package
+    /// ships only files, no entry-point class, no <see cref="Workload.Configure"/>
+    /// call, no contribution points. Built-in commands that know the package
+    /// id resolve the install directory directly when needed. The CLI loader
+    /// skips these entries entirely (workload-spec §3, §6.2 step 3).
+    /// </summary>
+    public bool ContentOnly { get; init; }
```
fabiocav: Noting this as non-blocking: We'll need to change this to a workload "kind" I have the proposal documenting that in PR.

src/Func/Hosting/WorkloadRegistration.cs
```diff
@@ -1,30 +1,164 @@
...
+            }
+            catch (Exception ex)
+            {
+                interaction.WriteWarning(
+                    $"Workload '{entry.PackageId}@{entry.PackageVersion}' could not be loaded: {ex.Message}");
+            }
+        }
+
+        IReadOnlyList<WorkloadInfo> publishedWorkloads = workloads;
+        services.AddSingleton(publishedWorkloads);
```
fabiocav: Left this in a previous review, but we shouldn't be registering a collection here.

src/Func/Commands/Workload/WorkloadListCommand.cs
```diff
@@ -8,33 +8,31 @@
...
+/// Lists workloads recorded in the global registry. The list is materialized
+/// at host startup by <see cref="Hosting.WorkloadRegistration"/> and
+/// published as a singleton, so display name and description come from each
+/// loaded <see cref="Workloads.Workload"/> instance and the file-backed
 /// registry isn't re-read on every invocation.
 /// </summary>
 internal sealed class WorkloadListCommand(
     IInteractionService interaction,
-    IWorkloadProvider workloads)
+    IReadOnlyList<WorkloadInfo> workloads)
```
fabiocav: The previous implementation had a more appropriate approach to capture this.

src/Func/Program.cs
```diff
@@ -10,58 +10,14 @@
...
-WorkloadRegistration.RegisterWorkloads(new DefaultFunctionsCliBuilder(hostBuilder.Services));
-
-using IHost host = hostBuilder.Build();
-
-// Start the host so hosted services (OTel providers) are running and
-// listeners are subscribed before any command code emits telemetry.
-await host.StartAsync();
+// Build + start the host (services, OpenTelemetry, workload load + Configure).
+// Extracted so tests can boot the same composition through CliHostFactory.
+using IHost host = await CliHostFactory.CreateHostAsync(interaction);
```
fabiocav: This path can throw a `GracefulException`, but that won't be caught here.

src/Func/Hosting/WorkloadRegistration.cs
```diff
@@ -1,30 +1,164 @@
...
+        IReadOnlyList<WorkloadEntry> liveEntries = SelectLiveEntries(allEntries);
+
+        var workloads = new List<WorkloadInfo>(liveEntries.Count);
+        foreach (WorkloadEntry entry in liveEntries)
+        {
+            try
+            {
+                // Per-entry call so a single failure surfaces as a warning
+                // without aborting discovery for the others.
+                IReadOnlyList<WorkloadInfo> loaded = loader.Load([entry]);
```
fabiocav, line 78: Workloads isolation is not transactional. We can flag this for now, but workloads may register services before throwing, which would leave partial registrations.

src/Func/Hosting/WorkloadRegistration.cs
```diff
@@ -1,30 +1,164 @@
...
+                // Isolate failures: a misbehaving workload should not brick
+                // the CLI. The user must still be able to invoke
+                // `func workload uninstall` to remove it.
+                interaction.WriteWarning(
+                    $"Workload '{workload.PackageId}@{workload.PackageVersion}' failed to initialize: {ex.Message}");
+            }
+        }
+
+        stopwatch.Stop();
+        CliTelemetry.Metric.RecordWorkloadBoot(workloads.Count, stopwatch.ElapsedMilliseconds);
```
fabiocav: The OTel pipeline won't be started at this point.

src/Func/Program.cs
fabiocav: Other paths also throw. The failure paths should be reviewed here.

src/Func/Commands/Workload/WorkloadListCommand.cs
liliankasem: Reverting back to IWorkloadProvider

src/Func/Hosting/WorkloadRegistration.cs
liliankasem: Will add a TODO here to handle this in another pr

src/Abstractions/Workloads/WorkloadKind.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Discriminator for the three shapes a workload package can take, declared
+/// in <c>workload.json</c>'s <c>kind</c> field
+/// (workload-package-layout §5.4). Determines whether the loader activates
+/// the package, skips it, or treats it as install-time-only.
+/// </summary>
+public enum WorkloadKind
```
jviau, line 12: Does this need to be public? Is anyone expected to consume this outside of our own tooling? Workloads will use it in msbuild, but will they use it in code?

src/Abstractions/Workloads/WorkloadMetadata.cs
```diff
@@ -4,21 +4,27 @@
...
-///
-/// Read at install time to discover the entry-point assembly and type, which
-/// the CLI then records in its global workload registry. Distinct from the
-/// global registry itself: this file describes a single workload as authored;
-/// the registry tracks every workload installed on the machine.
+/// Contents of a workload's <c>workload.json</c> manifest. Every workload
+/// package must ship one at its root; packages without it are not valid
+/// workloads and are rejected at install time.
 /// </summary>
 public sealed class WorkloadMetadata
```
jviau, line 13: This as well - does the workload metadata API surface need to be public?

src/Func/Hosting/CliHostFactory.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// Creates and starts the CLI host. Registers built-in commands and
+    /// workload storage, loads installed workloads, and invokes each
+    /// workload's <see cref="Workloads.Workload.Configure"/>. The returned
+    /// host is already started so OpenTelemetry listeners are subscribed and
+    /// hosted services are running before any command code executes.
+    /// </summary>
+    /// <param name="interaction">The interaction service used for warnings, errors, and prompts. Registered as a singleton in the host.</param>
+    /// <param name="configureConfiguration">Optional hook for tests to override configuration sources (e.g. point <c>Workloads:Home</c> at a temp directory).</param>
+    /// <param name="cancellationToken">Cancellation propagated to <see cref="IHost.StartAsync"/> and the workload registration step.</param>
+    public static async Task<IHost> CreateHostAsync(
```
jviau, line 29: Can this be an extension method for `HostApplicationBuilder`, or return `HostApplicationBuilder`? Then you don't need to take in the `Action`. Usage can be:

``` csharp
HostApplicationBuilder builder = CliHost.CreateBuilder(interaction);
// any extract builder configuration

await host.StartAsync(...);
```

src/Func/Hosting/WorkloadRegistration.cs
```diff
@@ -1,30 +1,150 @@
...
+        IConfiguration configuration,
+        IInteractionService interaction,
+        CancellationToken cancellationToken = default)
     {
-        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(services);
+        ArgumentNullException.ThrowIfNull(configuration);
+        ArgumentNullException.ThrowIfNull(interaction);
+
+        var stopwatch = Stopwatch.StartNew();
```
jviau, line 58: Wonder if we should start using diagnostic activities for this? That will have a duration built into it. We can then have an activity on-end hook which will translate them to metrics for us. This way we get 2 pieces of telemetry: the trace and the metric.

If not a trace, then should consider implementing a disposable struct for the duration-based metrics, so you can call:

``` CSHarp
using MyMetric metric = CliTelemetry.Metric.StartWorkloadBoot(...);
```

src/Func/Hosting/CliHostFactory.cs
```diff
@@ -0,0 +1,85 @@
...
+            interaction,
+            cancellationToken);
+
+        IHost host = hostBuilder.Build();
+
+        // Start before any command code emits telemetry so OTel listeners are
+        // subscribed.
+        await host.StartAsync(cancellationToken);
+
+        CliTelemetry.Metric.RecordWorkloadBoot(
```
jviau, line 79: Curious why this metric is here and not before `hostBuilder.Build()`. The values submitted here are not recording anything from the host build and start.

We also probably want to capture failed starts here and have `error` dimensions on the metric for failure.

src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
```diff
@@ -67,11 +65,51 @@ public WorkloadMetadata Read(string packageDirectory)
...
-        // only have to handle one.
+        // Reject unknown $schema values strictly rather than partially
+        // interpret a future schema.
+        if (!WorkloadManifestSchema.IsPackageManifestSupported(metadata.Schema))
+        {
+            string supported = string.Join(
+                ", ",
+                WorkloadManifestSchema.SupportedPackageManifestSchemas);
+            throw new InvalidWorkloadException(
+                $"'{metadataPath}' has unsupported or missing $schema '{metadata.Schema}'. " +
```
jviau, line 76: We have an existing schema exception message for another scenario. Can we make sure they are worded/structured similarly.

src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
jviau, line 116: Do we need to add the `(required for kind 'workload')`?

src/Func/Workloads/Loading/WorkloadLoader.cs
```diff
@@ -33,42 +33,49 @@ public IReadOnlyList<WorkloadInfo> Load(IReadOnlyList<WorkloadEntry> entries)
...
+        // Defense-in-depth: the metadata reader already rejects rooted paths
+        // and `..` segments, but the on-disk registry stores the same value
+        // and could be tampered with. Refuse anything resolving outside
+        // the content root.
         string contentRootWithSeparator = contentRoot.EndsWith(Path.DirectorySeparatorChar)
             ? contentRoot
             : contentRoot + Path.DirectorySeparatorChar;
         if (!assemblyPath.StartsWith(contentRootWithSeparator, StringComparison.Ordinal))
         {
             throw new GracefulException(
```
jviau, line 56: Should these be `InvalidWorkloadException`?

src/Func/Workloads/Storage/WorkloadJsonContext.cs
```diff
@@ -1,17 +1,26 @@
...
 /// <summary>
-/// System.Text.Json source-generated serialization context for the workload
-/// storage shapes. Build-time generation keeps the JSON path reflection-free
-/// (AOT/trim-friendly) and centralizes naming policy so individual properties
-/// don't carry <c>[JsonPropertyName]</c> attributes.
+/// Source-generated <see cref="JsonSerializerContext"/> for workload storage
+/// shapes. Keeps the JSON path reflection-free (AOT/trim-friendly).
 /// </summary>
-[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+[JsonSourceGenerationOptions(
```
jviau, line 14: There is a `UseStringEnumConverter = true` property -- does that not work?

src/Func/Workloads/Storage/WorkloadRegistry.cs
```diff
@@ -47,15 +50,79 @@ internal sealed class WorkloadEntry
...
     /// </summary>
     public IReadOnlyList<string> Aliases { get; init; } = [];
 
     /// <summary>
-    /// Where to find the <see cref="Workload"/> implementation. Copied from
-    /// the package's <see cref="WorkloadMetadata"/> at install time so the
-    /// CLI doesn't have to re-read the package on every load.
+    /// Catalog feed URL or local path the package was installed from.
+    /// </summary>
+    public string Source { get; init; } = string.Empty;
```
jviau, line 69: What is the purpose of tracking `Source`?

src/Func/Workloads/Storage/WorkloadRegistry.cs
```diff
@@ -47,15 +50,79 @@ internal sealed class WorkloadEntry
...
+
+    /// <summary>
+    /// True when installed directly via <c>func workload install &lt;id&gt;</c>;
+    /// false when pulled in as a meta-package member. Defaults to true for legacy entries.
+    /// </summary>
+    public bool InstalledExplicitly { get; init; } = true;
```
jviau, line 75: Should we track exactly which meta package installed this? What if there is overlap between two meta packages, what is the expected behavior when only one is uninstalled?

Or maybe this should have an install ref-count. Both explicit install and meta package install increments ref count:

meta uninstall: decreases ref count, uninstalls if refcount hits 0.
explicit uninstall: always uninstalls, ignores ref count (but increments on install!)

src/Func/Workloads/Storage/WorkloadRegistry.cs
liliankasem: Keeping track for scenarios such as

- Update / repair (func workload update, func workload repair): re-query the same feed/nupkg without the user re-typing it.
- Provenance / audit: distinguish a sideloaded local nupkg from a trusted feed install when troubleshooting.
- Diagnostics: show in a future func workload list --verbose.

src/Abstractions/Workloads/WorkloadKind.cs
liliankasem: should be internal - thanks for catching!

liliankasem: Good call. Moved to Func and made internal in #4953. (Addressed in #4953)

src/Abstractions/Workloads/WorkloadMetadata.cs
liliankasem: Same: moved to Func and made internal alongside WorkloadKind. (Addressed in #4953)

src/Func/Hosting/CliHostFactory.cs
liliankasem: Done in #4953. `CliHostFactory.CreateHostAsync` is now `CliHost.CreateBuilder` returning `HostApplicationBuilder`, with `builder.RegisterWorkloadsAsync()` as a follow-up extension call. Tests no longer need a config callback - they just mutate `builder.Configuration` directly.

src/Func/Hosting/WorkloadRegistration.cs
liliankasem: Adopted the activity option in #4953. `WorkloadRegistration` now opens a `cli.workload.boot` activity tagged with the workload count, and a new `WorkloadBootMetricListener` (an `ActivityListener`) translates the stop into the existing histogram with `outcome=success|error`. Both signals stay in sync.

src/Func/Hosting/CliHostFactory.cs
liliankasem: Fixed in #4953 via the activity bridge: the metric is recorded by the listener when the activity stops, and the activity now also captures failures (`outcome=error` tag, exception attached). The activity itself wraps just the registration phase rather than Build/StartAsync, since registration is the variable-cost work.

src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
liliankasem: Aligned in #4953. The `$schema` rejection now uses the same wording structure as `WorkloadStore.ReadRegistryAsync`'s registry-schema message: "The schema 'X' declared by manifest 'path' is not supported. Supported schemas are: ... Check for spelling or try updating the CLI to the latest version."

src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
liliankasem: Dropped in #4953 - the kind=workload branch is the only caller of `ValidateWorkloadEntryPoint`, so the parenthetical was redundant. Message is now just `'path' is missing entryPoint.`

src/Func/Workloads/Loading/WorkloadLoader.cs
liliankasem: Yes, switched in #4953. Workload-load failures now use `InvalidWorkloadException` consistently with discovery and registration.

src/Func/Workloads/Storage/WorkloadJsonContext.cs
liliankasem: Tried it. The SG-generated converter for `UseStringEnumConverter = true` does not apply `PropertyNamingPolicy.CamelCase` to enum value names, so it would emit `"Workload"` instead of the schema-required `"workload"`. Kept the custom converter and added a doc-comment in `WorkloadJsonContext` explaining why.

src/Func/Workloads/Storage/WorkloadRegistry.cs
liliankasem: Already consumed by the install/uninstall PR (#4926) - `WorkloadInstaller.cs` writes it on install and `WorkloadInstallCommand.cs` prints it back as part of the success message for content-only installs. Also reserved for future `func workload update` / `repair` and `workload list --verbose` provenance display, so removing it now would be a schema migration we'd just have to pay back. Keeping as-is.

src/Func/Workloads/Storage/WorkloadRegistry.cs
liliankasem: Adopted the ref-count model in #4953. Replaced `InstalledExplicitly` (bool) with `InstallRefCount` (int): explicit install and each meta-package install bump the count; explicit uninstall always removes; meta uninstall decrements and removes only at zero. Wiring up the meta-uninstall side comes with the meta install command.

src/Func/Hosting/CliHostFactory.cs
liliankasem: Update: walked back the factory removal. `CliHostFactory.CreateHostAsync` is back as the production entry point so Program.cs and tests don't repeat the Build/StartAsync ceremony. Internally it calls the new `builder.RegisterWorkloadsAsync()` extension. Tests that need to mutate the builder still get `CliHostFactory.CreateBuilder` + the extension.

PR #4942, opened 2026-05-06
Adds func pack and func publish 

harshivcodes:
Adds `func pack` and `func publish` support for Go (a first-class language in
 Core Tools), plus follow-up fixes uncovered while testing the end-to-end flow.
 
 **`func pack` / `func publish` for Go**
 - New Go-aware code path that builds the project via `go build`, stages the
   resulting binary alongside `host.json` / function metadata, and produces the
   deployable zip / publish payload.
 - Honors `GOOS` / `GOARCH` so users can cross-compile from any dev machine to
   the target Function App's OS/architecture.
 
 **macOS E2E pipeline fixes**
 - Switched the macOS agent over to Homebrew Go instead of the `GoTool@0` task,
   which ships amd64 binaries that don't run on Apple Silicon agents.
 - Pinned `go@1.24` and overrode `GOROOT` so the brew toolchain is the sole Go
   on PATH during the run.
 
 **Shared `Executable` reliability fix**
 - `Executable.RunAsync` could return before the async stdout/stderr pumps had
   drained for very short-lived processes (e.g. `go version`), causing
   intermittent "Could not find a Go installation" failures on WSL and CI.
   Added a `Process.WaitForExit()` drain after the exit task completes, with a
   regression test in `ExecutableTests`.
 
 **Misc**
 - Re-ordered the language list so Go appears after Python in the `func init`
   / `func new` wizard.

### DIFF
--- a/eng/build/Minified.targets
+++ b/eng/build/Minified.targets
@@ -6,6 +6,7 @@
     <_LanguageWorkers Include="$(PublishDir)\workers\java" />
     <_LanguageWorkers Include="$(PublishDir)\workers\powershell" />
     <_LanguageWorkers Include="$(PublishDir)\workers\node" />
+    <_LanguageWorkers Include="$(PublishDir)\workers\native" />
   </ItemGroup>
 
   <!-- Remove worker directories from minified builds -->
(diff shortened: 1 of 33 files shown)


harshivcodes:
should have been against a go feature support branch.

PR #4943, opened 2026-05-06
Adds func pack and func publish support for Golang

harshivcodes:
Adds `func pack` and `func publish` support for Go (a first-class language in Core Tools), plus follow-up fixes uncovered while testing the end-to-end flow.
 
 **`func pack` / `func publish` for Go**
 - New Go-aware code path that builds the project via `go build`, stages the
   resulting binary alongside `host.json` / function metadata, and produces the
   deployable zip / publish payload.
 - Honors `GOOS` / `GOARCH` so users can cross-compile from any dev machine to
   the target Function App's OS/architecture.
 
 **macOS E2E pipeline fixes**
 - Switched the macOS agent over to Homebrew Go instead of the `GoTool@0` task,
   which ships amd64 binaries that don't run on Apple Silicon agents.
 - Pinned `go@1.24` and overrode `GOROOT` so the brew toolchain is the sole Go
   on PATH during the run.
 
 **Shared `Executable` reliability fix**
 - `Executable.RunAsync` could return before the async stdout/stderr pumps had
   drained for very short-lived processes (e.g. `go version`), causing
   intermittent "Could not find a Go installation" failures on WSL and CI.
   Added a `Process.WaitForExit()` drain after the exit task completes, with a
   regression test in `ExecutableTests`.
 
 **Misc**
 - Re-ordered the language list so Go appears after Python in the `func init`
   / `func new` wizard.


src/Cli/func/Common/Executable.cs
```diff
@@ -122,11 +122,32 @@ public async Task<int> RunAsync(Action<string> outputCallback = null, Action<str
...
                 else
                 {
-                    return await exitCodeTask.WaitAsync(timeout.Value).ConfigureAwait(false);
+                    var exitCode = await exitCodeTask.WaitAsync(timeout.Value).ConfigureAwait(false);
+
+                    if (_streamOutput)
+                    {
+                        // See comment above: drain async output handlers before returning.
+                        Process.WaitForExit();
+                    }
```
liliankasem, line 154: When a caller passes a timeout, we honor it for exitCodeTask but then fall into a parameterless WaitForExit() that has no bound. If the async output pump never completes, this blocks forever even though the caller asked for a timeout. Can we cap the drain (e.g. WaitForExit(5000)) and just move on if it doesn't flush?

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -92,19 +94,100 @@ public static async Task BuildProject(string workingDirectory = null)
...
+                || stderr.Contains("inconsistent vendoring", StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> when <paramref name="binaryName"/> exists in
+        /// <paramref name="workingDirectory"/> and is newer than every <c>.go</c>
+        /// source file plus <c>go.mod</c>/<c>go.sum</c> in the same directory tree.
+        /// Lets <c>func start</c> skip a redundant <c>go build</c> on warm reruns.
+        /// </summary>
+        internal static bool IsBinaryUpToDate(string workingDirectory, string binaryName)
```
liliankasem: This only looks at *.go, go.mod, and go.sum. If someone edits a file referenced via //go:embed (templates, JSON, static assets), the binary will be considered up to date and func start runs stale code with no warning. Either drop the optimization or call out the limitation in the "skipping build" message so users know to delete app to force a rebuild.

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -92,19 +94,100 @@ public static async Task BuildProject(string workingDirectory = null)
...
+            try
+            {
+                binaryWriteTime = File.GetLastWriteTimeUtc(binaryPath);
+            }
+            catch (IOException)
+            {
+                return false;
+            }
+
+            var sources = Directory.EnumerateFiles(workingDirectory, "*.go", SearchOption.AllDirectories);
```
liliankasem: SearchOption.AllDirectories will pick up vendor/, testdata/, and *_test.go. go mod vendor or running tests can bump those mtimes and trigger spurious rebuilds, and test files shouldn't invalidate the runtime binary anyway. Can we exclude vendor/ and *_test.go from the scan?

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        protected internal override void ValidateFunctionApp(string functionAppRoot, PackOptions options)
+        {
+            GoHelpers.AssertGoFunctionAppLayout(functionAppRoot);
+        }
+
+        protected override Task<string> GetPackingRootAsync(string functionAppRoot, PackOptions options)
+        {
+            // Go packs from the function app root without a separate staging dir.
+            return Task.FromResult(functionAppRoot);
```
liliankasem, line 25: We build app directly inside the user's source tree and never clean it up, so pack/publish leaves a binary in their repo that they'll almost certainly commit by accident. Could we either build into a staging dir and zip from there, or have func init for Go drop a .gitignore with app in it?

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -1331,6 +1387,20 @@ private static void InternalListIgnoredFiles(GitIgnoreParser ignoreParser)
 
         private static void InternalListIncludedFiles(GitIgnoreParser ignoreParser)
         {
+            if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Go)
+            {
+                // Go zip is built from the explicit allowlist in GoHelpers.GetPackFiles; print those
+                // files rooted at the function app root so output matches what is actually packaged
+                // even when 'func publish' is run from a subdirectory.
+                var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
+                foreach (var file in GoHelpers.GetPackFiles(functionAppRoot))
```
liliankasem, line 1406: GoHelpers.GetPackFiles returns a hardcoded host.json + app regardless of whether app exists yet. If a user runs --list-included-files before a build, we'll print a path that isn't on disk. Maybe filter by existence, or annotate the missing one.

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        // PackFunctionAsync intentionally not overridden — the default flow goes through
+        // PackHelpers.CreatePackage → ZipHelper.GetAppZipFile, which has a Go branch
+        // that emits the explicit allowlist (host.json + app).
+        public override Task RunAsync()
+        {
+            // Keep this since this subcommand is not meant to be run directly; dispatch happens
+            // via PackAction.RunRuntimeSpecificPackAsync, which calls RunAsync(PackOptions) above.
+            return Task.CompletedTask;
```
liliankasem: This silently returns Task.CompletedTask. If a future refactor ever dispatches the subcommand directly we'd just no-op with no signal. Instead, throw NotSupportedException("Use RunAsync(PackOptions). Dispatch happens via PackAction.RunRuntimeSpecificPackAsync.") so misuse is loud.

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -124,6 +207,167 @@ public static void AssertBinaryExists(string workingDirectory = null)
...
+        /// by <c>func pack</c> and <c>func publish</c> — the produced binary is the
+        /// one that runs on the Linux Functions host, regardless of the dev OS.
+        /// </summary>
+        /// <remarks>
+        /// TODO(arm64): Currently hardcoded to linux/amd64. Linux ARM64 Function Apps
+        /// are not yet supported by Core Tools' Go pipeline. When the ARM resource
+        /// model exposes the target instance architecture, this method should accept
+        /// a target arch and the publish path should reject mismatches up front.
+        /// </remarks>
+        internal static async Task BuildForLinux(string workingDirectory = null)
```
liliankasem, line 168: Just confirming: BuildProject got the IsBinaryUpToDate short-circuit but BuildForLinux deliberately doesn't, right? Want to make sure pack/publish always produces a fresh binary, intentional?

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
```diff
@@ -21,21 +21,29 @@ public static string ResolveFunctionAppRoot(string folderPath)
...
+                return Path.Combine(Environment.CurrentDirectory, $"{Path.GetFileName(functionAppRoot)}.zip");
             }
-            else
+
+            string resolvedPath = Path.Combine(Environment.CurrentDirectory, outputPath);
+
+            // If the user supplied a path ending in .zip, treat it as the destination file
+            // (matching the --help text: "file path where the packed ZIP archive will be created").
+            // Otherwise treat it as a directory and place <functionAppRoot>.zip inside it.
+            if (outputPath.EndsWith(".zip", StringComparison.OrdinalIgnoreCase))
```
liliankasem: Decision is purely on .EndsWith(".zip"). If a user passes --output mydir where mydir already exists as a file, we'll hit Directory.CreateDirectory on a file path and throw an opaque IOException. Worth a clearer error message for that case.

src/Cli/func/Common/Executable.cs
liliankasem: private const int OutputDrainTimeoutMs = 5000;
...
if (!Process.WaitForExit(OutputDrainTimeoutMs))
{
    // log a warning and continue
}

eng/ci/templates/jobs/test-e2e-osx.yml
```diff
@@ -40,6 +40,19 @@ jobs:
...
+  # On macOS, GoTool@0 (in install-tools.yml) is skipped because it only
+  # ships an amd64 Go binary, which can't run on Apple Silicon agents.
+  # Install an arm64-native Go via Homebrew, pinned to the same major.minor
+  # (1.24) used by GoTool@0 on Linux/Windows, so all three OSes test against
+  # a consistent Go version.
+  - bash: |
+      set -euo pipefail
+      brew install go@1.24
+      GO_BIN="$(brew --prefix go@1.24)/bin"
+      echo "##vso[task.prependpath]$GO_BIN"
```
liliankasem, line 57: PR description says we override GOROOT, but the step only prepends to PATH. It works today because GoTool@0 is skipped on Darwin, but if anything else ever sets GOROOT (or a self-hosted agent has one cached) the brew Go will silently use the wrong root. Worth an explicit unset GOROOT (or setting it to $(brew --prefix go@1.24)/libexec) in the same step.

src/Cli/func/Helpers/GoHelpers.cs
liliankasem: Honestly, I don't think we should ship with this optimization at all. It can silently run stale code during local dev/test and that's not worth the saved seconds on go build.

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
```diff
@@ -21,21 +21,29 @@ public static string ResolveFunctionAppRoot(string folderPath)
...
-                // Create directory if it doesn't exist
-                Directory.CreateDirectory(resolvedPath);
-                resolvedPath = Path.Combine(resolvedPath, $"{Path.GetFileName(functionAppRoot)}");
+                return resolvedPath;
             }
 
-            return resolvedPath + ".zip";
+            Directory.CreateDirectory(resolvedPath);
+            return Path.Combine(resolvedPath, $"{Path.GetFileName(functionAppRoot)}.zip");
         }
```
liliankasem, line 41: The change to ResolveOutputPath changes behavior for non-go users as well.

Before: --output myfile.zip → produced cwd/myfile.zip/<appname>.zip (a directory named myfile.zip containing the zip).

After: --output myfile.zip → produces cwd/myfile.zip directly.

The new behavior is obviously what users meant, but it's still a silent behavioral change for all runtimes, not just Go. If anyone has a script depending on the old "directory named foo.zip" quirk, they break with no warning. We should not do this

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
liliankasem: I would say a temp dir is best practice here

liliankasem (changes requested):

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
harshivcodes: i put it in the bin/ directory similar to how dotnet-isolated does it, and that path is added to gitignore by default

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: - BuildProject (func start) — has IsBinaryUpToDate check at line 101, skips rebuild if sources haven't changed. Makes warm func start fast.
 - BuildForLinux (func pack/publish) — no cache check. Always rebuilds because the local cached binary may be for the dev OS (e.g. windows/amd64) while pack/publish needs linux/amd64

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
harshivcodes: good point, this is the current flow:
```
# No --output: zip goes to cwd named after app folder
 func pack
 → ./myapp.zip
 
 # --output with directory: zip placed inside
 func pack --output dist
 → dist/myapp.zip  (creates dist/ if needed)
 
 # --output pointing at an existing file (not a dir)
 func pack --output somefile.txt
 → ERROR: --output 'somefile.txt' refers to an existing file.
    Pass an existing or new directory path instead.
```

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
harshivcodes: Reverted this.

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: ack - removed this since it was only a couple of seconds.

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
harshivcodes: annotating on the missing one.

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
harshivcodes: fixed.

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: removed. binary is rebuilt every time func start is run.

test/Cli/Func.UnitTests/HelperTests/ZipHelperTests.cs
```diff
@@ -214,5 +214,40 @@ private void WriteOutput(string output)
         {
             _output.WriteLine(output);
         }
+
+        [Fact]
+        public async Task GetAppZipFile_GoRuntime_ContainsOnlyHostJsonAndApp()
```
liliankasem, line 219: The new GetAppZipFile_GoRuntime_ContainsOnlyHostJsonAndApp test verifies entries but never checks entry.ExternalAttributes for the 0755/0777 Unix exec bit on app. Without this bit, the binary won't run on the Linux Functions host. This is the most important assertion to add since it guards core deployment behavior.

liliankasem (approved):

src/Cli/func/Actions/LocalActions/PackAction/GoPackSubcommandAction.cs
```diff
@@ -0,0 +1,72 @@
...
+            // directory), surface a clear error rather than letting Directory.CreateDirectory
+            // fail with an opaque IOException ("Cannot create ... because a file ... exists").
+            if (File.Exists(resolvedPath))
+            {
+                throw new CliException(
+                    $"--output '{outputPath}' refers to an existing file. " +
+                    "Pass an existing or new directory path instead.");
+            }
+
+            return base.ResolveOutputPath(functionAppRoot, outputPath);
```
liliankasem, line 53: Minor: when `outputPath` is non-empty we compute `resolvedPath = Path.Combine(Environment.CurrentDirectory, outputPath)` just for the `File.Exists` guard, then `base.ResolveOutputPath` recomputes the exact same path. Consider keeping the local for the guard but inlining the base logic, or just dropping the unused local. Not a correctness issue, just dead duplication.

test/Cli/Func.UnitTests/ActionsTests/InitActionGoTests.cs
```diff
@@ -81,5 +81,91 @@ public void ProgrammingModelHelper_Go_DefaultsToV1()
...
+            });
+
+            var previousEnv = Environment.GetEnvironmentVariable(Constants.FunctionsWorkerRuntime);
+            var previousDir = Environment.CurrentDirectory;
+            var tempDir = Path.Combine(Path.GetTempPath(), "func-test-native-" + Path.GetRandomFileName());
+            Directory.CreateDirectory(tempDir);
+            try
+            {
+                File.WriteAllText(Path.Combine(tempDir, "go.mod"), "module test");
+                Environment.CurrentDirectory = tempDir;
```
liliankasem: These new tests mutate process-global state (`Environment.CurrentDirectory` and the `FUNCTIONS_WORKER_RUNTIME` env var). If xUnit assembly-level parallelization is ever enabled across these test classes, they'll flake against each other and against any other test that reads cwd or that env var.

The pre-existing `StartHostActionTests.ResolveNativeWorkerRuntime_*` tests use `FileSystemHelpers.Override(fileSystem)` for the same scenario. Worth using that pattern here for consistency and isolation.

test/Cli/Func.UnitTests/HelperTests/ZipHelperTests.cs
```diff
@@ -214,5 +214,40 @@ private void WriteOutput(string output)
...
+
+                GlobalCoreToolsSettings.CurrentWorkerRuntime = WorkerRuntime.Go;
+
+                var stream = await ZipHelper.GetAppZipFile(dir, buildNativeDeps: false, BuildOption.Default, noBuild: false);
+
+                Assert.NotNull(stream);
+                using var archive = new ZipArchive(stream, ZipArchiveMode.Read);
+                var entries = archive.Entries.Select(e => e.FullName).OrderBy(n => n).ToArray();
+
+                Assert.Equal(new[] { "bin/app", "host.json" }, entries);
```
liliankasem: This test guards what's packaged, but not the most important regression risk: the exec bit on `bin/app`. Without `0755` (or similar) on the zip entry, the Linux Functions host won't be able to launch the binary after extraction.

Worth adding an assertion like:
```csharp
var appEntry = archive.Entries.Single(e => e.FullName == "bin/app");
// upper 16 bits of ExternalAttributes are Unix mode; check u+x/g+x/o+x
((appEntry.ExternalAttributes >> 16) & 0x49).Should().Be(0x49);
```
The current code looks correct (`executables` is passed through), but a regression here would silently break Linux deploys with no other test catching it.

liliankasem (commented):
A few small things, mostly style and test hygiene. Nice work on the ELF validation and the up-front Go-specific guardrails in publish.

test/Cli/Func.UnitTests/HelperTests/ZipHelperTests.cs
harshivcodes: added.

test/Cli/Func.UnitTests/ActionsTests/InitActionGoTests.cs
harshivcodes: fixed.

liliankasem (approved):

src/Cli/func/Helpers/GoHelpers.cs
```diff
@@ -114,14 +145,180 @@ public static void AssertBinaryExists(string workingDirectory = null)
...
+
+        /// <summary>
+        /// Cross-compiles the user's Go project into a linux/amd64 binary named
+        /// <see cref="GoBinaryName"/> in <paramref name="workingDirectory"/>. Used
+        /// by <c>func pack</c> and <c>func publish</c> — the produced binary is the
+        /// one that runs on the Linux Functions host, regardless of the dev OS.
+        /// </summary>
+        /// <remarks>
+        /// TODO(arm64): Currently hardcoded to linux/amd64. Linux ARM64 Function Apps
+        /// are not yet supported by Core Tools' Go pipeline. When the ARM resource
```
liliankasem: What does this mean? Why wouldn't arm64 be supported?

src/Cli/func/Actions/LocalActions/CreateFunctionAction.cs
```diff
@@ -127,6 +127,18 @@ public override async Task RunAsync()
                 return;
             }
 
+            // Go uses programmatic registration in main.go (app.HTTP, app.Timer, ...);
+            // there are no per-function templates to scaffold, so 'func new' is intentionally not wired up.
+            if (_workerRuntime == WorkerRuntime.Go)
+            {
+                ColoredConsole
+                    .Error
+                    .WriteLine(ErrorColor("The 'func new' command is not supported for the Go runtime."))
```
liliankasem: Is there plans to support this? If so, we should say not yet
```suggestion
                    .WriteLine(ErrorColor("The 'func new' command is not yet supported for the Go runtime."))
```

src/Cli/func/Helpers/GoHelpers.cs
harshivcodes: just that we do an opinionated build today - the TODO was misleading, I just removed that all together.

eng/build/Minified.targets
```diff
@@ -6,7 +6,6 @@
     <_LanguageWorkers Include="$(PublishDir)\workers\java" />
     <_LanguageWorkers Include="$(PublishDir)\workers\powershell" />
     <_LanguageWorkers Include="$(PublishDir)\workers\node" />
-    <_LanguageWorkers Include="$(PublishDir)\workers\native" />
```
liliankasem, line 9: Why removed?

src/Cli/func/Common/Executable.cs
```diff
@@ -139,6 +166,24 @@ public async Task<int> RunAsync(Action<string> outputCallback = null, Action<str
...
+                    Colors.Net.ColoredConsole.WriteLine(VerboseColor(
+                        $"Output drain for '{_exeName}' did not complete within {OutputDrainTimeoutMs}ms; returning exit code with possibly truncated output."));
+                }
+
+                return;
+            }
+
+            // Process exited within timeout; the no-arg overload flushes the async
+            // stdout/stderr event handlers that WaitForExit(int) does not drain.
+            Process.WaitForExit();
```
liliankasem: Wrap in a try/catch

}
catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)
{
    {
        ColoredConsole.WriteLine(VerboseColor($"Output drain for '{_exeName}' failed: {ex.Message}. Returning captured exit code."));
    }
}

liliankasem (approved):
How are the e2e tests going -do you need any help with those?

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -164,13 +164,43 @@ public override async Task RunAsync()
                 throw new CliException($"--build {PublishBuildOption} is not supported for Windows Elastic Premium Function Apps.");
             }
 
+            // Get the WorkerRuntime
+            WorkerRuntimeLanguageHelper.ResolveNativeWorkerRuntime(_secretsManager);
```
liliankasem: Was just thinking about this, and I think its cleaner to fold this into `GetCurrentWorkerRuntimeLanguage` so callers only call one method:

C#
```
public static WorkerRuntime GetCurrentWorkerRuntimeLanguage(ISecretsManager secretsManager, bool refreshSecrets = false)
{
    var setting = Environment.GetEnvironmentVariable(Constants.FunctionsWorkerRuntime)
                  ?? secretsManager.GetSecrets(refreshSecrets)?.FirstOrDefault(...).Value;

    // Map "native" to a concrete runtime via project markers before normal normalization.
    if (string.Equals(setting, "native", StringComparison.OrdinalIgnoreCase))
    {
        return ResolveNativeFromProjectMarkers();  // go.mod -> Go, else throw
    }

    try { return NormalizeWorkerRuntime(setting); }
    catch { return WorkerRuntime.None; }
}
```

Then PackAction / PublishFunctionAppAction / CreateFunctionAction go back to a single call site with no try/catch dance.

This allows us to have just one settings read instead of two when env var is unset, no behavior change for non-Go, and you delete the try/catch (CliException) workaround that exists today only because GetSecrets() is being called twice.

eng/build/Minified.targets
harshivcodes: the native worker config is a single JSON file the host needs for func start

src/Cli/func/Common/Executable.cs
harshivcodes: done.

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
harshivcodes: done.

liliankasem (approved):

satvu (approved):

PR #4949, opened 2026-05-11
Align workload-spec with package-layout doc (kind discriminator)

liliankasem:
## Why

`workload-spec.md` and `workload-package-layout.md` disagreed on the
`workload.json` and `workloads.json` shape. The package-layout doc
introduced a `kind` discriminator (`workload`, `content`, `meta`) and
required `$schema` URLs; the spec still described the older
`contentOnly: true` boolean.

Per the prior decision (package-layout wins), this PR aligns the spec
to match. No behaviour changes; docs only.

## What

- **§3 Definitions:** replace the "Content-only workload" row with a
  single "Workload kind" row covering all three kinds; update the host
  runtime row to reference `kind: content`.
- **§4.1 `workload list`:** describe the kind column instead of
  "normal or content-only".
- **§5.1:** rewrite the content-only paragraph to cover both `content`
  and `meta` non-workload kinds.
- **§5.3 Per-workload manifest:** rewrite to use `$schema` + `kind`
  with three example manifests; defer to the layout doc §5.4 for the
  full schema; document that `entryPoint` is **forbidden** for
  `content` and `meta` (not just ignored).
- **§6.1 install steps 5/7:** validate `$schema`, switch on `kind`,
  reject non-workload entries that declare an entryPoint, stamp
  `kind` and `installedExplicitly` into the registry row, mention
  `metas[]`.
- **§6.2 discovery steps 1/3/7/8:** skip non-`workload` kinds; refer
  to the eager-loaded `IWorkloadProvider` singleton fed by the
  startup load.
- **§6.4, §8, §12:** terminology cleanups ("content-only workloads" ->
  `content` workloads / `kind: content`).
- **§10.1 Registry schema:** add `metas[]` top-level array and
  `MetaEntry` shape; replace `contentOnly` with `kind`; add
  `installedExplicitly` to `WorkloadEntry`.

## What this does NOT change

- `workload-package-layout.md` is the source of truth and is
  unchanged here.
- Behaviour: this is a docs-only PR. The implementation in #4941
  already follows the layout doc's schema.

## Notes

- The two schema URL families (`aka.ms/func/workloads/v1/...` for the
  registry, `aka.ms/func-workloads/package/v1/...` for the package
  manifest) are intentional and match what the layout doc declares;
  unifying them is a separate decision.

### DIFF
--- a/proposed/workload-spec.md
+++ b/proposed/workload-spec.md
@@ -41,9 +41,9 @@
 | Term                  | Meaning                                                                 |
 |-----------------------|-------------------------------------------------------------------------|
 | **Func CLI**          | The `func` binary itself. Owns command parsing, UX, workload lifecycle. Distinct from the Functions host runtime, which is itself delivered as a workload (see §4.6). |
-| **Host runtime**      | `Microsoft.Azure.WebJobs.Script.WebHost`, the process that actually executes Functions. Acquired and versioned as a content-only workload (see below); `func start` knows how to launch it directly. |
+| **Host runtime**      | `Microsoft.Azure.WebJobs.Script.WebHost`, the process that actually executes Functions. Acquired and versioned as a `content` workload (see below); `func start` knows how to launch it directly. |
 | **Workload**          | An installable extension that contributes init/new/pack support and/or commands. Concretely: a NuGet package whose entry-point assembly contains a class deriving from the abstract `Workload` base in `Func.Cli.Abstractions` (see §5). |
-| **Content-only workload** | A workload that ships only files (no entry-point class, no `Configure` call, no contribution points). The CLI installs and tracks it like any other workload, but the loader skips it: it is consumed by built-in commands that know its package id and disk layout. The host-runtime workload is the canonical example. |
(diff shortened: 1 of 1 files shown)


proposed/workload-spec.md
```diff
@@ -984,14 +1048,14 @@ no separate active-version pointer in the registry.
...
+8. **Non-host `content` consumption.** `func start` knows the host
    workload's package id (`Azure.Functions.Cli.Workload.Host`) and
    resolves its install directory directly. There is no general
-   contract today for other built-in commands to consume
-   content-only workloads. If a second content-only workload appears,
-   we'll need either a name-based registry lookup helper or a
-   content-only contribution mechanism (e.g. registering a content
-   directory under a stable key).
+   contract today for other built-in commands to consume `content`
+   workloads. If a second `content` workload appears, we'll need
```
fabiocav: The only tweak I'd make here is that we expect the command (or workload that has a dependency on the content only workload) to define the expected contract with the workload. 

The CLI's responsibility is simple to manage installation/updates/removal and ensure that workload is present as a WorloadEntry.

fabiocav (approved):

PR #4951, opened 2026-05-11
Resolve telemetry command name from ParseResult

fabiocav:
Replace the raw string[] walk in Program.cs with a dedicated CommandNameResolver that walks the matched command up the Parents chain on the ParseResult. The root command is now parsed once and the same ParseResult is reused for InvokeAsync, removing a parallel source of truth for "what command was invoked".

Behavior corrections (telemetry command name):
- `func`                       "unknown" -> "help"
- `func --verbose`             "unknown" -> "version"
- `func badcommand`            "badcommand" -> "unknown"
- `func workload remote add`   "workload remote" -> "workload remote add"
- `func --opt value init`      unchanged: "init" (no more skip-`-` heuristic)

Telemetry cardinality stays bounded: only command paths defined in the parse tree, plus the "unknown" / "version" / "help" sentinels, can be recorded.

Adds 9 xUnit tests covering null-arg guards, root-only fallbacks, top-level / nested / deeply nested subcommands, recursive verbose after a subcommand, and unknown-token parse errors.

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/Func/Telemetry/ActivityExtensions.cs
+++ b/src/Func/Telemetry/ActivityExtensions.cs
@@ -11,31 +11,52 @@ namespace Azure.Functions.Cli.Telemetry;
 /// </summary>
 /// <remarks>
 /// Naming and span shape follow the (experimental) OTel CLI semantic
-/// conventions: <see cref="ActivityKind.Internal"/> kind, span name set to
-/// the subcommand path, and the <c>cli.command.name</c> attribute.
-/// Resource-level attributes (<c>service.name</c>, <c>service.version</c>,
-/// <c>os.*</c>, <c>process.runtime.*</c>) are configured once on the
+/// conventions: <see cref="ActivityKind.Internal"/> kind, fixed operation
+/// name <c>"cli.command"</c>, and the <c>cli.command.name</c> attribute
+/// applied via <see cref="SetCommandName"/> once parsing resolves the
+/// invoked command path. Resource-level attributes
+/// (<c>service.name</c>, <c>service.version</c>, <c>os.*</c>,
+/// <c>process.runtime.*</c>) are configured once on the
 /// <see cref="OpenTelemetry.Resources.ResourceBuilder"/> and inherited by
 /// every span — they should not be set per span here.
 /// </remarks>
 internal static class ActivityExtensions
 {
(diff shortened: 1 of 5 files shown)


liliankasem (approved):

liliankasem (approved):

src/Func/Telemetry/ActivityExtensions.cs
```diff
@@ -11,31 +11,52 @@ namespace Azure.Functions.Cli.Telemetry;
...
-            return activity;
+            return source.StartActivity(CommandActivityName, ActivityKind.Internal);
         }
     }
 
     extension(Activity activity)
     {
+        /// <summary>
+        /// Tags the activity with the resolved CLI command path and sets the
+        /// display name so trace explorers show <c>"workload list"</c> rather
```
jviau, line 49: Want to confirm that app insights will use the `DisplayName` as the mechanism for grouping `DependencyTelemetry` together?

src/Func/Telemetry/CommandNameResolver.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            return string.Join(' ', parts);
+        }
+
+        if (parseResult.Errors.Count > 0)
+        {
+            return "unknown";
+        }
+
+        return parseResult.GetValue(rootCommand.VerboseOption) ? "version" : "help";
```
jviau, line 52: How does `VerboseOption` map to `version` or `help`? Can you explain this?

fabiocav: This simply reflects the current implementation here:

https://github.com/Azure/azure-functions-core-tools/blob/9bdfbbe14592e0533308cc969c3bbecefb988b62/src/Func/Parser.cs#L153-L161

The scope of the PR was to improve the logic to avoid duplication and ensure accuracy of what we end up resolving here for telemetry. We can review this behavior in a separate PR.

src/Func/Telemetry/ActivityExtensions.cs
fabiocav: Confirmed from the Azure Monitor OTel exporter mapping; not yet live-validated in the portal.

PR #4952, opened 2026-05-11
Add NuGet catalog plumbing for workloads

liliankasem:
Adds the catalog/feed client used by upcoming `func workload install <id>`, `func workload update`, and `func workload search`. **Plumbing only**, no command or installer is wired to it yet; that lands in a follow-up PR.

## What's new

- `Workloads/Catalog/`:
  - `IPackageSourceProvider` / `PackageSourceProvider`: `--source` override > configured `Workloads:Catalog:Source` > nuget.org fallback. Single source, not a list. Returns `NuGet.Configuration.PackageSource` directly.
  - `IWorkloadCatalog` / `WorkloadCatalog`: thin coordinator that resolves the source via the provider, caches per-source clients, delegates. `SearchAsync`, `ResolveLatestVersionAsync`, `ResolveVersionAsync` (exact `--version`), and `DownloadAsync`.
  - `NuGetProtocolSourceClient`: wrapper over `SourceRepository`. Handles both HTTPS v3 feeds and local directories through the same v3 protocol stack.
  - `CatalogSearchQuery` record: `Filter`, `IncludePrerelease`, `Skip`, `Take`, `OverrideSource`.
- `Hosting/WorkloadCatalogRegistration.AddWorkloadCatalog()`.
- 233 tests covering source precedence, the v3 wire (substituted `PackageSearchResource` / `FindPackageByIdResource` wired into a real `SourceRepository`), latest-version + exact-version resolve, and download.

## Decisions worth a look

- **Single source at a time, not a list.** The configured source can be overridden per-call with `--source` (governance scenario: private feed, or a local nupkg path for testing). Multi-source aggregation isn't needed; with under ~10 published workloads, "search the configured source, not a fan-out" is simpler and matches the user mental model. The `--source` override (URL or local directory path) is the only knob.
- **Built on `NuGet.Protocol`, aligned with the [workload-package-layout doc](https://github.com/Azure/azure-functions-core-tools/blob/docs/proposed/workload-package-layout.md) §5.5 ("What we use NuGet for, what we own"):** `Repository.Factory.GetCoreV3(PackageSource)` dispatches HTTPS and local sources through the same v3 resource stack; `PackageSearchResource` for search, `FindPackageByIdResource` for download. No `NuGet.Resolver` / `NuGet.PackageManagement` / `NuGet.DependencyResolver.Core` (the doc-listed exclusions). `NuGet.Configuration.PackageSource` for source modeling.
- **`SearchAsync` returns `Task<IReadOnlyList<CatalogSearchResult>>`.** No `AsyncPageable<T>` and no `Azure.Core` dependency: consumers will call once with `Skip`/`Take`, not iterate pages over an aggregated stream.
- **`ResolveVersionAsync(packageId, NuGetVersion version, ...)`** for `install --version <v>`. Returns `null` if the configured source doesn't advertise the version, so the installer can construct a `ResolvedPackage` without leaking per-source clients.
- **`ResolveLatestVersionAsync` returns `ResolvedPackage(PackageId, Version, Source)`** so `DownloadAsync` takes that record directly, no re-resolve needed.
- **Downloads spool to a `DeleteOnClose` temp file** so callers get a seekable `Stream` that `PackageArchiveReader` can consume. Ownership transfers to the caller, which is why `OpenPackageAsync` uses try/catch instead of `await using`.
- **`IWorkloadCatalog` documents that NuGet is a permanent dependency.** The abstraction exists for testability, not for swapping providers later.
- **Per-source-client construction is a `Func<PackageSource, NuGetProtocolSourceClient>`** registered in DI. `WorkloadCatalog` caches by `source.Name` in a `ConcurrentDictionary` so an override + default in the same process don't both rebuild a `SourceRepository`.
- **`AliasTagPrefix = "alias:"`** is duplicated as a private const here vs depending on `WorkloadInstaller.AliasTagPrefix`; keeps this PR pure plumbing; the wiring PR is the right place to centralize it.

## Out of scope (deferred)

- Wiring catalog into `IWorkloadInstaller` or any command (next PR).
- Alias-vs-`--exact` resolution flow (§6.1 step 2 of the spec): orchestration on top of `SearchAsync`; belongs in the command-layer PR.
- `nuget.config` discovery, `NuGet.Credentials` auth/private feeds, signing verification, background staleness check, `PackageMetadataResource` (used by the installer's meta-expansion pipeline per layout doc §6.1).


src/Func/Workloads/Catalog/ISourceClient.cs
```diff
@@ -0,0 +1,32 @@
...
+/// <summary>
+/// Per-source surface the catalog talks to, uniform across v3 NuGet feeds and local folders.
+/// Kept as an interface to give the aggregator a clean substitution seam in tests; production
+/// has exactly two implementations.
+/// </summary>
+internal interface ISourceClient
+{
+    public PackageSource Source { get; }
+
+    public Task<IReadOnlyList<CatalogSearchResult>> SearchAsync(
```
jviau: I recommend either copying Azure.Core's `AsyncPageable` (https://github.com/Azure/azure-sdk-for-net/blob/Azure.Core_1.55.0/sdk/core/Azure.Core/src/AsyncPageable.cs), or just using `Azure.Core` directly to get access to it.

`AsyncPageable` is a great abstraction over `IAsyncEnumerable` to do async pagenation.

src/Func/Workloads/Catalog/ISourceClient.cs
```diff
@@ -0,0 +1,32 @@
...
+        string? query,
+        bool includePrerelease,
+        int skip,
+        int take,
+        CancellationToken cancellationToken);
+
+    public Task<IReadOnlyList<NuGetVersion>> ListVersionsAsync(
```
jviau: We might want `AsyncPageable` here as well.

src/Func/Workloads/Catalog/IWorkloadCatalog.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// Searches the configured sources for workload packages. Results are deduped by
+    /// package id (highest version wins) and ordered highest-version-first.
+    /// </summary>
+    /// <param name="query">Optional free-form query; null or empty returns all matching packages.</param>
+    /// <param name="includePrerelease"><c>true</c> to include prerelease versions.</param>
+    /// <param name="skip">Pagination offset applied per source before merge.</param>
+    /// <param name="take">Maximum number of results requested per source.</param>
+    /// <param name="overrideSource">Optional <c>--source</c> override. When set, only this source is consulted.</param>
+    /// <param name="cancellationToken">Cancellation propagated to all per-source requests.</param>
+    public Task<IReadOnlyList<CatalogSearchResult>> SearchAsync(
```
jviau, line 25: Same here, lets use `AsyncPageable` over `Task<List>` in all places that are paged.

src/Func/Workloads/Catalog/PackageSource.cs
```diff
@@ -0,0 +1,19 @@
...
+/// A configured workload feed: either a v3 NuGet service-index URL or a
+/// local directory containing <c>.nupkg</c> files.
+/// </summary>
+/// <param name="Name">Human-readable identifier surfaced in errors and telemetry.</param>
+/// <param name="Location">
+/// For remote sources, the absolute <c>index.json</c> URI. For local sources,
+/// the absolute directory path expressed as a <c>file://</c> URI.
+/// </param>
+/// <param name="IsLocal">
+/// <c>true</c> when the source is an on-disk directory; <c>false</c> for a
```
jviau: Can the URI not tell you if it is local or not?

``` csharp
file://folder/on/disk
```

The scheme for the URI will be `file`. Or you can make up your own scheme. Then `IsLocal` can be a property that inspects `Location.Scheme`

src/Func/Workloads/Catalog/LocalFolderSourceClient.cs
```diff
@@ -0,0 +1,222 @@
...
+using NuGet.Packaging;
+using NuGet.Versioning;
+
+namespace Azure.Functions.Cli.Workloads.Catalog;
+
+/// <summary>
+/// Local-directory <see cref="ISourceClient"/>: serves <c>.nupkg</c>s from
+/// <see cref="PackageSource.Location"/>, filtered to the <c>FuncCliWorkload</c> package type.
+/// </summary>
+internal sealed class LocalFolderSourceClient(PackageSource source) : ISourceClient
```
jviau: add some argument validation to ensure `PackageSource source` is a local feed.

jviau: File-system based nuget feeds already have a formal definition and structure to them via nuget. They are more than just a flat list of .nupkgs.

We should leverage the existing nuget semantics for file system feeds and not define our own. Their client probably supports it out of the box.

src/Func/Workloads/Catalog/PackageSource.cs
```diff
@@ -0,0 +1,19 @@
...
+/// remote v3 feed.
+/// </param>
+internal sealed record PackageSource(string Name, Uri Location, bool IsLocal);
```
jviau: nuget already has a `PackageSource`, are we able to re-use that?

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,146 @@
...
+namespace Azure.Functions.Cli.Workloads.Catalog;
+
+/// <summary>
+/// HTTP-backed <see cref="ISourceClient"/> built on <c>NuGet.Protocol</c>.
+/// Drives <see cref="PackageSearchResource"/> and
+/// <see cref="FindPackageByIdResource"/> against the v3 service index of
+/// the supplied <see cref="PackageSource"/>, restricting search to the
+/// <c>FuncCliWorkload</c> package type.
+/// </summary>
+internal sealed class NuGetProtocolSourceClient(PackageSource source, SourceRepository repository) : ISourceClient
```
jviau: `SourceRepository` already has a `PackageSource` property, is having the `PackageSource` in the ctor redundant?

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,146 @@
...
+        CancellationToken cancellationToken)
+    {
+        PackageSearchResource searchResource = await _repository.GetResourceAsync<PackageSearchResource>(cancellationToken);
+
+        SearchFilter filter = new(includePrerelease)
+        {
+            PackageTypes = [PackageType],
+        };
+
+        IEnumerable<IPackageSearchMetadata> hits = await searchResource.SearchAsync(
```
jviau, line 39: With the API as is, caller needs to handle pagination and must always supply `take` (page size). With `AsyncPageable` the API and usage becomes much more intuitive. You will have to write your own `AsyncPageable` implementation which wraps the `PackageSearchResource`, but it shouldn't be too hard.

src/Func/Workloads/Catalog/ISourceClient.cs
jviau: Also for these query APIs, make a "Filter" or "Query" record as the input instead of individual parameters:

"take" and "skip" are become `PageSize` and `ContinuationToken` respectively, to follow the `AsyncPageable` format.
``` CSharp
internal record CatalogSearchQuery
{
    public const in DefaultPageSize = 100; // default value of 'take' if not supplied.

    public string? Filter { get; init; }

    public int? PageSize { get; init; } // nullable as you will need to differentiate between default vs user supplied

    public string? ContinuationToken { get; init; }
}
```

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public async Task<IReadOnlyList<NuGetVersion>> ListVersionsAsync(
+        string packageId,
+        CancellationToken cancellationToken)
+    {
+        FindPackageByIdResource findResource = await _repository.GetResourceAsync<FindPackageByIdResource>(cancellationToken);
+        using var cache = new SourceCacheContext();
+
+        IEnumerable<NuGetVersion> versions = await findResource.GetAllVersionsAsync(
```
jviau, line 67: No paging here offered by nuget? Okay, interesting.

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,146 @@
...
+        NuGetVersion version,
+        CancellationToken cancellationToken)
+    {
+        FindPackageByIdResource findResource = await _repository.GetResourceAsync<FindPackageByIdResource>(cancellationToken);
+        using var cache = new SourceCacheContext();
+
+        // Spool to a temp file so the caller gets a seekable stream that
+        // PackageArchiveReader can consume without keeping NuGet.Protocol's
+        // internal buffers alive.
+        string tempPath = Path.Combine(Path.GetTempPath(), $"funccli-workload-{Guid.NewGuid():N}.nupkg");
```
jviau: What about `Path.GetTempFile()`?
```suggestion
        string tempPath = Path.Combine(Path.GetTempPath(), $"funccli-workload-{Path.GetTempFile()}.nupkg");
```

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,146 @@
...
+                throw new WorkloadPackageNotFoundException(
+                    $"Package '{packageId}' {version} was not found on source '{Source.Name}'.");
+            }
+
+            fileStream.Position = 0;
+            return fileStream;
+        }
+        catch
+        {
+            await fileStream.DisposeAsync();
```
jviau, line 113: Any reason you are not doing `await using FileStream fs = new(...);`?

jviau (changes requested):
Biggest changes will be:

1. Can we use nuget types for `PackageSource` and `IPackageSourceProvider`?
2. Local feed behavior - leverage nuget for this as well, as they already support local feeds
3. Adopt `AsyncPageable<T>` from Azure.Core. This gives a much cleaner approach to paged queries.

jviau:
Here is some code that should work for `AsyncPageable`.

``` CSharp
internal interface ISourceClient
{
    // updated API surface
    public AsyncPageable<CatalogSearchResult> SearchAsync(
        CatalogSearchQuery? query = null,
        CancellationToken cancellationToken = default);
}

// the type used by `ISourceClient`
internal record CatalogSearchQuery
{
    public const int DefaultPageSize = 100; // change this value as needed.

    public string? Query { get; init; }

    public bool IncludePrerelease { get; init; }

    public int? PageSize { get; init; }  // we do not 'Take', as take is done via System.Linq.AsyncEnumerable.

    public int? Skip { get; init; } // could also be `string? ContinuationToken` to stay true to the AsyncPageable API
}

// nested class within NuGetProtocolSourceClient
private sealed class PackageSearchPageable(
        SourceRepository repository, CatalogSearchQuery? query, CancellationToken cancellationToken)
        : AsyncPageable<CatalogSearchResult>(cancellationToken)
    {
        public override IAsyncEnumerable<Page<CatalogSearchResult>> AsPages(
            string? continuationToken = null, int? pageSizeHint = null)
        {
            int skip = GetSkip(continuationToken);
            int take = GetTake(pageSizeHint);
            return GetPages(skip, take, CancellationToken);
        }

        private async IAsyncEnumerable<Page<CatalogSearchResult>> GetPages(
            int skip, int take, [EnumeratorCancellation] CancellationToken cancellationToken)
        {
            SearchFilter filter = new(query?.IncludePrerelease ?? false)
            {
                PackageTypes = [PackageType],
            };

            string? nextContinuationToken;
            PackageSearchResource search = await repository.GetResourceAsync<PackageSearchResource>(cancellationToken);
            do
            {
                IEnumerable<IPackageSearchMetadata> hits = await search.SearchAsync(
                    query?.Query,
                    filter,
                    skip,
                    take,
                    NullLogger.Instance,
                    cancellationToken);

                List<CatalogSearchResult> results = [];
                foreach (IPackageSearchMetadata hit in hits)
                {
                    if (hit.Identity?.Id is null || hit.Identity.Version is null)
                    {
                        continue;
                    }

                    results.Add(new CatalogSearchResult(
                        PackageId: hit.Identity.Id.ToLowerInvariant(),
                        LatestVersion: hit.Identity.Version,
                        Title: hit.Title,
                        Description: hit.Description,
                        Aliases: ParseAliases(hit.Tags),
                        Source: repository.PackageSource));
                }

                skip += results.Count;
                nextContinuationToken = results.Count == take ? skip.ToString() : null;
                yield return Page<CatalogSearchResult>.FromValues(results, nextContinuationToken, null!);
            } while (nextContinuationToken is not null);
        }

        private int GetSkip(string? continuationToken)
        {
            if (continuationToken is not null && int.TryParse(continuationToken, out int parsed) && parsed >= 0)
            {
                return parsed;
            }

            return query?.Skip ?? 0;
        }

        private int GetTake(int? pageSizeHint)
        {
            return pageSizeHint ?? query?.PageSize ?? CatalogSearchQuery.DefaultPageSize;
        }
    }
```

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -0,0 +1,197 @@
...
+        return aliases;
+    }
+
+    private sealed class SearchPageable(
+        SourceRepository repository,
+        PackageSource source,
+        CatalogSearchQuery query,
+        CancellationToken cancellationToken) : AsyncPageable<CatalogSearchResult>(cancellationToken)
+    {
+        private readonly SourceRepository _repository = repository;
```
jviau: You don't need to make these fields if you aren't doing any arg validation. The methods can directly reference the ctor values.

src/Func/Workloads/Catalog/WorkloadCatalog.cs
```diff
@@ -0,0 +1,209 @@
...
+            return byId.Values
+                .OrderByDescending(r => r.LatestVersion)
+                .ThenBy(r => r.PackageId, StringComparer.Ordinal)
+                .ToList();
+        }
+
+        private static async Task<IEnumerable<CatalogSearchResult>> DrainAsync(AsyncPageable<CatalogSearchResult> pageable)
+        {
+            List<CatalogSearchResult> results = [];
+            await foreach (CatalogSearchResult item in pageable.ConfigureAwait(false))
```
jviau: Using `System.Linq.AsyncEnumerable` this is a `.ToListAsync()` call.

src/Func/Workloads/Catalog/WorkloadCatalog.cs
```diff
@@ -0,0 +1,209 @@
...
+                yield return Page<CatalogSearchResult>.FromValues(slice, nextToken, response: null!);
+
+                if (nextToken is null)
+                {
+                    yield break;
+                }
+            }
+        }
+
+        private async Task<IReadOnlyList<CatalogSearchResult>> MaterialiseAsync()
```
jviau: One of the purposes of `AsyncPageable` (and `IAsyncEnumerable`), is to avoid having to realize these large collections in memory all at once. Instead, you are ever only realizing single pages. Additionally, you want to be able to process results as the next page is being asynchronously fetched.

Realizing them all here defeats both of those goals. Is there a way we can avoid that? What is the experience that necessetates eagerly loading them all?

fabiocav (commented):
Not yet reviewed, but for all NuGet actions, let's please ensure we're using the NuGet packages/types defined in https://github.com/Azure/azure-functions-core-tools/blob/docs/proposed/workload-package-layout.md

src/Func/Workloads/Catalog/WorkloadCatalog.cs
liliankasem: Going to revert the pageable changes

src/Func/Workloads/Catalog/CatalogSearchQuery.cs
```diff
@@ -0,0 +1,28 @@
...
+    public bool IncludePrerelease { get; init; }
+
+    /// <summary>Number of results to skip; defaults to 0.</summary>
+    public int Skip { get; init; }
+
+    /// <summary>Maximum number of results to return; defaults to <see cref="DefaultTake"/>.</summary>
+    public int? Take { get; init; }
+
+    /// <summary>Optional <c>--source</c> override.</summary>
+    public string? OverrideSource { get; init; }
```
jviau: nit: name as `Source`

```suggestion
    public string? Source { get; init; }
```

src/Func/Workloads/Catalog/IPackageSourceProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// then the nuget.org default.
+    /// </summary>
+    /// <param name="overrideSource">
+    /// Optional explicit source from <c>--source</c>: a v3 <c>index.json</c>
+    /// URL or a local directory path.
+    /// </param>
+    /// <exception cref="ArgumentException">
+    /// The resolved source is not a recognisable URL or existing directory.
+    /// </exception>
+    public PackageSource GetSource(string? overrideSource = null);
```
jviau: nit: drop `override`. Add a comment that `null` will use a default/configured source.
```suggestion
    public PackageSource GetSource(string? source = null);
```

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
jviau: ah only noticed now the stream is being returned out

src/Func/Workloads/Catalog/WorkloadCatalog.cs
```diff
@@ -0,0 +1,99 @@
...
+/// delegates the catalog operations workload commands need: search,
+/// version resolution, and package download.
+/// </summary>
+internal sealed class WorkloadCatalog(
+    IPackageSourceProvider sourceProvider,
+    Func<PackageSource, NuGetProtocolSourceClient> clientFactory) : IWorkloadCatalog
+{
+    private readonly IPackageSourceProvider _sourceProvider = sourceProvider ?? throw new ArgumentNullException(nameof(sourceProvider));
+    private readonly Func<PackageSource, NuGetProtocolSourceClient> _clientFactory = clientFactory ?? throw new ArgumentNullException(nameof(clientFactory));
+    private readonly ConcurrentDictionary<string, NuGetProtocolSourceClient> _clients = new(StringComparer.OrdinalIgnoreCase);
```
jviau: A CLI is typically short lived. Is caching going to offer any benefits?

jviau (approved):

src/Func/Workloads/Catalog/WorkloadCatalog.cs
liliankasem: Yeah no need for this - thanks for catching

PR #4953, opened 2026-05-12
Address PR #4941 review feedback

liliankasem:
Follow-up on Jacob's post-merge review of #4941.

### Changes
- **Internal types.** `WorkloadKind`, `WorkloadMetadata`, `EntryPointSpec` move from `Abstractions` to `Func` and become `internal`. Nothing outside the CLI consumes them.
- **`CliHost.CreateBuilder`.** Replaces `CliHostFactory.CreateHostAsync(interaction, configHook, ct)`. Callers now own the boot sequence:
  ```csharp
  HostApplicationBuilder builder = CliHost.CreateBuilder(interaction);
  // any extra builder configuration here
  await builder.RegisterWorkloadsAsync(ct);
  using IHost host = builder.Build();
  await host.StartAsync(ct);
  ```
  Tests no longer need a `configureConfiguration` callback; they mutate `builder.Configuration` directly.
- **Activity-based boot telemetry.** `WorkloadRegistration` opens a `cli.workload.boot` activity tagged with `cli.workload.count`. New `WorkloadBootMetricListener` bridges the activity stop into the existing `cli.workload.boot_duration` histogram with `outcome=success|error`. Trace and metric stay in sync, and failures are now captured.
- **Ref-counted workload entries.** `InstalledExplicitly` (bool) becomes `InstallRefCount` (int). Explicit and meta-package installs both bump the count; explicit uninstall always removes; meta uninstall decrements and removes only at zero. Lays the groundwork for meta install/uninstall.
- **Schema error wording.** `WorkloadMetadataReader`'s `$schema` rejection now mirrors `WorkloadStore.ReadRegistryAsync`'s wording for consistency. Dropped the redundant `(required for kind 'workload')` from the entryPoint missing-field error since the kind branch already gates that path.
- **`InvalidWorkloadException`.** `WorkloadLoader` no longer throws `GracefulException`; it uses `InvalidWorkloadException` like the rest of the workload subsystem.

### Notes / responses left on Jacob's threads
- `Source` is intentionally kept (consumed today by `WorkloadInstallCommand` and reserved for `update`/`repair`/diagnostics).
- `UseStringEnumConverter = true` was tried and not adopted: the SG-generated converter does not apply `JsonNamingPolicy.CamelCase` to enum values, so the schema-required lowercase form would regress. Doc-comment in `WorkloadJsonContext` records why.

### Tests
207/207 pass. Loader and metadata-reader tests updated for the new exception type and schema wording. `CliHostFactoryTests` renamed to `CliHostTests` and updated to drive the new `CreateBuilder` shape.


src/Func/Telemetry/WorkloadBootMetricListener.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    private static void OnActivityStopped(Activity activity)
+    {
+        if (!string.Equals(activity.OperationName, TelemetryConventions.WorkloadBootActivityName, StringComparison.Ordinal))
+        {
+            return;
+        }
+
+        int count = activity.GetTagItem(TelemetryConventions.CliWorkloadCount) is int value ? value : 0;
+        string outcome = activity.Status == ActivityStatusCode.Error
```
jviau: It may be helpful to have more than just success/fail, but also what _kind_ of failure. We can use `error.type` tag (an existing OTel convention: https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/) and populate with the exception recorded on the activity. When non-null / empty string, we infer it as a failed entry.

I am not sure what the `Activity.Fail` extension method we have will all end up setting on the end DependencyTelemetry in app insights. Will we need another lower-cardinality `error.type` for the span that is re-used here?

src/Func/Hosting/CliHostFactory.cs
```diff
@@ -12,21 +12,34 @@
...
+/// extension directly.
 /// </summary>
-internal static class CliHost
+internal static class CliHostFactory
 {
+    /// <summary>
+    /// Creates the builder, registers installed workloads, builds, and starts
+    /// the host.
+    /// </summary>
+    public static async Task<IHost> CreateHostAsync(IInteractionService interaction, CancellationToken cancellationToken = default)
```
jviau, line 28: small naming nit: this just says `Create`, but it also starts the host.

Do we rename this, or have the caller be responsible for starting?

liliankasem: Will introduce a new caller for start

PR #4954, opened 2026-05-12
Add workload catalog consumers: install-by-id, update, search

liliankasem:
Stacks on #4952. **Base branch: `liliankasem/feat/workload-nuget-catalog`.** Rebase to `vnext` once #4952 merges.

## What

Wires the catalog plumbing into three user-facing commands:

1. **`func workload install <package-id>`** now accepts a NuGet package id in addition to a local `.nupkg` path. Dispatch is by extension: `.nupkg` then existing local pipeline, otherwise catalog. New options: `--version|-v`, `--source`, `--include-prereleases` (rejected on local `.nupkg` installs via cross-validator).
2. **`func workload update <package-id>`** (new). Implements the in-place atomic swap from spec §6.4: resolve installed row, resolve new version (catalog), stage + validate, atomic `Directory.Move`, registry remove-old + save-new, delete old install. Failure during staging leaves the existing install untouched. `--major` opts in to crossing major versions; `--version` selects which installed row to update when multiple are present. Per scope, `--all` is omitted (follow-up).
3. **`func workload search [query]`** (new). Renders catalog results as a table (Id, Latest, Description) with `--skip`/`--take` pagination, `--source`, `--include-prereleases`.

## How

- All orchestration lives in `IWorkloadInstaller`. Two new methods: `InstallFromCatalogAsync` (resolve, download, reuse `InstallFromPackageAsync`) and `UpdateAsync` (returns `WorkloadUpdateResult` carrying the new entry, previous version, and a `NoUpdateAvailable` flag for the "already latest" case).
- `WorkloadInstaller` gains an `IWorkloadCatalog` constructor dep.
- Centralized `WorkloadInstaller.AliasTagPrefix`; removed the duplicate `private const` from `LocalFolderSourceClient` and `NuGetProtocolSourceClient`.
- DI: `CliHostFactory` now calls `AddWorkloadCatalog()`; `BuiltInCommands` registers the two new commands; `WorkloadCommand` lists them as subcommands.

## Out of scope (called out in the task)

- `--exact` (alias resolution) on install: separate PR.
- `update --all`: separate PR.
- `--json` output across commands: separate cross-cutting concern.
- `prune`, background staleness, project detection.
- Single-write registry transaction. §6.4 step 5 documents the orphaned-directory case as recoverable via `prune`; an inline comment notes this in `UpdateAsync`.

## Tests

`dotnet test test/Func.Tests/Func.Tests.csproj` -> 267 passed, 0 failed.

New coverage:
- `WorkloadInstallerTests`: catalog resolve+delegate, no-candidate throw, explicit-version download, download-failure cleanup, update happy path / no-update / not-installed / staging-failure rollback / orphaned-old-dir tolerance.
- `WorkloadInstallCommandTests`: dispatch by extension, options forwarding, cross-validator rejection on local installs, package-not-found to `GracefulException`.
- `WorkloadUpdateCommandTests`: arg/option shape, success rendering, "already latest" hint, not-installed graceful exception, missing-arg validation.
- `WorkloadSearchCommandTests`: arg/option shape, empty-results warning, table rendering, option forwarding, default skip/take, description truncation.

## Worker versions

No host dependency change.

### DIFF
--- a/src/Func/Commands/Workload/WorkloadCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadCommand.cs
@@ -17,15 +17,21 @@ internal sealed class WorkloadCommand : FuncCliCommand, IBuiltInCommand
     public WorkloadCommand(
         WorkloadListCommand listCommand,
         WorkloadInstallCommand installCommand,
-        WorkloadUninstallCommand uninstallCommand)
+        WorkloadUninstallCommand uninstallCommand,
+        WorkloadUpdateCommand updateCommand,
+        WorkloadSearchCommand searchCommand)
         : base("workload", "Manage Func CLI workloads.")
     {
         ArgumentNullException.ThrowIfNull(listCommand);
         ArgumentNullException.ThrowIfNull(installCommand);
         ArgumentNullException.ThrowIfNull(uninstallCommand);
+        ArgumentNullException.ThrowIfNull(updateCommand);
+        ArgumentNullException.ThrowIfNull(searchCommand);
 
         Subcommands.Add(listCommand);
         Subcommands.Add(installCommand);
         Subcommands.Add(uninstallCommand);
+        Subcommands.Add(updateCommand);
+        Subcommands.Add(searchCommand);
     }
 }
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -25,7 +25,7 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
 
     public Argument<string> WorkloadArgument { get; } = new("id")
     {
-        Description = "Workload package id or alias to install.",
+        Description = "Workload package id, alias, or path to a local .nupkg.",
     };
 
     public Option<string?> VersionOption { get; } = new("--version", "-v")
@@ -38,9 +38,9 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         Description = "Catalog source URL or local directory to resolve from. Default: the configured catalog.",
     };
 
-    public Option<bool> IncludePrereleasesOption { get; } = new("--include-prereleases")
+    public Option<bool> IncludePrereleaseOption { get; } = new("--prerelease")
     {
-        Description = "Allow prerelease versions when resolving from the catalog.",
+        Description = "Allow prerelease versions when resolving from the catalog. Default: stable only.",
     };
 
     public Option<bool> ForceOption { get; } = new("--force", "-f")
@@ -59,28 +59,13 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         _interaction = interaction ?? throw new ArgumentNullException(nameof(interaction));
         _installer = installer ?? throw new ArgumentNullException(nameof(installer));
 
-        WorkloadArgument.Validators.Add(result =>
-        {
-            string? value = result.GetValue(WorkloadArgument);
-            if (string.IsNullOrWhiteSpace(value))
-            {
-                result.AddError("A workload id is required.");
-            }
-        });
-
-        VersionOption.Validators.Add(result =>
-        {
-            string? value = result.GetValue(VersionOption);
-            if (!string.IsNullOrWhiteSpace(value) && !NuGetVersion.TryParse(value, out _))
-            {
-                result.AddError($"'{value}' is not a valid semver version.");
-            }
-        });
+        WorkloadArgument.AddRequiredIdValidator();
+        VersionOption.AddSemVerValidator();
 
         Arguments.Add(WorkloadArgument);
         Options.Add(VersionOption);
         Options.Add(SourceOption);
-        Options.Add(IncludePrereleasesOption);
+        Options.Add(IncludePrereleaseOption);
         Options.Add(ExactOption);
         Options.Add(ForceOption);
     }
@@ -90,20 +75,22 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         string workload = parseResult.GetValue(WorkloadArgument)!;
         string? versionText = parseResult.GetValue(VersionOption);
         string? source = parseResult.GetValue(SourceOption);
-        bool includePrereleases = parseResult.GetValue(IncludePrereleasesOption);
+        bool includePrerelease = parseResult.GetValue(IncludePrereleaseOption);
         bool exact = parseResult.GetValue(ExactOption);
         bool force = parseResult.GetValue(ForceOption);
 
         try
         {
-            WorkloadInstallResult result = await _installer.InstallFromCatalogAsync(
-                workload,
-                string.IsNullOrEmpty(versionText) ? null : NuGetVersion.Parse(versionText),
-                source,
-                includePrereleases,
-                exact,
-                force,
-                cancellationToken);
+            WorkloadInstallResult result = LooksLikeLocalPackagePath(workload)
+                ? await _installer.InstallFromPackageAsync(workload, force, cancellationToken)
+                : await _installer.InstallFromCatalogAsync(
+                    workload,
+                    string.IsNullOrEmpty(versionText) ? null : NuGetVersion.Parse(versionText),
+                    source,
+                    includePrerelease,
+                    exact,
+                    force,
+                    cancellationToken);
 
             _interaction.WriteSuccess(BuildSuccessMessage(result));
             return 0;
@@ -112,11 +99,15 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         {
             throw new GracefulException(ex.Message, isUserError: true);
         }
-        catch (AmbiguousAliasException ex)
+        catch (AmbiguousPackageMatchException ex)
         {
             throw new GracefulException(ex.Message, isUserError: true);
         }
         catch (InvalidWorkloadException ex)
+        {
+            throw new GracefulException(ex.Message, isUserError: true);
+        }
+        catch (FileNotFoundException ex)
         {
             throw new GracefulException(ex.Message, isUserError: true);
         }
@@ -127,6 +118,16 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
                 isUserError: true);
         }
     }
+
+    /// <summary>
+    /// Returns <c>true</c> when the positional argument looks like a path
+    /// to a <c>.nupkg</c> on disk (i.e. ends in <c>.nupkg</c> and points to
+    /// an existing file). Lets the user install a local package without
+    /// going through the catalog. NuGet package ids cannot legally end in
+    /// <c>.nupkg</c>, so this is unambiguous.
+    /// </summary>
+    private static bool LooksLikeLocalPackagePath(string value)
+        => value.EndsWith(".nupkg", StringComparison.OrdinalIgnoreCase) && File.Exists(value);
 
     private static string BuildSuccessMessage(WorkloadInstallResult result)
     {
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -5,31 +5,52 @@ using System.CommandLine;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Console;
 using Azure.Functions.Cli.Workloads;
+using Azure.Functions.Cli.Workloads.Catalog;
 using Azure.Functions.Cli.Workloads.Discovery;
 using Azure.Functions.Cli.Workloads.Install;
 using Azure.Functions.Cli.Workloads.Storage;
+using NuGet.Versioning;
 
 namespace Azure.Functions.Cli.Commands.Workload;
 
 /// <summary>
-/// <c>func workload install &lt;package&gt;</c>. Installs a workload from a
-/// <c>.nupkg</c> on disk. Feed-based acquisition is not yet supported.
+/// <c>func workload install &lt;package&gt;</c>. Resolves a workload package
+/// id (or alias) through the configured catalog and installs it. Use
+/// <c>--source</c> to point at a local folder or alternate feed.
 /// </summary>
 internal sealed class WorkloadInstallCommand : FuncCliCommand
 {
-    private const string NupkgExtension = ".nupkg";
-
     private readonly IInteractionService _interaction;
     private readonly IWorkloadInstaller _installer;
 
     public Argument<string> WorkloadArgument { get; } = new("id")
     {
-        Description = "Workload to install. Currently must be a path to a .nupkg on disk.",
+        Description = "Workload package id or alias to install.",
+    };
+
+    public Option<string?> VersionOption { get; } = new("--version", "-v")
+    {
+        Description = "Specific semver version to install. Default: the latest stable version in the catalog.",
+    };
+
+    public Option<string?> SourceOption { get; } = new("--source")
+    {
+        Description = "Catalog source URL or local directory to resolve from. Default: the configured catalog.",
+    };
+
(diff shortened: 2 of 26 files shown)


src/Func/Commands/Workload/WorkloadCommand.cs
```diff
@@ -17,15 +17,21 @@ internal sealed class WorkloadCommand : FuncCliCommand, IBuiltInCommand
     public WorkloadCommand(
```
jviau, line 17: I wonder if we should switch to some kind of collection based import:

`IEnumerable<WorkloadSubCommand>`

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -5,33 +5,54 @@
...
+    {
+        Description = "Specific semver version to install. Default: the latest stable version in the catalog.",
+    };
+
+    public Option<string?> SourceOption { get; } = new("--source")
+    {
+        Description = "Catalog source URL or local directory to resolve from. Default: the configured catalog.",
+    };
+
+    public Option<bool> IncludePrereleasesOption { get; } = new("--include-prereleases")
```
jviau: how about `--prerelease`? This aligns with what `dotnet tool install` and `dotnet package add` have.

src/Func/Commands/Workload/WorkloadSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+        Arity = ArgumentArity.ZeroOrOne,
+        Description = "Optional free-form query. Omit to list all workloads.",
+    };
+
+    public Option<string?> SourceOption { get; } = new("--source")
+    {
+        Description = "Catalog source URL or local directory to search.",
+    };
+
+    public Option<bool> IncludePrereleasesOption { get; } = new("--include-prereleases")
```
jviau: nit: no pluralization:
```suggestion
    public Option<bool> IncludePrereleaseOption { get; } = new("--include-prerelease")
```

src/Func/Commands/Workload/WorkloadSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+        Options.Add(IncludePrereleasesOption);
+    }
+
+    protected override async Task<int> ExecuteAsync(ParseResult parseResult, CancellationToken cancellationToken)
+    {
+        string? query = parseResult.GetValue(QueryArgument);
+        string? source = parseResult.GetValue(SourceOption);
+        bool includePrereleases = parseResult.GetValue(IncludePrereleasesOption);
+
+        var searchQuery = new CatalogSearchQuery
```
jviau, line 59: Do we want to add support for `--skip` and `--take` now, or add a TODO for paging support later?

src/Func/Commands/Workload/WorkloadSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public Argument<string?> QueryArgument { get; } = new("query")
+    {
+        Arity = ArgumentArity.ZeroOrOne,
+        Description = "Optional free-form query. Omit to list all workloads.",
+    };
+
+    public Option<string?> SourceOption { get; } = new("--source")
+    {
+        Description = "Catalog source URL or local directory to search.",
```
jviau: I think we should explain that this specifically is a nuget feed only. Not sure if you want to call out that local feeds are also supported, but to me it is kind of implied because that is how nuget feeds have worked for a long time (remote API or local folder)

```suggestion
        Description = "NuGet feed URI to search.",
```

src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -44,33 +65,54 @@ public WorkloadInstallCommand(IInteractionService interaction, IWorkloadInstalle
...
         string workload = parseResult.GetValue(WorkloadArgument)!;
+        string? versionText = parseResult.GetValue(VersionOption);
+        string? source = parseResult.GetValue(SourceOption);
+        bool includePrereleases = parseResult.GetValue(IncludePrereleasesOption);
+        bool exact = parseResult.GetValue(ExactOption);
         bool force = parseResult.GetValue(ForceOption);
 
         try
         {
-            WorkloadInstallResult result = await _installer.InstallFromPackageAsync(workload, force, cancellationToken);
```
jviau, line 69: I think we should still support installing from a `nupkg` on disk. Can we keep this functionality? Specifically, check if `WorkloadArgument` points to a `.nupkg` on disk, if so install it directly

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+        Description = "Allow crossing a major-version boundary. Default: same major only.",
+    };
+
+    public Option<string?> SourceOption { get; } = new("--source")
+    {
+        Description = "Catalog source URL or local directory to resolve from.",
+    };
+
+    public Option<bool> IncludePrereleasesOption { get; } = new("--include-prereleases")
```
jviau: Same feedback here: use just `--prerelease`

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
```diff
@@ -0,0 +1,275 @@
...
+    {
+        try
+        {
+            WorkloadUpdateResult result = await _installer.UpdateAsync(
+                packageId, targetVersion, source, includePrereleases, allowMajor, cancellationToken);
+
+            RenderSingle(result);
+            return 0;
+        }
+        catch (WorkloadPackageNotFoundException ex)
```
jviau, line 188: Maybe we can create an exception helper: `catch (Exception ex) when (ex.IsUserError())`

Or is that making it too general for determining when an exception is user error on a per-command basis?

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
```diff
@@ -0,0 +1,275 @@
...
+            catch (Exception ex) when (
+                ex is WorkloadPackageNotFoundException
+                or FileNotFoundException
+                or InvalidWorkloadException
+                or InvalidOperationException)
+            {
+                // Per-id failure must not block other workloads. Surface
+                // the message and keep going; the final exit code reflects
+                // whether any failed.
+                failed++;
```
jviau: Do you need an int, or can it be a bool?

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -30,23 +33,54 @@ public async Task<IReadOnlyList<CatalogSearchResult>> SearchAsync(
     {
         ArgumentNullException.ThrowIfNull(query);
 
-        PackageSearchResource searchResource = await _repository.GetResourceAsync<PackageSearchResource>(cancellationToken);
         SearchFilter filter = new(query.IncludePrerelease)
         {
-            PackageTypes = [PackageType],
+            // nuget.org and most V3 servers ignore custom package types in
+            // this filter (only Dependency / DotnetTool / Template are
```
jviau: 👀 is there a nuget issue for this? Or is it called out in docs anywhere?

This is _really_ problematic if this is true. We have to search every single package to find all CLI workloads. Filtering client side won't be acceptable. We would have to also page through every package ever trying to find CLI packages. We need SOMETHING to filter on.

src/Func/Commands/Workload/WorkloadSearchCommand.cs
liliankasem: Was going to defer for later - I can add a comment here

src/Func/Workloads/Install/AmbiguousAliasException.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads.Install;
+
+/// <summary>
+/// Thrown when an alias provided to <c>func workload install</c> matches
+/// more than one package id in the catalog (spec §6.1 step 2). The user
+/// must re-run with <c>--exact</c> and a specific package id.
+/// </summary>
+internal sealed class AmbiguousAliasException : Exception
```
jviau:
```suggestion
internal sealed class AmbiguousPackageMatchException : Exception
```

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
liliankasem: Yeah this was the agent solution but I want to double check on this to see if its hallucinating again, let me confirm

src/Func/Commands/Workload/WorkloadInstallCommand.cs
liliankasem: Renamed to `--prerelease` to match `dotnet tool install` / `dotnet package add`. Will update the spec in a follow-up PR.

src/Func/Commands/Workload/WorkloadSearchCommand.cs
liliankasem: Done, renamed to `IncludePrereleaseOption` / `--prerelease`.

src/Func/Commands/Workload/WorkloadSearchCommand.cs
liliankasem: Done, updated to `NuGet feed URI to search.`

src/Func/Commands/Workload/WorkloadInstallCommand.cs
liliankasem: Re-added. Install dispatches to `InstallFromPackageAsync` when the arg ends in `.nupkg` and exists on disk.

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
liliankasem: Done, renamed to `--prerelease` here too.

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
liliankasem: Keeping the per-command narrow catches deliberately. Per AGENTS.md, commands are the boundary that decides which exceptions from their dependencies are user errors; a generic `IsUserError()` helper would flatten that intent and risk hiding genuine bugs as user errors. Happy to revisit if a clear pattern emerges across more commands.

src/Func/Commands/Workload/WorkloadUpdateCommand.cs
liliankasem: Switched to `bool`.

src/Func/Workloads/Install/AmbiguousAliasException.cs
liliankasem: Renamed.

src/Func/Commands/Workload/WorkloadCommand.cs
liliankasem: Deferring for now. Sub-commands today have differing constructor shapes (some need shared services like `IWorkloadStore`, some don't), so a flat `IEnumerable<WorkloadSubCommand>` import would force them all to a common base. Happy to revisit once we have more workload sub-commands and the right shared base shakes out.

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
liliankasem: Confirmed empirically (probe script in session artifacts): nuget.org silently ignores `packageTypeFilter=` (the param NuGet.Client serialises `SearchFilter.PackageTypes` as) and only honours `packageType=` (singular), which matches the [wiki you linked](https://github.com/NuGet/Home/wiki/Search-by-Package-Type-and-Query-Language-Surfacing). No typed `NuGet.Protocol` API exposes the singular form, so the fix hand-builds the V3 `SearchQueryService` URL with `packageType=FuncCliWorkload` and routes through `HttpSource` (auth/retry/proxy/cache via NuGet's pipeline). Search is now V3-only — local folder feeds throw at `PackageSourceProvider`; direct `.nupkg` install still works via the path-vs-id dispatch on `WorkloadInstallCommand`.

src/Func/Commands/Workload/WorkloadOptionValidators.cs
```diff
@@ -0,0 +1,71 @@
...
+using NuGet.Versioning;
+
+namespace Azure.Functions.Cli.Commands.Workload;
+
+/// <summary>
+/// Shared parse-time validators for <c>func workload</c> options and
+/// arguments. Centralized so message wording and parsing rules stay
+/// consistent across install/uninstall/update/search.
+/// </summary>
+internal static class WorkloadOptionValidators
```
jviau, line 14: This is a good justification for having options/arguments just be re-usable statics instead of per-command properties. I believe this is what dotnet cli does (re-usable statics).

But we can leave that for a later PR.

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
```diff
@@ -30,31 +41,17 @@ public async Task<IReadOnlyList<CatalogSearchResult>> SearchAsync(
...
-
-        IEnumerable<IPackageSearchMetadata> hits = await searchResource.SearchAsync(
-            query.Filter ?? string.Empty,
-            filter,
-            query.Skip,
-            query.Take ?? CatalogSearchQuery.DefaultTake,
-            NullLogger.Instance,
-            cancellationToken);
+        int take = query.Take ?? CatalogSearchQuery.DefaultTake;
+        Uri? searchUri = await TryBuildV3SearchUriAsync(query, take, cancellationToken);
```
jviau, line 45: Oh you are building the query request manually now. Can you not use the previous code, but switch from `PackageTypes = "..."` to:

``` CSharp
const string packageTypeFilter = "packageType=FuncCliWorkload";
string query = string.IsNullOrEmpty(query.Filter) ? packageTypeFilter : $"{packageTypeFilter} {query.Filter}";
IEnumerable<IPackageSearchMetadata> hits = await searchResource.SearchAsync(query, ...);
```

src/Func/Commands/Workload/WorkloadOptionValidators.cs
liliankasem: Will defer for a future convo/pr

src/Func/Workloads/Catalog/NuGetProtocolSourceClient.cs
liliankasem: Tried this against nuget.org just now — it actively breaks search:

```
q='cowsay'                                 → totalHits=11   (baseline, unfiltered)
q='packageType=DotnetTool cowsay'          → totalHits=0    (your suggestion)
q='packagetype:DotnetTool cowsay'          → totalHits=11   (colon form, token ignored)
q='packageType:DotnetTool cowsay'          → totalHits=11   (colon form, token ignored)
q='cowsay' &packageType=DotnetTool         → totalHits=6    (separate URL param — what we ship)
```

The server treats the entire q= value as a literal phrase, so packageType=DotnetTool cowsay matches packages whose tokens contain that exact string (which is none). Neither the = nor the : query-language form is honoured for packageType — only the separate URL parameter &packageType= works.

The other constraint is that even if a query-language form did work, we still couldn't reach it through PackageSearchResource.SearchAsync without injecting it into the q value: NuGet.Client's typed SearchFilter.PackageTypes is hardcoded to serialise as &packageTypeFilter= (the param nuget.org silently ignores). So the typed API has no path to &packageType= (singular) at all.

That's why I went with the hand-built URL. It still routes through HttpSource so credential providers / retries / proxy / cache all apply — just bypasses the typed PackageSearchResource because that resource can't emit the right wire format.

PR #4956, opened 2026-05-12
Add workload resolution, project detection, and invocation error contract

liliankasem:
Implements the runtime mechanism for "which workload owns this command invocation?" per spec §5.1, §5.2, and §6.3. **Mechanism only**, no built-in command is rewired to use it yet.

## What's new

**Public** (`Func.Cli.Abstractions`)
- `IProjectResolver` — contribution point a workload registers to claim a directory.
- `EvaluationResult` — `(bool IsMatch, string? Reason, string? WorkerRuntime)` with `Match(...)` / `NoMatch(...)` factories.
- `FunctionsCliBuilder.RegisterProjectResolver(...)`.

**Internal** (`src/Func/Workloads/`)
- `IWorkloadResolver` — §5.2: `--stack` → `FUNCTIONS_WORKER_RUNTIME` → project resolvers. Returns `Resolved` or `None`.
- `IWorkloadInvoker` + `WorkloadProtocolException` — §6.3 error contract: `[<packageId>]` prefix, `GracefulException` passes through verbatim, other exceptions wrap with a "file an issue" hint, `OperationCanceledException` flows through.
- `ILocalSettingsReader` behind an interface (tests don't touch the real filesystem).
- `WorkloadProjectResolverContribution(WorkloadInfo, IProjectResolver)` — owner-aware wrapper, mirrors the command-ownership pattern.

**Cleanup:** deleted the legacy `src/Func/Commands/ProjectDetector.cs`, zero callers on `vnext`.

**Tests:** 265 passed / 0 failed. Sln-wide build clean.

## Decisions

- **Resolution is binary (`Resolved | None`).** Multi-claim outcomes surface as a `None` with a candidate list and a `--stack` hint, not as a separate `Ambiguous` case. Multi-claim implies a registry inconsistency (install should enforce one workload per alias / runtime), not a normal user state.
- **Project-resolver outcome is binary (`Match` / `NoMatch`).** Spec §5.2 listed `yes`/`no`/`maybe`; `maybe` was treated identically to `no` everywhere, so dropped.
- **Runtime-miss does not fall through to project resolvers.** Honour the user's explicit `FUNCTIONS_WORKER_RUNTIME` declaration; return `None` with a runtime-specific message.
- **`WorkloadResolutionContext.SkipDirectoryDetection`** for `func init` (no project to evaluate against, spec §4.2).

## Spec deviations (doc PR to follow)

User-facing behaviour matches spec §4.3 / §4.4 / §5.2 / §7: multi-claim is an error listing candidates with a `--stack` hint. The resolver's *return shape* deviates from §5.2's prose:

- The contribution point is `IProjectResolver` (returning `EvaluationResult` with `Match`/`NoMatch`), not `IProjectDetector` returning `yes`/`no`/`maybe`.
- Multi-claim is modelled as a regular `None` outcome rather than a distinct `Ambiguous` case.
- Selector / runtime multi-claim handling is specified (spec only addresses resolver-path ambiguity today).

## Out of scope

Rewiring `init` / `new` / `pack` / `start`. First real workload consumer. §4.5 verbose-help labelling. Install-time catalog-side alias ambiguity is handled by #4954 per spec §6.1 step 2; per spec §5.3, alias uniqueness in the installed set is by convention only, so multi-claim is disambiguated at the use site via `--stack` rather than blocked at install.


src/Abstractions/Workloads/IProjectDetector.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Workloads;
+
+/// <summary>
+/// Decides whether a directory is "this workload's" project. Registered by a
+/// workload from <see cref="Workload.Configure"/> via
+/// <see cref="FunctionsCliBuilder.RegisterDetector"/> and consumed by built-in
+/// commands that bind to a workload from a directory (init, new, pack, start).
```
fabiocav: Small nits on comment not being super clear here.

src/Abstractions/Workloads/IProjectDetector.cs
```diff
@@ -0,0 +1,41 @@
...
+///
+/// Public so workload authors can implement it.
```
fabiocav: nit: This comment is not necessary.

src/Abstractions/Workloads/IProjectDetector.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Globs (relative to the working directory) the CLI evaluates as a
+    /// pre-filter before invoking <see cref="DetectAsync"/>. If at least one
+    /// glob matches a file under the working directory, the detector runs;
+    /// if none match, it is skipped. An empty list means the detector is
+    /// always a candidate (used by detectors whose decision is purely
+    /// content- or runtime-based).
+    /// </summary>
+    public IReadOnlyList<string> ProjectMarkers { get; }
```
fabiocav: Is this needed? Shouldn't the CLI delegate all this to the implementer of this interface?

src/Abstractions/Workloads/IProjectDetector.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// <summary>
+    /// Worker-runtime ids this detector claims (e.g. <c>"dotnet-isolated"</c>,
+    /// <c>"python"</c>). Used to short-circuit resolution when
+    /// <c>FUNCTIONS_WORKER_RUNTIME</c> is set in <c>local.settings.json</c>.
+    /// Empty means the detector does not claim any runtime id and is reached
+    /// only via the marker + <see cref="DetectAsync"/> path.
+    /// </summary>
+    public IReadOnlyList<string> WorkerRuntimes { get; }
```
fabiocav: Why a list? When do expect multiple? 

Is this needed on the interface? Can we get this information from the detection/evaluation result?

src/Func/Workloads/Resolution/IDirectoryMarkerMatcher.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Azure.Functions.Cli.Workloads.Resolution;
+
+/// <summary>
+/// Evaluates a detector's <see cref="IProjectDetector.ProjectMarkers"/> globs
+/// against a working directory. Behind an interface so resolver tests do not
+/// touch the real filesystem; production wires up
+/// <see cref="DirectoryMarkerMatcher"/>.
+/// </summary>
+internal interface IDirectoryMarkerMatcher
```
fabiocav: Is this interface needed?

src/Abstractions/Workloads/DetectionResult.cs
```diff
@@ -0,0 +1,20 @@
...
+/// user can see why each detector answered the way it did.
+/// </summary>
+/// <param name="Claimed"><c>true</c> if the detector claims the directory.</param>
+/// <param name="Reason">Optional human-readable justification (e.g. <c>"matched *.csproj"</c>).</param>
+public sealed record DetectionResult(bool Claimed, string? Reason = null)
+{
+    /// <summary>Convenience factory: detector does not claim the directory.</summary>
+    public static DetectionResult No(string? reason = null) => new(false, reason);
+
+    /// <summary>Convenience factory: detector claims the directory.</summary>
```
fabiocav: nit on the `Yes` / `No` names, but no need to block.

src/Abstractions/Workloads/FunctionsCliBuilder.cs
```diff
@@ -98,6 +98,25 @@ public void RegisterCommand(Type commandType)
...
     protected abstract void OnRegisterCommand(Func<IServiceProvider, FuncCommand> factory);
+
+    /// <summary>
+    /// Registers an <see cref="IProjectDetector"/> that participates in
+    /// workload resolution for the current project (init, new, pack, start).
+    /// The host tags the registration with the calling workload so resolver
+    /// diagnostics can attribute results to their owner.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="detector"/> is <c>null</c>.</exception>
+    public void RegisterDetector(IProjectDetector detector)
```
fabiocav: nit: `RegisterProjectDetector`

src/Abstractions/Workloads/FunctionsCliBuilder.cs
```diff
@@ -98,6 +98,25 @@ public void RegisterCommand(Type commandType)
...
+    {
+        ArgumentNullException.ThrowIfNull(detector);
+        OnRegisterDetector(detector);
+    }
+
+    /// <summary>
+    /// Hook implemented by the host to wire <paramref name="detector"/> into
+    /// the workload resolver, tagged with the calling workload's identity.
+    /// </summary>
+    protected abstract void OnRegisterDetector(IProjectDetector detector);
```
fabiocav: Similar naming feedback

src/Abstractions/Workloads/FunctionsCliBuilder.cs
```diff
@@ -98,6 +98,25 @@ public void RegisterCommand(Type commandType)
...
+    /// Registers an <see cref="IProjectDetector"/> that participates in
+    /// workload resolution for the current project (init, new, pack, start).
+    /// The host tags the registration with the calling workload so resolver
+    /// diagnostics can attribute results to their owner.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"><paramref name="detector"/> is <c>null</c>.</exception>
+    public void RegisterDetector(IProjectDetector detector)
+    {
+        ArgumentNullException.ThrowIfNull(detector);
+        OnRegisterDetector(detector);
```
fabiocav: Are we following this `X()` -> `OnX()` pattern for validation?

fabiocav (commented):
Initial comments... we can also sync on a review.

src/Abstractions/Workloads/IProjectDetector.cs
liliankasem: Done in 3f80d4f. Tightened the summary and dropped the visibility-rationale line.

src/Abstractions/Workloads/IProjectDetector.cs
liliankasem: Done in 3f80d4f. Tightened the summary and dropped the visibility-rationale line.

src/Abstractions/Workloads/IProjectDetector.cs
liliankasem: Agreed. Dropped ProjectMarkers in 3f80d4f. The interface is now just DetectAsync; implementers handle their own pre-filtering (file globs, marker presence, etc.) inside the method.

src/Abstractions/Workloads/IProjectDetector.cs
liliankasem: Dropped WorkerRuntimes from the interface in 3f80d4f. The matched runtime now travels back on DetectionResult.WorkerRuntime, so a detector reports it once per directory inspection rather than declaring a static list. The resolver applies FUNCTIONS_WORKER_RUNTIME as a tie-breaker against that value.

src/Func/Workloads/Resolution/IDirectoryMarkerMatcher.cs
liliankasem: No longer needed. Deleted in 3f80d4f along with DirectoryMarkerMatcher (the CLI-side glob pre-filter is gone now that detectors own their own filtering).

src/Abstractions/Workloads/FunctionsCliBuilder.cs
liliankasem: Renamed to RegisterProjectDetector in 3f80d4f.

src/Abstractions/Workloads/FunctionsCliBuilder.cs
liliankasem: Hook is gone in 3f80d4f (see reply on the X/OnX thread).

src/Abstractions/Workloads/FunctionsCliBuilder.cs
liliankasem: No, it's not a pattern we use anywhere else. Inlined in 3f80d4f: RegisterCommand and RegisterProjectDetector are now abstract directly on the base class, with validation done in the concrete DefaultFunctionsCliBuilder.

liliankasem: To follow up: we discussed adopting it broadly and decided against it as a default. The pattern earns its keep when multiple impls share non-trivial validation, or when we want a stable public surface to hang telemetry/tracing on (e.g. a 'workload registered command' activity). Neither applies today. Easy to reintroduce on this class specifically if/when one of those shows up.

src/Abstractions/Workloads/DetectionResult.cs
liliankasem: Renamed in 65434a8: `Yes`/`No` → `Match`/`NoMatch`. Also reframed the type: `IProjectDetector.DetectAsync` → `IProjectResolver.EvaluateAsync` returning `EvaluationResult` (with `IsMatch` instead of `Claimed`). Reads better at the call site.

fabiocav (approved):
Let's get the changes in. We can continue to iterate once merged.

PR #4961, opened 2026-05-13
Expand Dependabot coverage across ecosystems

liliankasem:
## Why

We were missing a proactive feed for new advisories against our pinned dependencies. Dependabot already runs for `dotnet-sdk`; this expands coverage so:

- Security advisories surface under `Security > Dependabot alerts` (org-private until publicly disclosed).
- Dependabot can auto-open PRs to address them.
- Routine version drift gets a weekly PR per group instead of being silently ignored.

## Changes

`.github/dependabot.yml` adds the following ecosystems:

| Ecosystem | Path | Notes |
|-----------|------|-------|
| `nuget` | `/` | All NuGet (CPM transitives included). Grouped: `powershell-workers`, `opentelemetry`, `microsoft-bcl` (`Microsoft.Extensions.*` + `System.*`) |
| `npm` | `/eng/tools/publish-tools/npm` | The `azure-functions-core-tools` npm package |
| `pip` | `/eng/tools/publish-tools` | Publish tooling Python deps |
| `docker` | `/docker` | Container base image |
| `github-actions` | `/` | Workflow action versions |

All on a weekly Wednesday cadence with `open-pull-requests-limit` caps to keep noise down. Existing `dotnet-sdk` ecosystem entry untouched.

**Intentionally skipped:** `test/TestFunctionApps/*` (test fixtures, noisy churn) and `src/Cli/func/StaticResources/requirements.txt` (template embedded in CLI for end users, not our dep).

## Scope changes after review

Dropped from this PR per discussion:

- **Trivy artifact scan**: moved to `liliankasem/feat/trivy-artifact-scan` for a separate team conversation. Trivy scans inside shipped artifact contents; the existing tools (Dependency Review, NuGet audit, Component Governance, Dependabot) are primarily manifest-driven. Worth a broader call on whether to address that gap.
- **Promoting `NU1902` / `NU1903` to CI errors**: reverted. We don't want to halt CI on existing packages with newly-disclosed vulnerabilities. Dependabot will track those instead.
- **OpenTelemetry CPM transitive-pinning overrides**: never landed (would force host transitives to versions the host wasn't built against). Tracked separately at https://github.com/Azure/azure-functions-host/issues/11773.

## What this PR does *not* cover

Dependabot reads manifests we own at the repo level. It does not unzip another ecosystem's package to look at content inside it. Concretely, the following classes of finding will not be flagged:

- `commons-lang3-3.9.jar` packaged as content inside `Microsoft.Azure.Functions.JavaWorker.nupkg`. Dependabot's NuGet scanner sees `Microsoft.Azure.Functions.JavaWorker` and its `.nuspec` dependencies. It does not unzip the nupkg, find the JAR, and identify the Maven coordinates inside.
- `gozip` Go binary inside the NuGet package (well, was, that's gone now).
- Python wheels under `workers/python/.../site-packages/` inside `Microsoft.Azure.Functions.PythonWorker.nupkg`. Dependabot pip scanner only reads `requirements.txt` files we author. It does not reach into `.whl` METADATA inside a NuGet package.

Component Governance has the same shape: its language detectors (NuGet, Maven, Pip, Npm, etc.) are manifest-triggered, so a bare `.jar` or `.whl` shipped without a `pom.xml` / `requirements.txt` next to it is invisible to them. The empirical evidence is that `commons-lang3 3.9` sat unflagged inside the JavaWorker JAR for 7 months while CG was running.

CG does include a `Linux` detector backed by [syft](https://github.com/anchore/syft) that *can* read JAR and wheel contents directly, but it is primarily scoped to scanning container layers (deb/rpm/apk databases). Whether we can point it at our unzipped artifact tree to close the same gap that Trivy would close is worth investigating in the broader team conversation, alongside Trivy.

That investigation is what the parked Trivy branch is a backstop for.

## Out of scope / follow-ups

- Team discussion on artifact-content scanning (Trivy, CG syft path, or otherwise).
- Host package age gate.
- SBOM emission per release.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,3 +32,35 @@ updates:
         patterns:
           - "Microsoft.Extensions.*"
           - "System.*"
+
+  # Publish-tooling npm package (eng/tools/publish-tools/npm).
+  - package-ecosystem: "npm"
+    directory: "/eng/tools/publish-tools/npm"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 5
+
+  # Publish-tooling Python requirements.
+  - package-ecosystem: "pip"
+    directory: "/eng/tools/publish-tools"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 5
+
+  # Container base image.
+  - package-ecosystem: "docker"
+    directory: "/docker"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 5
+
+  # GitHub Actions workflow versions.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 5
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,7 @@ updates:
         patterns:
           - "OpenTelemetry*"
           - "Azure.Monitor.OpenTelemetry.*"
-      microsoft-extensions:
+      microsoft-bcl:
         patterns:
           - "Microsoft.Extensions.*"
+          - "System.*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,24 @@ updates:
         update-types: 
           - "version-update:semver-major"
           - "version-update:semver-minor"
+
+  # NuGet packages (transitives included via central package management).
+  # Security advisories are surfaced through the Security tab regardless of schedule;
+  # the weekly cadence is for routine version drift PRs.
+  - package-ecosystem: "nuget"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 10
+    groups:
+      powershell-workers:
+        patterns:
+          - "Microsoft.Azure.Functions.PowerShellWorker.*"
+      opentelemetry:
+        patterns:
+          - "OpenTelemetry*"
+          - "Azure.Monitor.OpenTelemetry.*"
+      microsoft-extensions:
+        patterns:
+          - "Microsoft.Extensions.*"


jviau:
@liliankasem the existing _Dependency Review_ github workflow was intended to catch PRs with net-new vulnerabilities. But not block CI if an existing package was determined to have a vulnerability.

Additionally, we already have controls for verifying nuget package vulnerabilities. Both our regular build will check for vulnerabilities, and we also have component governance that runs.

I don't know if we should be bringing in yet another tool for this. I'd rather see if we can get the existing tools to work. I also want to make sure we can intentionally differentiate between a PR which introduces a NEW package with a vulnerability, vs an existing package that has newly disclosed vulnerability. We shouldn't halt all of our CI for the later there.

liliankasem:
> @liliankasem the existing _Dependency Review_ github workflow was intended to catch PRs with net-new vulnerabilities. But not block CI if an existing package was determined to have a vulnerability.
> 
> Additionally, we already have controls for verifying nuget package vulnerabilities. Both our regular build will check for vulnerabilities, and we also have component governance that runs.
> 
> I don't know if we should be bringing in yet another tool for this. I'd rather see if we can get the existing tools to work. I also want to make sure we can intentionally differentiate between a PR which introduces a NEW package with a vulnerability, vs an existing package that has newly disclosed vulnerability. We shouldn't halt all of our CI for the later there.

That is a fair point, I just feel like they keep getting ignored or skipped. If a CI raising something when caught, even if it does not block CI, it needs to do more to ensure there is an action item - maybe it can open a PR or GH issue to address it? Thoughts?

jviau:
> maybe it can open a PR or GH issue to address it? Thoughts?

Dependabot should be configurable to auto-open PRs to address disclosed CVEs. We also get Azure Devops workitems, but not GH issues. Using GH issues for this is questionable, as we would want disclosures to not be immediately public until we have a chance to look at it. 

But yes an improved experience is always ideal. I would start with looking at Dependabot. Ensure it is set up correctly. It is supposed to report vulnerabilities here also: https://github.com/Azure/azure-functions-core-tools/security/dependabot

try adding this, or some variation of, to `.github/dependabot.yml`

``` yml
version: 2
updates:
  - package-ecosystem: "nuget"
    directory: "/"
    schedule:
      interval: "weekly"
```

.github/dependabot.yml
```diff
@@ -10,3 +10,24 @@ updates:
...
+      powershell-workers:
+        patterns:
+          - "Microsoft.Azure.Functions.PowerShellWorker.*"
+      opentelemetry:
+        patterns:
+          - "OpenTelemetry*"
+          - "Azure.Monitor.OpenTelemetry.*"
+      microsoft-extensions:
+        patterns:
+          - "Microsoft.Extensions.*"
```
jviau, line 33: consider adding `System.*` here and renaming the group to `microsoft-bcl`

eng/build/Engineering.props
```diff
@@ -3,12 +3,12 @@
...
-    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU5123</WarningsNotAsErrors>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);NU1901;NU5123</WarningsNotAsErrors>
 
-    <!-- Nuget audit as warnings only, even in TreatWarningsAsErrors. -->
-    <!-- Except for in CI, critical will fail the build. -->
-    <WarningsNotAsErrors Condition="'$(CI)' == 'false'">$(WarningsNotAsErrors);NU1904</WarningsNotAsErrors>
-    <WarningsAsErrors Condition="'$(CI)' == 'true'">$(WarningsAsErrors);NU1904</WarningsAsErrors>
+    <!-- Nuget audit failures all break the build in CI: NU1902 (moderate), NU1903 (high), NU1904 (critical). -->
+    <!-- Outside CI they remain warnings to keep local dev unblocked. -->
+    <WarningsNotAsErrors Condition="'$(CI)' == 'false'">$(WarningsNotAsErrors);NU1902;NU1903;NU1904</WarningsNotAsErrors>
```
jviau: Are you sure you want moderate and high severaty of existing packages to grind our CI to a halt? Ideally we want a runway with this, we have X days based on severity to fix it before it starts blocking CI.

Component Governance is supposed to help with that, but it might be for ADO PRs only, not GitHub.

eng/build/Engineering.props
```diff
@@ -18,6 +18,37 @@
     <NuGetAuditMode>all</NuGetAuditMode> <!-- audit transitive dependencies. -->
   </PropertyGroup>
 
+  <!-- Suppress audit advisories that don't apply to net8.0 builds. -->
+  <ItemGroup>
```
jviau: Can we move suppressions to their own props file.

eng/ci/templates/steps/scan-artifact-trivy.yml
jviau, line 1: Do you intend to keep trivy or can we remove it? I'd prefer we keep the same approach to CVEs across all our repos. So if you want to keep it we should have a broader team discussion on that.

eng/build/Engineering.props
liliankasem: > We have X days based on severity to fix it before it starts blocking CI.

That makes sense. But yeah, I do want CI to come to a halt if we are addressing issues in a timely manner. If there is a way to have a timed gate (must resolve in x days or CI will be red) I would love to implement that 

> Component Governance is supposed to help with that, but it might be for ADO PRs only, not GitHub.

ADO is just not front and centre unless your CI is red and you have to go look at the pipeline, I can check the docs to see if there is anything we can go for GH

eng/ci/templates/steps/scan-artifact-trivy.yml
liliankasem: That is fair - we should chat about it. I feel like Trivy fills in a gap here for non-NuGet content shipped inside our packages (the JARs, native binaries, Python wheels); that's the gap that let commons-lang3 3.9 sit unflagged in the JavaWorker JAR for 7 months.

Worth a broader team discussion on whether to address that gap centrally (Trivy or otherwise). For this PR I'll revert the Trivy changes and keep just Dependabot + the NuGet audit gate.

.github/dependabot.yml
liliankasem: Done in 0235f146a — group renamed to `microsoft-bcl` with `System.*` added.

eng/build/Engineering.props
liliankasem: Reverted the NU1902/NU1903 promotion. Only `NU1904` (critical) remains a CI error, matching the original behaviour. Dependabot will pick up moderate/high advisories without halting CI.

A timed gate (suppress for N days then break the build) would be ideal but isn't something `NuGetAudit` or `dependabot.yml` supports natively. Will look into whether Component Governance has a GitHub-surface equivalent.

eng/build/Engineering.props
liliankasem: Moot now — all `NuGetAuditSuppress` entries removed in the scope reduction (since NU1902/1903 are warnings only, suppressions aren't needed to make the build pass). Will revisit if we ever need to suppress critical advisories.

eng/ci/templates/steps/scan-artifact-trivy.yml
liliankasem: Trivy moved out to `liliankasem/feat/trivy-artifact-scan` for that broader team conversation. This PR is now Dependabot-only.

jviau (approved):

PR #5013, opened 2026-05-15
Implement IProjectInitializer for Python, Node, and Go workloads

liliankasem:
Replaces the `NotImplementedException` stubs in the Python, Node, and Go `IProjectInitializer` implementations with real scaffolding. Templates ship as embedded resources in each workload assembly.

The base `func init` flow (PR #5012) already writes a minimal `host.json` (`{ "version": "2.0" }`) and `.func/config.json`; each workload's `InitializeAsync` runs after that and lays down the stack-specific files.

## Python (`--stack python`)

- `function_app.py` (v2 model, anonymous auth)
- `requirements.txt`, `getting_started.md`, `.gitignore`, `local.settings.json` (`runtime=python`)
- Merges `extensionBundle` into `host.json` only when missing.

**Workload-contributed options:**
- `--no-bundle` skips the `extensionBundle` merge entirely.
- `--bundles-channel <GA|Preview|Experimental>` (`-c`) picks the bundle id (default GA).

## Node (`--stack node`, v4 programming model)

- Branches on `context.Language` for JavaScript vs TypeScript.
- `package.json` (with `__PROJECT_NAME__` substituted from `--name` or the directory, sanitised to npm-name shape), `tsconfig.json` for TS.
- `.funcignore`, `.gitignore`, `local.settings.json` (`runtime=node`), empty `src/functions/` dir.
- Merges `extensionBundle` into `host.json` only when missing.
- Runs `npm install` after scaffold by default. Failure is non-fatal so a missing `npm` install does not break init.

**Workload-contributed options:**
- `--no-bundle`, `--bundles-channel <GA|Preview|Experimental>` (`-c`), as above.
- `--skip-npm-install` bypasses the post-scaffold `npm install`.

## Go (`--stack go`)

- `go.mod` pinning `go 1.24` and `github.com/azure/azure-functions-golang-worker v0.2.0-preview` (template placeholder `{ModuleName}` substituted from `--name` or the directory).
- `main.go` boilerplate around `sdk.FunctionApp()` + `worker.Start(app)`. No sample function (`func init` scaffolds the project, not functions).
- `.funcignore`, `.gitignore`, `local.settings.json` (`runtime=native`, matching `GoProjectResolver`).

**Workload-contributed options:**
- `--skip-go-mod-tidy` bypasses the post-scaffold `go mod tidy` (run by default; failure is non-fatal).

## `--language` resolution (InitCommand)

`--language` is now resolved against the selected initializer's `SupportedLanguages`:

- Explicit value: case-insensitive match; rejected with a clear error if not in the list.
- Single supported language: auto-selected.
- Multiple supported languages, interactive: prompted.
- Multiple supported languages, non-interactive: error with a `Re-run with --language <a|b>` hint.

Practical effect: `func init --stack node` now prompts for JavaScript vs TypeScript instead of silently defaulting.

## Implementation notes

- `IProjectInitializer` itself is unchanged.
- Each workload owns a small internal `ProjectFiles` helper (write-if-missing, embedded-template reader, `host.json` JSON-merge). Duplicated per workload to keep the `Abstractions` surface lean.
- `npm install` and `go mod tidy` are exposed as `internal Func<string, CancellationToken, Task<(int, string)>>` test seams so unit tests don't spawn real processes.
- All file writes respect `--force`. Re-running `func init` without `--force` is a no-op for files that already exist; user-customised `extensionBundle` / `host.json` blocks are left alone.
- Bumped `release_notes.md` for each workload.

## Testing

- `dotnet build` (release, full solution): clean, 0 warnings.
- `dotnet test` (full solution): 456 passed (Python 32, Node 36, Go 22, Func 358 incl. 3 new `--language` resolution tests, plus existing DotNet/Workloads suites).

## Open questions

- **Verbose logging from workloads.** Right now only `InitCommand` logs file writes (`· Wrote host.json.`, `· Wrote .func/config.json.`); workload-written files are silent. Adding a `--verbose` per-file log requires either dragging `Spectre.Console` into `Abstractions` (so workloads can take `IInteractionService`), carving a slim console abstraction into `Abstractions`, or threading a callback / progress event through `InitContext`. Worth a team discussion before picking a direction. Not blocking this PR.
- **Go worker runtime moniker.** `GoProjectResolver.WorkerRuntime` is `"native"` and the Go `local.settings.json` template matches. The legacy `feature/go-support` branch maps `WorkerRuntime.Go` to `"go"` in the host. Worth confirming end-to-end whether `native` actually works with the Go SDK worker, or whether `go` is the correct moniker once the host is updated.


src/Workloads/Stacks/Go/GoProjectInitializer.cs
```diff
@@ -2,30 +2,120 @@
...
-/// Stub <see cref="IProjectInitializer"/> for Go. Scaffolding logic
-/// lands in a follow-up PR; this only claims the stack so the host can
-/// route <c>func init --stack go</c> here.
+/// Scaffolds a Go Functions project (<c>go.mod</c>, <c>main.go</c>, <c>.funcignore</c>, <c>.gitignore</c>,
+/// <c>local.settings.json</c>) targeting the Go worker SDK, then optionally runs <c>go mod tidy</c>.
 /// </summary>
 internal sealed class GoProjectInitializer : IProjectInitializer
 {
+    private const string ModuleNamePlaceholder = "{ModuleName}";
+    private const string DefaultModuleName = "function-app";
```
liliankasem: I think this needs to be app for preview go worker, right @harshivcodes

src/Workloads/Stacks/Go/Templates/funcignore
```diff
@@ -0,0 +1,10 @@
+*.js.map
+*.ts
+.git*
+.vscode
+local.settings.json
+test
+getting_started.md
```
swapnil-nagar, line 5: Remove Node Specifics entries and add Go entries.

src/Workloads/Stacks/Node/NodeProjectInitializer.cs
```diff
@@ -2,30 +2,176 @@
...
+        bool noBundle = parseResult.GetValue(NoBundleOption);
+        BundleChannel channel = parseResult.GetValue(BundlesChannelOption);
+        bool skipNpmInstall = parseResult.GetValue(SkipNpmInstallOption);
+
+        string projectName = ResolveProjectName(context);
+        string packageJsonTemplate = isTypeScript
+            ? ProjectFiles.ReadTemplate("package-ts.json")
+            : ProjectFiles.ReadTemplate("package-js.json");
+
+        ProjectFiles.WriteIfMissing(
```
swapnil-nagar, line 72: I can notice NodeProjectInitializer and PythonInitializer has duplicated code. It would be nice to implement with Strategy or Template pattern

src/Workloads/Stacks/Go/GoProjectInitializer.cs
```diff
@@ -2,30 +2,120 @@
...
+            ProjectFiles.ReadTemplate("local.settings.json"),
+            force);
+
+        if (!parseResult.GetValue(SkipGoModTidyOption))
+        {
+            await RunGoModTidy(root, cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    private static string ResolveModuleName(InitContext context)
```
swapnil-nagar, line 88: Same duplicated code across different language, see if it can be optimize.

src/Workloads/Stacks/Go/Templates/go.mod
```diff
@@ -0,0 +1,5 @@
+module {ModuleName}
+
+go 1.24
+
+require github.com/azure/azure-functions-golang-worker v0.2.0-preview
```
swapnil-nagar, line 5: [Checking] We are pinning version?

src/Workloads/Stacks/Go/Templates/go.mod
```diff
@@ -0,0 +1,5 @@
+module {ModuleName}
+
+go 1.24
```
swapnil-nagar, line 3: [Checking] We are pinning version? Is there way to have way to have relaxed version updates like 1.x?

src/Workloads/Stacks/Node/NodeProjectInitializer.cs
liliankasem: Moved some duplicate code to abstractions. We can revisit a heavier base initializer in the future

src/Workloads/Stacks/Go/Templates/go.mod
liliankasem: go 1.24 isn't actually a pin. The go directive specifies a minimum toolchain; users with Go 1.25+ build fine, they just need ≥1.24. This is the standard convention.

src/Workloads/Stacks/Go/Templates/go.mod
liliankasem: I think this is the only way to express a dep in go.mod — Go modules don't support npm-style ranges like 1.x, every dep has to be a single version. It's a starting point: go mod tidy and go get -u resolve updates within SemVer rules, and the version gets re-resolved against the module proxy on go get.

We could run go get github.com/azure/azure-functions-golang-worker@v0 (or @latest) as part of func init to always pull the newest, but that's not offline-friendly and makes scaffolds non-reproducible.

Let's pin for now and revisit a "stay current" strategy post-Build of the Go worker.

src/Workloads/Stacks/Go/Templates/go.mod
harshivcodes: this is a side effect of defining the version in the [worker code](https://github.com/[Azure/azure-functions-golang-worker](https://github.com/Azure/azure-functions-golang-worker))

src/Workloads/Stacks/Go/Templates/funcignore
```diff
@@ -0,0 +1,7 @@
...
```
harshivcodes, line 5: could you also add /bin, func start will produce the go executable and puts it in /bin

src/Workloads/Stacks/Go/Templates/go.mod
harshivcodes: I'd actually push back on pinning here. The reproducibility argument cuts both ways — Go devs expect their project (the one they commit) to be reproducible via go.mod + go.sum, not for func init itself to produce the same SDK version across days. Once they commit, they're pinned; that's the whole point of go.sum.
 
The offline-friendly concern is real but I think it's already moot — func init runs go mod tidy by default, which needs network anyway. Anyone running offline is already opting out with --skip-go-mod-tidy, and at that point they can go get whatever version they want themselves.

Happy to defer to post-Build if you'd rather not churn this PR, just wanted to flag it before the template constant becomes a thing people copy

src/Workloads/Stacks/Go/GoProjectInitializer.cs
harshivcodes: correct

harshivcodes (approved):

src/Workloads/Stacks/Go/Templates/go.mod
liliankasem: Assuming this is the go worker version we are releasing/using for build, let's keep it like this and I will add a TODO here for reworking this. I'll probably hand this off to lang worker owners to maintain and do as they see fit for their stack

swapnil-nagar (approved):

PR #5014, opened 2026-05-15
Implement IProjectInitializer for the DotNet workload

satvu:
Simple `func init` functionality. Does not include everything from legacy [InitAction](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Cli/func/Actions/LocalActions/InitAction/InitAction.cs) and is missing support for Docker (which likely needs a broader design across workloads/CLI project). CLI abstraction added based on instructions from [AGENTS.md](https://github.com/Azure/azure-functions-core-tools/blob/vnext/AGENTS.md).

### Summary

Implements the previously-stubbed `DotNetProjectInitializer.InitializeAsync` to scaffold .NET Azure Functions projects using `dotnet new func` from the `Microsoft.Azure.Functions.Worker.ProjectTemplates` NuGet package. Introduces an `IDotnetCliRunner` abstraction for testable process execution.

### Changes

**New files:**
| File | Description |
|------|-------------|
| `IDotnetCliRunner.cs` | Interface wrapping `dotnet` CLI execution |
| `DotnetCliRunner.cs` | Production implementation with concurrent stdout/stderr reads and process-tree cleanup on cancellation |

**Modified files:**
| File | Description |
|------|-------------|
| `DotNetProjectInitializer.cs` | Replaces `NotImplementedException` stub with full implementation: installs templates, runs `dotnet new func` with language/framework/force options, accepts `IDotnetCliRunner` via primary constructor |
| `DotNetWorkload.cs` | Registers `IDotnetCliRunner` singleton |
| `Workload.DotNet.csproj` | Adds `InternalsVisibleTo` for `DynamicProxyGenAssembly2` (NSubstitute proxy support) |
| `DotNetProjectInitializerTests.cs` | Rewrites tests to use NSubstitute mock; adds `--target-framework` option test, `NormalizeLanguage` theory (10 cases), constructor null-guard test |

### Key design decisions

- **`IDotnetCliRunner`** — process I/O behind an interface per repo conventions (AGENTS.md). Kept internal to the DotNet workload since only this stack needs it.
- **`--force` is conditional** — only passed to `dotnet new` when `context.Force` is true (the stub had no logic; the legacy `InitAction` always passed `--force`).
- **Concurrent stream reads** — `DotnetCliRunner` uses `Task.WhenAll` for stdout/stderr to prevent deadlock on large output.
- **Cancellation safety** — kills the entire process tree on `OperationCanceledException` to avoid orphaned `dotnet` processes.

### Testing

- 21 tests pass, 0 warnings
- All tests use `Substitute.For<IDotnetCliRunner>()` — no real `dotnet` process needed
- Manually tested CLI and created project as expected, hive and workload are created at expected locations.

### DIFF
--- a/src/Abstractions/Common/Constants.cs
+++ b/src/Abstractions/Common/Constants.cs
@@ -1,12 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace Azure.Functions.Cli.Common;
+namespace Azure.Functions.Cli.Abstractions.Common;
 
 /// <summary>
 /// Common constants used across the Azure Functions CLI.
 /// </summary>
-public static class SharedConstants
+public static class Constants
 {
     /// <summary>
     /// Directory name (under the user profile) the func CLI persists state in.
--- a/src/Abstractions/Common/SharedConstants.cs
+++ b/src/Abstractions/Common/SharedConstants.cs
@@ -1,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Common;
+
+/// <summary>
+/// Common constants used across the Azure Functions CLI.
+/// </summary>
+public static class SharedConstants
+{
+    /// <summary>
+    /// Directory name (under the user profile) the func CLI persists state in.
+    /// </summary>
+    public const string FuncHomeDirectoryName = ".azure-functions";
+}
--- a/src/Func/Common/Constants.cs
+++ b/src/Func/Common/Constants.cs
@@ -34,7 +34,6 @@ internal static class Constants
 
     /// <summary>
     /// Prefix for the CLI environment variables, including those that bind into <c>IConfiguration</c>.
-    /// e.g. <c>FUNC_CLI_Workloads__Home</c> binds <c>Workloads:Home</c>.
     /// </summary>
     public const string EnvironmentVariablePrefix = "FUNC_CLI_";
 
--- a/src/Func/Common/Constants.cs
+++ b/src/Func/Common/Constants.cs
@@ -28,11 +28,6 @@ internal static class Constants
     public const string VersionCacheFileName = ".version-check";
 
     /// <summary>
-    /// Directory name (under the user profile) the func CLI persists state in.
-    /// </summary>
-    public const string FuncHomeDirectoryName = ".azure-functions";
-
-    /// <summary>
     /// Prefix for the CLI environment variables, including those that bind into <c>IConfiguration</c>.
     /// </summary>
     public const string EnvironmentVariablePrefix = "FUNC_CLI_";
--- a/src/Func/Common/Constants.cs
+++ b/src/Func/Common/Constants.cs
@@ -34,6 +34,7 @@ internal static class Constants
 
     /// <summary>
     /// Prefix for the CLI environment variables, including those that bind into <c>IConfiguration</c>.
+    /// e.g. <c>FUNC_CLI_Workloads__Home</c> binds <c>Workloads:Home</c>.
     /// </summary>
     public const string EnvironmentVariablePrefix = "FUNC_CLI_";
 
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -139,12 +139,6 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli, ITemp
                 $"The dotnet CLI operation timed out after {OperationTimeout.TotalMinutes:0} minutes. Check your network connection and try again.",
                 isUserError: true);
         }
-        catch (Win32Exception ex)
-        {
-            throw new GracefulException(
-                $"Failed to terminate the dotnet process (Win32Exception: {ex.Message}). You may need to end it manually and retry.",
-                isUserError: true);
-        }
     }
 
     private bool IsHiveFresh()
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -19,7 +19,7 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
     internal const string TemplateShortName = "func";
     internal const string DefaultFramework = "net10.0";
 
-    internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(7);
+    internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(30);
 
     private static readonly string _hivePath = Path.Combine(
         Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -18,10 +18,14 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
     internal const string TemplateShortName = "func";
     internal const string DefaultFramework = "net10.0";
 
+    internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(7);
+
     private static readonly string _hivePath = Path.Combine(
         Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
         "azure-functions-core-tools",
         "dotnet-template-hive");
+
+    private static readonly string _timestampPath = Path.Combine(_hivePath, ".installed");
 
     private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
 
@@ -60,22 +64,56 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
             "--language", language,
             "--Framework", framework,
             "--debug:custom-hive", _hivePath,
+            "--force", // CLI creates host.json before calling this so we need to pass force for the the template to overwrite it.
         ];
-
-        if (context.Force)
-        {
-            args.Add("--force");
-        }
 
         await _dotnetCli.RunAsync(args, projectPath, cancellationToken);
     }
 
     internal async Task EnsureTemplatesInstalledAsync(CancellationToken cancellationToken)
     {
+        if (IsHiveFresh())
+        {
+            return;
+        }
+
         await _dotnetCli.RunAsync(
             ["new", "install", TemplatesPackageName, "--debug:custom-hive", _hivePath],
             workingDirectory: null,
             cancellationToken);
+
+        WriteTimestamp();
+    }
+
+    private static bool IsHiveFresh()
+    {
+        try
+        {
+            if (!File.Exists(_timestampPath))
+            {
+                return false;
+            }
+
+            DateTime installedUtc = File.GetLastWriteTimeUtc(_timestampPath);
+            return DateTime.UtcNow - installedUtc < HiveTtl;
+        }
+        catch (IOException)
+        {
+            return false;
+        }
+    }
+
+    private static void WriteTimestamp()
+    {
+        try
+        {
+            Directory.CreateDirectory(_hivePath);
+            File.WriteAllText(_timestampPath, string.Empty);
+        }
+        catch (IOException)
+        {
+            // Non-fatal; next run will simply reinstall.
+        }
     }
 
     internal static string NormalizeLanguage(string? language)
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -3,8 +3,8 @@
 
 using System.CommandLine;
 using Azure.Functions.Cli.Commands;
-using Azure.Functions.Cli.Abstractions.Common;
 using Azure.Functions.Cli.Projects;
+using Azure.Functions.Cli.Common;
 
 namespace Azure.Functions.Cli.Workloads.DotNet;
 
@@ -13,28 +13,26 @@ namespace Azure.Functions.Cli.Workloads.DotNet;
 /// Uses the <c>Microsoft.Azure.Functions.Worker.ProjectTemplates</c> NuGet
 /// template package to scaffold projects via <c>dotnet new func</c>.
 /// </summary>
-internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IProjectInitializer
+internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli, ITemplateHivePathProvider hivePathProvider) : IProjectInitializer
 {
     internal const string TemplatesPackageName = "Microsoft.Azure.Functions.Worker.ProjectTemplates";
     internal const string TemplateShortName = "func";
     internal const string DefaultFramework = "net10.0";
 
+    internal static readonly IReadOnlyList<string> SupportedFrameworks = ["net10.0"];
+
     internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(30);
-
-    private static readonly string _hivePath = Path.Combine(
-        Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
-        Constants.FuncHomeDirectoryName,
-        "dotnet-template-hive");
-
-    private static readonly string _timestampPath = Path.Combine(_hivePath, ".installed");
+    internal static readonly TimeSpan OperationTimeout = TimeSpan.FromMinutes(5);
 
     private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
+    private readonly string _hivePath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).HivePath;
+    private readonly string _timestampPath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).TimestampPath;
 
     public string Stack => "dotnet";
 
     public IReadOnlyList<string> SupportedLanguages => ["C#", "F#", "csharp", "fsharp"];
 
-    public Option<string> FrameworkOption { get; } = new("--target-framework")
+    public Option<string> FrameworkOption { get; } = new("--tfm")
     {
         Description = "The target framework for the project (e.g. net10.0).",
         DefaultValueFactory = _ => DefaultFramework
@@ -51,11 +49,25 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
         ArgumentNullException.ThrowIfNull(parseResult);
 
         string projectPath = context.WorkingDirectory.Info.FullName;
-        string? projectName = context.ProjectName ?? Path.GetFileName(projectPath);
+
         string language = NormalizeLanguage(context.Language);
+        if (!SupportedLanguages.Contains(language, StringComparer.OrdinalIgnoreCase))
+        {
+            throw new ArgumentException($"Language '{context.Language}' is not supported. Supported languages: {string.Join(", ", SupportedLanguages)}.", nameof(context));
+        }
+
+        string projectName = context.ProjectName ?? Path.GetFileName(projectPath);
+        ArgumentException.ThrowIfNullOrWhiteSpace(projectName, nameof(context.ProjectName));
+
         string framework = parseResult.GetValue(FrameworkOption) ?? DefaultFramework;
+        if (!SupportedFrameworks.Contains(framework, StringComparer.OrdinalIgnoreCase))
+        {
+            throw new ArgumentException($"Framework '{framework}' is not supported. Supported frameworks: {string.Join(", ", SupportedFrameworks)}.", nameof(framework));
+        }
 
         await EnsureTemplatesInstalledAsync(cancellationToken);
+
+        Directory.CreateDirectory(projectPath);
 
         List<string> args =
         [
@@ -63,12 +75,16 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
             "--name", projectName,
             "--output", projectPath,
             "--language", language,
-            "--Framework", framework,
+            "--framework", framework,
             "--debug:custom-hive", _hivePath,
-            "--force", // CLI creates host.json before calling this so we need to pass force for the the template to overwrite it.
         ];
 
-        await _dotnetCli.RunAsync(args, projectPath, cancellationToken);
+        if (context.Force)
+        {
+            args.Add("--force");
+        }
+
+        await RunWithTimeoutAsync(args, projectPath, cancellationToken);
     }
 
     internal async Task EnsureTemplatesInstalledAsync(CancellationToken cancellationToken)
@@ -78,7 +94,7 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
             return;
         }
 
-        await _dotnetCli.RunAsync(
+        await RunWithTimeoutAsync(
             ["new", "install", TemplatesPackageName, "--debug:custom-hive", _hivePath],
             workingDirectory: null,
             cancellationToken);
@@ -86,12 +102,39 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
         WriteTimestamp();
     }
 
-    private static bool IsHiveFresh()
+    private async Task RunWithTimeoutAsync(
+        IReadOnlyList<string> args,
+        string? workingDirectory,
+        CancellationToken cancellationToken)
+    {
+        using CancellationTokenSource timeoutCts = new(OperationTimeout);
+        using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(
+            cancellationToken, timeoutCts.Token);
+
+        try
+        {
+            await _dotnetCli.RunAsync(args, workingDirectory, linkedCts.Token);
+        }
+        catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
+        {
+            throw new GracefulException(
+                $"The dotnet CLI operation timed out after {OperationTimeout.TotalMinutes:0} minutes. Check your network connection and try again.",
+                isUserError: true);
+        }
+    }
+
+    private bool IsHiveFresh()
     {
         try
         {
             if (!File.Exists(_timestampPath))
             {
+                return false;
+            }
+
+            if (!Directory.Exists(_hivePath) || !Directory.EnumerateFileSystemEntries(_hivePath).Skip(1).Any())
+            {
+                // Hive directory is missing or effectively empty; force reinstall.
                 return false;
             }
 
@@ -104,7 +147,7 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
         }
     }
 
-    private static void WriteTimestamp()
+    private void WriteTimestamp()
     {
         try
         {
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -18,6 +18,11 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
     internal const string TemplateShortName = "func";
     internal const string DefaultFramework = "net10.0";
 
+    private static readonly string _hivePath = Path.Combine(
+        Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+        "azure-functions-core-tools",
+        "dotnet-template-hive");
+
     private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
 
     public string Stack => "dotnet";
@@ -47,42 +52,28 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
 
         await EnsureTemplatesInstalledAsync(cancellationToken);
 
-        try
-        {
-            List<string> args =
-            [
-                "new", TemplateShortName,
-                "--name", projectName,
-                "--output", projectPath,
-                "--language", language,
-                "--Framework", framework,
-            ];
+        List<string> args =
+        [
+            "new", TemplateShortName,
+            "--name", projectName,
+            "--output", projectPath,
+            "--language", language,
+            "--Framework", framework,
+            "--debug:custom-hive", _hivePath,
+        ];
 
-            if (context.Force)
-            {
-                args.Add("--force");
-            }
-
-            await _dotnetCli.RunAsync(args, projectPath, cancellationToken);
-        }
-        finally
+        if (context.Force)
         {
-            await UninstallTemplatesAsync(cancellationToken);
+            args.Add("--force");
         }
+
+        await _dotnetCli.RunAsync(args, projectPath, cancellationToken);
     }
 
     internal async Task EnsureTemplatesInstalledAsync(CancellationToken cancellationToken)
     {
         await _dotnetCli.RunAsync(
-            ["new", "install", TemplatesPackageName],
-            workingDirectory: null,
-            cancellationToken);
-    }
-
-    internal async Task UninstallTemplatesAsync(CancellationToken cancellationToken)
-    {
-        await _dotnetCli.RunAsync(
-            ["new", "uninstall", TemplatesPackageName],
+            ["new", "install", TemplatesPackageName, "--debug:custom-hive", _hivePath],
             workingDirectory: null,
             cancellationToken);
     }
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -3,7 +3,7 @@
 
 using System.CommandLine;
 using Azure.Functions.Cli.Commands;
-using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Abstractions.Common;
 using Azure.Functions.Cli.Projects;
 
 namespace Azure.Functions.Cli.Workloads.DotNet;
@@ -23,7 +23,7 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IPr
 
     private static readonly string _hivePath = Path.Combine(
         Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
-        SharedConstants.FuncHomeDirectoryName,
+        Constants.FuncHomeDirectoryName,
         "dotnet-template-hive");
 
     private static readonly string _timestampPath = Path.Combine(_hivePath, ".installed");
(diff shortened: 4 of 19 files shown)


src/Workload/DotNet/DotNetProjectInitializer.cs
```diff
@@ -8,21 +8,82 @@ namespace Azure.Functions.Cli.Workload.DotNet;
...
 /// <summary>
 /// Project initializer for .NET (C# and F#) Azure Functions.
+/// Uses the <c>Microsoft.Azure.Functions.Worker.ProjectTemplates</c> NuGet
+/// template package to scaffold projects via <c>dotnet new func</c>.
 /// </summary>
-internal sealed class DotNetProjectInitializer : IProjectInitializer
+internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IProjectInitializer
 {
+    internal const string TemplatesPackageName = "Microsoft.Azure.Functions.Worker.ProjectTemplates";
+    internal const string TemplatesPackageVersion = "4.0.5544";
```
satvu: For preview scope, not sure if we should hard code (I think legacy always required updating templates references) or implement cache + grab latest?

src/Workload/DotNet/DotNetProjectInitializer.cs
```diff
@@ -8,21 +8,82 @@ namespace Azure.Functions.Cli.Workload.DotNet;
...
+    }
+
+    internal static string NormalizeLanguage(string? language)
+    {
+        if (string.IsNullOrWhiteSpace(language))
+        {
+            return "C#";
+        }
+
+        return language.ToUpperInvariant() switch
```
liliankasem: I think the init command does to lower on lang - might want to double check and align

src/Workload/DotNet/Workload.DotNet.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
-    <Title>Azure Functions CLI .NET Workload</Title>
+    <Title>DotNet</Title>
     <Description>Azure Functions tooling for .NET (C#, F#) projects.</Description>
```
liliankasem:
```suggestion
    <Description>Azure Functions CLI tooling for .NET (C#, F#) projects.</Description>
```

src/Workload/DotNet/DotNetWorkload.cs
liliankasem:
```suggestion
    public override string Description => "Azure Functions CLI tooling for .NET (C#) projects.";
```

src/Workload/DotNet/DotNetProjectInitializer.cs
liliankasem: legacy used to install latest -> run -> then uninstall every time you ran func init or new

there was also a hardcoded version for the templates that were packaged with the CLI (for offline support)

src/Workload/DotNet/DotNetProjectInitializer.cs
```diff
@@ -9,21 +9,96 @@ namespace Azure.Functions.Cli.Workload.DotNet;
...
+    private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
+
     public string Stack => "dotnet";
 
     public IReadOnlyList<string> SupportedLanguages => ["C#", "F#", "csharp", "fsharp"];
 
-    public IReadOnlyList<Option> GetInitOptions() => [];
+    public Option<string> FrameworkOption { get; } = new("--target-framework")
+    {
+        Description = "The target framework for the project (e.g. net8.0, net10.0).",
```
liliankasem:
```suggestion
        Description = "The target framework for the project (e.g. net10.0).",
```

src/Workload/DotNet/DotNetProjectInitializer.cs
```diff
@@ -9,21 +9,96 @@ namespace Azure.Functions.Cli.Workload.DotNet;
...
+            if (context.Force)
+            {
+                args.Add("--force");
+            }
+
+            await _dotnetCli.RunAsync(args, projectPath, cancellationToken);
+        }
+        finally
+        {
+            await UninstallTemplatesAsync(cancellationToken);
```
liliankasem: Do we want to try using hives to avoid this install uninstall situation?

src/Workload/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -9,21 +9,87 @@ namespace Azure.Functions.Cli.Workload.DotNet;
...
-internal sealed class DotNetProjectInitializer : IProjectInitializer
+internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli) : IProjectInitializer
 {
+    internal const string TemplatesPackageName = "Microsoft.Azure.Functions.Worker.ProjectTemplates";
+    internal const string TemplateShortName = "func";
+    internal const string DefaultFramework = "net10.0";
+
+    private static readonly string _hivePath = Path.Combine(
+        Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+        "azure-functions-core-tools",
```
liliankasem: we have a clear path for the cli already

    /// <summary>
    /// Directory name (under the user profile) the func CLI persists state in.
    /// </summary>
    public const string FuncHomeDirectoryName = ".azure-functions";

satvu: Moved this to a file named `SharedConstants` in abstractions. The file needed a different name from `Constants` which is internal to Func but in the same namespace as `SharedConstants` (imported by `using Azure.Functions.Cli.Common`). Don't love the naming, but had to avoid conflict somehow.

src/Abstractions/Common/SharedConstants.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Common;
+
+/// <summary>
+/// Common constants used across the Azure Functions CLI.
+/// </summary>
+public static class SharedConstants
```
liliankasem: Can this remain Constants we change change the namespace to `Azure.Functions.Cli.Abstractions.Common`

src/Func/Common/Constants.cs
```diff
@@ -34,6 +34,7 @@ internal static class Constants
 
     /// <summary>
     /// Prefix for the CLI environment variables, including those that bind into <c>IConfiguration</c>.
+    /// e.g. <c>FUNC_CLI_Workloads__Home</c> binds <c>Workloads:Home</c>.
```
liliankasem: remove this comment (`FUNC_CLI_Workloads__Home`) is invalid these days

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,72 @@
...
+            string stdout = stdoutTask.Result;
+            string stderr = stderrTask.Result;
+
+            if (process.ExitCode != 0)
+            {
+                throw new InvalidOperationException(
+                    $"'dotnet {string.Join(' ', arguments)}' failed with exit code {process.ExitCode}.{Environment.NewLine}{stderr}{stdout}");
+            }
+        }
+        catch (OperationCanceledException)
```
liliankasem: This only kills the child on cancel. Any other exception falls through; using disposes the Process handle but the OS process keeps running orphaned. We should use `try { ... } finally { if (!process.HasExited) SafeKill(process); }` or similar to safely ensure a process is killed and not an orphan

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            // Ensure the child process does not outlive the caller.
+            if (!process.HasExited)
+            {
+                process.Kill(entireProcessTree: true);
```
liliankasem, line 79: This can throw an `InvalidOperationException` exception - we should wrap this in a try/catch (and/or the finally I mentioned above should help take care of it)

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
     {
-        throw new NotImplementedException(
-            ".NET project initialization is not implemented yet.");
+        ArgumentNullException.ThrowIfNull(context);
+        ArgumentNullException.ThrowIfNull(parseResult);
+
+        string projectPath = context.WorkingDirectory.Info.FullName;
+        string? projectName = context.ProjectName ?? Path.GetFileName(projectPath);
+        string language = NormalizeLanguage(context.Language);
+        string framework = parseResult.GetValue(FrameworkOption) ?? DefaultFramework;
```
liliankasem, line 62: Do we need any validators for any of these args?

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+
+        List<string> args =
+        [
+            "new", TemplateShortName,
+            "--name", projectName,
+            "--output", projectPath,
+            "--language", language,
+            "--Framework", framework,
+            "--debug:custom-hive", _hivePath,
+            "--force", // CLI creates host.json before calling this so we need to pass force for the the template to overwrite it.
```
liliankasem: This is concerning, we will ways force install? even if the user didn't request force? what are we trying to overwrite?

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+
+    internal async Task EnsureTemplatesInstalledAsync(CancellationToken cancellationToken)
+    {
+        if (IsHiveFresh())
+        {
+            return;
+        }
+
+        await _dotnetCli.RunAsync(
+            ["new", "install", TemplatesPackageName, "--debug:custom-hive", _hivePath],
```
liliankasem: This can run for a long time on slow networks - can we use a progress bar or something to indicate that something is happening rather than leave users hanging . We might also want to consider a timeout

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,72 @@
...
+    public async Task RunAsync(
+        IReadOnlyList<string> arguments,
+        string? workingDirectory,
+        CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(arguments);
+
+        ProcessStartInfo psi = new()
+        {
+            FileName = "dotnet",
```
liliankasem: This is path dependent - we need to resolve where the dotnet cli is properly rather than assume it is on the path. See https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Cli.Utils/Muxer.cs

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+        string language = NormalizeLanguage(context.Language);
+        string framework = parseResult.GetValue(FrameworkOption) ?? DefaultFramework;
+
+        await EnsureTemplatesInstalledAsync(cancellationToken);
+
+        List<string> args =
+        [
+            "new", TemplateShortName,
+            "--name", projectName,
+            "--output", projectPath,
```
liliankasem, line 85: projectPath is used as the child's WorkingDirectory but may not exist yet (dotnet new --output is supposed to create it). On Unix Process.Start fails with errno 2. Either Directory.CreateDirectory(projectPath) first, or pass null since --output is absolute.

src/Abstractions/Common/Constants.cs
```diff
@@ -0,0 +1,15 @@
...
+
+/// <summary>
+/// Common constants used across the Azure Functions CLI.
+/// </summary>
+public static class Constants
+{
+    /// <summary>
+    /// Directory name (under the user profile) the func CLI persists state in.
+    /// </summary>
+    public const string FuncHomeDirectoryName = ".azure-functions";
```
liliankasem, line 14: Duplicate FuncHomeDirectoryName. The new Abstractions/Common/Constants.cs adds it, two callsites are repointed, but the original in src/Func/Common/Constants.cs (line 36) is not removed.

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+    private static readonly string _timestampPath = Path.Combine(_hivePath, ".installed");
+
+    private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
+
     public string Stack => "dotnet";
 
     public IReadOnlyList<string> SupportedLanguages => ["C#", "F#", "csharp", "fsharp"];
 
-    public IReadOnlyList<Option> GetInitOptions() => [];
+    public Option<string> FrameworkOption { get; } = new("--target-framework")
```
liliankasem: short option should be -tfm

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+
+        await EnsureTemplatesInstalledAsync(cancellationToken);
+
+        List<string> args =
+        [
+            "new", TemplateShortName,
+            "--name", projectName,
+            "--output", projectPath,
+            "--language", language,
+            "--Framework", framework,
```
liliankasem:
```suggestion
            "--framework", framework,
```

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+    internal const string DefaultFramework = "net10.0";
+
+    internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(30);
+
+    private static readonly string _hivePath = Path.Combine(
+        Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+        Constants.FuncHomeDirectoryName,
+        "dotnet-template-hive");
+
+    private static readonly string _timestampPath = Path.Combine(_hivePath, ".installed");
```
liliankasem: _hivePath / _timestampPath are static readonly baked against the real user profile. Not injectable → not testable in isolation, not redirectable for sandboxed/CI builds, and doesn't honor the FUNC_CLI_WORKLOADS_HOME override that exists in Constants.cs. Move to instance state, injected (alongside an IFileSystem or a small HivePathProvider).

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -3,27 +3,132 @@
...
+
+        await _dotnetCli.RunAsync(
+            ["new", "install", TemplatesPackageName, "--debug:custom-hive", _hivePath],
+            workingDirectory: null,
+            cancellationToken);
+
+        WriteTimestamp();
+    }
+
+    private static bool IsHiveFresh()
```
liliankasem: IsHiveFresh checks only the timestamp file, not that the hive directory still contains the templates. If something deletes the hive but not the marker (AV, cleanup script, partial install), the next dotnet new func fails with a confusing "template not found". Also assert Directory.Exists(_hivePath) and non-empty, or better, dotnet new list --debug:custom-hive _hivePath to verify the func template is present.

test/Workloads/Stacks/DotNet.Tests/DotNetProjectInitializerTests.cs
```diff
@@ -2,45 +2,123 @@
...
-        Assert.Equal(["C#", "F#", "csharp", "fsharp"], new DotNetProjectInitializer().SupportedLanguages);
+        Assert.Throws<ArgumentNullException>(() => new DotNetProjectInitializer(null!));
+    }
+
+    [Fact]
+    public async Task EnsureTemplatesInstalledAsync_CallsInstallWhenTimestampMissing()
+    {
+        DotNetProjectInitializer initializer = CreateInitializer();
+
+        string hivePath = Path.Combine(
```
liliankasem: We should be using tmp directories for test. As it stands, this mutates real user state which we want to avoid

EnsureTemplatesInstalledAsync_CallsInstallWhenTimestampMissing deletes the developer's actual ~/.azure-functions/dotnet-template-hive/.installed; the "fresh" test creates one. Under xUnit parallelism they race each other and corrupt one another's state, and they spook devs running tests locally. Fix flows from #8 — make the hive path injectable and point tests at a per-test temp dir.

test/Workloads/Stacks/DotNet.Tests/DotNetProjectInitializerTests.cs
```diff
@@ -2,45 +2,123 @@
...
+            Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+            ".azure-functions",
+            "dotnet-template-hive");
```
liliankasem: We should use a guid or similar for test hives to keep things seperate

src/Workloads/Stacks/DotNet/Workloads.DotNet.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="Azure.Functions.Cli.Workloads.DotNet.Tests" />
+    <InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
```
liliankasem: Consider hoisting into Directory.Build.props (or a shared <ItemGroup>) so it doesn't get re-added per workload.

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
satvu: In the main branch (v4), the flow is:

 1. dotnet new func runs first (creates everything including host.json)
 2. Then the CLI writes additional files on top

So there's no conflict — dotnet new creates host.json and the CLI doesn't overwrite it.

In v5 (vnext), the order is reversed:

 1. InitCommand writes host.json first
 2. Then calls the workload's InitializeAsync → dotnet new

Without --force, dotnet new sees the existing host.json and fails with an error like:

 Creating this template will make changes to existing files:
   Overwrite   ./host.json
 To create the template anyway, rerun the command with --force.
 
We could avoid writing host.json first?

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -58,15 +65,101 @@ public async Task RunAsync(
...
+                catch { }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Resolves the full path to the <c>dotnet</c> muxer using well-known environment
+    /// variables and conventions, rather than relying on PATH resolution.
+    /// </summary>
+    private static string ResolveDotnetPath()
```
liliankasem: This should be its own file and not part of the runner

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -13,28 +13,26 @@ namespace Azure.Functions.Cli.Workloads.DotNet;
...
     private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
+    private readonly string _hivePath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).HivePath;
+    private readonly string _timestampPath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).TimestampPath;
 
     public string Stack => "dotnet";
 
     public IReadOnlyList<string> SupportedLanguages => ["C#", "F#", "csharp", "fsharp"];
 
-    public Option<string> FrameworkOption { get; } = new("--target-framework")
+    public Option<string> FrameworkOption { get; } = new("--tfm")
```
liliankasem: No I meant both `--target-framework, -tfm`

src/Workloads/Stacks/DotNet/TemplateHivePathProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Azure.Functions.Cli.Workloads.DotNet;
+
+/// <summary>
+/// Default implementation that resolves the template hive path, honoring the
+/// <c>FUNC_CLI_WORKLOADS_HOME</c> environment variable when set, and falling
+/// back to <c>~/.azure-functions/dotnet-template-hive</c>.
+/// </summary>
+internal sealed class TemplateHivePathProvider : ITemplateHivePathProvider
+{
+    private const string WorkloadsHomeEnvVar = "FUNC_CLI_WORKLOADS_HOME";
```
liliankasem: Why are you using workloads home for the hive? And why are we cloning this const instead of moving to abstractions?

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
satvu: Timeout added, can't have progress bar until `IInteractionService` is public

src/Workloads/Stacks/DotNet/TemplateHivePathProvider.cs
satvu: Agent based it on this: https://github.com/Azure/azure-functions-core-tools/pull/5014#discussion_r3276397584 - could you clarify that comment regarding honoring the workloads home and how it relates to the hive?

liliankasem: discussed offline - workload path has nothing to do with this change so removing. adding hive home config opt

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,75 @@
...
+            await Task.WhenAll(stdoutTask, stderrTask);
+            await process.WaitForExitAsync(cancellationToken);
+
+            string stdout = stdoutTask.Result;
+            string stderr = stderrTask.Result;
+
+            if (process.ExitCode != 0)
+            {
+                throw new InvalidOperationException(
+                    $"'dotnet {string.Join(' ', arguments)}' failed with exit code {process.ExitCode}.{Environment.NewLine}{stderr}{stdout}");
```
liliankasem: **`InvalidOperationException` overloads two distinct failure modes; callers can't distinguish them.**

Both "failed to start" (line 42) and "non-zero exit code" (here) throw `InvalidOperationException`. From a command's perspective these are very different:
- Failed to start: usually a missing/broken SDK install (developer environment problem).
- Non-zero exit: `dotnet new` rejected the user's args, template missing, project conflict, etc. (user error, often a `GracefulException` candidate).

`DotNetProjectInitializer` currently doesn't catch either, so both propagate raw to `Program.Main`, which only catches `GracefulException`. Net effect: a perfectly normal `dotnet new` failure (e.g. directory already contains a project) surfaces as a stack trace.

Suggest:
- Introduce `DotnetCliException(int exitCode, string stderr, string stdout)` (or similar) for the non-zero-exit case.
- Have `DotNetProjectInitializer` catch it and wrap in `GracefulException(... isUserError: true)`, preserving the inner exception per `AGENTS.md`.
- Keep "failed to start" as `InvalidOperationException` (or wrap in a more specific type that `DotNetProjectInitializer` also converts to a user-facing message).

Bonus: include the redacted command, exit code, and stderr separately on the exception so logging/telemetry can use them without parsing the message.

src/Workloads/Stacks/DotNet/DotnetPathResolver.cs
```diff
@@ -0,0 +1,101 @@
...
+
+using System.Runtime.InteropServices;
+
+namespace Azure.Functions.Cli.Workloads.DotNet;
+
+/// <summary>
+/// Resolves the full path to the <c>dotnet</c> muxer using well-known environment
+/// variables and conventions, rather than relying on PATH resolution.
+/// </summary>
+internal static class DotnetPathResolver
```
liliankasem: **Static class touching env + filesystem violates the repo's testability rule.**

From `AGENTS.md`:
> Anything that touches I/O, the filesystem, the environment, the network, the clock, or process state must be a regular class behind an interface (or abstract base) so it can be substituted in tests.

`DotnetPathResolver` reads env vars and probes the filesystem, yet it's `static` and called from a `Lazy<string>` static field in `DotnetCliRunner`. That means:

- The resolved path is process-global and cached forever. Any test that depends on `IDotnetCliRunner` indirectly pulls in the host machine's `DOTNET_HOST_PATH` / install layout.
- There's no seam to test the resolution logic itself (priority order, arch-specific vars, default paths).

Suggest: introduce `IDotnetPathResolver` with a `DotnetPathResolver : IDotnetPathResolver` singleton, inject it into `DotnetCliRunner`, and let it cache internally. Tests then substitute it.

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        finally
+        {
+            if (!process.HasExited)
+            {
+                try
+                {
+                    process.Kill(entireProcessTree: true);
+                }
+                catch { }
```
liliankasem: **Empty `catch { }` violates `AGENTS.md`.**

> Swallowing an exception silently is allowed only for genuine best-effort cleanup paths, and the catch block must include a comment explaining why losing the exception is acceptable. Never `catch (Exception)` in business logic.

This is cleanup, so swallowing is defensible, but the bare `catch { }` (no exception type, no comment) needs to be:

```csharp
catch (Exception)
{
    // Best-effort cleanup: the process may have exited between HasExited and Kill,
    // or we may lack permission. Either way, nothing useful to do here.
}
```

Also worth narrowing the catch to `InvalidOperationException` / `Win32Exception` (the documented `Process.Kill` failure modes) rather than catching everything.

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -4,26 +4,174 @@
...
+            "--framework", framework,
+            "--debug:custom-hive", _hivePath,
+        ];
+
+        if (context.Force)
+        {
+            args.Add("--force");
+        }
+
+        await RunWithTimeoutAsync(args, projectPath, cancellationToken);
```
liliankasem: **`InvalidOperationException` from the CLI runner reaches `Program.Main` as a raw stack trace.**

`Program.Main` only catches `GracefulException`. If `dotnet new func` exits non-zero (template missing, output conflict, bad framework, etc.) the `InvalidOperationException` thrown by `DotnetCliRunner` propagates unwrapped and the user sees a stack trace instead of a CLI error.

This is the command boundary for this subsystem (per `AGENTS.md`: "Commands are the boundary"). Catch the expected dotnet failures here and wrap them:

```csharp
try
{
    await RunWithTimeoutAsync(args, projectPath, cancellationToken);
}
catch (InvalidOperationException ex) // or DotnetCliException once introduced
{
    throw new GracefulException(
        $"Failed to scaffold project: {ex.Message}",
        isUserError: true,
        innerException: ex);
}
```

Same applies to `EnsureTemplatesInstalledAsync` (template install can fail for network reasons, which is a user-actionable error).

src/Workloads/Stacks/DotNet/DotnetPathResolver.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private static IReadOnlyList<string> GetDotnetRootVariables()
+    {
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return RuntimeInformation.OSArchitecture switch
+            {
+                Architecture.X86 => ["DOTNET_ROOT(x86)", "DOTNET_ROOT"],
+                Architecture.Arm64 => ["DOTNET_ROOT", "DOTNET_ROOT(ARM64)"],
```
liliankasem: **`DOTNET_ROOT(ARM64)` may be the wrong variant.**

The .NET host's documented arch-suffixed variants are `DOTNET_ROOT_<ARCH>` (uppercase, underscore, no parens) for everything except the legacy `DOTNET_ROOT(x86)` form. So on Windows ARM64 the host looks at `DOTNET_ROOT_ARM64`, not `DOTNET_ROOT(ARM64)`.

Worth double-checking against the [host config docs](https://learn.microsoft.com/dotnet/core/tools/dotnet-environment-variables#dotnet_root-dotnet_rootx86) and aligning, otherwise users with the standard ARM64 install layout will silently fall through to default install paths or PATH.

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Diagnostics;
+
+namespace Azure.Functions.Cli.Workloads.DotNet;
+
+/// <summary>
+/// Executes <c>dotnet</c> CLI commands in a child process.
+/// </summary>
+internal sealed class DotnetCliRunner : IDotnetCliRunner
+{
+    private static readonly Lazy<string> _dotnetPath = new(DotnetPathResolver.Resolve);
```
liliankasem: **Static `Lazy<string>` on a DI-registered singleton is a lifecycle smell.**

The class is registered as a singleton in `DotNetWorkload.Configure`, so this `Lazy` is effectively process-global. That's not strictly wrong (the dotnet host doesn't move), but:

- The static field can't be reset between integration tests, so any test that touches the real `DotnetCliRunner` is one-shot per process.
- It hides the dependency on `DotnetPathResolver` (no way to substitute it in tests, see related comment on `DotnetPathResolver`).

If you adopt `IDotnetPathResolver` as suggested, drop this static `Lazy` and let the resolver own caching as an instance field on a singleton. That keeps lifetime where DI can see it.

src/Workloads/Stacks/DotNet/DotnetPathResolver.cs
```diff
@@ -0,0 +1,101 @@
...
+        foreach (string defaultPath in GetDefaultInstallPaths())
+        {
+            if (File.Exists(defaultPath))
+            {
+                return defaultPath;
+            }
+        }
+
+        // 5. Fallback – rely on PATH (best effort).
+        return _dotnetExeName;
```
liliankasem: **Silent fallback to bare `"dotnet"` hides a real misconfiguration.**

If none of the resolution steps succeed, we hand `"dotnet"` to `Process.Start` and let PATH resolution take over. When that also fails, the user gets a `Win32Exception`/`ComponentModelException` from `Process.Start` with no hint that the muxer wasn't found.

Consider throwing a `GracefulException` here ("Could not locate the dotnet host. Install the .NET SDK or set DOTNET_HOST_PATH / DOTNET_ROOT.") so the failure mode is actionable. The PATH fallback can stay, but only if `which dotnet` actually resolves it (so we can detect the missing-host case).

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -4,26 +4,174 @@
...
+    internal const string DefaultFramework = "net10.0";
+
+    internal static readonly IReadOnlyList<string> SupportedFrameworks = ["net10.0"];
+
+    internal static readonly TimeSpan HiveTtl = TimeSpan.FromDays(30);
+    internal static readonly TimeSpan OperationTimeout = TimeSpan.FromMinutes(5);
+
+    private readonly IDotnetCliRunner _dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
+    private readonly string _hivePath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).HivePath;
+    private readonly string _timestampPath = (hivePathProvider ?? throw new ArgumentNullException(nameof(hivePathProvider))).TimestampPath;
```
liliankasem: **Duplicate null check; `hivePathProvider` is dereferenced twice.**

Both lines evaluate `(hivePathProvider ?? throw ...)`. Capture once at the top of the constructor body:

```csharp
ArgumentNullException.ThrowIfNull(hivePathProvider);
_dotnetCli = dotnetCli ?? throw new ArgumentNullException(nameof(dotnetCli));
_hivePath = hivePathProvider.HivePath;
_timestampPath = hivePathProvider.TimestampPath;
```

Also: caching the two string properties at construction time effectively freezes the hive path for the lifetime of the singleton. That's fine for a CLI process, but if `TemplateHivePathProvider` ever becomes anything other than "read env var once at startup," this initializer won't see the change. Holding `ITemplateHivePathProvider` directly is cheaper and safer.

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -4,26 +4,174 @@
...
+    private bool IsHiveFresh()
+    {
+        try
+        {
+            if (!File.Exists(_timestampPath))
+            {
+                return false;
+            }
+
+            if (!Directory.Exists(_hivePath) || !Directory.EnumerateFileSystemEntries(_hivePath).Skip(1).Any())
```
liliankasem: **`.Skip(1).Any()` is a brittle way to ignore the timestamp file.**

This relies on the timestamp file being one of the entries and the directory containing at least one other entry. If `.installed` doesn't exist yet but other files do, we silently skip a valid entry. If enumeration order changes (it isn't guaranteed), behaviour shifts.

Be explicit:

```csharp
bool hasTemplates = Directory
    .EnumerateFileSystemEntries(_hivePath)
    .Any(e => !string.Equals(Path.GetFileName(e), ".installed", StringComparison.Ordinal));

if (!hasTemplates)
{
    return false;
}
```

src/Workloads/Stacks/DotNet/TemplateHivePathProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    internal const string HivePathEnvironmentVariable = "FUNC_CLI_DOTNET_TEMPLATE_HIVE";
+    private const string HiveDirectoryName = "dotnet-template-hive";
+
+    public string HivePath { get; }
+    public string TimestampPath { get; }
+
+    public TemplateHivePathProvider()
+    {
+        string? configured = Environment.GetEnvironmentVariable(HivePathEnvironmentVariable);
```
liliankasem, line 26: **Security intent: please make it explicit (and verify it holds).**

Reading via `Environment.GetEnvironmentVariable` directly is the right call: `FUNC_CLI_DOTNET_TEMPLATE_HIVE` shares the `FUNC_CLI_` prefix that `CliConfigurationSourceBuilder` binds into `IConfiguration`, so if this were ever read through `IOptions`/`IConfiguration` it could also be set via `~/.azure-functions/config.json` or `.func/config.json` (project-local file). Pulling from the process env var bypasses all of those file-backed config sources.

Two follow-ups:

1. Mirror the `WorkloadHomeResolver` doc comment so the security intent is recorded next to the code (not just inferred):
   ```csharp
   // Read directly (not through IConfiguration) so the template hive cannot be
   // redirected by host config, global config, or project .func/config.json.
   ```
2. Consider validating the resolved value (e.g. reject relative paths, reject paths under unexpected roots) before handing it to `dotnet new ... --debug:custom-hive`. As written, a stray `FUNC_CLI_DOTNET_TEMPLATE_HIVE=/etc` from a sourced shell profile would silently retarget the hive there. `Path.GetFullPath` normalises but doesn't validate intent.

src/Workloads/Stacks/DotNet/DotnetCliRunner.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            // Read both streams concurrently to avoid deadlock when either
+            // buffer fills while the other is being awaited.
+            Task<string> stdoutTask = process.StandardOutput.ReadToEndAsync(cancellationToken);
+            Task<string> stderrTask = process.StandardError.ReadToEndAsync(cancellationToken);
+
+            await Task.WhenAll(stdoutTask, stderrTask);
+            await process.WaitForExitAsync(cancellationToken);
+
+            string stdout = stdoutTask.Result;
```
liliankasem: **`.Result` after `WhenAll` can throw `AggregateException`; cancellation path is fragile.**

When `cancellationToken` cancels mid-read, `ReadToEndAsync` throws `OperationCanceledException`. `Task.WhenAll` then rethrows the first exception (good), but if we somehow get past `WhenAll` with a faulted task, `.Result` rewraps in `AggregateException`. Use `await stdoutTask` / `await stderrTask` instead of `.Result` for consistent exception unwrapping:

```csharp
string stdout = await stdoutTask;
string stderr = await stderrTask;
```

Minor, but it keeps the cancellation/exception story uniform.

liliankasem (approved):

PR #5017, opened 2026-05-18
Restructure workloads under Stacks subfolder

fabiocav:
Move workload and test projects to src/Workload/Stacks and test/Workload/Stacks. Update all project references, solution entries, CI pipeline triggers, and documentation to match the new directory structure for improved organization and maintainability.

resolves #issue_for_this_pr

### Pull request checklist


jviau (commented):
I have a slight preference for keeping folder structure reflecting the namespace in a way.

_Workloads/{Name}_ becomes _Azure.Functions.Cli.Workloads.{Name}_
_Abstractions_ becomes _Azure.Functions.Cli.Abstractions_

Do you want to include `Stacks` in the name? Should we drop Workloads and reorganize as `Azure.Functions.Cli.Stacks.{Name}`? Then non-stacks can remain as `Workloads`, like durable.

fabiocav:
> I have a slight preference for keeping folder structure reflecting the namespace in a way.
> 
> _Workloads/{Name}_ becomes _Azure.Functions.Cli.Workloads.{Name}_ _Abstractions_ becomes _Azure.Functions.Cli.Abstractions_
> 
> Do you want to include `Stacks` in the name? Should we drop Workloads and reorganize as `Azure.Functions.Cli.Stacks.{Name}`? Then non-stacks can remain as `Workloads`, like durable.

I generally agree that folder and namespace alignment is useful, but I’d avoid making it mechanical here. `src/Workload/<kind>/...` is meant as repository organization for workload categories, while the public/API namespace should remain shaped around the product concepts.
 
That’s also why I don’t want Abstractions to become a namespace segment. ASP.NET Core commonly uses `*.Abstractions` in package/project/assembly names while placing public types in the functional namespace: 
- `Microsoft.AspNetCore.Http.Abstractions` exposes types under `Microsoft.AspNetCore.Http`
- `Microsoft.AspNetCore.Http.Features`, and `Microsoft.AspNetCore.Routing`
-  `Microsoft.AspNetCore.Authentication.Abstractions` uses `Microsoft.AspNetCore.Authentication`
- `Microsoft.Extensions.Localization.Abstractions` uses `Microsoft.Extensions.Localization`
- `Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions` uses `Microsoft.Extensions.Diagnostics.HealthChecks`.
 
I think we should be careful not to create artificial namespace segments just because a repo folder exists. That can pollute using lists, make namespaces longer than the API concept requires, and make types harder to discover in the namespace where users naturally expect them.
 
For stack workloads specifically, I’m not strongly opposed to adding Stacks to the implementation namespaces, but I’d treat that as a deliberate API-shape decision rather than a consequence of this folder move. My preference for this PR is to keep namespaces stable and let Stacks remain the physical workload kind grouping. I can follow up in a different PR for `Stacks`, but also want to eventually pluralize `Workloads` for consistency.

liliankasem (approved):
I'm fine with any approach we take here as long as we remain consistent

PR #5023, opened 2026-05-18
ci(vnext): add release pipelines

jviau:
This PR introduces release pipelines for:

1. `func` cli -- currently staging only, stages to public/pre-release as a universal artifact. Each runtime is its own artifact.
2. All nuget packages (Workloads, Abstractions). Stages to public/pre-release, releases to nuget via partner pipeline

Each release is its own pipeline which builds and tests the discrete unit being released.

This PR also deletes all of the per-workload official & public CI files, they are not needed.


liliankasem (approved):
🚀

PR #5025, opened 2026-05-18
Refactor: IOptions<StackOptions> + IWorkloadProvider.FindByStack

liliankasem:
## Summary

Adds a thin `IOptions<StackOptions>` layer + a stack lookup primitive on `IWorkloadProvider`, and reshapes `IWorkloadResolver` to use them. No changes to `IProjectResolver`, `EvaluationResult`, or `WorkloadInfo` (they stay exactly as they are in `vnext`).

## What changed

- **`StackOptions { Stack }`**: the user-expressed stack pin (e.g. "dotnet-isolated", "python", "native"). Populated via two `IConfigureOptions<StackOptions>` implementations registered in `WorkloadRegistration`:
  - `LocalSettingsStackConfigure` — `FUNCTIONS_WORKER_RUNTIME` from `local.settings.json`
  - `FuncProjectStackConfigure` — `stack` from `.func/config.json` (registered last, so it wins)
- **`IWorkloadProvider.FindByStack(string) → WorkloadInfo?`**: case-insensitive alias lookup. Aliases are the sole identifier; `WorkloadInfo.WorkerRuntime` is not introduced.
- **`IWorkloadResolver.ResolveAsync(DirectoryInfo, ct)`**:
  - If `StackOptions.Stack` is set, try `FindByStack`; on hit, return.
  - Otherwise iterate registered `IProjectResolver`s and return the unique claimant (or `None` if zero/multiple).
  - No host.json gate; no candidate narrowing.
- **Deleted**: `WorkloadResolutionContext` (resolver takes a directory and reads pin from `IOptions`).

## Out of scope

A pinned stack that matches no alias falls through to project resolvers — this lets `FUNCTIONS_WORKER_RUNTIME=native` reach the Go/Rust/etc. resolvers. Adding `--stack` to commands and richer per-command directory scoping are follow-ups; nothing in `vnext` consumes `IWorkloadResolver` yet.

## Tests

459 passing in `Func.Tests`, plus existing workload package tests. New coverage:
- `WorkloadResolverTests` — alias hit, alias miss → resolver fallback, no-pin claim resolution, ambiguity.
- `WorkloadProviderTests` — `FindByStack` alias/no-match/null.
- `StackOptionsTests` — `IConfigureOptions` pipeline (last-registered wins, skip-on-null).


src/Func/Hosting/WorkloadRegistration.cs
```diff
@@ -122,6 +123,9 @@ private static async Task<int> RegisterCoreAsync(
 
         services.AddSingleton<ILocalSettingsReader, LocalSettingsReader>();
         services.AddSingleton<IFuncProjectConfigReader, FuncProjectConfigReader>();
+        services.AddOptions<StackOptions>();
```
fabiocav: Why two setup types?

fabiocav (commented):
I needed to start plumbing the configuration (stack information in particular for the start flow), so I got that going here: https://github.com/Azure/azure-functions-core-tools/pull/5032

That will reduce the scope of the current PR (remove the IOptions<StackOptions> changes).

liliankasem:
Superseded by #5032 (unified configuration system), which landed the `StackOptions` + `IConfigureNamedOptions` setup that this PR was building toward. Closing.

PR #5026, opened 2026-05-18
vnext: Proposal - Scaffolding end to end templates from GitHub repos

manvkaur:
# vnext: Proposal - Scaffolding end to end apps from GitHub repos

## Design: `func quickstart` for vNext

Adds a design proposal for `func quickstart` — a new command that scaffolds complete function apps from GitHub-hosted templates discovered via a CDN manifest.

**Design doc:** [proposed/templates-e2e.md](https://github.com/Azure/azure-functions-core-tools/blob/manvkaur/doc/phoenixtemplates/proposed/templates-e2e.md)

> **Note:** The command verb is still under discussion. Development is proceeding with `quickstart` as the working name.

### What's proposed

- `func quickstart`, `func quickstart list`, `func quickstart info <id>` — scaffolds a complete function app from a curated GitHub repo
- CDN manifest with ETag caching (24h TTL), bundled fallback, and staging CDN support (`FUNC_TEMPLATE_MANIFEST_URL`)
- `gitRef` required per manifest entry — templates pinned to immutable tags/SHAs, no default-branch resolution
- Release/compliance model: manifest lives in `azure-functions-templates` repo, requires PR + approval, staging → production promotion
- Template download via `--fetch git|http` (git default, graceful fallback to HTTP when git not on PATH)
- Git process safety: `IGitRunner` abstraction, timeout, cancellation, non-interactive, argument-array invocation
- Interactive Spectre prompts (runtime → language → trigger) with full CI/non-TTY fallback (`--language`, `--template`, `--yes`)
- Runtime-to-language mapping: user-facing labels (`.NET`, `Node`, `Python`, `Java`, `Powershell`) with sub-prompts for .NET (C#/F#) and Node (JS/TS)
- Path traversal prevention, trusted-org validation, HTTPS-only scheme enforcement
- `--path` for target directory control (must be empty or absent)
- Standardised error messaging table covering 15+ failure modes
- Cache and file layout scoped under `~/.azure-functions/<verb>/`

### Explicitly deferred

- `--force` / file conflict handling (tracked as follow-up)

### Open questions for reviewers

- Command verb naming (`quickstart`, `sample`, `bootstrap`, or on `func new`)
- Default language strategy when `--language` is omitted interactively
- Whether bare `func new` should redirect to this flow when no workloads are installed

### Template Manifest Updates
https://github.com/Azure/azure-functions-templates/pull/1787


proposed/templates-e2e.md
```diff
@@ -0,0 +1,883 @@
...
+
+This design introduces `func new templates`: a command that downloads **complete, immediately runnable function app templates** directly from GitHub. Templates are discovered via a live manifest hosted on the Azure Functions CDN. Adding a new template is as simple as publishing a GitHub repo and adding an entry to the manifest — no CLI release required. The manifest is versioned independently; the CLI picks up new templates automatically on the next manifest refresh.
+
+Three pain points this solves:
+
+- **Release cycle friction** — new templates are available as soon as the manifest is updated, not after a 6–8 week CLI release cycle
+- **Maintenance burden** — templates live as standalone GitHub repos, not embedded in the CLI binary; they can be updated, PR'd, and iterated on independently
+- **Scaffolding gap in vnext** — the vnext `func new` stub exits with code 1 if no workload is installed; `func new templates` gives developers a working template experience from a clean install, no workload required
```
liliankasem: Workloads will always be required to work with the CLI so I wouldn't propose this as a pain point this resolves. We do plan to have an onboarding experience so users can install all required workloads during initial use of the CLI

proposed/templates-e2e.md
```diff
@@ -0,0 +1,883 @@
...
+  - [vnext Architecture](#vnext-architecture-what-already-exists)
+  - [fnx init — Source Implementation](#fnx-init--source-implementation)
+
+---
+
+## Problem Statement
+
+Today's function app templates are scattered across multiple sources — the CLI binary, extension bundle, VS Code extension, and Maven repository — each shipping on its own cadence. Getting a template in front of a developer requires a coordinated release across several of these channels, with a minimum turnaround of 6–8 weeks from merge to broad availability. Templates are also bundled inside the CLI binary itself, meaning every addition or update requires a full CLI release.
+
+This design introduces `func new templates`: a command that downloads **complete, immediately runnable function app templates** directly from GitHub. Templates are discovered via a live manifest hosted on the Azure Functions CDN. Adding a new template is as simple as publishing a GitHub repo and adding an entry to the manifest — no CLI release required. The manifest is versioned independently; the CLI picks up new templates automatically on the next manifest refresh.
```
liliankasem: This doesn't provide templating behaviour so I am reluctant for us to call these templates. There's no parameterization, no conditional inclusion, no post-actions. It clones a GitHub repo into a folder. We have a separate templates workload in flight that will provide real templating (engine + manifest + unified sources), and naming this templates will collide with it both conceptually and in the CLI surface. Let's pick a name that reflects what it actually is: bootstrapping a project from a curated sample repo

For this new experience, I recommend:

- `func new sample`
- `func new bootstrap`
- `func quickstart`
- `func clone`

I think `func new sample` (with func new sample list / func new sample info <id>) makes the most sense as it's the most accurate, and doesn't overload "template. I also like `func quickstart`

Thoughts?

proposed/templates-e2e.md
```diff
@@ -0,0 +1,883 @@
...
+- [Appendix](#appendix)
+  - [References](#references)
+  - [vnext Architecture](#vnext-architecture-what-already-exists)
+  - [fnx init — Source Implementation](#fnx-init--source-implementation)
+
+---
+
+## Problem Statement
+
+Today's function app templates are scattered across multiple sources — the CLI binary, extension bundle, VS Code extension, and Maven repository — each shipping on its own cadence. Getting a template in front of a developer requires a coordinated release across several of these channels, with a minimum turnaround of 6–8 weeks from merge to broad availability. Templates are also bundled inside the CLI binary itself, meaning every addition or update requires a full CLI release.
```
liliankasem: Our new workload templates package aims to solve this problem, so this problem statement doesn't really reflect that state of the CLI in v5 and is more of a problem with v4 - I would suggest reframing this

proposed/templates-e2e.md
```diff
@@ -0,0 +1,883 @@
...
+- Download and create a **complete, runnable function app** from a GitHub template — all function code, config, and dependencies included; `func start` works immediately after scaffolding
+- CDN-backed template manifest with ETag caching (24h TTL)
+- `func new templates list` — non-interactive table of available templates, filterable by `--language`, `--resource`, `--iac`, and keyword (`--search`); `--search` is a **case-insensitive substring match** (not semantic/fuzzy) against `id`, `displayName`, `resource`, `tags`, and `shortDescription`
+- `func new templates` (bare invocation) — interactive flow: worker runtime → (Node sub-prompt: JS/TS) → trigger → scaffold; the runtime prompt matches the existing `func new` UX (`dotnet (isolated worker model)`, `Node`, `Python`, `Java`, `Powershell`); trigger selection supports incremental keyword filtering; powered by the existing `IInteractionService` / Spectre.Console already in vnext
+- **.NET isolated worker model only** — the manifest `CSharp` language maps exclusively to the .NET isolated worker model; the in-process model is not supported and will not be added (it is on a deprecation path)
```
liliankasem: What about FSharp?

manvkaur: like node, I will add sub prompt for dotnet isolated to csharp and fsharp but we don't have any e2e fsharp sample

proposed/templates-e2e.md
manvkaur: added this to the top of the doc, keeping this comment unresolved for discussion with other reviewers. @vrdmr , @fabiocav

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+
+| Candidate | Full Command | Subcommands | Pros | Cons |
+| --- | --- | --- | --- | --- |
+| **`templates`** (proposed) | `func new templates` | `list`, `info` | Familiar term; matches industry use for e2e prod ready apps | Collides with templates workload; |
+| **`sample`** | `func new sample` | `list`, `info` | These are sample repos; no overloading of "template" | Might imply "not production-ready" |
+| **`quickstart`** | `func quickstart` | `list`, `info` | Clear intent; aligns with Azure Samples naming (`*-quickstart-*-azd`) | Longer to type; top-level only (not under `func new`) |
+| **`bootstrap`** | `func new bootstrap` | `list`, `info` | Accurate for "set up a project from scratch" | Less discoverable; not a common Azure CLI verb |
+| **`clone`** | `func new clone` | `list`, `info` | Technically accurate (it clones a repo) | Implies git; doesn't convey curation or discovery |
+
+**Recommendation:** `templates` — this is the industry used term for scaffolding a new project, including production-ready starting points. We will need resolve conflict with the in-flight templates workload and establish distinction.
```
fabiocav: In addition to the concerns brought up about the conflict and confusion with the existing experience, this scaffolding is not providing templating behavior, the command shape `func new templates` reads unnaturally to me. In a `new <noun>` pattern, the noun after new is usually the thing being created, so this reads like “create a new template” not “create a new app from a template.” Since the template/sample is the source rather than the target artifact, I’d avoid this shape. `func quickstart`, `func new --sample <id>`, or another source-oriented name would be clearer.

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+- [Appendix](#appendix)
+  - [References](#references)
+  - [vnext Architecture](#vnext-architecture-what-already-exists)
+  - [fnx init — Source Implementation](#fnx-init--source-implementation)
+
+---
+
+## Problem Statement
+
+In vnext, workload templates solve the per-function scaffolding story: `func new` adds a single function file to an existing project using templates shipped in workload packages. A complementary scenario is **complete-app scaffolding** — downloading a fully runnable function app (function code, host.json, local.settings.json, IaC configuration, and all dependencies) in a single step so that `func start` works immediately.
```
fabiocav: > in a single step so that `func start` works immediately.

This seems to depend on the `--env` work mentioned below? In the current state, manual configuration will be required before you can have a working app, so it's worth rewording this to capture the proposed state.

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+services.AddSingleton<TemplatesInfoCommand>();
+services.AddSingleton<TemplatesCommand>();          // not as FuncCliCommand — it's a subcommand
+services.AddSingleton<ITemplateManifestService, TemplateManifestService>();
+services.AddSingleton<ITemplateFunctionScaffolder, TemplateFunctionScaffolder>();
+// NewCommand's registration already exists; it now also resolves TemplatesCommand via DI
+```
+
+---
+
+### File Layout
```
fabiocav: Git execution needs a process abstraction, timeout, cancellation, non-interactive environment, and argument-array invocation. Ensure you do not shell-concatenate repo URLs or folder paths.

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+}
+```
+
+`TemplateFunctionScaffolder` implementation (ported from `fnx/lib/init/scaffold.js`):
+
+| Strategy | Condition | Detail |
+| ---------- | ----------- | -------- |
+| git sparse-checkout | `git` available on PATH | `git clone --filter=blob:none --sparse <repoUrl>` then `git sparse-checkout set <folderPath>` |
+| Tree API + raw URLs | No git (subfolder) | One API call to `GET /repos/{owner}/{repo}/git/trees/main?recursive=1` (60 req/hr) to enumerate files, then download each via `raw.githubusercontent.com` (~5,000 req/hr). Tree response capped at 5 MB and cached. |
+| GitHub zip API | No git (whole repo) | `GET https://api.github.com/repos/<owner>/<repo>/zipball/<ref>` → extract and strip `repo-main/` prefix |
```
fabiocav: This is not a stable contract. The branch information should be in the manifest so we don't may assumptions about it here.

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+
+When `--env` succeeds, the post-scaffold banner drops all install/activate steps — the environment is ready, so only `cd` (if `--path` was used) and `func start` remain. If `--env` partially fails (e.g. `pip install` fails), the banner falls back to showing the full manual steps with a warning that automatic setup was incomplete.
+
+---
+
+## Open Questions
+
+- [ ] **Default language** — `--language` is currently required with no fallback (target dir is empty, nothing to detect from). Should there be a default so repeat users don't have to supply it every time? Options:
+  - **Detect from existing files** — run `ProjectDetector.DetectStackAndLanguage(targetDir)` on the resolved target before prompting; if detectable files exist (e.g. `requirements.txt`, `package.json`, `*.csproj`), use the detected language as the default; if target is empty, fall through to prompt or error. Already exists at `src/Func/Commands/ProjectDetector.cs`. Becomes more useful when `--force` is added in a future iteration (target dir may have project files).
+  - **Persist last-used language** — store in `~/.azure-functions-core-tools/config` after first scaffold; use as default on next run
```
fabiocav: Please align with the home directory documented in other specs. (e.g., https://github.com/Azure/azure-functions-core-tools/blob/docs/proposed/cli-profiles.md#6-profile-sources-and-layering)

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+
+## Goals / Non-Goals
+
+### Goals
+
+- Add `func new templates` as a built-in subcommand of `NewCommand` on the vnext branch
+- Download and create a **complete, runnable function app** from a GitHub template — all function code, config, and dependencies included; `func start` works immediately after scaffolding
+- CDN-backed template manifest with ETag caching (24h TTL)
+- `func new templates list` — non-interactive table of available templates, filterable by `--language`, `--resource`, `--iac`, and keyword (`--search`); `--search` is a **case-insensitive substring match** (not semantic/fuzzy) against `id`, `displayName`, `resource`, `tags`, and `shortDescription`
+- `func new templates` (bare invocation) — interactive flow: worker runtime → (dotnet sub-prompt: C#/F#, Node sub-prompt: JS/TS) → trigger → scaffold; the runtime prompt matches the existing `func new` UX (`dotnet (isolated worker model)`, `Node`, `Python`, `Java`, `Powershell`); trigger selection supports incremental keyword filtering; powered by the existing `IInteractionService` / Spectre.Console already in vnext
```
fabiocav: User facing labels for .NET should be simply `.NET` instead of `dotnet`, with no mention of `isolated worker model` (the new CLI only supports isolated)

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
+# Core Tools vNext: `func new templates` — CDN-Backed Template Discovery
+
+**Author:** manvkaur
```
fabiocav: I think the release/compliance model needs to be specified here as well. As written, template content is downloaded directly from GitHub repos/branches, so any commit landing on the referenced branch can become live CLI output without a CLI release, workload release, or obvious validation gate. That effectively makes repo branch updates production artifact releases. It's fine if that's the documented plan, but it needs to be captured here.
 
- What is the staging and validation mechanism before content becomes visible to users?
- Do we have separate dev/staging/prod manifests or CDN paths?
- Can the CLI resolve templates from a non-prod manifest/branch for testing?
- Who approves manifest changes and repo content changes?

I’d expect the manifest to pin an immutable commit/tag, plus a promotion process where validated entries move from staging to production.

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+
+> **This decision drives the rest of the document.** The verb chosen here determines the command or subcommand name, class names, CLI surface, and whether it collides with the in-flight templates workload.
+
+This design proposes **`templates`** as the subcommand verb (`func new templates`). However, there is a naming conflict: a separate templates workload is in flight that provides templating (engine, parameterization, etc.). Using `templates` for this command risks conceptual and CLI-surface collision with that workload.
+
+| Candidate | Full Command | Subcommands | Pros | Cons |
+| --- | --- | --- | --- | --- |
+| **`templates`** (proposed) | `func new templates` | `list`, `info` | Familiar term; matches industry use for e2e prod ready apps | Collides with templates workload; |
+| **`sample`** | `func new sample` | `list`, `info` | These are sample repos; no overloading of "template" | Might imply "not production-ready" |
+| **`quickstart`** | `func quickstart` | `list`, `info` | Clear intent; aligns with Azure Samples naming (`*-quickstart-*-azd`) | Longer to type; top-level only (not under `func new`) |
```
liliankasem: `top-level only (not under `func new`)
 Doesn't have to be, could also live on func new if folks prefer that

proposed/templates-e2e.md
```diff
@@ -0,0 +1,921 @@
...
+| **`bootstrap`** | `func new bootstrap` | `list`, `info` | Accurate for "set up a project from scratch" | Less discoverable; not a common Azure CLI verb |
```
liliankasem: Would love to understand why this would be less discoverable than say `func new templates` templates on `new` doesn't exist today either so I think the discoverability journey is exactly the same. When customers run `func --help` or `func new --help` they will be able to see all of the subcommands/commands available to them so I don't think naming has anything to do with discoverability

proposed/templates-e2e.md
liliankasem: My recommendation is we go with `quickstart` or `bootstrap` (whether on `func` root or `func new`). 

---

The claim that `templates` is the industry term for this doesn't match either CLI conventions or Azure's own vocabulary.

CLI verbs: Across the most-used scaffolding tools (create-react-app, npm create *, ng new, rails new, nest new, dotnet new, cargo new, yo, hugo new, nuxi init, cookiecutter, degit), the primary verbs are create, new, and init. "Template" almost always appears as a noun or flag (--template, dotnet new <template>), not the command itself. 

**Microsoft/Azure usage**: Azure docs already draw clean distinctions:

Quickstart → "Ready-to-run, minimal-fuss starting point for a project."
Sample → "End-to-end application/code repos demonstrating real scenarios. Often production-shaped."
Template → "A parameterized, declarative IaC definition. ARM/Bicep templates have a parameters section that is rendered against user input."

When Azure uses "template", the artifact is **parameterized** and rendered. So by Azure's own established vocabulary it is a sample or quickstart, not a template. "Bootstrap" and "quickstart" are at least as defensible as "templates" here.

proposed/templates-e2e.md
vrdmr: The reasoning of using the word `templates` is mainly following what vercel, bolt.new, AI Template Gallery, and AZD Templates call it.

image

image

image

image

image


Lambda folks use "blueprints" but I don't want to use the same word as AWS Lambda.

> For this new experience, I recommend:
> 
> - `func new sample`
> - `func new bootstrap`
> - `func quickstart`
> - `func clone`
>
> I think func new sample (with func new sample list / func new sample info ) makes the most sense as it's the most accurate, and doesn't overload "template. I also like func quickstart

suggestions are good but limiting the feature to "quickstart" would be incorrect and leaving options for advanced samples.

We are calling it Template Gallery in VSCode as well - as it is a template gallery
image

vrdmr: >There's no parameterization, no conditional inclusion, no post-actions

I don't think core tools today does all the features mentioned for all the _templates X languages_ matrix - i think it is only done in few cases, and we can explore how to add support for it.

proposed/templates-e2e.md
vrdmr: Provided my comments above.

Repasting it 
--------------------

The reasoning of using the word `templates` is mainly following what vercel, bolt.new, AI Template Gallery, and AZD Templates call it.

image

image

image

image

image


Lambda folks use "blueprints" but I don't want to use the same word as AWS Lambda.

> For this new experience, I recommend:
> 
> - `func new sample`
> - `func new bootstrap`
> - `func quickstart`
> - `func clone`
>
> I think func new sample (with func new sample list / func new sample info ) makes the most sense as it's the most accurate, and doesn't overload "template. I also like func quickstart

suggestions are good but limiting the feature to "quickstart" would be incorrect and leaving options for advanced samples.

We are calling it Template Gallery in VSCode as well - as it is a template gallery
image

proposed/templates-e2e.md
manvkaur: added a section for this

proposed/templates-e2e.md
manvkaur: udpated

proposed/templates-e2e.md
manvkaur: added a section explaining the Release & Compliance Model answering the questions and also updating the requirements on tags. Please review again

Example PR for manifest - https://github.com/Azure/azure-functions-templates/pull/1787

proposed/templates-e2e.md
manvkaur: done

proposed/templates-e2e.md
manvkaur: done, added this to gitref and also Release and Compliance model section

jviau:
@manvkaur we are planning on integrating `Microsoft.TemplateEngine` directly into func v5. This engine will have an extensibility point where we can source templates directly from GitHub (but of course we need to ensure quality gates on it).

@soninaren is working on the integration. We should reframe this design proposal with how it can plug into the template engine. We may be able to use all the repos _as is_, but it also opens the door for these repos to include a `template.json` for any customization and tokenization they need.

manvkaur:
> @manvkaur we are planning on integrating `Microsoft.TemplateEngine` directly into func v5. This engine will have an extensibility point where we can source templates directly from GitHub (but of course we need to ensure quality gates on it).
> 
> @soninaren is working on the integration. We should reframe this design proposal with how it can plug into the template engine. We may be able to use all the repos _as is_, but it also opens the door for these repos to include a `template.json` for any customization and tokenization they need.

@jviau Quickstart onboards repos using - https://github.com/Azure/azure-functions-templates/blob/95486674fd9bfab2daf2e0fc25c288d0a8def0c1/Functions.Templates/Template-Manifest/manifest.json#L79

This is also used by VS Code Template Gallery and Functions MCP Tools

jviau:
@manvkaur yeah I see how it is done today. I believe the process will be very similar for `Microsoft.TemplateEngine` (publish a manifest the CLI can consume). The shape of this manifest will most likely change though to align with the `template.json` model from `Microsoft.TemplateEngine`. Once we have that manifest in place, then it will be pretty straight forward for us to implement a custom package provider for the template engine which will be able to reference that manifest and then go download the zip from github to get the template locally.

jviau:
@manvkaur quickstart integration to template engine spec PRs are available here:

#5555 -- pipeline for ingesting samples, and how we can use sample repos as-is, or they can optionally leverage the full template engine by including a `.template.config/template.json`
#5556 -- how `func init` will use the same engine for both 'basic' project creation as well as onboarded samples. This makes the sample experience: (1) acquire (`func new install <sample-name>`) and (2) execute: `func init --template <sample-name>`

Going to close this PR as it is superseded. Please leave comments/questions on the other spec PRs.

PR #5044, opened 2026-05-19
Restrict workload Home override to FUNC_CLI_WORKLOADS_HOME env var

liliankasem:
## What

`WorkloadPathsOptions.Home` was bound from any `IConfiguration` source (json config, `local.settings.json`, in-memory). That meant the workload root could be silently redirected by per-project settings, which we don't want for a process-level concern.

This PR makes the override path explicit:

- `Home` is read directly from the `FUNC_CLI_WORKLOADS_HOME` env var (when explicitly set, non-empty), otherwise falls back to `~/.azure-functions`.
- `WorkloadStorageRegistration` no longer calls `BindConfiguration("Workloads")`.
- `WorkloadRegistration.RegisterWorkloadsAsync` no longer takes an `IConfiguration`; it constructs `WorkloadPathsOptions` directly.
- Renamed the env var from `FUNC_CLI_Workloads__Home` to `FUNC_CLI_WORKLOADS_HOME` (the double-underscore casing only mattered when it was a config section binding).

The `Workloads:Catalog` section is still bound through `IConfiguration` (unchanged).

## Why

The workload home directory is a process/machine-level concern, not a per-project setting. Allowing arbitrary config sources to override it made it easy to misconfigure and hard to reason about.

## Tests

- New `WorkloadPathsOptionsTests` coverage: env-var override applied, empty value falls back to default, unset falls back to default.
- `CliHostFactoryTests` updated to set/restore the env var instead of pushing `Workloads:Home` via in-memory config.
- Full `Func.Tests` suite: 447/447 passing.


test/Func.Tests/Common/EnvironmentVariableScope.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Azure.Functions.Cli.Tests.Common;
+
+/// <summary>
+/// Sets an environment variable for the lifetime of the scope and restores
+/// the previous value on dispose. Tests rely on this to mutate process-global
+/// env vars (e.g. <c>FUNC_CLI_WORKLOADS_HOME</c>) without leaking state to
+/// later tests in the same process.
+/// </summary>
+internal sealed class EnvironmentVariableScope : IDisposable
```
jviau: I'd prefer we don't use any env var's in tests, it will still set this env for all other test cases running concurrently.

We will need an abstraction layer between the env var and consuming it in the product code, so the test can shim that abstraction.

src/Func/Common/IEnvironmentVariables.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Azure.Functions.Cli.Common;
+
+/// <summary>
+/// Thin abstraction over process environment-variable reads. Exists so
+/// product code never touches <see cref="System.Environment.GetEnvironmentVariable(string)"/>
+/// directly and tests can substitute their own values without mutating
+/// process-global state (which would leak across xUnit parallel test runs).
+/// </summary>
+internal interface IEnvironmentVariables
```
jviau: Sorry didn't comment earlier. But we can still use `IConfiguration` as this abstraction, but keep it as a separate instance of the configuration (so regular `IConfiguration` import isn't this).

The previous IHostBuilder had a concept of "host" configuration vs "app" configuration (two config sources). Not sure what that is like in the new app-builder approach.

Simplest idea would be to have:

``` CSharp
internal interface IHostConfiguration : IConfiguration
{
}

internal sealed class HostConfiguration(IConfiguration inner) : IHostConfiguration
{
    // implement everything through `IConfiguration` inner. 
}

// setup
internal static IHostConfiguration BuildHostConfiguration()
{
    IConfiguration inner = ... // build this as needed
    return new HostConfiguration(inner);
}
```

But with all that said, does the app builder already have some concept of host-vs-app config?

liliankasem: Looks like the new app builder doesn't have that config split so I think we would have to go with your suggestion here

---

In HostApplicationBuilder there is a single ConfigurationManager exposed as builder.Configuration. What used to be "host config" vs "app config" is now just layering order into that one instance:

- HostApplicationBuilder's constructor first adds the "host" sources (content root, DOTNET_ env vars) into Configuration.
- Then it calls ApplyDefaultAppConfiguration(...), which layers appsettings.json, env vars (no prefix), user secrets, CLI args, etc. on top of the same ConfigurationManager.

fabiocav (commented):
I'm not sure `IHostConfiguration` is the right name here. In generic host terminology, host configuration has specific semantics: it initializes `IHostEnvironment`, is later folded into app configuration, has lower precedence than app configuration, and is not registered separately as a DI-resolvable “host-only” configuration. This PR’s type is instead an allowlisted bootstrap/env-var settings bag. Reusing `HostBuilder` terminology suggests common behavior and may cause confusion in the future.
 
The approach also feels broader than what this change needs. If the only current setting is workload home, I’d rather keep the abstraction named after the actual behavior with something like `IWorkloadHomeProvider`, or even just a `WorkloadHomeResolver` plus a narrow test override for the boot-time registration path.
 
Do we have use for this broader abstraction in any other scenario right now? If not, we can take an even simpler approach to avoid setting environment variables in tests, which I agree we should not do, with something like:
 
 ```csharp
 internal sealed class WorkloadPathsOptions : IWorkloadPaths
 {
     public WorkloadPathsOptions()
         : this(WorkloadHomeResolver.Resolve())
     {
     }
 
     internal WorkloadPathsOptions(string home)
     {
         Home = Path.GetFullPath(home);
     }
 
     public string Home { get; }
 }
 ```
 
`WorkloadHomeResolver.Resolve` can contain the current logic to resolve `FUNC_CLI_WORKLOADS_HOME` or fall back to the default. For integration tests that need the full host boot path, we can pass a narrow service into that path directly instead of introducing a general `IConfiguration` host configuration service.
 
This is startup-path code, and with the current `IHostConfiguration` model, `HostConfiguration`, `HostConfigurationBuilder`, service registration, descriptor scanning before host build, and `WorkloadPathsOptionsSetup` are a lot of machinery to supply one value.
 
If we later need a broader “restricted” configuration concept (or if we really feel we need this now), we can design and name that explicitly rather than reusing `HostBuilder` terminology for a different concept.

fabiocav:
For clarity, I’m okay not blocking this if we want the fix in, but I’d strongly prefer we avoid the `IHostConfiguration` name at least or narrow this to workload home specific code. My concern is less about current correctness and more about the design and terminology.

liliankasem:
Going with the basic appraoch for now, we can expand and revisit this at another date

PR #5046, opened 2026-05-20
ci(vnext): fix cli release process

jviau:
1. Add strong-name signing to vnext projects
2. Adds a post-publish target that will error out if our publish ever starts including loose files
3. Fix macos signing (zip and unzip, because macos signing is a picky signer)
4. Fix universal artifacts upload
    - 1ESPT blocks universal upload in release jobs. This was just a gap, they are adding support for it soon. Until then I am using a regular job.


eng/ci/templates/jobs/build.yml
```diff
@@ -117,8 +120,36 @@ jobs:
...
+      # requires the payload to be a .zip or .dmg, so we zip each osx-* rid folder, sign the zips,
+      # then expand them back over the publish output.
+      - pwsh: |
+          $ErrorActionPreference = 'Stop'
+          $stage = Join-Path '$(publish_dir)' '_macsign'
+          New-Item -ItemType Directory -Force -Path $stage | Out-Null
+          Get-ChildItem -Path '$(publish_dir)' -Directory -Filter 'osx-*' | ForEach-Object {
+            $payload = Get-ChildItem -Path $_.FullName -Exclude '*.pdb'
+            $zip = Join-Path $stage "$($_.Name).zip"
+            Compress-Archive -Path $payload -DestinationPath $zip -Force
```
liliankasem, line 133: Agent: Compress-Archive/Expand-Archive on the Windows agent uses the BCL ZipArchive API, which doesn't preserve Unix mode bits. Same constraint exists on main (which uses ArchiveFiles@2), so this is not a regression — but please confirm whatever consumes the universal package (func-cli-osx-*) chmod +x's the apphost. If consumers expect mode bits in the zip, switch the zip step to tar -czf or use a tool that preserves permissions.

Lilian: I did test the installation on mac and the `func` executable is missing permissions, we usually ship this with chmod applied so that customers dont have to do that themselves - I believe this is a missing step in the release  pipeline

```log
> ./func
zsh: permission denied: ./func
```

jviau:
ignore the recent push. Will revert it. Meant to push to ADO for testing only.

eng/ci/templates/jobs/build.yml
jviau: Linux and MacOS are now uploaded as a .tar.gz

liliankasem (approved):

PR #5054, opened 2026-05-20
Pack worker payloads into Node/Python/Go worker workloads

liliankasem:
Targets `vnext`.

Wires up the scaffolded `Workloads.Workers.{Node,Python,Go}` projects so packing produces a usable worker payload under `content/workers/<lang>/`.

## Approach

**Node and Python**
- `PackageReference` to the upstream worker package now uses `GeneratePathProperty="true"` with `IncludeAssets=none`, so we restore the nupkg purely to read it off disk, without pulling its build targets / contentFiles into the workload project.
- New MSBuild target (`BeforeTargets="_GetPackageFiles"`) globs the restored package payload and emits `<None Pack="true">` items:
  - Node: `$(PkgMicrosoft_Azure_Functions_NodeJsWorker)/contentFiles/any/any/workers/node/**/*`
  - Python: `$(PkgMicrosoft_Azure_Functions_PythonWorker)/tools/**/*`
- Items use explicit `TargetPath="%(RecursiveDir)%(Filename)%(Extension)"` with trailing-slash `PackagePath="content/workers/<lang>/"`. This sidesteps NuGet pack's folder-detection (any `PackagePath` whose last segment has no extension, e.g. `LICENSE`, is treated as a folder and appends the source's project-relative path, which doubled paths in earlier attempts).

**Go**
- No upstream worker NuGet. Ships a static native `worker.config.json` (from `feature/go-support`) at `content/workers/go/worker.config.json`.


src/Workloads/Workers/Go/content/workers/native/worker.config.json
```diff
@@ -0,0 +1,11 @@
+{
```
liliankasem, line 1: @harshivcodes does this path match where the host will need to look for worker config? Or should it be 
`src/Workloads/Workers/Go/content/workers/native/worker.config.json`

src/Workloads/Workers/Node/Workloads.Workers.Node.csproj
```diff
@@ -13,15 +13,33 @@
...
   <ItemGroup>
     <None Include="workload.json" Pack="true" PackagePath="/" />
-    <Content Include="content/**" Pack="true" PackagePath="content/" />
   </ItemGroup>
 
+  <Target Name="IncludeNodeWorkerContentInPack" BeforeTargets="_GetPackageFiles">
+    <Error Condition="'$(PkgMicrosoft_Azure_Functions_NodeJsWorker)' == ''"
+           Text="Microsoft.Azure.Functions.NodeJsWorker package path not resolved. Run 'dotnet restore' first." />
+    <ItemGroup>
+      <_NodeWorkerFile Include="$(PkgMicrosoft_Azure_Functions_NodeJsWorker)/contentFiles/any/any/workers/node/**/*" />
```
jviau, line 42: This is coupled with the precise packaging layout of the node worker. Can we add a TODO comment to decouple this. I think we should be able to include the assets of the nodejsworker in this project and let it wire up its files, then this project transforms the items it adds and repacks them.

jviau (approved):
I'd prefer a decoupled approach, where we don't manually pick files from the nuget packages, but instead let it tell us what files it needs. But this will do for now.

src/Workloads/Workers/Node/Workloads.Workers.Node.csproj
liliankasem: Yup that's fair, I will add a todo for now and will eventually have the agent grab all our TODOs and open work items

src/Workloads/Workers/Go/content/workers/native/worker.config.json
liliankasem: Confirmed we use the native path in v4 - updating this here

PR #5055, opened 2026-05-20
Bundles workload becomes content-only

liliankasem:
Converts the `Azure.Functions.Cli.Workloads.ExtensionBundles` package to a **content workload** per the bundles spec ([PR #5033](https://github.com/Azure/azure-functions-core-tools/pull/5033)). The CLI-side resolver lands in [PR #5036](https://github.com/Azure/azure-functions-core-tools/pull/5036) stacked on this branch.

## Changes

- `workload.json`: `kind` = `content`, no `entryPoint`.
- `Workloads.ExtensionBundles.csproj`: assembly-free packaging project. Drops:
  - `ExtensionBundlesWorkload` class (no longer needed)
  - `Microsoft.Extensions.DependencyInjection.Abstractions`, `Microsoft.Extensions.Logging`, `NuGet.Versioning` package references
  - `ProjectReference` to `Abstractions` and `InternalsVisibleTo`
  - `tools/any/<dll>` pack item
- Bundle payload now packs under `tools/any/` via `DownloadDefaultBundle.targets` next to the csproj (CDN download at pack time, retried, gated by `SkipDefaultBundleDownload` for offline dev builds).
- Package tags: `kind:content alias:bundles func-workload func-tool`.
- Drop `test/Workloads/Tools/ExtensionBundles.Tests/` and its slnx + CI YAML references (no assembly to test).
- README rewritten for the content-only model; `release_notes.md` kept empty (first release).

## Version scheme

The workload pkg has no code of its own, so its version maps **1:1** to the bundle payload version it carries. `Directory.Version.props` derives `$(VersionPrefix)` from two props:

| Prop                  | Meaning                                                                                       |
|-----------------------|-----------------------------------------------------------------------------------------------|
| `$(BundleVersion)`    | Bundle payload version. Drives the CDN URL and the user-facing version in `func` logs.        |
| `$(BundleChannel)`    | `stable` \| `preview` \| `experimental`. Selects the CDN bundle id and the prerelease label.  |

| Channel        | Workload pkg version    | CDN bundle id                                              |
|----------------|-------------------------|------------------------------------------------------------|
| `stable`       | `4.35.0`                | `Microsoft.Azure.Functions.ExtensionBundle`                |
| `preview`      | `4.35.0-preview`        | `Microsoft.Azure.Functions.ExtensionBundle.Preview`        |
| `experimental` | `4.35.0-experimental`   | `Microsoft.Azure.Functions.ExtensionBundle.Experimental`   |

Preview/experimental are NuGet prereleases, so `func workload install` won't pick them up without an explicit version pin or `--prerelease`. Stable carries no prerelease label and is treated as a release. The CDN URL always uses the bare `$(BundleVersion)`.

## Verification

- `CI=true dotnet build -c Release` — 0 warnings, 0 errors.
- `CI=true dotnet test -c Release --no-build` — all tests pass.
- `dotnet pack` produces:
  - `Azure.Functions.Cli.Workloads.ExtensionBundles.4.35.0.nupkg` (default, stable channel)
  - `... .4.35.0-preview.nupkg` with `/p:BundleChannel=preview`
  - `... .4.35.0-experimental.nupkg` with `/p:BundleChannel=experimental`

## Follow-up

PR #5036 implements `IExtensionBundleResolver` and the three-channel scanner filter that maps host.json `extensionBundle.id` values to the correct workload pkg.


src/Workloads/Tools/ExtensionBundles/Workloads.ExtensionBundles.csproj
```diff
@@ -3,28 +3,20 @@
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
     <Title>Extension Bundles</Title>
```
liliankasem: Azure Functions CLI Extension Bundles

src/Workloads/Tools/ExtensionBundles/DownloadDefaultBundle.targets
```diff
@@ -0,0 +1,55 @@
...
+
+    TODO(spec §5.4): branch on prerelease label so the preview/experimental
+    workload pulls from the matching CDN id. For now hardcoded to stable.
+  -->
+
+  <PropertyGroup>
+    <_BundleCdnId>Microsoft.Azure.Functions.ExtensionBundle</_BundleCdnId>
+  </PropertyGroup>
+
+  <Target Name="DownloadDefaultBundle"
```
liliankasem, line 35: Need to add a TODO that we will be getting the bundle pkg from CI build artifacts and not the CDN.

src/Workloads/Tools/ExtensionBundles/Directory.Version.props
```diff
@@ -1,6 +1,37 @@
...
+                              fix. Lives in the 4th segment regardless of
+                              channel.
+      $(BundleChannel)      - stable | preview | experimental. Selects the
+                              CDN bundle id (see DownloadDefaultBundle.targets)
+                              and the workload pkg's prerelease label.
+
+    Derived $(VersionPrefix):
+      stable        -> 4.35.0.1
+      preview       -> 4.35.0.1-preview
+      experimental  -> 4.35.0.1-experimental
```
liliankasem: Or does 4.35.0-experimental.1 make more sense?

soninaren: As discussed offline, we don't need the build number since the bundle resolution lives in core cli.

src/Workloads/Tools/ExtensionBundles/Directory.Version.props
```diff
@@ -1,6 +1,37 @@
...
   <PropertyGroup>
-    <VersionPrefix>1.0.0</VersionPrefix>
-    <VersionSuffix>preview.1</VersionSuffix>
+    <BundleVersion>4.35.0</BundleVersion>
+    <WorkloadIteration>1</WorkloadIteration>
+    <BundleChannel>stable</BundleChannel>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(BundleChannel)' == 'stable'">
+    <VersionPrefix>$(BundleVersion).$(WorkloadIteration)</VersionPrefix>
```
jviau: nuget semver is 3 sections only, we can't have the fourth section.

What we can do is follow dotnet and the func host. Make the patch version a 3 digit value to show the state.

`4.35.100` -> maps to `4.35.0`. Need to iterate the package here, we release `4.35.101`

But honestly, do we need to worry? Can we just go with `4.35.0`. Eventually I imagine this package will move to the bundles repo itself, and then they can just rev the patch together as needed.

liliankasem: I really wanted a cry emoji but a thumbs down will have to do

liliankasem: Thoughts on how we can version the workload whlst keeping the bundle version match?

liliankasem: Actually we don't need this - this is now content only so there is no workload code/logic we need versioning for. a 1-1 map with bundles is sufficient

src/Workloads/Tools/ExtensionBundles/Directory.Version.props
liliankasem: not applicable

soninaren (changes requested):
One minor change requested. otherwise looks good to me

src/Workloads/Tools/ExtensionBundles/DownloadDefaultBundle.targets
```diff
@@ -0,0 +1,71 @@
...
+
+    The CDN URL always carries the bare $(BundleVersion); the workload pkg's
+    own prerelease label (preview/experimental.N) is not part of the URL.
+
+    TODO: switch the pack-time payload source from the public CDN to the
+    bundle CI build artifacts so the workload always ships the exact
+    bundle drop we own, instead of whatever happens to be on the CDN.
+  -->
+
+  <PropertyGroup Condition="'$(BundleChannel)' == 'stable'">
```
jviau: Use 1 property group, put the condition on the individual properties.

src/Workloads/Tools/ExtensionBundles/DownloadDefaultBundle.targets
```diff
@@ -0,0 +1,71 @@
...
+
+    The CDN bundle id is selected from $(BundleChannel):
+      stable        -> Microsoft.Azure.Functions.ExtensionBundle
+      preview       -> Microsoft.Azure.Functions.ExtensionBundle.Preview
+      experimental  -> Microsoft.Azure.Functions.ExtensionBundle.Experimental
+
+    The CDN URL always carries the bare $(BundleVersion); the workload pkg's
+    own prerelease label (preview/experimental.N) is not part of the URL.
+
+    TODO: switch the pack-time payload source from the public CDN to the
```
jviau, line 24: I recommend running this during restore, so we can keep all our network related operations confined to restore only. To do this use `AfterTargets="CollectPackageReferences"`. There might also be a condition we should check that it is an actual restore being called. I'll take a look soon and follow up.

src/Workloads/Tools/ExtensionBundles/README.md
```diff
@@ -1,24 +1,58 @@
...
+`Directory.Version.props`:
+
+```bash
+dotnet pack ... /p:BundleChannel=preview
+dotnet pack ... /p:BundleChannel=experimental
+```
+
+## Layout
+
+The package places the bundle payload under `tools/any/` (Workload Spec
```
jviau: Can we reuse `contentFiles` for content workloads? seems more fitting honestly.

jviau (commented):
I may have missed this, but how are content workloads processed post-install? Is there content copied or linked anywhere?

liliankasem:
> I may have missed this, but how are content workloads processed post-install? Is there content copied or linked anywhere?

No it's not copied or linked - we just preference the content directly from the workloads/ install path

src/Workloads/Tools/ExtensionBundles/README.md
liliankasem: I think that makes sense - but out of scope for this PR. If we want to make a change we should add to the spec and update all content workloads in one PR. Edit: actually the spec talks about why contentFiles isn't a good fit for workloads in general: https://github.com/Azure/azure-functions-core-tools/blob/docs/proposed/workload-package-layout.md#3-pre-existing-patterns-considered

jviau:
> > I may have missed this, but how are content workloads processed post-install? Is there content copied or linked anywhere?
> 
> No it's not copied or linked - we just preference the content directly from the workloads/ install path

How does the cli know to include these folders for bundle probing? Are we assuming all `kind:content` workloads are bundles only?

src/Workloads/Tools/ExtensionBundles/DownloadDefaultBundle.targets
jviau: How will a local build work?

liliankasem:
> > > I may have missed this, but how are content workloads processed post-install? Is there content copied or linked anywhere?
> > 
> > 
> > No it's not copied or linked - we just preference the content directly from the workloads/ install path
> 
> How does the cli know to include these folders for bundle probing? Are we assuming all `kind:content` workloads are bundles only?

No `kind:content` workloads are not bundles only, we do not make that assumption. However, we do know that bundles will be in the bundles workload id `Azure.Functions.Cli.Workloads.ExtensionBundles`

PR #5070, opened 2026-05-21
bug: dedup workload-contributed init options

liliankasem:
Closes the `// TODO: detect option-name collisions across workloads` in `InitCommand`.

## Why
Multiple workloads can legitimately contribute the same init option, today every extension-bundle stack (Node, Python, Go) defines both `--no-bundle` and `--bundles-channel`/`-c`. Before this change `InitCommand` naively added each duplicate to its `Options` list. That had two user-visible problems:

1. `func init --help` listed `--no-bundle` once per installed stack.
2. `parseResult.GetValue(option)` is identity-based in System.CommandLine 2.x. Only the **first** registered instance binds to the parsed value, so a workload that "lost" the registration race silently read `default(T)` even when the user passed the flag. Whether `func init --stack python --no-bundle` actually skipped the bundle depended on registration order across workloads.

## What
- New `IInitOptionRegistry` (abstraction) + `InitOptionRegistry` (concrete) in `src/Abstractions/Projects/`. Workloads register options through `registry.GetOrAdd(...)` and stash the returned canonical instance. Same-named contributions across workloads collapse to one instance; type mismatches and alias collisions throw with both workload ids in the message.
- `IProjectInitializer.GetInitOptions()` becomes `GetInitOptions(IInitOptionRegistry registry)` so the registry is the single place options enter the command.
- New `CommonInitOptions` factory (`NoBundle()`, `BundlesChannel()`) so the wording and defaults of cross-stack options live in one file.
- Node / Python / Go / DotNet initializers updated to use the registry and (where applicable) `CommonInitOptions`. Workload tests updated to drive `GetInitOptions` through an `InitOptionRegistry`.
- New `InitCommand` tests cover the four interesting cases: dedup-by-name, correct reads from the "loser" workload, type-mismatch error, alias collision error.
- Docs (`docs/cli-architecture.md`, `docs/building-a-workload.md`, `.github/skills/create-workload/SKILL.md`) updated to match the new signature and call out `CommonInitOptions`.

## Verification
- `dotnet build -c Release`: 0 warnings, 0 errors.
- `dotnet test -c Release`: 603 / 603 passing across Func.Tests + four workload test projects.

## Follow-up (out of scope)
Dynamic per-stack option visibility, hide a stack's options on workload-aware commands once `.func/config.json` pins a stack. Tracked separately in #TBD (will open after this lands).

### DIFF
--- a/src/Abstractions/Projects/CommonInitOptions.cs
+++ b/src/Abstractions/Projects/CommonInitOptions.cs
@@ -18,10 +18,10 @@ namespace Azure.Functions.Cli.Projects;
 public static class CommonInitOptions
 {
     /// <summary>
-    /// <c>--no-bundle</c>. Used by every stack that emits a default extension bundle entry
+    /// <c>--no-bundles</c>. Used by every stack that emits a default extension bundle entry
     /// in <c>host.json</c> (Node, Python, Go, ...).
     /// </summary>
-    public static Option<bool> NoBundle() => new("--no-bundle")
+    public static Option<bool> NoBundle() => new("--no-bundles")
     {
         Description = "Skip writing the default extensionBundle block in host.json.",
         DefaultValueFactory = _ => false,
--- a/src/Abstractions/Projects/CommonInitOptions.cs
+++ b/src/Abstractions/Projects/CommonInitOptions.cs
@@ -1,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.CommandLine;
+
+namespace Azure.Functions.Cli.Projects;
+
+/// <summary>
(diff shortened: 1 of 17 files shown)


src/Abstractions/Projects/CommonInitOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+/// Each method returns a fresh instance per call. The registry collapses duplicates so the
+/// option appears once in <c>--help</c> and every workload sees the same parsed value.
+/// </remarks>
+public static class CommonInitOptions
+{
+    /// <summary>
+    /// <c>--no-bundle</c>. Used by every stack that emits a default extension bundle entry
+    /// in <c>host.json</c> (Node, Python, Go, ...).
+    /// </summary>
+    public static Option<bool> NoBundle() => new("--no-bundle")
```
liliankasem: @copilot 

```suggestion
    public static Option<bool> NoBundles() => new("--no-bundles")
```

PR #5078, opened 2026-05-22
Add DotNetProjectFactory

satvu:
## Summary

Introduces DotNetProjectFactory — the .NET workload's implementation of IFunctionsProjectFactory — which detects .NET Functions projects from both source directories and pre-built output directories. Also adds a generic AddProjectFactory<TFactory>() overload to FunctionsCliBuilder so factories can be DI-resolved with constructor dependencies.

### Project Detection (DotNetProjectFactory)
- Detects **source projects**: single .csproj/.fsproj in the working directory → DotNetSourceProject
- Detects **output directories**: presence of host.json + worker.config.json + .azurefunctions/ → DotNetOutputProject
- Returns NotCreated when no match, or when multiple project files are ambiguous

### Project Hierarchy
- **DotNetProject** (abstract base): common stack metadata (StackName = "dotnet", SupportsExtensionBundles = false, worker reference)
- **DotNetSourceProject**: requires build before host startup — calls dotnet build then resolves OutputPath via MSBuild
- **DotNetOutputProject**: already compiled, no build step needed

### DI-based Factory Registration
- New FunctionsCliBuilder.AddProjectFactory<TFactory>() abstract method
- DefaultFunctionsCliBuilder implementation uses ActivatorUtilities so factories can take constructor dependencies (e.g., IDotnetCliRunner)
- DotNetWorkload.Configure now calls uilder.AddProjectFactory<DotNetProjectFactory>()

### DotnetCliRunner Enhancements
- Added RunWithOutputAsync to IDotnetCliRunner for capturing stdout (used by DotNetSourceProject to read OutputPath)
- Implementation reads stdout/stderr concurrently to avoid deadlock

### Folder Reorganization
- Moved DotnetCliRunner, DotnetPathResolver, and related interfaces into DotNetCli/ subfolder

### Tests
- DotNetProjectFactoryTests: empty dir, single csproj/fsproj, multiple projects, nonexistent dir, output directory detection, non-.NET files ignored
- DotNetSourceProjectTests: build + output path resolution, cancellation, null argument validation
- DotNetWorkloadTests: verifies factory registration


src/Workloads/Stacks/DotNet/DotNetProjectFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+            FunctionsWorkerResolutionResult.Resolved resolved =>
+                Created(new DotNetFunctionsProject(context.WorkingDirectory, resolved.Worker), reason),
+            FunctionsWorkerResolutionResult.NotResolved notResolved =>
+                Failed(ProjectCreationFailures.WorkerNotResolved(notResolved.Failure)),
+            _ => throw new InvalidOperationException(
+                $"Unsupported worker resolution result: {workerResult.GetType().FullName}"),
+        };
+    }
+
+    private static string? TryFindSingleProjectFile(DirectoryInfo directory, out string? notCreatedReason)
```
satvu: Do we want to exclude `extensions.csproj` and `extensions.fsproj`? 

Legacy has [this](https://github.com/Azure/azure-functions-core-tools/blob/b6cac7872569134d6309fda6ba3ae945b38ad69c/src/Cli/func/Helpers/DotnetHelpers.cs#L280) specific case where it's excluded unless directory is named that?

src/Workloads/Stacks/DotNet/DotNetFunctionsProject.cs
```diff
@@ -0,0 +1,27 @@
...
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Projects;
+using Azure.Functions.Cli.Workers;
+
+namespace Azure.Functions.Cli.Workloads.DotNet;
+
+/// <summary>
+/// Represents a resolved .NET Functions project.
+/// </summary>
+internal sealed class DotNetFunctionsProject(WorkingDirectory workingDirectory, IFunctionsWorker worker) : FunctionsProject
```
satvu: Will add overrides for `PrepareForHostRunAsync` and other methods in future iterations - this is where behavior such using/switching to an output directory would live if that strategy is implemented.

src/Workloads/Stacks/DotNet/DotNetOutputProject.cs
```diff
@@ -0,0 +1,28 @@
...
+using Azure.Functions.Cli.Projects;
+using Azure.Functions.Cli.Workers;
+
+namespace Azure.Functions.Cli.Workloads.DotNet;
+
+/// <summary>
+/// A .NET Functions project detected from build output (has host.json, worker.config.json, and an .exe).
+/// Already compiled; no build step needed before host startup.
+/// </summary>
+internal sealed class DotNetOutputProject(WorkingDirectory workingDirectory, IFunctionsWorker worker) : FunctionsProject
```
satvu: Dotnet has two `FunctionsProject` types, one for running from source project and the other from build output with slightly different behavior. 

Build output doesn't need a build step or extra path checking. Splitting into two projects types may (?) enable other differentiations later on (aggent suggests hot reload as a possibility when running from source project).

fabiocav (approved):
Discussed small enhancements offline, but flagging as approved.

src/Workloads/Stacks/DotNet/DotNetProjectFactory.cs
satvu: No longer needed - legacy concern.

src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
```diff
@@ -0,0 +1,66 @@
...
+            await BuildAsync(projectDir, cancellationToken);
+        }
+
+        DirectoryInfo outputDirectory = await GetOutputDirectoryAsync(projectDir, cancellationToken);
+        context.StartupDirectory = outputDirectory;
+    }
+
+    private async Task<DirectoryInfo> GetOutputDirectoryAsync(string projectDir, CancellationToken cancellationToken)
+    {
+        string output = await _dotnetCli.RunWithOutputAsync(
```
satvu, line 40: We assume that if the user skips build that the project is already built and this should work. If it doesn't, an error will be thrown below.

src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
```diff
@@ -0,0 +1,66 @@
...
+            ["msbuild", ProjectFilePath, "--getProperty:OutputPath"],
```
satvu, line 41: `OutputPath` here should return the directory path so it is correct.

liliankasem (approved):

PR #5093, opened 2026-05-26
Add go support (preview)

harshivcodes:
Go Language Support (Preview)

This PR adds preview support for the Go programming language to Azure Functions Core Tools, enabling developers to create, build, and deploy Go-based Azure Functions.

Features Added

 - func init - Initialize Go projects with --worker-runtime go (also supports --go/--golang shortcuts)
 - func start - Run Go functions locally with hot reload support
 - func pack - Package Go functions for deployment (Linux AMD64 binary)
 - func publish - Deploy Go functions to Azure (Flex Consumption plan only)

Key Implementation Details

 - Go detected via go.mod presence (registered as FUNCTIONS_WORKER_RUNTIME=native)
 - Native worker configuration included in minified builds
 - Opt-in preview flag: FUNCTIONS_CLI_GO_PREVIEW=1
 - Requires Go 1.24 or later
 - Uses 1ES centralized Go module proxy in CI pipelines
 - Linux-only deployment (Windows host builds cross-compile to linux/amd64)

Related PRs

 - #4875 - func init support
 - #4892 - func start support  
 - #4943 - func pack and func publish support
 - #4959 - Remove legacy gozip binary
 - #4966 - Add preview opt-in flag
 - #5062 - Include native worker in minified builds

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,6 +1,4 @@
-# Azure Functions CLI 4.11.1
-# Azure Functions CLI 4.11.0
-# Azure Functions CLI 4.12.0-preview.1
+# Azure Functions CLI 4.12.0
 
 #### Host Version
 
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,4 +1,6 @@
 # Azure Functions CLI 4.11.1
+# Azure Functions CLI 4.11.0
+# Azure Functions CLI 4.12.0-preview.1
 
 #### Host Version
 
@@ -11,3 +13,4 @@
 
 - Fix `HttpsProxyAgent is not a constructor` error in `install.js` when installing behind a proxy (https-proxy-agent v9 requires a named import).
 - Remove `azure-functions-core-tools` from the `devDependencies` of generated function app templates (`package-js.json`, `package-js-v4.json`, `package-ts.json`, `package-ts-v4.json`). Users get the CLI from their system install; pulling it in again via `npm install` doubled disk usage and made every project's install fragile to Core Tools postinstall regressions.
+- Add Go language support to `func init`, `func start`, `func pack`, and `func publish` (#4875, #4892, #4943)
--- a/release_notes.md
+++ b/release_notes.md
@@ -11,4 +11,4 @@
 
 - Fix `HttpsProxyAgent is not a constructor` error in `install.js` when installing behind a proxy (https-proxy-agent v9 requires a named import).
 - Remove `azure-functions-core-tools` from the `devDependencies` of generated function app templates (`package-js.json`, `package-js-v4.json`, `package-ts.json`, `package-ts-v4.json`). Users get the CLI from their system install; pulling it in again via `npm install` doubled disk usage and made every project's install fragile to Core Tools postinstall regressions.
-- Add Go language support to `func init`, `func start`, `func pack`, and `func publish` (#4875, #4892, #4943)
+- Add **preview** Go language support to `func init`, `func start`, `func pack`, and `func publish` (#4875, #4892, #4943). Behaviour, build/publish flags, and deployment layout may change before GA.
--- a/src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
+++ b/src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
@@ -164,12 +164,49 @@ namespace Azure.Functions.Cli.Actions.AzureActions
                 throw new CliException($"--build {PublishBuildOption} is not supported for Windows Elastic Premium Function Apps.");
             }
 
+            // Get the WorkerRuntime
+            var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager, refreshSecrets: true);
+            if (workerRuntime != WorkerRuntime.None)
+            {
+                GlobalCoreToolsSettings.CurrentWorkerRuntime = workerRuntime;
+            }
+            else
+            {
+                workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntime;
+            }
+
+            // Go is cross-compiled to linux/amd64 and currently only supported on Flex Consumption.
+            // Reject Windows targets, non-Flex SKUs, and unsupported build modes up front.
+            if (workerRuntime == WorkerRuntime.Go)
+            {
+                if (!functionApp.IsLinux)
+                {
+                    throw new CliException("Go is only supported for Linux Function Apps.");
+                }
+
+                if (!functionApp.IsFlex)
+                {
+                    throw new CliException("Go is only supported on Flex Consumption Function Apps.");
+                }
+
+                if (PublishBuildOption == BuildOption.Remote || PublishBuildOption == BuildOption.Container)
+                {
+                    throw new CliException(
+                        $"--build {PublishBuildOption} is not supported for Go. Run 'func publish' without '--build {PublishBuildOption}'.");
+                }
+
+                // BuildNativeDeps is checked separately because ResolveBuildOption (called later) flips
+                // PublishBuildOption to BuildOption.Container when --build-native-deps is set, which would
+                // bypass the Go build branch and ship a zip with a missing or stale 'app' binary.
+                if (BuildNativeDeps)
+                {
+                    throw new CliException("--build-native-deps is not supported for Go. Run 'func publish' without '--build-native-deps'.");
+                }
+            }
+
             // Get the GitIgnoreParser from the functionApp root
             var functionAppRoot = ScriptHostHelpers.GetFunctionAppRootDirectory(Environment.CurrentDirectory);
             var ignoreParser = PublishHelper.GetIgnoreParser(functionAppRoot);
-
-            // Get the WorkerRuntime
-            var workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntime;
 
             // Get Stacks once for both .NET version determination and EOL checking
             var stacks = await AzureHelper.GetFunctionsStacks(AccessToken, ManagementURL);
@@ -223,6 +260,26 @@ namespace Azure.Functions.Cli.Actions.AzureActions
             if (isNonCsxDotnetRuntime && !NoBuild && PublishBuildOption != BuildOption.Remote)
             {
                 await DotnetHelpers.BuildAndChangeDirectory(Path.Combine("bin", "publish"), DotnetCliParameters);
+            }
+            else if (workerRuntime == WorkerRuntime.Go &&
+                     PublishBuildOption != BuildOption.Remote &&
+                     PublishBuildOption != BuildOption.Container)
+            {
+                // Go always builds locally — there is no remote/container path for Go publishes.
+                // The remote/container conditions are redundant given the up-front rejection above;
+                // keeping them documents intent and stays defensive against future changes to ResolveBuildOption.
+                GoHelpers.AssertGoFunctionAppLayout(functionAppRoot);
+
+                if (NoBuild)
+                {
+                    // Trust the user's pre-built binary, but verify it is actually a linux/amd64 ELF
+                    // so we don't silently upload a host-OS binary to a Linux Flex host.
+                    GoHelpers.AssertLinuxAmd64Binary(functionAppRoot);
+                }
+                else
+                {
+                    await GoHelpers.BuildForLinux(functionAppRoot);
+                }
             }
 
             if (!isNonCsxDotnetRuntime)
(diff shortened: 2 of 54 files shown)


src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -45,7 +45,7 @@ public static WorkerRuntime CurrentWorkerRuntime
             {
                 if (_currentWorkerRuntime == WorkerRuntime.None)
                 {
-                    ColoredConsole.Error.WriteLine(QuietWarningColor("Can't determine project language from files. Please use one of [--dotnet-isolated, --dotnet, --javascript, --typescript, --python, --powershell, --custom]"));
+                    ColoredConsole.Error.WriteLine(QuietWarningColor("Can't determine project language from files. Please use one of [--dotnet-isolated, --dotnet, --javascript, --typescript, --python, --powershell, --custom, --go]"));
```
liliankasem: If this is going into main, we need clear signposting that this is a preview


```suggestion
                    ColoredConsole.Error.WriteLine(QuietWarningColor("Can't determine project language from files. Please use one of [--dotnet-isolated, --dotnet, --javascript, --typescript, --python, --powershell, --custom, --go (preview)]"));
```

src/Cli/func/Helpers/GlobalCoreToolsSettings.cs
```diff
@@ -143,6 +143,10 @@ public static void Init(ISecretsManager secretsManager, string[] args)
                 {
                     _currentWorkerRuntime = WorkerRuntime.Custom;
                 }
+                else if (args.Contains("--go") || args.Contains("--golang"))
+                {
+                    _currentWorkerRuntime = WorkerRuntime.Go;
```
liliankasem, line 148: A user running func init --go bypasses preview gating entirely -> ensure it still sets `FUNCTIONS_CLI_GO_PREVIEW` in local settings json so that other commands can pick up this is a go project

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -25,7 +25,9 @@ public enum WorkerRuntime
         [DisplayString("powershell")]
         Powershell,
         [DisplayString("custom")]
-        Custom
+        Custom,
+        [DisplayString("go")]
```
liliankasem:
```suggestion
        [DisplayString("go (preview)")]
```

src/Cli/func/Common/Executable.cs
```diff
@@ -139,6 +166,39 @@ public async Task<int> RunAsync(Action<string> outputCallback = null, Action<str
             }
         }
 
+        private void DrainAsyncOutput()
+        {
+            try
+            {
+                if (!Process.WaitForExit(OutputDrainTimeoutMs))
```
liliankasem: The 5s timeout here doesn't bound what the comments say it bounds. By this point the process has already exited (we just awaited exitCodeTask), so Process.WaitForExit(5000) returns true immediately. The thing we actually want to bound — the no-arg Process.WaitForExit() on the next line that drains the async event handlers — has no timeout. If an OutputDataReceived/ErrorDataReceived handler ever hangs (e.g. a deadlock writing to Console.Out), the CLI hangs indefinitely.

Recommend switching to WaitForExitAsync(CancellationToken) so the cap actually applies to the drain, and the wait stops blocking a thread-pool thread:

```C#
private async Task DrainAsyncOutputAsync()
{
    try
    {
        using var cts = new CancellationTokenSource(OutputDrainTimeout);
        await Process.WaitForExitAsync(cts.Token).ConfigureAwait(false);
    }
    catch (OperationCanceledException)
    {
        {
            ColoredConsole.WriteLine(VerboseColor(
                $"Output drain for '{_exeName}' did not complete within {OutputDrainTimeout.TotalMilliseconds}ms; returning exit code with possibly truncated output."));
        }
    }
    catch (Exception ex) when (ex is InvalidOperationException or Win32Exception)
    {
        // unchanged
    }
}
```

And call sites become await DrainAsyncOutputAsync().ConfigureAwait(false);.

WaitForExitAsync is documented as draining async output handlers the same way the no-arg sync overload does, so the fix to the original race is preserved.

liliankasem: Second, with this change we should also update PythonHelpers.

This retry loop was a workaround for the exact race that Executable.DrainAsyncOutput now fixes (process exits before OutputDataReceived flushes). With the drain in place, sb is guaranteed populated by the time exe.RunAsync returns, so this loop is always a no-op on iteration 0.

Suggest removing it in this PR — leaving the workaround in will mislead future readers into thinking the race still exists somewhere else, and we'd want a regression to surface as a failure, not be silently masked here.

Replace  https://github.com/Azure/azure-functions-core-tools/blob/8b8f38641/src/Cli/func/Helpers/PythonHelpers.cs#L331-L342

With

```C#
if (exitCode == 0)
{
    return sb.ToString().Trim();
}
```

src/Cli/func/Common/Utilities.cs
```diff
@@ -64,6 +64,19 @@ internal static void WarnIfPreviewVersion()
             ColoredConsole.WriteLine();
         }
 
+        internal static void WarnIfGoWorkerRuntime(Helpers.WorkerRuntime workerRuntime)
```
liliankasem, line 67: WarnIfGoWorkerRuntime is defined in Utilities.cs:67 and called from only one place: InitAction.cs:218.

func start, func pack, func publish do not call it. A user who initialized a Go app a week ago, then runs func start today, will see no preview warning at any point during today's session.

Call Utilities.WarnIfGoWorkerRuntime(workerRuntime) from:

- StartHostAction.PreRunConditions (right after the runtime is resolved)
- PublishFunctionAppAction.RunAsync (right after the runtime is resolved)
- PackAction.ExecuteAsync (right after the runtime is resolved)

release_notes.md
```diff
@@ -8,3 +8,5 @@
   - Host Runtime Version: 4.48.100 (includes 4.848.100, 4.648.100)
 
 #### Changes
+
+- Add Go language support to `func init`, `func start`, `func pack`, and `func publish` (#4875, #4892, #4943)
```
```suggestion
- [Preview Feature] Add Go language support to `func init`, `func start`, `func pack`, and `func publish` (#4875, #4892, #4943)
```

src/Cli/func/Actions/AzureActions/PublishFunctionAppAction.cs
```diff
@@ -164,13 +164,50 @@ public override async Task RunAsync()
                 throw new CliException($"--build {PublishBuildOption} is not supported for Windows Elastic Premium Function Apps.");
             }
 
+            // Get the WorkerRuntime
+            var workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager, refreshSecrets: true);
```
liliankasem: On main, this line was `var workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntime`; (single global read set during Init). The new code re-resolves with refreshSecrets: true and writes back to the global for every publish, on every language. This is a behaviour change for Python/.NET/Node/PowerShell flows: if local.settings.json has been edited between Init and publish (scripted scenarios), the publish runtime can now differ from what Init selected.

The Go-specific motivation (resolving native → Go via project markers) only matters when the Init-time value was None. Suggested narrower fix:

```C#
var workerRuntime = GlobalCoreToolsSettings.CurrentWorkerRuntime;
if (workerRuntime == WorkerRuntime.None)
{
    workerRuntime = WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage(_secretsManager, refreshSecrets: true);
    if (workerRuntime != WorkerRuntime.None)
    {
        GlobalCoreToolsSettings.CurrentWorkerRuntime = workerRuntime;
    }
}
```

liliankasem: Same fix should be applied in StartHostAction.PreRunConditions (lines 736-742) for consistency.

src/Cli/func/Actions/HostActions/StartHostAction.cs
```diff
@@ -758,6 +768,20 @@ private async Task PreRunConditions()
             {
                 throw new CliException("Dotnet is required for PowerShell Functions. Please install dotnet (.NET Core SDK) for your system from https://www.microsoft.com/net/download");
             }
+            else if (GlobalCoreToolsSettings.CurrentWorkerRuntime == WorkerRuntime.Go)
```
liliankasem, line 778: EDIT:

The new Go branch depends on GlobalCoreToolsSettings.CurrentWorkerRuntime already being resolved from "native" → Go by the caller. The signature gives no hint of this coupling, so a future caller that doesn't go through PublishFunctionAppAction's re-resolve will silently take the wrong zip path. Suggest threading WorkerRuntime as an explicit parameter (the other branches at lines 34, 38 read the same global and have the same latent issue, but Go is the one being added now).

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -159,7 +164,7 @@ public static string NormalizeLanguage(string languageString)
             {
                 throw new ArgumentNullException(nameof(languageString), "language can't be empty");
             }
-            else if (_normalizeMap.ContainsKey(languageString))
+            else if (WorkerRuntimeStringToLanguage.ContainsKey(languageString))
```
liliankasem, line 167: Add a one-line test asserting NormalizeLanguage("golang") throws ArgumentException

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
```diff
@@ -169,15 +174,64 @@ public static string NormalizeLanguage(string languageString)
...
+            if (string.IsNullOrWhiteSpace(setting))
+            {
+                // SecretsManager.GetSecrets throws CliException when there is no project root
+                // (e.g. 'func pack' invoked from a parent directory before cwd is changed).
+                // Treat that as "no setting found" so callers can decide what to do.
+                try
+                {
+                    setting = secretsManager?.GetSecrets(refreshSecrets)?.FirstOrDefault(s => s.Key.Equals(Constants.FunctionsWorkerRuntime, StringComparison.OrdinalIgnoreCase)).Value;
+                }
+                catch (CliException)
```
liliankasem: Can you explain why you would need to swallow the exception here? The catch (CliException) { return WorkerRuntime.None; } discards two genuinely useful errors that SecretsManager.GetSecrets() raises:

1. "Unable to find project root. Expecting to find one of host.json, local.settings.json..." — thrown when the command is run outside any project.
2. "Failed to decrypt settings. Encrypted settings only be edited through 'func settings add'." — thrown when local.settings.json is encrypted and can't be decrypted.

The underlying failures still happen on this branch; the catch just hides them from the user. For func start, func azure functionapp publish, func new, and func init the original message resurfaces later because downstream code re-calls GetSecrets directly, so the user still sees the right error. But for func pack there is no second GetSecrets call, so the catch silently downgrades a specific actionable message into a generic one:

src/Cli/func/Common/Constants.cs
```diff
@@ -20,6 +20,12 @@ internal static partial class Constants
         public const string FuncIgnoreFile = ".funcignore";
         public const string FunctionsWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
         public const string FunctionsWorkerRuntimeVersion = "FUNCTIONS_WORKER_RUNTIME_VERSION";
+
+        // CLI-only preview opt-in for the Go worker. Read from env first, then local.settings.json
+        // (Values). Never sent to the host or to Azure (the CLI strips local.settings on publish).
+        // Intentionally Go-specific and short-lived: when the platform exposes a first-class
+        // mapping for "native" → concrete language, this flag and its consumers should be removed.
+        public const string FunctionsCliGoPreview = "FUNCTIONS_CLI_GO_PREVIEW";
```
liliankasem: I think we should rename this to `FUNCTIONS_CLI_NATIVE_LANGUAGE` - FUNCTIONS_CLI_GO_PREVIEW implies a feature flag and it is not a feature flag so I think this naming will be less confusign for customers.

We should also update the preview warning to say something like "Go support is in preview. The shape of FUNCTIONS_CLI_NATIVE_LANGUAGE, the build/publish behaviour, and the deployment layout may change before GA" or something to indicate that "FUNCTIONS_CLI_NATIVE_LANGUAGE" is a preview item and may change

liliankasem (changes requested):
1. Preview signposting is insufficient. Release notes, generated files, help text, and missing docs all let a user use Go without ever seeing the word "preview." - ensure that this is clearly sign posted.

2. Worth calling out in the PR description: the drain change in Executable runs for every _streamOutput: true caller (npm, dotnet, python, az, git, docker, kubectl, openssl, where, …), not just Go. I audited the call sites and the change is net-positive across stacks — no consumer breaks, and at least one (PythonHelpers.VerifyVersion, see other comment) has a documented workaround that this obsoletes. Per-call overhead is sub-millisecond on the normal path.

src/Cli/func/Helpers/WorkerRuntimeLanguageHelper.cs
satvu, line 167: Why the switch from using `_normalizeMap` to `WorkerRuntimeStringToLanguage`?

liliankasem: Will sync offline

liliankasem (approved):

vrdmr (approved):

PR #5094, opened 2026-05-26
build(vnext): per-RID pack for host workload

jviau:
This PR refactors the nuget packaging msbuild logic for the host workload. Now calling `dotnet pack` will fan out to packing each RID separately. Calling `dotnet pack -r {RID}` will pack that individual RID.

This PR also forces `self-contained` during pack.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -6,6 +6,7 @@
     <IsPackable>true</IsPackable>
     <IsPublishable>false</IsPublishable>
     <UpdateBuildNumber Condition="'$(IsReleaseCI)' == 'true'">true</UpdateBuildNumber>
+    <_CliRuntimeIdentifiers>win-x64;win-arm64;linux-x64;linux-arm64;osx-x64;osx-arm64</_CliRuntimeIdentifiers>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Func/Func.csproj
+++ b/src/Func/Func.csproj
@@ -5,18 +5,16 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net10.0</TargetFramework>
+    <RuntimeIdentifiers>$(_CliRuntimeIdentifiers)</RuntimeIdentifiers>
     <AssemblyName>func</AssemblyName>
     <RootNamespace>$(TopLevelNamespace)</RootNamespace>
     <InvariantGlobalization>false</InvariantGlobalization>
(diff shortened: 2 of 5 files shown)


fabiocav (commented):
Tried to validate this, need to double check...

I see the updated PR is green, but is the packaging issue fixed?  The Pack log shows only one host package being produced:
 
 `Azure.Functions.Cli.Workloads.Host.4.1048.200-preview...nupkg`
 
 It does not produce the expected RID-suffixed packages like:
 
 `Azure.Functions.Cli.Workloads.Host.linux-x64...nupkg`
 `Azure.Functions.Cli.Workloads.Host.win-x64...nupkg`
 etc.
 
 The build log also only shows the normal non-RID output:
 
 `out/bin/Workloads.Host/release/Azure.Functions.Cli.Workloads.Host.dll`
 
 not the RID-specific outputs under `release_linux-x64`, `release_win-x64`, etc.
 
The issue seems to be  `<SelfContained>true</SelfContained>` in `Workloads.Host.csproj`. With no explicit `RuntimeIdentifier`, the .NET SDK later infers a default RID for the current machine. Because of the import order:
 
 1. `Pack.Inner.targets` is checked before `Sdk.targets`, while `RuntimeIdentifier` is still empty, so it is not imported.
 2. `Sdk.targets` then infers `RuntimeIdentifier` because `SelfContained=true`.
 3. `Pack.Outer.targets` is checked after `Sdk.targets`, now sees `RuntimeIdentifier` as non-empty, and is skipped.
 
So neither custom target is imported in the normal `dotnet pack` case. The default SDK `Pack` target runs instead, producing a single package without the RID-specific fan-out or `StageHostWorkloadPayload`.

jviau:
Ah yeah I had self contained deferred, but it caused a build issue. Moving it fixed the build issue, but undid the pack logic.

jviau:
@fabiocav fixed the issue. It got more complicated than I anticipated because of how I needed to run a build for each RID with self-contained before pack. Unlike with the multi-RID publish for the CLI, which just builds & publishes together, this needed a separate build and pack so signing can occur in the middle for the dll's directly. But then it wasn't as simple as just building for each RID with self-contained. I needed the portable RID-less version still for tests!

In the end, I have a hook before build to go build all RID-specific versions with self-contained. Its ugly, but it works. We can look into adopting TFM-aliasing in the future, where we would have `<TargetFrameworks>net10.0;net10.0-win-x64;...</TargetFrameworks>` and so on.

liliankasem (approved):

PR #5097, opened 2026-05-26
Wire stack workloads into host start path

liliankasem:
Hook the Go, Node, and Python workloads into `func start` and have `PrepareProjectHostRunInitializationStep` populate real environment variables for the host process.

## Changes

**`PrepareProjectHostRunInitializationStep`**
Injects `ILocalSettingsProvider` and merges env vars in priority order:
1. `local.settings.json` `Values`
2. `state.HostEnvironmentVariables` (reserved for future steps)
3. `FUNCTIONS_WORKER_RUNTIME` from the resolved worker (always wins over stale `local.settings`)
4. `languageWorkers:<runtime>:workerDirectory` from the worker's config path (lights up installed workers without copying assets)
5. `state.BundleEnvVarsForHost`

**Per-stack `PrepareForHostRunAsync` hooks**
Each stack's `FunctionsProject` now extracts into its own file with an internal `Func<>` runner seam for testability.

- **Go**: runs `go build -o bin/<module> .` (module name read from `go.mod`, falls back to `bin/app`).
- **Node**: runs `npm install` when `node_modules` is missing, then `npm run build` when `package.json` declares a `build` script.
- **Python**: creates `.venv` if missing, runs `pip install -r requirements.txt`, sets `PYTHON_ISOLATE_WORKER_DEPENDENCIES=1`.

All failures bubble as `GracefulException(isUserError: true)` so the start command surfaces clean errors.

## Tests

- New `{Go,Node,Python}FunctionsProjectTests` exercise happy paths, idempotency (skip when deps already satisfied), and graceful failure on non-zero exit.
- `StartInitializationTests` extended with `DemoRunner_PopulatesEnvironmentVariablesForHostRun`: real `local.settings.json` on disk, asserts merge priority and that `languageWorkers:node:workerDirectory` resolves to the worker config's directory.
- Full suite green; clean `Release` build (no warnings).

### DIFF
--- a/src/Workloads/Stacks/Go/GoFunctionsProject.cs
+++ b/src/Workloads/Stacks/Go/GoFunctionsProject.cs
@@ -74,8 +74,16 @@ internal sealed class GoFunctionsProject : FunctionsProject
 
         try
         {
+            // The `module` directive is conventionally near the top of go.mod;
+            // bound the scan so a malformed file can't make us read megabytes.
+            int scanned = 0;
             foreach (string line in File.ReadLines(goModPath))
             {
+                if (++scanned > 100)
+                {
+                    break;
+                }
+
                 string trimmed = line.Trim();
                 if (!trimmed.StartsWith("module ", StringComparison.Ordinal))
                 {
@@ -108,7 +116,11 @@ internal sealed class GoFunctionsProject : FunctionsProject
     {
         try
         {
-            Directory.CreateDirectory(Path.GetDirectoryName(outputPath)!);
+            string? outputDir = Path.GetDirectoryName(outputPath);
+            if (!string.IsNullOrEmpty(outputDir))
+            {
+                Directory.CreateDirectory(outputDir);
+            }
 
             var psi = new ProcessStartInfo("go")
             {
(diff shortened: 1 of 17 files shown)


src/Workloads/Stacks/Go/GoFunctionsProject.cs
```diff
@@ -0,0 +1,141 @@
...
+            psi.ArgumentList.Add(outputPath);
+            psi.ArgumentList.Add(".");
+
+            using var process = Process.Start(psi);
+            if (process is null)
+            {
+                return (-1, "Failed to start 'go' process.");
+            }
+
+            string stderr = await process.StandardError.ReadToEndAsync(cancellationToken).ConfigureAwait(false);
```
satvu: AI review: 

**Potential deadlock in process I/O** - All three stacks read `StandardError.ReadToEndAsync()` *before* `WaitForExitAsync()`, but `StandardOutput` is also redirected and **never read**. If a child process writes enough to stdout to fill the OS pipe buffer (~4KB on Windows), it will block - and since nobody is draining stdout, the process hangs forever. Fix: either don't redirect stdout (`RedirectStandardOutput = false`) or also read it concurrently:

```csharp
Task<string> stderrTask = process.StandardError.ReadToEndAsync(ct);
Task<string> stdoutTask = process.StandardOutput.ReadToEndAsync(ct);
await Task.WhenAll(stderrTask, stdoutTask);
await process.WaitForExitAsync(ct);
```

liliankasem:
@copilot resolve the merge conflicts in this pull request

PR #5099, opened 2026-05-26
vnext: Add func quickstart command skeleton

manvkaur:
## Summary

Adds the bare-bones command skeleton for `func quickstart` — a new top-level command that will scaffold complete function apps from a CDN-hosted template manifest (designed in #5026). This PR is intentionally minimal: interfaces, models, and stub commands only. Functional implementation comes in follow-up PRs.

### What's in this PR

**Abstractions** (`src/Abstractions/Quickstart/`):
- `IQuickstartProvider` — per-stack interface workloads implement to opt into quickstart (analogous to `IProjectInitializer` for `func init`)
- `IQuickstartManifestService` — shared service interface for CDN manifest fetching
- `IQuickstartScaffolder` — shared service interface for template download + write
- `QuickstartEntry` — manifest entry model record
- `FetchMode` — enum: `Auto`, `Git`, `Http`

**Commands** (`src/Func/Commands/Quickstart/`):
- `QuickstartCommand` — parent `func quickstart` with `--language`, `--template`, `--resource`, `--iac`, `--search`, `--fetch` options
- `QuickstartListCommand` — `func quickstart list` with filter options
- `QuickstartInfoCommand` — `func quickstart info <id>`

**DI registration** — commands registered in `BuiltInCommands.cs`

All commands currently show "not yet available" until stack workloads register `IQuickstartProvider` implementations.

### Design decisions

- **Workload-extensible**: follows the `IProjectInitializer` pattern — each stack opts in by implementing `IQuickstartProvider`, shared helpers (manifest, scaffolder) are reusable
- **Sub-language prompting**: derived from `SupportedLanguages.Count > 1` (e.g. dotnet → C#/F#, node → JS/TS)
- **No implementations yet**: services and per-stack providers will be added in focused follow-up PRs

Design: #5026

### Pull request checklist

### Additional information

This is the first in a series of incremental PRs implementing `func quickstart`. Follow-ups:
1. Manifest service implementation (CDN fetch, ETag caching, TTL)
2. Scaffolder implementation (git clone, HTTP zip, path traversal prevention)
3. Per-stack `IQuickstartProvider` implementations (dotnet, go, node, python — java and powershell when vnext adds those stacks)
4. Full command logic (interactive prompts, filtering, non-TTY fallback)

### DIFF
--- a/src/Abstractions/Quickstart/FetchMode.cs
+++ b/src/Abstractions/Quickstart/FetchMode.cs
@@ -1,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Quickstart;
+
+/// <summary>
+/// Controls how template content is downloaded from GitHub.
+/// </summary>
+public enum FetchMode
+{
+    /// <summary>
+    /// Probe for git availability; use git if found, otherwise fall back to HTTP.
+    /// </summary>
+    Auto,
+
+    /// <summary>
+    /// Force git clone. Fails if git is not available.
+    /// </summary>
+    Git,
+
+    /// <summary>
+    /// Force HTTP zip download. Never invokes git.
+    /// </summary>
+    Http
+}
--- a/src/Abstractions/Quickstart/FetchMode.cs
+++ b/src/Abstractions/Quickstart/FetchMode.cs
@@ -14,12 +14,12 @@ public enum FetchMode
     Auto,
 
     /// <summary>
-    /// Force git clone. Fails if git is not available.
+    /// Use git clone. Fails if git is not available.
     /// </summary>
     Git,
 
     /// <summary>
-    /// Force HTTP zip download. Never invokes git.
+    /// Use HTTP zip download. Never invokes git.
     /// </summary>
     Http
 }
--- a/src/Func/Commands/Quickstart/QuickstartCommand.cs
+++ b/src/Func/Commands/Quickstart/QuickstartCommand.cs
@@ -1,0 +1,98 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.CommandLine;
+using Azure.Functions.Cli.Console;
+using Azure.Functions.Cli.Hosting;
+using Azure.Functions.Cli.Quickstart;
+
+namespace Azure.Functions.Cli.Commands.Quickstart;
+
+/// <summary>
+/// Top-level <c>func quickstart</c> command. Fetches a CDN-hosted manifest of
+/// complete function-app templates and scaffolds the selected one into a target
+/// directory. Delegates language resolution to registered <see cref="IQuickstartProvider"/>
+/// implementations contributed by each stack workload.
+/// </summary>
+internal sealed class QuickstartCommand : FuncCliCommand, IBuiltInCommand
+{
+    public Option<string?> LanguageOption { get; } = new("--language", "-l")
+    {
+        Description = "Worker runtime or language (e.g. python, node, java, dotnet, csharp, fsharp, javascript, typescript, powershell)"
+    };
+
+    public Option<string?> TemplateOption { get; } = new("--template", "-t")
+    {
+        Description = "Template id from the manifest (e.g. http-trigger-python-azd) — skips all prompts"
+    };
+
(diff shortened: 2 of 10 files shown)


manvkaur:
image

src/Abstractions/Quickstart/FetchMode.cs
```diff
@@ -0,0 +1,25 @@
...
+/// </summary>
+public enum FetchMode
+{
+    /// <summary>
+    /// Probe for git availability; use git if found, otherwise fall back to HTTP.
+    /// </summary>
+    Auto,
+
+    /// <summary>
+    /// Force git clone. Fails if git is not available.
```
liliankasem:
```suggestion
    /// Use git clone. Fails if git is not available.
```

src/Func/Commands/Quickstart/QuickstartCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+
+/// <summary>
+/// Top-level <c>func quickstart</c> command. Fetches a CDN-hosted manifest of
+/// complete function-app templates and scaffolds the selected one into a target
+/// directory. Delegates language resolution to registered <see cref="IQuickstartProvider"/>
+/// implementations contributed by each stack workload.
+/// </summary>
+internal sealed class QuickstartCommand : FuncCliCommand, IBuiltInCommand
+{
+    public Option<string?> LanguageOption { get; } = new("--language", "-l")
```
liliankasem, line 24: Do we want to add --stack for consistency with init? I think it might be weird to see stack and language in one and just language in the other?

src/Func/Commands/Quickstart/QuickstartCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        Description = "Worker runtime or language (e.g. python, node, java, dotnet, csharp, fsharp, javascript, typescript, powershell)"
```
liliankasem: Similar to init, this should dynamically populate from the available quickstart providers and not hard coded. You only see the options that you have installed/available to you

src/Func/Commands/Quickstart/QuickstartCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        Description = "Filter by trigger/binding resource (e.g. http, timer, blob, eventhub, servicebus, cosmos, sql, mcp, durable)"
+    };
+
+    public Option<string?> IacOption { get; } = new("--iac")
+    {
+        Description = "Filter by infrastructure-as-code type (e.g. bicep, terraform, none)"
+    };
+
+    public Option<string?> SearchOption { get; } = new("--search", "-s")
```
liliankasem: Does this make more sense as a subcommand? and -r and --iac make more sense as search options no?

src/Func/Commands/Quickstart/QuickstartInfoCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Azure.Functions.Cli.Commands.Quickstart;
+
+/// <summary>
+/// Displays detailed information about a specific template by id.
+/// </summary>
+internal sealed class QuickstartInfoCommand : FuncCliCommand
+{
+    public Argument<string> TemplateIdArgument { get; } = new("id")
+    {
+        Description = "Template id from the manifest (e.g. http-trigger-python-azd). Use 'func quickstart list' to see available ids."
```
liliankasem:
```suggestion
        Description = "Template ID from the manifest (e.g. http-trigger-python-azd). Use 'func quickstart list' to see available ids."
```

src/Func/Commands/Quickstart/QuickstartCommand.cs
manvkaur: Want to make sure I understand the suggestion correctly — are you saying:

 A) --resource and --iac should only live on func quickstart list (the browsing/discovery surface), and not on the bare func quickstart scaffold command?

 B) Or that there should be a separate func quickstart search subcommand (instead of using list with filter options)?

 If A — I kept them on the parent because they narrow the interactive template prompt before the user picks. e.g. func quickstart --language python --resource http skips showing timer/blob/queue triggers and goes straight to HTTP templates. Without them the user has to scroll the full list or alreadyknow the exact --template ID. They're optional — if you don't pass them you get the full unfiltered picker.

 If B — happy to discuss renaming list to search, though list felt more natural since the output is always a table of results (even without filters applied). Let me know which you had in mind!

liliankasem (approved):

PR #5103, opened 2026-05-26
workload install: require alias resolution by default

liliankasem:
## What

Without `--exact`, `func workload install <name>` now treats the positional argument as an alias **only**. If no package in the catalog declares the alias, the command fails with a hint to pass `--exact <packageId>` instead of silently installing a package whose id happens to match the input string.

## Why

Previously `ResolveAliasOrIdAsync` returned the raw input when no alias matched, so `InstallFromCatalogAsync` would then look it up as a literal package id. That's a squatting hole: an attacker who publishes a package literally named `node` to a configured source could satisfy `func workload install node`.

After this PR the install path always operates on a package id that was either:
- explicitly opted into by the user via `--exact <packageId>`, or
- resolved from an alias declared by a package the catalog returned.

## Notes

- `func workload uninstall` already operates on `WorkloadEntry.PackageId` from the local registry (the alias is only used to find the entry), so no logic change there.
- Local `.nupkg` installs are unaffected.
- This is PR 1. PR 2 will add publisher / signing / cert-bundle validation on top.

## Tests

- New `InstallFromCatalog_AliasResolution_NoMatch_Throws` covers the strict-failure path.
- Existing alias-fallback and `--exact` tests still pass.
- Full `Func.Tests` suite: 596 / 596 pass, clean release build with `TreatWarningsAsErrors=true`.

### DIFF
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -27,7 +27,7 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
 
     public Argument<string> WorkloadArgument { get; } = new("id")
     {
-        Description = "Workload package id, alias, or path to a local .nupkg.",
+        Description = "Workload alias (default) or path to a local .nupkg. Pass --exact to install by literal package id.",
     };
 
     public Option<string?> VersionOption { get; } = new("--version", "-v")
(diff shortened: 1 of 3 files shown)


liliankasem:
Closing — after digging into the existing pipeline, the alias-only failure I added in this PR doesn't actually close the squat hole on its own. The real gate is publisher trust (signing + cert bundle), which is PR 2. Will reopen against a fresh branch focused on that.

PR #5104, opened 2026-05-26
func setup: install stack workloads for the profile's supported runtimes

liliankasem:
## Gap

After a fresh `func setup`, running `func start` against a Node/Python/Go/.NET-isolated project still required a manual `func workload install <stack>` because the stack workloads (which own the project resolver and pre-run hooks) weren't installed.

## Fix

`func setup` now installs the matching stack workload alongside the worker. Selection is driven by the project's worker runtime, not the profile:

1. **Existing project:** detected from `local.settings.json` via the already-wired `StackOptions:Runtime` config key.
2. **Empty folder, interactive:** prompts the user to pick a stack.
3. **Non-interactive / scripted:** pass `--features <runtime>` (e.g. `--features node`).

Stack install happens inside the worker loop, so each worker runtime that has a matching stack package (`node`, `python`, `go`, `dotnet-isolated`) also gets its stack workload. Runtimes without a stack package (`java`, `powershell`, `custom`) skip silently.

## Decisions

- **Version range:** stacks resolve with no pinned range (`*`, latest stable, or latest prerelease with `--prerelease`). Profile schema doesn't currently model stacks; can be extended later.
- **Mapping:** kept inline on `SetupDependency` next to the existing `WorkerPackagePrefix` pattern. `SetupDependency.SupportsStack(runtime)` + `SetupDependency.Stack(runtime)` are the entry points.
- **Test seam:** `TestInteractionService.IsInteractive` is now `virtual` so the prompt path can be exercised by subclassing.

## Tests

- `RunAsync_ConfiguredDotnetIsolated_InstallsWorkerAndStackAndSkipsBundle` — local.settings detection installs the dotnet-isolated stack.
- `RunAsync_FeatureForRuntime_InstallsStackWorkload` — `--features node` installs the Node stack.
- `RunAsync_FeatureWithoutStackMapping_DoesNotInstallStack` — `--features java` doesn't try to install a stack.
- `RunAsync_InteractiveEmptyFolder_PromptsForStackAndInstalls` — empty folder + interactive prompts and installs the selected stack.

All existing tests pass, `CI=true dotnet build -c Release` is clean.


src/Func/Commands/Setup/SetupRunner.cs
```diff
@@ -283,6 +283,25 @@ private async Task<SetupDependencyPlan> BuildDependencyPlanAsync(
             dependencies.Add(SetupDependency.Worker(workerRuntime, workerRange));
         }
 
+        // Stack workloads own the project resolver and pre-run hooks. Install one per
+        // supported runtime in the profile so a fresh machine can run `func start`
+        // against any supported language after `func setup`. Runtimes without a
+        // matching stack workload (java, powershell, custom, dotnet in-proc) are
+        // skipped silently. Stacks aren't pinned in the profile today, so they
+        // resolve with no version range ("*", latest stable, or latest prerelease
+        // when --prerelease is passed). Ordered after workers, before bundles.
```
fabiocav: nit: Can we have a more succinct comment?

src/Func/Commands/Setup/SetupRunner.cs
```diff
@@ -283,6 +283,25 @@ private async Task<SetupDependencyPlan> BuildDependencyPlanAsync(
...
+        if (profileScope.Profile?.SupportedRuntimes is { } runtimes)
```
liliankasem: Not sure that using supported runtimes is the right approach, but might be fine for now?

The alternative is that we can prompt for a stack, and provide a --stack option for non-interactive

cc: @fabiocav

src/Func/Commands/Setup/SetupRunner.cs
```diff
@@ -601,6 +620,20 @@ internal sealed record SetupDependency(
     string? ResolvedPackageId)
 {
     private const string WorkerPackagePrefix = "Azure.Functions.Cli.Workloads.Workers.";
+    private const string StackPackagePrefix = "Azure.Functions.Cli.Workloads.";
+
+    // Maps a profile `supportedRuntimes` entry to the suffix of the matching stack
+    // workload package. Runtimes that don't have a corresponding stack workload
+    // (java, powershell, custom, dotnet in-proc) are intentionally absent and
+    // callers skip them silently.
+    private static readonly IReadOnlyDictionary<string, string> _runtimeToStackSuffix =
```
fabiocav: Stack info in the CLI starts to worry me :)

We should track this so we can derive from existing metadata or a different mechanism.

fabiocav (approved):

src/Func/Commands/Setup/SetupRunner.cs
liliankasem: Yeah not super happy about this but also OK to hardcode this for preview and figure it out later

src/Func/Commands/Setup/SetupRunner.cs
fabiocav: The design uses profiles for constraining installation only, not to derive what should be installed.

Is there a concrete scenario you had in mind that would require this?

fabiocav (changes requested):
See last comment about scenarios.

PR #5107, opened 2026-05-27
workload install: verify publisher trust before installing

liliankasem:
## What

Gates `func workload install` / `update` on publisher trust. Before any payload is extracted, the verifier checks two things:

1. **Package id prefix:** must start with `azure.functions.cli.workloads.` (the official Microsoft-published namespace).
2. **Signature chain:** the `.nupkg` must carry a primary signature whose signing certificate chains to a root in the bundled trust anchor store (`trusted-roots.pem`, embedded resource). The chain is built with `X509ChainTrustMode.CustomRootTrust`, so only the bundled roots are trusted, not whatever happens to be in the OS trust store.

Either failure rejects the install with a clear `GracefulException`. A single opt-out flag, `--allow-untrusted`, bypasses both gates and applies to catalog installs, local `.nupkg` installs, and `update`. Required for dev workflows where someone packs a workload locally and installs the unsigned `.nupkg`.

## Why a real cert bundle (not fingerprint pinning)

An earlier iteration of this PR pinned SHA-256 fingerprints of specific signing certs. That had two problems:

- No notion of expiry or chain. Anything that matched the hash was trusted forever, even if Microsoft rotated the cert or it was revoked.
- Brittle to legitimate cert rotation: rotating the leaf required shipping a CLI update.

A PEM bundle of *roots* (the same model cURL and OpenSSL use for the public web PKI) lets the publisher rotate intermediates and leafs freely; we only re-bundle when the root itself rotates. Revocation is intentionally `NoCheck` for v1 (OCSP/CRL on a CLI install path means every install needs network egress to a CA responder, hostile in offline/restricted environments). Opt-in revocation is a sensible follow-up.

## Out of scope

- OCSP / CRL checks (follow-up, behind a flag).
- Repository signatures (Microsoft publishes author-signed packages).
- Auto-rotating the bundle from a remote endpoint (future PR).

## Release blocker

`src/Func/Workloads/Install/Trust/trusted-roots.pem` ships as a placeholder (comment-only file). The mechanism is in place; the actual Microsoft NuGet author signing root must be added before a release that gates installs on this bundle. Until then the empty bundle causes every install of an `azure.functions.cli.workloads.*` package to fail trust verification unless `--allow-untrusted` is passed.

## Tests

- `WorkloadTrustVerifierTests`: allow-untrusted bypass, prefix mismatch, empty anchor store, unsigned package, null-arg validation, plus chain-validates-to-bundled-root and chain-does-not-validate paths using synthesized self-signed certs via `CertificateRequest`.
- `EmbeddedTrustAnchorStoreTests`: default load, caching, missing resource, populated bundle, empty/placeholder bundle.
- `WorkloadInstallerTests`: verifier wired into the installer; untrusted result short-circuits install; `allowUntrusted` flag forwards through the pipeline.
- `WorkloadInstallCommandTests` / `WorkloadUpdateCommandTests`: `--allow-untrusted` flag is plumbed.

All 612 tests pass; `CI=true dotnet build -c Release` is clean (0 warnings).

### DIFF
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -9,6 +9,7 @@ using Azure.Functions.Cli.Workloads;
 using Azure.Functions.Cli.Workloads.Catalog;
 using Azure.Functions.Cli.Workloads.Discovery;
 using Azure.Functions.Cli.Workloads.Install;
+using Azure.Functions.Cli.Workloads.Install.Trust;
 using Azure.Functions.Cli.Workloads.Storage;
 using NuGet.Versioning;
 
@@ -59,6 +60,11 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         Description = "Disable alias matching. <id> must be the literal package id.",
     };
 
+    public Option<bool> AllowUntrustedOption { get; } = new("--allow-untrusted")
+    {
+        Description = "Bypass the publisher trust check. Required for installing unsigned local development packs.",
+    };
+
     public WorkloadInstallCommand(
         IInteractionService interaction,
         IWorkloadInstaller installer,
@@ -79,6 +85,7 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         Options.Add(SourceOption);
         Options.Add(IncludePrereleaseOption);
         Options.Add(ExactOption);
+        Options.Add(AllowUntrustedOption);
         Options.Add(ForceOption);
     }
 
@@ -89,6 +96,7 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         string? source = parseResult.GetValue(SourceOption);
         bool includePrerelease = parseResult.GetValue(IncludePrereleaseOption);
         bool exact = parseResult.GetValue(ExactOption);
+        bool allowUntrusted = parseResult.GetValue(AllowUntrustedOption);
         bool force = parseResult.GetValue(ForceOption);
 
         if (includePrerelease)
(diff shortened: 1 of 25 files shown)


src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -55,6 +56,11 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
         Description = "Disable alias matching. <id> must be the literal package id.",
     };
 
+    public Option<bool> AllowUntrustedOption { get; } = new("--allow-untrusted")
```
jviau, line 63: Want to ensure we align with any prior-art of this kind of check. Is `--allow-untrusted` used in any other example?

src/Func/Func.csproj
```diff
@@ -30,6 +30,7 @@
 
   <ItemGroup>
     <EmbeddedResource Include="Profiles/BuiltInProfiles.json" LogicalName="Azure.Functions.Cli.Profiles.BuiltInProfiles.json" />
+    <EmbeddedResource Include="Workloads/Install/Trust/trusted-roots.pem" LogicalName="Azure.Functions.Cli.Workloads.Install.Trust.trusted-roots.pem" />
```
jviau, line 33: I wonder if we can just take this from the current dotnet SDK that is building the csproj itself.

I think the path would be:

``` xml
<_TrustedRoots>$([System.IO.Path]::GetDirectoryName('$(MSBuildToolsPath)'))/trustedroots</_TrustedRoots>
```

This will have two files `codesignctl.pem` and `timestampctl.pem`

liliankasem:
Closing in favor of issue #5430 for backlog tracking. The work will be picked up from there.

PR #5108, opened 2026-05-27
func start: match v4 local.settings.json env-var precedence and diagnostics

liliankasem:
Brings v5's `local.settings.json` -> host env var behavior into parity with v4. Three changes to `PrepareProjectHostRunInitializationStep.BuildEnvironmentVariables`:

1. **Real env wins over `local.settings.json`.** v4 (`StartHostAction.UpdateEnvironmentVariables` on `main`) skips a key when `Environment.GetEnvironmentVariable(key)` is already set in the CLI's process. v5 was letting `local.settings.json` overwrite anything. This was a real bug: shell-set env vars are explicit per-shell overrides (CI, debug, secret rotation) and must win over the file. Now they do.
2. **Empty-value handling.** `""` is preserved verbatim in the env-var dict (callers use it to clear an inherited value). The v5 snapshot layer already drops JSON-null values, so v4's null-skip branch has no work to do here.
3. **Warnings.** Skipped keys (empty key, already in env) are surfaced through `IInteractionService.WriteWarning` (no `Console.*` calls per AGENTS.md).

### Implementation

- New `IProcessEnvironment` seam in `src/Func/Common/` with a `string? Get(string)` wrapper over `Environment.GetEnvironmentVariable`, so the precedence check is testable without mutating real process state. No existing seam fit.
- `PrepareProjectHostRunInitializationStep` now takes `IProcessEnvironment` + `IInteractionService`. Priority comment in `BuildEnvironmentVariables` updated to call out the new precedence.
- `DemoStartInitializationRunner` threads the new deps through.
- `CliHostFactory` registers `IProcessEnvironment` as a singleton.

### Tests

`test/Func.Tests/Commands/Start/Initialization/PrepareProjectHostRunInitializationStepTests.cs` covers: normal pass-through, empty key (skip + warn), empty-string value (preserved), process env present (skip + warn), case-insensitive precedence, empty/null process env doesn't block, and that the precedence rule only applies to local.settings (host + bundle vars still flow through).

Validated with `CI=true dotnet build -c Release` (0 warnings, 0 errors) and `dotnet test test/Func.Tests/Func.Tests.csproj` (603 passing) plus the Node stack tests for safety.

### Notes

- Stacked on #5097 (`liliankasem/feat/stack-host-start-hooks`), which introduces the `BuildEnvironmentVariables` method this PR modifies. Once #5097 merges into `vnext`, this PR will retarget to `vnext` directly.
- No worker-version updates needed.
- Release notes entry added in `src/Func/release_notes.md`.

### DIFF
--- a/src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
+++ b/src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
@@ -114,7 +114,7 @@ internal sealed class PrepareProjectHostRunInitializationStep(
                 continue;
             }
 
-            if (!string.IsNullOrEmpty(_processEnvironment.Get(name)))
+            if (_processEnvironment.Get(name) is not null)
             {
                 _interaction.WriteWarning($"Skipping '{name}' from local.settings.json: already set in the current environment.");
                 continue;
--- a/src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
+++ b/src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Configuration;
+using Azure.Functions.Cli.Console;
 using Azure.Functions.Cli.Projects;
 using Azure.Functions.Cli.Workers;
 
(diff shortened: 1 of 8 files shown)


src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
```diff
@@ -92,4 +101,28 @@ private Dictionary<string, string> BuildEnvironmentVariables(StartInitialization
...
         return environmentVariables;
     }
+
+    private void MergeLocalSettingsValues(
+        IReadOnlyDictionary<string, string> values,
+        Dictionary<string, string> environmentVariables)
+    {
+        foreach ((string name, string value) in values)
+        {
+            if (string.IsNullOrEmpty(name))
```
satvu, line 111: Is there a chance we have values that a customer may want to set to an empty string? If so, we can just check for `null`. Same for line 117 below.

satvu (approved):

src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
liliankasem: Good catch on line 117, switched to a null-only check in 64089d5 so an explicit `FOO=""` in the shell still wins over local.settings.json. Added a test for that case.

Left line 111 (the empty-key skip) as `IsNullOrEmpty` since an empty JSON key isn't a meaningful env var name, but happy to switch that too if you'd rather be consistent.

PR #5109, opened 2026-05-27
vnext: implement quickstart manifest service

manvkaur:

resolves #5026 (partial - manifest service layer only)

This is PR 2 in the incremental `func quickstart` series (after skeleton #5099). It implements the manifest service that fetches, caches, validates, and filters CDN-hosted template entries.

**What is included:**
- `QuickstartManifestService` - fetch from CDN with ETag/304 support, stale cache fallback on network failure
- `IManifestCache` - filesystem abstraction for testability (follows `IProfileFileSystem` pattern)
- `QuickstartManifest` - rich wrapper with `Filter(language, resource, iac, search)` and priority ordering
- `QuickstartUrlValidator` - HTTPS + trusted host/org allow-listing for template URLs
- `HttpClientDefaults` - CLI-wide User-Agent header (`AzureFunctionsCli/{version} ({os})`) applied to all HTTP clients
- `QuickstartManifestOptions` - configurable CDN URL, cache dir, TTL, HTTP timeout
- Source-gen JSON context for AOT/trim-friendly deserialization
- 46 new tests (service, manifest filtering, URL validation, HTTP defaults)
- Architecture doc updated

### Pull request checklist

### Additional information

- Only `v`-prefixed `gitRef` entries are accepted (semver release tags), signed tag verification in scaffolding service
- URL allow-list: HTTPS only, `github.com` host, trusted orgs (`Azure`, `Azure-Samples`, `microsoft`)
- Free-text search covers: Id, DisplayName, Resource, Iac, LongDescription

### DIFF
--- a/src/Abstractions/Quickstart/QuickstartManifest.cs
+++ b/src/Abstractions/Quickstart/QuickstartManifest.cs
@@ -1,0 +1,90 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Quickstart;
+
+/// <summary>
+/// The deserialized CDN manifest with convenience filter methods.
+/// </summary>
+public sealed class QuickstartManifest
+{
+    /// <summary>
+    /// All entries after validation and trusted-org filtering has been applied.
+    /// </summary>
+    public IReadOnlyList<QuickstartEntry> Entries { get; }
+
+    public QuickstartManifest(IReadOnlyList<QuickstartEntry> entries)
+    {
+        Entries = entries ?? throw new ArgumentNullException(nameof(entries));
+    }
+
+    /// <summary>
+    /// Returns entries matching all supplied non-null/non-empty criteria,
+    /// ordered by <see cref="QuickstartEntry.Priority"/>.
+    /// </summary>
+    public IReadOnlyList<QuickstartEntry> Filter(
+        string? language = null,
+        string? resource = null,
+        string? iac = null,
+        string? search = null)
+    {
+        IEnumerable<QuickstartEntry> results = Entries;
+
+        if (!string.IsNullOrWhiteSpace(language))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Language, language, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(resource))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Resource, resource, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(iac))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Iac, iac, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(search))
+        {
+            results = results.Where(e => MatchesSearch(e, search));
+        }
+
+        return [.. results.OrderBy(e => e.Priority)];
+    }
+
+    private static bool MatchesSearch(QuickstartEntry entry, string search)
+    {
+        if (entry.Id.Contains(search, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        if (entry.DisplayName.Contains(search, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        if (entry.Resource.Contains(search, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        if (entry.Iac is not null &&
+            entry.Iac.Contains(search, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        if (entry.LongDescription is not null &&
+            entry.LongDescription.Contains(search, StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
+        return false;
+    }
+}
--- a/src/Abstractions/Quickstart/QuickstartManifest.cs
+++ b/src/Abstractions/Quickstart/QuickstartManifest.cs
@@ -13,7 +13,7 @@ public sealed class QuickstartManifest
     /// </summary>
     public IReadOnlyList<QuickstartEntry> Entries { get; }
 
-    public QuickstartManifest(IReadOnlyList<QuickstartEntry> entries)
+    internal QuickstartManifest(IReadOnlyList<QuickstartEntry> entries)
     {
         Entries = entries ?? throw new ArgumentNullException(nameof(entries));
     }
--- a/src/Abstractions/Quickstart/QuickstartManifest.cs
+++ b/src/Abstractions/Quickstart/QuickstartManifest.cs
@@ -15,7 +15,8 @@ public sealed class QuickstartManifest
 
     internal QuickstartManifest(IReadOnlyList<QuickstartEntry> entries)
     {
-        Entries = entries ?? throw new ArgumentNullException(nameof(entries));
+        Entries = [.. (entries ?? throw new ArgumentNullException(nameof(entries)))
+            .OrderBy(e => e.Priority)];
     }
 
     /// <summary>
@@ -53,7 +54,7 @@ public sealed class QuickstartManifest
             results = results.Where(e => MatchesSearch(e, search));
         }
 
-        return [.. results.OrderBy(e => e.Priority)];
+        return [.. results];
     }
 
     private static bool MatchesSearch(QuickstartEntry entry, string search)
--- a/src/Func/Http/HttpClientDefaults.cs
+++ b/src/Func/Http/HttpClientDefaults.cs
@@ -18,11 +18,13 @@ internal static class HttpClientDefaults
 
     public static IServiceCollection AddCliHttpDefaults(this IServiceCollection services)
     {
+        ArgumentNullException.ThrowIfNull(services);
+
         services.ConfigureHttpClientDefaults(builder =>
         {
             builder.ConfigureHttpClient(client =>
             {
-                client.DefaultRequestHeaders.UserAgent.ParseAdd(_userAgent);
+                client.DefaultRequestHeaders.TryAddWithoutValidation("User-Agent", _userAgent);
             });
         });
 
--- a/src/Func/Http/HttpClientDefaults.cs
+++ b/src/Func/Http/HttpClientDefaults.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Azure.Functions.Cli.Http;
+
+/// <summary>
+/// Configures CLI-wide HTTP client defaults (User-Agent, etc.) applied to all
+/// named and unnamed <see cref="HttpClient"/> instances created via
+/// <see cref="IHttpClientFactory"/>.
+/// </summary>
+internal static class HttpClientDefaults
+{
+    private static readonly string _userAgent = BuildUserAgent();
+
+    public static IServiceCollection AddCliHttpDefaults(this IServiceCollection services)
+    {
+        services.ConfigureHttpClientDefaults(builder =>
+        {
+            builder.ConfigureHttpClient(client =>
+            {
+                client.DefaultRequestHeaders.UserAgent.ParseAdd(_userAgent);
+            });
+        });
+
+        return services;
+    }
+
+    private static string BuildUserAgent()
+    {
+        string version = typeof(HttpClientDefaults).Assembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+            ?.InformationalVersion?.Split('+')[0] ?? "unknown";
+
+        string os = RuntimeInformation.OSDescription;
+        return $"AzureFunctionsCli/{version} ({os})";
+    }
+}
--- a/src/Func/Quickstart/IManifestCache.cs
+++ b/src/Func/Quickstart/IManifestCache.cs
@@ -37,9 +37,4 @@ internal interface IManifestCache
     /// Returns <see langword="true"/> if a cached manifest file exists on disk.
     /// </summary>
     public bool ManifestExists();
-
-    /// <summary>
-    /// Reads a local file override, or <see langword="null"/> if the file doesn't exist.
-    /// </summary>
-    public Task<string?> TryReadLocalFileAsync(string path, CancellationToken cancellationToken);
 }
--- a/src/Func/Quickstart/IManifestCache.cs
+++ b/src/Func/Quickstart/IManifestCache.cs
@@ -1,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Quickstart;
+
+/// <summary>
+/// Abstracts filesystem operations for the quickstart manifest cache.
+/// </summary>
+internal interface IManifestCache
+{
+    /// <summary>
+    /// Ensures the cache directory exists.
+    /// </summary>
+    public void EnsureDirectory();
+
+    /// <summary>
+    /// Reads the cached manifest JSON, or <see langword="null"/> if it doesn't exist or is unreadable.
+    /// </summary>
+    public string? TryReadManifest();
+
+    /// <summary>
+    /// Writes the manifest JSON to the cache.
+    /// </summary>
+    public void WriteManifest(string json);
+
+    /// <summary>
+    /// Reads the cache metadata, or <see langword="null"/> if it doesn't exist or is unreadable.
+    /// </summary>
+    public ManifestCacheMeta? TryReadMeta();
+
+    /// <summary>
+    /// Writes cache metadata.
+    /// </summary>
+    public void WriteMeta(ManifestCacheMeta meta);
+
+    /// <summary>
+    /// Returns <see langword="true"/> if a cached manifest file exists on disk.
+    /// </summary>
+    public bool ManifestExists();
+
+    /// <summary>
+    /// Reads a local file override, or <see langword="null"/> if the file doesn't exist.
+    /// </summary>
+    public Task<string?> TryReadLocalFileAsync(string path, CancellationToken cancellationToken);
+}
--- a/src/Func/Quickstart/ManifestCache.cs
+++ b/src/Func/Quickstart/ManifestCache.cs
@@ -70,8 +70,17 @@ internal sealed class ManifestCache(IOptions<QuickstartManifestOptions> options)
 
     private static void AtomicWrite(string destinationPath, string content)
     {
-        string tempPath = destinationPath + ".tmp";
-        File.WriteAllText(tempPath, content);
-        File.Move(tempPath, destinationPath, overwrite: true);
+        string tempPath = $"{destinationPath}.{Path.GetRandomFileName()}.tmp";
+        try
+        {
+            File.WriteAllText(tempPath, content);
+            File.Move(tempPath, destinationPath, overwrite: true);
+        }
+        catch
+        {
+            // Best-effort cleanup of the unique temp file.
+            try { File.Delete(tempPath); } catch { }
+            throw;
+        }
     }
 }
--- a/src/Func/Quickstart/ManifestCache.cs
+++ b/src/Func/Quickstart/ManifestCache.cs
@@ -37,7 +37,7 @@ internal sealed class ManifestCache(IOptions<QuickstartManifestOptions> options)
     }
 
     /// <inheritdoc/>
-    public void WriteManifest(string json) => File.WriteAllText(ManifestFilePath, json);
+    public void WriteManifest(string json) => AtomicWrite(ManifestFilePath, json);
 
     /// <inheritdoc/>
     public ManifestCacheMeta? TryReadMeta()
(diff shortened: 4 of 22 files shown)


src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        throw new InvalidOperationException(
+            "Unable to fetch the quickstart manifest and no cached copy is available. " +
+            "Check your network connection and try again.");
+    }
+
+    private async Task<QuickstartManifest?> TryFetchFromCdnAsync(string? etag, CancellationToken cancellationToken)
+    {
+        HttpClient httpClient = _httpClientFactory.CreateClient(QuickstartRegistration.HttpClientName);
+        httpClient.Timeout = _options.HttpTimeout;
```
liliankasem: 🔴 **Bug risk.** `HttpClient.Timeout` can only be set before the first request, and `IHttpClientFactory` pools/reuses the underlying `HttpClient`. On the 2nd+ call (or under concurrency) this throws `InvalidOperationException: This instance has already started one or more requests. Properties can only be modified before sending the first request.`

Configure timeout once at registration:

```csharp
services.AddHttpClient(HttpClientName)
        .ConfigureHttpClient((sp, c) =>
        {
            var opts = sp.GetRequiredService<IOptions<QuickstartManifestOptions>>().Value;
            c.Timeout = opts.HttpTimeout;
        });
```

Or use a linked `CancellationTokenSource` with `CancelAfter(opts.HttpTimeout)` in the service.

liliankasem (commented):
Code review focused on .NET design patterns. Overall solid: clean DI with primary constructors and null guards, `IManifestCache` mirrors the existing `IProfileFileSystem` pattern, source-gen JSON keeps it AOT/trim-friendly, and the 304/stale-fallback flow is well thought out. Comments below ordered by severity.

src/Func/Quickstart/ManifestCache.cs
```diff
@@ -0,0 +1,81 @@
...
+            return File.Exists(ManifestFilePath) ? File.ReadAllText(ManifestFilePath) : null;
+        }
+        catch (IOException)
+        {
+            return null;
+        }
+    }
+
+    /// <inheritdoc/>
+    public void WriteManifest(string json) => File.WriteAllText(ManifestFilePath, json);
```
liliankasem: 🔴 **Bug risk: non-atomic cache writes.** `File.WriteAllText` is not atomic. A crash mid-write, or a second `func` process racing on the same cache dir, leaves a torn/corrupt file. Writing meta after the manifest also opens a window where a fresh manifest sits on disk with stale (or no) meta.

Suggest write-to-temp + `File.Move(tmp, dest, overwrite: true)` for both `WriteManifest` and `WriteMeta`, and consider writing meta before the manifest so a crash leaves the cache marked stale rather than fresh-but-corrupt.

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,235 @@
...
+        {
+            request.Headers.TryAddWithoutValidation("If-None-Match", etag);
+        }
+
+        using HttpResponseMessage response = await httpClient.SendAsync(request, cancellationToken);
+
+        if (response.StatusCode == HttpStatusCode.NotModified)
+        {
+            _logger.LogDebug("Manifest not modified (304); refreshing cache timestamp.");
+            _cache.WriteMeta(new ManifestCacheMeta(etag!, _timeProvider.GetUtcNow()));
```
liliankasem: 🔴 **Nit / latent NRE.** `etag!` is only safe today because the 304 path is reachable only when we sent `If-None-Match` with a non-empty etag. A CDN returning 304 unsolicited would NRE. Use `etag ?? string.Empty` to match the OK path.

src/Func/Quickstart/QuickstartRegistration.cs
```diff
@@ -0,0 +1,46 @@
...
+                if (!string.IsNullOrWhiteSpace(overrideUrl))
+                {
+                    opts.ManifestUrl = overrideUrl;
+                }
+            });
+
+        services.AddHttpClient(HttpClientName);
+
+        services.AddSingleton<IManifestCache, ManifestCache>();
+        services.AddSingleton(TimeProvider.System);
```
liliankasem: 🟡 Prefer `TryAddSingleton(TimeProvider.System)` (from `Microsoft.Extensions.DependencyInjection.Extensions`). If anything else in the host registers `TimeProvider` (now or later), `AddSingleton` silently overrides it, which is exactly the surprise DI guidelines try to avoid for shared framework services.

src/Func/Quickstart/IManifestCache.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// <summary>
+    /// Returns <see langword="true"/> if a cached manifest file exists on disk.
+    /// </summary>
+    public bool ManifestExists();
+
+    /// <summary>
+    /// Reads a local file override, or <see langword="null"/> if the file doesn't exist.
+    /// </summary>
+    public Task<string?> TryReadLocalFileAsync(string path, CancellationToken cancellationToken);
```
liliankasem: 🟡 **SRP nit.** Reading a user-supplied override path has nothing to do with "the manifest cache." It's only here because both happen to hit the filesystem. Tests for caching end up mocking a method they don't care about, and vice versa.

Consider splitting into a tiny `IFileReader` (or just inlining `File.ReadAllTextAsync` in the service behind a small protected virtual). Keeps `IManifestCache` cohesive.

src/Abstractions/Quickstart/QuickstartManifest.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        if (!string.IsNullOrWhiteSpace(iac))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Iac, iac, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(search))
+        {
+            results = results.Where(e => MatchesSearch(e, search));
```
liliankasem, line 54: 🟡 `QuickstartEntry.Tags` exists for exactly this kind of search but isn't included in `MatchesSearch`. Either fold tags into the search predicate or drop the `Tags` property to avoid a dead-weight field.

src/Func/Quickstart/QuickstartUrlValidator.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        if (!string.IsNullOrEmpty(uri.UserInfo))
+        {
+            return false;
+        }
+
+        if (!string.Equals(uri.Host, RequiredHost, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
```
liliankasem, line 53: 🟡 Two cheap belt-and-suspenders checks worth adding given this is a security boundary:

1. `if (!uri.IsDefaultPort) return false;` so `https://github.com:8080/Azure/x` doesn't pass.
2. Require a non-empty repo segment, otherwise `https://github.com/Azure` (org root, no repo) passes.

src/Func/Http/HttpClientDefaults.cs
```diff
@@ -0,0 +1,41 @@
...
+/// <summary>
+/// Configures CLI-wide HTTP client defaults (User-Agent, etc.) applied to all
+/// named and unnamed <see cref="HttpClient"/> instances created via
+/// <see cref="IHttpClientFactory"/>.
+/// </summary>
+internal static class HttpClientDefaults
+{
+    private static readonly string _userAgent = BuildUserAgent();
+
+    public static IServiceCollection AddCliHttpDefaults(this IServiceCollection services)
```
liliankasem, line 19: 🟢 Missing `ArgumentNullException.ThrowIfNull(services);` for parity with `AddQuickstartManifest` and the rest of the CLI's DI extensions.

src/Func/Http/HttpClientDefaults.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    private static string BuildUserAgent()
+    {
+        string version = typeof(HttpClientDefaults).Assembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+            ?.InformationalVersion?.Split('+')[0] ?? "unknown";
+
+        string os = RuntimeInformation.OSDescription;
+        return $"AzureFunctionsCli/{version} ({os})";
```
liliankasem, line 41: 🟢 `UserAgent.ParseAdd` throws `FormatException` on malformed input, and `RuntimeInformation.OSDescription` can contain odd characters on exotic platforms. Since this runs at startup and is applied to every `HttpClient` in the CLI, a bad OS string would break all outgoing HTTP. Worth a try/catch with an `AzureFunctionsCli/{version}` fallback.

src/Abstractions/Quickstart/QuickstartManifest.cs
```diff
@@ -0,0 +1,90 @@
...
+/// The deserialized CDN manifest with convenience filter methods.
+/// </summary>
+public sealed class QuickstartManifest
+{
+    /// <summary>
+    /// All entries after validation and trusted-org filtering has been applied.
+    /// </summary>
+    public IReadOnlyList<QuickstartEntry> Entries { get; }
+
+    public QuickstartManifest(IReadOnlyList<QuickstartEntry> entries)
```
liliankasem: 🟢 Per the repo's "internal by default" guideline: `QuickstartManifest` and its public constructor are only used by `QuickstartManifestService` and the test project today. If the Abstractions assembly intentionally surfaces these for the in-proc host, ignore. Otherwise consider `internal` + `[InternalsVisibleTo]` for the test assembly.

src/Func/Quickstart/ManifestCache.cs
manvkaur: Good point on atomic writes - done via temp + File.Move(overwrite: true) now, so no torn/corrupt files.

On write order though, I think manifest-first, then meta is actually safer:

 - With atomic writes, there's no "half-written" state - the file is either fully old or fully new.
 - If we crash after writing meta but before manifest: meta has a fresh timestamp -> IsCacheFresh returns true -> serves old content for up to 24h without re-fetching.
 - If we crash after writing manifest but before meta: meta is stale/missing -> IsCacheFresh returns false -> triggers a CDN fetch on next call -> self-heals immediately.

So manifest-first gives us a "fail-open to re-fetch" behavior, while meta-first gives "fail-closed on stale data." I'll keep manifest -> metaorder if that sounds good?

src/Abstractions/Quickstart/QuickstartManifest.cs
manvkaur: Good call - made the constructor internal and added InternalsVisibleTo for func and Azure.Functions.Cli.Tests in Abstractions.csproj. The class itself stays public since it's the return type of IQuickstartManifestService.GetManifestAsync() (public interface), but consumers can no longer construct arbitrary instances

src/Func/Quickstart/ManifestCache.cs
```diff
@@ -0,0 +1,77 @@
...
+        string json = JsonSerializer.Serialize(meta, QuickstartJsonContext.Default.ManifestCacheMeta);
+        AtomicWrite(MetaFilePath, json);
+    }
+
+    /// <inheritdoc/>
+    public bool ManifestExists() => File.Exists(ManifestFilePath);
+
+    private static void AtomicWrite(string destinationPath, string content)
+    {
+        string tempPath = destinationPath + ".tmp";
```
liliankasem: Two func processes (CI runner, dev shell, two parallel commands) using the same cache dir will both write to manifest.json.tmp concurrently. One overwrites the other mid-stream, both call File.Move, and the destination ends up with interleaved/torn content, defeating the atomic-write goal. The temp file also leaks if File.WriteAllText succeeds but File.Move throws.

use a unique temp name and clean up on failure

```csharp
string tempPath = $"{destinationPath}.{Path.GetRandomFileName()}.tmp";
try
{
    File.WriteAllText(tempPath, content);
    File.Move(tempPath, destinationPath, overwrite: true);
}
catch
{
    try { File.Delete(tempPath); } catch { /* best effort */ }
    throw;
}
```

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Azure.Functions.Cli.Quickstart;
+
+/// <summary>
+/// Fetches the quickstart manifest from the CDN with ETag-based caching
+/// and stale-cache fallback on network failures.
+/// </summary>
+internal sealed class QuickstartManifestService(
```
liliankasem, line 15: I thought this was a shared service for the stack workloads or no longer? Doesn't make sense to have multiple implementations of this (I noticed the interface is in abstractions)

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+        if (Path.IsPathFullyQualified(url) && !url.StartsWith(HttpSchemePrefix, StringComparison.OrdinalIgnoreCase))
+        {
+            localPath = url;
+            return true;
+        }
+
+        return false;
+    }
+
+    private async Task<QuickstartManifest> LoadFromLocalFileAsync(string path, CancellationToken cancellationToken)
```
liliankasem, line 134: the CDN path catches JsonException in the outer try (line 57), but LoadFromLocalFileAsync has no such try. If a user points FUNC_QUICKSTART_MANIFEST_URL at a file that exists but contains invalid JSON, DeserializeEntries throws and the user sees a stack trace instead of the carefully crafted "Manifest override file '{path}' is empty or malformed." message. The current code only hits that message when the JSON is syntactically valid but missing the templates array.

To address this, you can wrap DeserializeEntries in a try/catch or move the friendly-error rethrow into a catch (JsonException)

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+    private const string HttpSchemePrefix = "http";
+
+    private readonly IHttpClientFactory _httpClientFactory = httpClientFactory ?? throw new ArgumentNullException(nameof(httpClientFactory));
+    private readonly IManifestCache _cache = cache ?? throw new ArgumentNullException(nameof(cache));
+    private readonly QuickstartManifestOptions _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
+    private readonly TimeProvider _timeProvider = timeProvider ?? throw new ArgumentNullException(nameof(timeProvider));
+    private readonly ILogger<QuickstartManifestService> _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+    /// <inheritdoc/>
+    public async Task<QuickstartManifest> GetManifestAsync(CancellationToken cancellationToken = default)
```
liliankasem, line 32: Cache the Task of the in-flight fetch. Concurrent callers await the same Task<QuickstartManifest>, you only hit the CDN once, and once it completes you clear the field so the next call re-evaluates cache freshness.

```C#
private Task<QuickstartManifest>? _inFlight;
private readonly object _gate = new();

public Task<QuickstartManifest> GetManifestAsync(CancellationToken ct = default)
{
    lock (_gate)
    {
        // lock is fine here: we don't await inside it, we just hand back a Task
        _inFlight ??= FetchAndClearAsync(ct);
        return _inFlight;
    }
}

private async Task<QuickstartManifest> FetchAndClearAsync(CancellationToken ct)
{
    try
    {
        return await FetchAsync(ct); // your current GetManifestAsync body
    }
    finally
    {
        lock (_gate) { _inFlight = null; }
    }
}
```

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+
+    private bool HasRequiredFields(QuickstartEntry entry)
+    {
+        if (string.IsNullOrWhiteSpace(entry.Id) ||
+            string.IsNullOrWhiteSpace(entry.Language) ||
+            string.IsNullOrWhiteSpace(entry.Resource) ||
+            string.IsNullOrWhiteSpace(entry.RepositoryUrl) ||
+            string.IsNullOrWhiteSpace(entry.FolderPath))
+        {
+            _logger.LogDebug("Dropping quickstart entry with missing required field: '{Id}'.", entry.Id);
```
liliankasem: nit: `"Dropping quickstart entry with missing required field: '{Id}'."` logs empty/whitespace when `Id` itself is blank, making the message useless for diagnosis. Either log which field was missing, or include a fallback like `entry.DisplayName` / `entry.RepositoryUrl`.

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+                continue;
+            }
+
+            filtered.Add(entry);
+        }
+
+        return new QuickstartManifest(filtered);
+    }
+
+    private bool HasRequiredFields(QuickstartEntry entry)
```
liliankasem: nit: `HasRequiredFields` is an instance method despite being pure validation. It uses `_logger`. Either keep it instance and move the dropping-reason variant to a static helper that returns the missing-field name (see the comment on the log line below), or accept the logger as a parameter so the method becomes static (matches `HasValidGitRef`'s shape). Stylistic.

src/Func/Quickstart/QuickstartManifestService.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        // Support file:// URIs
+        if (Uri.TryCreate(url, UriKind.Absolute, out Uri? uri) && uri.IsFile)
+        {
+            localPath = uri.LocalPath;
+            return true;
+        }
+
+        // Support raw absolute file paths (Windows or Unix)
+        if (Path.IsPathFullyQualified(url) && !url.StartsWith(HttpSchemePrefix, StringComparison.OrdinalIgnoreCase))
```
liliankasem: nit: `TryGetLocalFilePath` heuristic is fragile. `Path.IsPathFullyQualified("https://cdn.example.com/foo")` is `false` on both Windows and Unix, so the `HttpSchemePrefix` guard is belt-and-suspenders. But `Path.IsPathFullyQualified("//server/share")` returns `true` on Windows (UNC path), which would then be returned as a "local file path" and passed to `File.ReadAllTextAsync`, which is valid but probably not what the user expected from `FUNC_QUICKSTART_MANIFEST_URL`. Low likelihood; mention in the env-var docs or accept only `file://` + drive-letter / `/`-rooted paths explicitly.

src/Abstractions/Quickstart/QuickstartManifest.cs
```diff
@@ -0,0 +1,90 @@
...
+            results = results.Where(e =>
+                string.Equals(e.Iac, iac, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(search))
+        {
+            results = results.Where(e => MatchesSearch(e, search));
+        }
+
+        return [.. results.OrderBy(e => e.Priority)];
```
liliankasem: nit: `Filter` materializes a new list on every call. `return [.. results.OrderBy(...)];` is fine for current scale, but if `Filter` is called per-keystroke from an interactive picker, caching the sorted `Entries` once and re-filtering with a single `Where` chain is faster. Premature today.

src/Abstractions/Quickstart/QuickstartManifest.cs
```diff
@@ -0,0 +1,90 @@
...
+        if (!string.IsNullOrWhiteSpace(resource))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Resource, resource, StringComparison.OrdinalIgnoreCase));
+        }
+
+        if (!string.IsNullOrWhiteSpace(iac))
+        {
+            results = results.Where(e =>
+                string.Equals(e.Iac, iac, StringComparison.OrdinalIgnoreCase));
```
liliankasem, line 49: nit: `QuickstartManifest.Filter`'s `iac` filter mismatches `null`. `entry.Iac == null` rows are dropped when `iac` is supplied, even if the user passed `"none"`. If the manifest represents "no IaC" as the JSON value `"none"` (per the `QuickstartEntry.Iac` doc comment), this is fine. If it's represented as a missing field / `null`, the user can never filter for "templates without IaC". Worth confirming with the manifest schema.

src/Func/Quickstart/ManifestCache.cs
liliankasem: Yup totally fair - sounds good to me

src/Func/Quickstart/QuickstartManifestService.cs
manvkaur: Yes, it's a shared service. A single CDN manifest covers all languages and workloads, so each workload consumes the same manifest via Filter(language: ...) rather than needing its own implementation. The interface is in Abstractions so the command handler depends on the contract rather than the concrete service - same pattern as IProfileFileSystem. The single implementation is intentional; the interface is for DI and testability (NSubstitute substitution in tests), not for polymorphism.

src/Func/Quickstart/QuickstartManifestService.cs
manvkaur: Good defensive pattern for shared/server-side services. For the CLI though, GetManifestAsync is called once per process invocation from the command handler - there are no concurrent in-process callers today.

There's also a subtle correctness concern with the pattern: the first caller's CancellationToken gets captured in FetchAndClearAsync. If a second caller awaits the same Task with a different token and the first caller cancels (Ctrl+C), both callers receive OperationCanceledException - even though the second caller didn't cancel. Fixing that properly (linked tokens, per-caller cancellation) adds meaningful complexity.

 Given the single-call-per-process execution model and the token-sharing subtlety, keeping it simple for now and adding coalescing when concurrent call sites emerge seems like the safer trade-off. What do you think?

src/Func/Quickstart/QuickstartManifestService.cs
manvkaur: Refactored to a static GetMissingFields helper that returns the identifier and list of missing field names — logging stays in BuildManifest. This addresses both this comment and the log message nit.

 Worth noting the templates repo CI pipeline does schema validation and ensures required properties are present before merging the manifest changes, so this is a defensive fallback rather than the primary validation gate.

src/Abstractions/Quickstart/QuickstartManifest.cs
manvkaur: Yeah, none means that template doesn't have infra support for deployment, it is only function app code.
Manifest schema enforces allowed values - "arm", "bicep", "terraform" or "none" and CI checks for schema adherence

src/Func/Quickstart/QuickstartManifestService.cs
liliankasem: I think fair to keep simple for now

liliankasem (approved):

PR #5111, opened 2026-05-27
vnext: add quickstart scaffolder with git/HTTP template fetching

manvkaur:

Part 3 of incremental `func quickstart` implementation ([PR 1: skeleton](https://github.com/Azure/azure-functions-core-tools/pull/5026), [PR 2: manifest service](https://github.com/Azure/azure-functions-core-tools/pull/5109)).

Adds the **QuickstartScaffolder** service — responsible for downloading templates (via git clone or HTTP zip), verifying tag integrity, cleaning metadata, promoting subfolders, and copying files to the target directory.

**Key design decisions:**
- **Strategy pattern** — `ITemplateFetcher` with `GitTemplateFetcher` and `HttpTemplateFetcher`, resolved via `IEnumerable<ITemplateFetcher>` multi-registration
- **Tag-only enforcement** — `ls-remote --exit-code` rejects branch refs; `cat-file -t` output verified as `"tag"` (rejects lightweight tags)
- **Dependency Inversion** — scaffolder depends only on abstractions (`ITemplateFetcher`, `IFetchModeResolver`); zero knowledge of git/HTTP internals
- **Security hardening** — flag injection prevention, path traversal checks, credential-prompt suppression via per-process env vars, 60s timeout, zip-slip protection
- **Shared `DirectoryGuard`** — filesystem helpers extracted from `InitCommand` and reused by scaffolder

### Pull request checklist

### Additional information

**New files (12):**
- `QuickstartScaffolder.cs` — pure orchestrator (~148 lines)
- `ITemplateFetcher.cs` / `GitTemplateFetcher.cs` / `HttpTemplateFetcher.cs` — strategy pattern
- `IGitRunner.cs` / `GitRunner.cs` / `GitRunnerException.cs` — git process abstraction
- `IFetchModeResolver.cs` / `FetchModeResolver.cs` — auto-mode resolution
- `QuickstartScaffolderRegistration.cs` — DI extension
- `FakeGitRunner.cs` / `QuickstartScaffolderTests.cs` — test infrastructure

**Shared extractions (3):**
- `DirectoryGuard.cs` + `DirectoryGuardTests.cs` — extracted from InitCommand
- `QuickstartConstants.cs` — `GitHubHost` shared with manifest service (PR 2)

**Tests:** 24 new (16 scaffolder + 8 DirectoryGuard), 762 total pass, 0 warnings.

**Depends on:** PR #5109 (manifest service) for shared HttpClient defaults (tracked via inline TODO).


src/Func/Quickstart/HttpTemplateFetcher.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    private static string BuildArchiveUrl(QuickstartEntry entry)
+    {
+        var uri = new Uri(entry.RepositoryUrl);
+        string path = uri.AbsolutePath.TrimEnd('/');
+
+        string tag = entry.GitRef ?? throw new InvalidOperationException(
+            $"Template '{entry.Id}' has no GitRef. Cannot build archive URL.");
+
+        return $"{QuickstartConstants.GitHubHost}{path}{ArchiveUrlPathSegment}{tag}{ArchiveUrlExtension}";
```
liliankasem: `uri.Host` from `entry.RepositoryUrl` is dropped here — the archive URL is always rebuilt against `QuickstartConstants.GitHubHost`. If a non-GitHub `RepositoryUrl` ever slips through manifest validation, the HTTP path silently goes to github.com while `GitTemplateFetcher` would still use the real host. Either assert `uri.Host == "github.com"` and throw a clear error, or build the URL from `uri.Host` directly so both fetchers agree.

src/Func/Quickstart/QuickstartScaffolder.cs
```diff
@@ -0,0 +1,148 @@
...
+        if (Directory.Exists(gitDir))
+        {
+            DirectoryGuard.ClearReadOnlyRecursive(new DirectoryInfo(gitDir));
+            Directory.Delete(gitDir, recursive: true);
+        }
+
+        string githubDir = Path.Combine(directory, ".github");
+        if (Directory.Exists(githubDir))
+        {
+            Directory.Delete(githubDir, recursive: true);
```
liliankasem, line 81: Deleting `.github` unconditionally is risky. Quickstart-style template repos commonly ship intentional content under `.github/workflows/` (e.g. `azure-dev.yml` for `azd up`); wiping it silently breaks those scenarios. Consider only removing `.git`, or making `.github` removal opt-in via the manifest entry, or scoping the deletion to known non-template files (CODEOWNERS, ISSUE_TEMPLATE, ...).

src/Func/Quickstart/HttpTemplateFetcher.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    private static string BuildArchiveUrl(QuickstartEntry entry)
+    {
+        var uri = new Uri(entry.RepositoryUrl);
+        string path = uri.AbsolutePath.TrimEnd('/');
+
+        string tag = entry.GitRef ?? throw new InvalidOperationException(
+            $"Template '{entry.Id}' has no GitRef. Cannot build archive URL.");
```
liliankasem: `HttpTemplateFetcher` requires a non-null `GitRef`, but `GitTemplateFetcher` accepts null. Under `FetchMode.Auto` the same entry would succeed when git is available and hard-fail when it isn't. Either enforce non-null `GitRef` up front in `QuickstartScaffolder.ValidateEntry` so both modes agree, or document the asymmetry and surface a clearer error.

src/Func/Quickstart/GitRunner.cs
```diff
@@ -0,0 +1,136 @@
...
+            ProcessStartInfo psi = CreateStartInfo(workingDirectory: null);
+            psi.ArgumentList.Add("--version");
+
+            using Process process = Process.Start(psi)
+                ?? throw new InvalidOperationException("Failed to start git process.");
+
+            try
+            {
+                string output = await process.StandardOutput.ReadToEndAsync(cancellationToken);
+                await process.WaitForExitAsync(cancellationToken);
```
liliankasem: `TryGetVersionAsync` has no timeout. A hung `git --version` (slow shim, network-mounted PATH, hostile global config) will block `FetchModeResolver` for the lifetime of the caller's token. Add a short timeout (a few seconds) like `RunCoreAsync` does — falling back to HTTP is the right behavior when the probe is slow.

src/Func/Quickstart/QuickstartScaffolder.cs
```diff
@@ -0,0 +1,148 @@
...
+                $"Invalid GitRef '{entry.GitRef}': must not start with '-' (potential flag injection).",
+                nameof(entry));
+        }
+
+        if (entry.FolderPath.Contains("..", StringComparison.Ordinal))
+        {
+            throw new ArgumentException(
+                $"Invalid FolderPath '{entry.FolderPath}': must not contain '..' (path traversal).",
+                nameof(entry));
+        }
```
liliankasem, line 159: `Contains("..")` blocks traversal, but rooted paths slip through. `Path.Combine(tempDir, "C:\\evil")` returns `C:\evil` on Windows, and `/etc/passwd` would behave similarly on Unix. Add a `Path.IsPathRooted(entry.FolderPath)` check (and reject) alongside the `..` guard.

src/Func/Quickstart/QuickstartScaffolderRegistration.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// available. Wire the named client through that path instead of bare
+    /// IHttpClientFactory.
+    /// </remarks>
+    public static IServiceCollection AddQuickstartScaffolder(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        services.AddHttpClient(HttpTemplateFetcher.HttpClientName, client =>
+        {
+            // TODO: Use shared HttpClient registration with user-agent and other defaults once available. Post merge of PR 5109.
```
liliankasem: PR #5109 is already in this PR's base history, so `AddCliHttpDefaults` should be available now — worth wiring it up here instead of leaving the TODO, since this fetcher currently issues archive downloads with no user-agent.

src/Func/Quickstart/IGitRunner.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Quickstart;
+
+/// <summary>
+/// Runs <c>git</c> CLI commands in a child process with security-hardened
+/// environment variables that prevent credential prompts and network leaks.
```
liliankasem: Minor wording: the env vars set in `GitRunner` suppress credential prompts and interactive SSH, they don't really prevent "network leaks". Consider tightening the doc to match what they actually do.

src/Func/Quickstart/QuickstartScaffolder.cs
manvkaur: Good call flagging this -  I checked the actual template repos to make sure.

.github/workflows/ is where GitHub Actions CI/CD pipelines live (e.g. build-on-push, deploy-on-merge). The azure-dev.yml you might be thinking of is actually azure.yaml - that's the azd project definition file and it lives at the repo root (fixed structure required by azd), so it's preserved by thescaffolder since we only clean .github/, not root-level files.

  Looking at what's actually in .github/ across our template repos today (e.g. functions-quickstart-dotnet-azd):

   -  ISSUE_TEMPLATE.md, PULL_REQUEST_TEMPLATE.md - repo management for the Azure-Samples org
   - .github/prompts/ - propagation tooling for managing the template repo

None of this is intended for the user's scaffolded project - it's all maintenance content for the source repo. Leaving it would just add noise and confusion. What do you think?

I would like to delete some more meaningless files for customers which are only for repo maintenance and management but restricting to this folder for now.

src/Func/Quickstart/GitTemplateFetcher.cs
```diff
@@ -0,0 +1,103 @@
...
+    private readonly IGitRunner _gitRunner = gitRunner ?? throw new ArgumentNullException(nameof(gitRunner));
+    private readonly ILogger<GitTemplateFetcher> _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+    /// <inheritdoc />
+    public FetchMode Mode => FetchMode.Git;
+
+    /// <inheritdoc />
+    public async Task FetchAsync(QuickstartEntry entry, string tempDirectory, CancellationToken cancellationToken)
+    {
+        if (entry.GitRef is not null)
```
fabiocav: This flow won't enforce cloning from a tag.

The verification is validating that a tag exists, but a branch with the same name could end up being cloned. 

The safe pattern would be:

```
git init <tempDirectory>
 git -C <tempDirectory> remote add origin <repositoryUrl>
 git -C <tempDirectory> fetch --depth 1 --no-tags origin refs/tags/<gitRef>:refs/tags/<gitRef>
 git -C <tempDirectory> checkout --detach refs/tags/<gitRef>
 git -C <tempDirectory> rev-parse HEAD
 git -C <tempDirectory> rev-list -n 1 refs/tags/<gitRef>
```
Then compare the results of the ref and fail if they don't match.

The key thing is avoiding `git clone --branch <tag>`, so a branch with the same name wouldn't end up being cloned.

src/Func/Quickstart/QuickstartScaffolder.cs
```diff
@@ -0,0 +1,167 @@
...
+        string targetDirectory,
+        FetchMode fetchMode,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(entry);
+        ArgumentException.ThrowIfNullOrWhiteSpace(targetDirectory);
+
+        ValidateEntry(entry);
+
+        string tempDir = Path.Combine(Path.GetTempPath(), $"{TempDirectoryPrefix}{Guid.NewGuid():N}");
```
fabiocav: Don't use `Path.GetTempPath()` for this.

Instead, use `Directory.CreateTempSubdirectory()`. which will create a random, unpredictable directory under the system temp path, with permissions restricted to the user/owner.

PR #5120, opened 2026-05-27
Fix offline bundle E2E test flakiness

harshivcodes:
Fix offline bundle E2E test flakiness

Bug: Start_NodeV4_OfflineWithCachedBundles_UsesCachedVersion fails when it runs before any other func start test has populated the bundle cache. The test fixture only runs func init + func new — neither downloads extension bundles. Since xUnit doesn't guarantee test order, the test can run on a clean cache, causing GetCachedBundleOrThrow to throw instead of emitting the expected stdout message.

Fix: Added an EnsureBundlesCached helper that checks if bundles already exist and, if not, runs func bundles download before the offline test executes. The check is a no-op when bundles are already cached.


liliankasem:
I think this only addresses part of the problem; this fixes the clean-cache-on-first-run case, but the CI failure looks like a concurrency race between the two helpers in this file:

1. RunOfflineWithCachedBundlesTest calls EnsureBundlesCached → sees bundles, returns.
2. RunOfflineWithoutCachedBundlesTest (different language fixture, running in parallel) does Directory.Move(defaultBundlePath, backupPath) on the same global path.
3. Step 1's func start --offline subprocess then reads ~/.azure-functions-core-tools/Functions/ExtensionBundles/..., finds nothing, and throws at ExtensionBundleHelper.cs:102.
EnsureBundlesCached is a check-then-use against process-global state, so it can't close this race on its own.

Could we instead isolate RunOfflineWithoutCachedBundlesTest from the real user cache? E.g. point the bundle download path at a per-test temp directory (via ExtensionBundleOptions.DownloadPath or the equivalent env var the CLI honors) so the "without cached" scenario never touches ~/.azure-functions-core-tools. That removes the race entirely and is also crash-safe: today, if the test process dies between the Move and the finally restore, the dev/CI machine is left with no bundle cache.

PR #5122, opened 2026-05-27
Add Azurite HTTP probe service

liliankasem:
Slice 2 of the managed-Azurite feature. Implements the HTTP probe service from §7 / §16.2 of the design doc. No wiring into `func start`, no process / Docker management; those are later slices.

## What's in the slice

Under `src/Func/Commands/Start/Azurite/`:

- `AzuriteProbeStatus` enum: `Ready`, `NotListening`, `PortConflict`, `Partial` (matches the §7.6 table).
- `AzuriteService` + `AzuriteEndpointStatus` + `AzuriteEndpointProbeOutcome` record: per-endpoint result with HTTP status, request id, error code, and optional detail.
- `AzuriteProbeResult` record with `From(...)` factory that classifies the aggregate per §7.6.
- `IAzuriteProbe` / `AzuriteProbe`: `HttpClient`-only probe (no Azure Storage SDK). Uses `IHttpClientFactory` with a named `"AzuriteProbe"` client. Builds the three probes per §7.2–§7.4 (`?comp=list` for blob/queue, `/Tables` for table) with `x-ms-version: 2021-12-02`. Storage-shaped detection per §7.5 covers `x-ms-request-id` / `x-ms-error-code` headers, `Server: Azurite*`, and storage-style XML or JSON error bodies. Connection-refused / unreachable sockets map to `NotListening`; cancellation propagates. The three probes run in parallel via `Task.WhenAll`.
- `AzuriteServiceCollectionExtensions.AddAzuriteProbe`: DI helper. Defined but **not wired** into the CLI composition root yet.

## Tests

`test/Func.Tests/Commands/Start/Azurite/AzuriteProbeTests.cs` uses an in-process `HttpListener` (dependency-free, cross-platform) to cover the §16.2 cases:

- `x-ms-request-id` response → storage-shaped → `Ready`.
- `x-ms-error-code` response → storage-shaped.
- `Server: Azurite-Blob/3.35.0` header → storage-shaped.
- Plain HTTP 200 with no storage markers on all three → `PortConflict`.
- All three endpoints on unused ports → `NotListening`.
- Two ready + one refused → `Partial`.
- Already-cancelled token → `OperationCanceledException`.

## Verification

- `CI=true dotnet build -c Release`: clean (0 warnings, 0 errors).
- `dotnet test test/Func.Tests/Func.Tests.csproj -c Release`: 697 passed, 0 failed.

## Out of scope

- Wiring into `StartCommand` or any startup step.
- Process / Docker discovery and startup.
- Scope id, lease, metadata files.
- `--no-azurite` flag.
- Release notes.

### DIFF
--- a/src/Func/Commands/Start/Azurite/AzuriteEndpointProbeOutcome.cs
+++ b/src/Func/Commands/Start/Azurite/AzuriteEndpointProbeOutcome.cs
@@ -1,0 +1,60 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Commands.Start.Azurite;
+
+/// <summary>
+/// Identifies the Azurite service an endpoint was probed for.
+/// </summary>
+internal enum AzuriteService
+{
+    Blob,
+    Queue,
+    Table,
+}
+
+/// <summary>
+/// Per-endpoint probe classification. Mirrors <see cref="AzuriteProbeStatus"/>
+/// minus <see cref="AzuriteProbeStatus.Partial"/>, which only makes sense as
+/// an aggregate.
+/// </summary>
+internal enum AzuriteEndpointStatus
+{
+    /// <summary>
+    /// The endpoint responded with markers that identify it as Azure Storage
+    /// (Azurite or otherwise): an <c>x-ms-request-id</c> or
+    /// <c>x-ms-error-code</c> header, a <c>Server</c> header that contains
+    /// "Azurite", or a storage-style error payload.
+    /// </summary>
+    Ready,
+
+    /// <summary>
+    /// The TCP connection was refused or otherwise unreachable.
+    /// </summary>
+    NotListening,
+
+    /// <summary>
+    /// The endpoint responded, but the response did not look like Azure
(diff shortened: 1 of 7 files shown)


src/Func/Commands/Start/Azurite/AzuriteEndpointProbeOutcome.cs
```diff
@@ -0,0 +1,58 @@
...
+/// </summary>
+internal enum AzuriteEndpointStatus
+{
+    /// <summary>
+    /// The endpoint responded with markers that identify it as Azure Storage
+    /// (Azurite or otherwise): an <c>x-ms-request-id</c> or
+    /// <c>x-ms-error-code</c> header, a <c>Server</c> header that contains
+    /// "Azurite", or a storage-style error payload.
+    /// </summary>
+    StorageShaped,
```
satvu: Not really into the "Shaped" part of the name but could be convinced otherwise.

satvu (commented):
lgtm

src/Func/Commands/Start/Azurite/AzuriteEndpointProbeOutcome.cs
liliankasem: I don't like it either :D

PR #5125, opened 2026-05-28
Fix bug-bash findings across CLI, setup, and workloads

liliankasem:
Bundle of small fixes from an e2e bug bash of the v5 CLI.

## Fixes

1. **`-?` removed from help options.** It was advertised but unsupported by `zsh`. Dropped from `HelpCommand` and `SpectreHelpAction` text.
2. **Table column centering removed.** `IInteractionService.WriteTable` now left-aligns headers and cells.
3. **Extension bundle install message simplified.** `SetupRunner.Bundle` no longer embeds the package id in the DisplayName, so the message reads `Installed extension bundle 4.35.0` instead of `Installed extension bundle Microsoft.Azure.Functions.ExtensionBundle 4.35.0`.
4. **`func setup` uses a multi-select prompt.** Added `PromptForMultiSelectionAsync` to `IInteractionService` / `SpectreInteractionService` and wired it into `SetupRunner.GetDefaultFeaturesAsync` so users can pick multiple stacks with space.
6. **`dotnet new func` uses `--Framework`.** The Worker template's option is capital `F`; lowercase failed with exit code 127 (#6).
7. **Skip language in `.func/config.json` when redundant.** `InitCommand` omits `stack.language` when the stack only supports one language (e.g. Python, Go), avoiding the noisy `"runtime": "python", "language": "python"` shape.
8. **Workload Title/Description normalized.** Updated all 9 workload csprojs for consistent display names (.NET, Go, Node.js, Python, .NET Worker, Functions Host, Extension Bundles, etc.).
9. **Worker install confirm prompt on its own line.** `ResolveFunctionsWorkerInitializationStep` prepends a newline so the prompt doesn't get clobbered by the Spectre progress live region.

## Deferred

- #5 (first-run setup with onboarding/telemetry opt-in) is intentionally left for a follow-up; it's a bigger UX change than the others.

## Verification

- `dotnet test`: 657/657 pass.
- `CI=true dotnet build -c Release`: 0 warnings, 0 errors.

### DIFF
--- a/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
@@ -98,7 +98,7 @@ internal sealed class DotNetProjectInitializer(IDotnetCliRunner dotnetCli, ITemp
             "--name", projectName,
             "--output", projectPath,
             "--language", language,
-            "--framework", framework,
+            "--Framework", framework,
             "--debug:custom-hive", _hivePathProvider.HivePath,
         ];
 
--- a/src/Workloads/Stacks/Go/Workloads.Go.csproj
+++ b/src/Workloads/Stacks/Go/Workloads.Go.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
     <Title>Go</Title>
-    <Description>Azure Functions tooling for Go projects.</Description>
+    <Description>Azure Functions CLI tooling for Go projects.</Description>
     <PackageType>FuncCliWorkload</PackageType>
     <PackageTags>kind:workload alias:go func-workload</PackageTags>
     <IncludeBuildOutput>false</IncludeBuildOutput>
--- a/src/Workloads/Stacks/Go/Workloads.Go.csproj
+++ b/src/Workloads/Stacks/Go/Workloads.Go.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
-    <Title>Go</Title>
+    <Title>Go Stack</Title>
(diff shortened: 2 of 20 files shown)


src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
```diff
@@ -98,7 +98,7 @@ public async Task InitializeAsync(
             "--name", projectName,
             "--output", projectPath,
             "--language", language,
-            "--framework", framework,
+            "--Framework", framework,
```
satvu, line 101: huh this worked for me on windows - is this something regarding our template package? b/c the documentation is not capitalized: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new#options

src/Workloads/Stacks/Go/Workloads.Go.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
     <Title>Go</Title>
-    <Description>Azure Functions CLI tooling for Go projects.</Description>
+    <Description>Azure Functions tooling for Go projects.</Description>
```
satvu: I'm ok w/ it just wondering why we are removing CLI references across the files

liliankasem: No I don't actually want to do this - will undo :D

src/Workloads/Stacks/DotNet/DotNetProjectInitializer.cs
liliankasem: Doesn't work for me :(

Confirmed on this Mac: lowercase fails, capital succeeds. Repro just now in /tmp:

Plain text
--framework net10.0  →  exit 127, "'--framework' is not a valid option"
--Framework net10.0  →  Restore succeeded.

PR #5126, opened 2026-05-28
Fabiocav/host integration

fabiocav:
## Summary
 
 This brings `func start` together with the host workload so the CLI now resolves, installs, launches, and manages the real host workload process by default, while preserving the hidden demo mode for demos and iteration.
 
 Key changes:
 - Add CLI-side host workload launch infrastructure for `func start`
   - Resolve the RID-specific host workload package.
   - Auto-install the host workload when needed.
   - Launch the workload executable from `tools\any`.
   - Pass the prepared `HostRunContext` startup directory and environment variables into the host process.
   - Validate port availability before launch.
   - Preserve host process lifecycle and exit code through the dashboard pipeline.
 - Add structured host-to-CLI logging
   - Emit workload NDJSON records for host logs, function metadata, HTTP events, invocation events, and lifecycle signals.
   - Parse structured host records in the CLI before falling back to line-based stdout/stderr parsing.
   - Register structured logging for both the outer workload host and inner ScriptHost logging pipeline.
   - Shorten long host log categories at the workload boundary.
 - Improve dashboard rendering
   - Surface structured log fields in plain, compact, and JSON output.
   - Handle wrapped compact log rows without cropping dashboard sections.
   - Suppress empty console-format header rows when fallback parsing is used.
 - Add host workload diagnostics and cleanup
   - Add opt-in raw host log capture for diagnostics.
   - Add Windows job-object cleanup for launched host process trees.
   - Add host workload packaging marker content.
 - Update the DotNet stack workload to use constants and dynamic worker resolution.
 - Add coverage for host process launch, output parsing, lifecycle handling, workload resolution/install, structured host logging, raw capture, compact rendering, and start initialization behavior.


liliankasem:
Ship it 🚀

PR #5131, opened 2026-05-28
Add VS Code launch + script for debugging a workload alongside func

liliankasem:
## What

A one-keystroke (F5) debug loop for stepping into both `func` and a workload's source.

- New `Run func cli (debug workload)` launch config in `.vscode/launch.json`. Sets `FUNC_CLI_WORKLOADS_HOME` to `${workspaceFolder}/.debug-workloads` so debugging never touches the real workload home.
- New `build-and-deploy-workload` composite task in `.vscode/tasks.json` (with a `pickString` for the workload csproj) that builds Func and then runs the new script.
- New `eng/scripts/deploy-workload-for-debug.ps1`. Builds the chosen workload, reads `OutputPath`/`AssemblyName`/`PackageId`/`PackageVersion` via `dotnet msbuild -getProperty`, and copies the fresh `.dll` + `.pdb` over the already-installed bits at `<home>/workloads/<id>/<version>/tools/any/`.
- `.gitignore`: ignore `.debug-workloads/`.
- `docs/building-a-workload.md`: short "Debugging" section explaining the flow.

## Prereq

The workload must be installed once into the debug home before F5 will redeploy bits over it:

```pwsh
$env:FUNC_CLI_WORKLOADS_HOME = "$PWD/.debug-workloads"
func workload install <path-to-nupkg>
```

After that, edit workload code, F5, and breakpoints in both Func and the workload source bind once `WorkloadLoadContext.LoadFromAssemblyPath` fires.

## Notes

- Picker is limited to the stack workloads (Node/Python/DotNet/Go) since pure content workloads (host, workers, extension bundles) have no entry assembly to step into.
- Script smoke-tested against `Workloads.Node` (`dotnet msbuild -getProperty` resolves correctly).
- `justMyCode: false` on the launch config so the debugger steps into workload code that lives outside the Func project tree.


fabiocav (commented):
Adding this logic to an MSBuild target would make it reusable across IDEs. Simpler to expose to VS Code, VS and CLI

PR #5133, opened 2026-05-28
Set executable bit on extracted host workload binary

liliankasem:
## Problem

On macOS / Linux, starting func against an installed workload fails with:

```
Error: Failed to start host process '.../Azure.Functions.Cli.Workloads.Host':
... Permission denied
```

NuGet zip extraction does not preserve Unix mode bits, so the extracted host binary lands without the executable bit set.

## Fix

Narrow workaround: in `WorkloadInstaller`, after extracting a package whose id starts with `Azure.Functions.Cli.Workloads.Host.` (the host workload family), chmod the single well-known host binary at `tools/any/Azure.Functions.Cli.Workloads.Host` on Unix. No-op on Windows. No other package is affected.

### Why this shape

- Reviewer feedback on the prior revision (which sniffed ELF / Mach-O / shebang magic bytes everywhere) flagged that it was too broad and "scary."
- This revision touches exactly one path in exactly one package family. The package id check is the safety gate.
- A general per-package executables mechanism (e.g. `executables` in `workload.json`) can replace this later when more packages need it.

### Changes

- `HostWorkloadPackage.PackageIdPrefix` promoted to `public const` so the installer can match against it.
- `WorkloadInstaller.EnsureHostExecutableBit` called after extraction in both the install and update/staging paths.
- Tests cover the host package (gets +x) and a non-host package shipping the same relative file name (does not get +x).

## Validation

- `dotnet test test/Func.Tests/Func.Tests.csproj`: 820 passed, 0 failed.
- Reproduced the original `Permission denied` locally and confirmed it goes away with a freshly installed host workload built from this branch.

### DIFF
--- a/src/Func/Workloads/HostWorkloadPackage.cs
+++ b/src/Func/Workloads/HostWorkloadPackage.cs
@@ -7,7 +7,7 @@ namespace Azure.Functions.Cli.Workloads;
 
 internal static class HostWorkloadPackage
 {
-    private const string PackageIdPrefix = "Azure.Functions.Cli.Workloads.Host.";
+    internal const string PackageIdPrefix = "Azure.Functions.Cli.Workloads.Host.";
 
     public static string CurrentPackageId => FromRuntimeIdentifier(CurrentRuntimeIdentifier);
 
--- a/src/Func/Workloads/Install/WorkloadInstaller.cs
+++ b/src/Func/Workloads/Install/WorkloadInstaller.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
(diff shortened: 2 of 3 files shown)


src/Func/Workloads/WorkloadMetadata.cs
```diff
@@ -30,6 +30,15 @@ internal sealed class WorkloadMetadata
...
     public EntryPointSpec? EntryPoint { get; init; }
 
+    /// <summary>
+    /// Package-relative paths to files inside the workload that must be
+    /// marked executable on Unix after install (e.g. native apphost
+    /// binaries). Each entry must be a forward-slash relative path that
+    /// stays within the package. Null / empty means the workload ships no
+    /// native executables.
+    /// </summary>
+    public IReadOnlyList<string>? Executables { get; init; }
```
liliankasem: @fabiocav thoughts on this for now?

PR #5160, opened 2026-05-28
fix(vnext): improve bundle channel resolution logic

jviau:
- Update bundle channel resolution logic to allow for release segments mapping to a stable channel. This allows for preview semantics as we iterate on stable releases, such as `4.35.0-rc.1`. This will unblock using the local nupkg with the `-dev` suffix as is, as well as non-release CI packages.
- Refactor template <--> bundle matching
- Fix template package version comparison
    - previous code was using a simple string ordinal comparison, which is not correct when comparing SemVer2 (`1.10.0` would be considered less than `1.9.0` with string comparison).

### DIFF
--- a/src/Func/Templates/TemplatesChannelMapper.cs
+++ b/src/Func/Templates/TemplatesChannelMapper.cs
@@ -1,6 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Azure.Functions.Cli.Bundles;
+using Azure.Functions.Cli.Projects;
+using NuGet.Versioning;
+
 namespace Azure.Functions.Cli.Templates;
 
 /// <summary>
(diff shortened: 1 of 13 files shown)


liliankasem (approved):

liliankasem (changes requested):
We would also need to fix the templates workloads

```
> f5 new
Error: No installed templates workload matches this project's bundle channel (Microsoft.Azure.Functions.ExtensionBundle -> channel 'stable').
Install one with: func workload install Azure.Functions.Cli.Workloads.Templates.Node --version <version>.
```

src/Func/Templates/TemplatesChannelMapper.cs
```diff
@@ -16,100 +20,40 @@ namespace Azure.Functions.Cli.Templates;
...
         ArgumentNullException.ThrowIfNull(rows);
-        ArgumentNullException.ThrowIfNull(channelLabel);
+        if (channel == BundleChannel.Unknown)
+        {
+            return null;
+        }
 
         InstalledTemplatesWorkload? best = null;
+        NuGetVersion? bestVersion = null;
         foreach (InstalledTemplatesWorkload row in rows)
```
jviau, line 44: I am wondering if `InstalledTemplatesWorkload` needs to be in abstractions? Being in there means I couldn't change `PackageVersion` to type `NuGetVersion`, and I need to do a parse here.

liliankasem (approved):

src/Func/Templates/TemplatesChannelMapper.cs
jviau: going to leave as is for now.

PR #5164, opened 2026-05-28
func new --template <id> --help: render hydrated, deduped, resource-resolved options

soninaren:
## What

Three follow-up fixes to `func new --template <id> --help` against `vnext`, all surfaced by running against the real packed Node/Python templates workloads:

| # | Commit | Fix |
|---|---|---|
| 1 | `75220c46` | `func new --template <id> --help` actually renders the hydrated per-template options (was silently dropped) |
| 2 | `f4100d3c` | Per-prompt `$xxx_label` resource keys (e.g. `$trigger_functionName_label`) get resolved to the user-facing English text |
| 2 | `f4100d3c` | Same prompt id referenced by N jobs of the same template (e.g. CreateNewApp + AppendToFile + CreateNewBlueprint + AppendToBlueprint all want `trigger-functionName`) is now surfaced once, not N times |

## Bug 1 — stage-B `--help` wasn't wired

`NewCommandRunner.cs:130` called `_optionHydrator.Hydrate(template)` and immediately discarded the result (`_ = hydrated;`). The hydrated `Option<T>` list never reached the parser, so `func new --template <id> --help` looked identical to top-level help no matter which template was specified.

**Fix:** introduce `ITemplateAwareHelpProvider`, implemented by `NewCommand`. `SpectreHelpAction.Invoke` calls it after matching the command but before rendering, so the command can dynamically attach options that depend on a stage-A value. `NewCommand.AttachHydratedOptionsForHelp` drops any previously-attached dynamic options (so reruns start clean), reads `--template <id>` from the parse result, runs the resolution + hydration pipeline synchronously via `GetAwaiter().GetResult()`, and attaches the resulting `Option<T>` instances to the command's `Options` collection before help renders.

A new `NewCommandRunner.HydrateOptionsForTemplateAsync` re-uses the same `ResolveContextAsync` + `ListTemplatesAsync` pipeline `ListAsync` runs, finds the matching template, and returns `_optionHydrator.Hydrate(template)`. Resolution failures (no project, no workload, no matching template) return `null` so the help renderer falls back to built-ins-only output instead of crashing.

## Bug 2a — `$xxx_label` resource keys not resolved

`V2TemplateProjection.ProjectInput` set the prompt description to `doc?.Label` directly. The label in real `userPrompts.json` files is a resource key like `"label": "$trigger_functionName_label"` and needs to be looked up in `Resources.json`. Project-level Name and Description were already going through `ResolveResource` — the same helper just wasn't being applied to prompts.

**Fix:** route `doc.Label` through `ResolveResource(doc?.Label, payload.Resources)`.

## Bug 2b — same prompt repeated across jobs

A v2 template typically has 4 jobs (`CreateNewApp`, `AppendToFile`, `CreateNewBlueprint`, `AppendToBlueprint`) that share the same inputs. The projection iterated every job's `Inputs` and pushed each prompt without deduping. Result: `func new -t HttpTrigger-Python --help` listed `--trigger-function-name` 4×.

**Fix:** a `HashSet<string>` (case-insensitive) guards the prompt list against duplicate `paramId`s.

## Before / after

```
$ func new --template HttpTrigger-Python --help
```

**Before:**
```
--app-file-name                   $app_filename_label
--trigger-function-name           $trigger_functionName_label
--http-trigger-auth-level         $httpTrigger_authLevel_label
--app-selected-file-name          $app_selected_filename_label
--trigger-function-name           $trigger_functionName_label
--http-trigger-auth-level         $httpTrigger_authLevel_label
--blueprint-file-name             $blueprint_filename_label
--trigger-function-name           $trigger_functionName_label
--http-trigger-auth-level         $httpTrigger_authLevel_label
--blueprint-existing-file-name    $blueprint_selected_filename_label
--trigger-function-name           $trigger_functionName_label
--http-trigger-auth-level         $httpTrigger_authLevel_label
```

**After:**
```
--app-file-name                   Provide a filename for your function
--trigger-function-name           Provide a function name
--http-trigger-auth-level         Authorization level
--app-selected-file-name          File Name
--blueprint-file-name             Blueprint function file Name
--blueprint-existing-file-name    Select a file to save for your blueprint function
```

12 rows → 6 unique rows, all with English labels.

## Scope

- Help-rendering only. **The hydrated options are still ignored on non-help invocations** — supporting user-supplied per-prompt argv values (e.g. `func new -t HttpTrigger -n MyHttp --auth-level anonymous`) needs a stage-B re-parse in `ExecuteAsync`; that's tracked as a follow-up. This PR closes the discoverability gap so users at least see what each template accepts.
- Existing **9 / 9** V2 unit tests still pass. Release build clean (`CI=true`, `TreatWarningsAsErrors`).

## Validation

End-to-end against the real packed Node + Python templates workloads. Both stacks now render hydrated, deduped, English-labelled options under `--help`.

### DIFF
--- a/src/Func/Commands/SpectreHelpAction.cs
+++ b/src/Func/Commands/SpectreHelpAction.cs
@@ -24,6 +24,17 @@ internal sealed class SpectreHelpAction(HelpCommand helpCommand) : SynchronousCo
             return _helpCommand.ShowGeneralHelp();
         }
 
+        // Stage-B help hydration: when the matched command is a
+        // template-aware verb (e.g. `func new`) and the user supplied a
+        // stage-A value the help renderer needs to know about
+        // (`--template <id>`), let the command attach the relevant
+        // dynamic options before we hand it to the renderer. Falls
+        // through to a built-ins-only render when the hook returns 0.
+        if (command is ITemplateAwareHelpProvider templateAware)
+        {
+            templateAware.AttachHydratedOptionsForHelp(parseResult);
+        }
+
         _helpCommand.RenderCommandHelp(command);
         return 0;
     }
--- a/src/Func/Commands/NewCommand.cs
+++ b/src/Func/Commands/NewCommand.cs
@@ -16,7 +16,7 @@ namespace Azure.Functions.Cli.Commands;
 /// the engine-dispatch step until <see cref="ITemplateEngineProvider"/>
 /// implementations exist.
 /// </summary>
-internal sealed class NewCommand : FuncCliCommand, IBuiltInCommand
+internal sealed class NewCommand : FuncCliCommand, IBuiltInCommand, ITemplateAwareHelpProvider
 {
     public Option<string?> NameOption { get; } = new("--name", "-n")
     {
@@ -44,6 +44,7 @@ internal sealed class NewCommand : FuncCliCommand, IBuiltInCommand
     };
 
     private readonly NewCommandRunner _runner;
+    private readonly HashSet<string> _builtInOptionNames;
 
     public NewCommand(NewCommandRunner runner)
         : base("new", "Create a new function from a template.")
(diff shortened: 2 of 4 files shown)


satvu (approved):

src/Func/Commands/SpectreHelpAction.cs
```diff
@@ -24,6 +24,17 @@ public override int Invoke(ParseResult parseResult)
             return _helpCommand.ShowGeneralHelp();
         }
 
+        // Stage-B help hydration: when the matched command is a
+        // template-aware verb (e.g. `func new`) and the user supplied a
+        // stage-A value the help renderer needs to know about
+        // (`--template <id>`), let the command attach the relevant
+        // dynamic options before we hand it to the renderer. Falls
+        // through to a built-ins-only render when the hook returns 0.
+        if (command is ITemplateAwareHelpProvider templateAware)
```
fabiocav, line 33: Discussed offline. Let's follow up with a fix to this.

PR #5184, opened 2026-05-29
Enable workload prerelease environment override

fabiocav:
## Summary
- Add environment-backed workload catalog options setup for `FUNC_CLI_WORKLOADS_SOURCE` and `FUNC_CLI_WORKLOADS_PRERELEASE`.
- Apply the prerelease override centrally in workload catalog and installer resolution paths, while restoring command/start callers to stable-only defaults.
- Remove `WorkloadSourceResolver` and update workload tests, including the Windows-safe Azurite Unix path lookup assertion.


liliankasem (approved):

liliankasem:
@fabiocav - all of our packages during preview will have -preview.1 (minus bundles and templates I think?) So should we introduce this change if it means that customers will have to set this environment variable? Maybe it's easier if they don't need to set anything and we keep the default as prerelease is visible always

PR #5189, opened 2026-05-29
Suppress existing Azurite endpoint status output

fabiocav:
## Summary
- Suppress the existing Azurite endpoint status line during managed Azurite initialization to avoid live display rendering issues.

## Tests
- dotnet test test\Func.Tests\Func.Tests.csproj --filter FullyQualifiedName~ManagedAzuriteOrchestratorTests


liliankasem (approved):
Lets defo revisit this

PR #5220, opened 2026-05-29
enhancement: fix first-run experience

liliankasem:
Implements the first-run experience documented in
[`docs/first-run-experience.md`](docs/first-run-experience.md). The spec
itself is added as part of this PR; treat the doc as the source of truth
for behavior.

## What the user sees

- **First-time use.** On the user's first invocation (including bare
  `func`), the CLI shows a multi-paragraph explanation of the new
  workload model and asks `Run \`func setup\` now? [Y/n]`. Yes runs
  setup inline; No writes a marker and continues; Ctrl+C writes the
  marker and aborts.
- **`func setup` interactive selection.** Multi-select shows the stacks
  that are not already installed. When some stacks are already
  installed, a single muted `Already installed: node, python` line
  surfaces above the prompt. Empty selection is a clean opt-out: marker
  written, hint shown, exit 0.
- **`func init` / `func new` first-run.** The prompt fires, setup runs
  inline, and the coordinator short-circuits with
  `Setup complete. Re-run \`func init\` to use the new stacks.` so the
  user does not hit a confusing "no templates found" error from the
  cached workload loader. A proper in-process reload is tracked as
  follow-up work.
- **Subsequent runs after declining.** When the marker is present but
  no workloads are installed, every non-skipped command shows one muted
  line: `Tip: run \`func setup\` to install your dev environment and
  language stack dependencies.` Suppressed in non-interactive contexts.
- **Always-skipped surfaces.** `func setup`, `func --help` / `-h` /
  `--version` / `-v`, and parse-error invocations stay quiet.

## State

- Marker file: `~/.azure-functions/.first-run-complete` (override via
  `FUNC_HOME`). Presence is the signal; contents are informational.
- Workload registry: `~/.azure-functions/workloads.json` via
  `IWorkloadStore`.

A user is first-run only when **both** the marker is missing and the
workload registry is empty. The workload-aware check covers users who
installed workloads through `func workload install` or an earlier setup
before the marker existed.

## Type / API changes

- `IFirstRunStateStore` is async (`IsFirstRunAsync`) and gains
  `GetStateAsync` returning `FirstRunState` (`NeverPrompted`,
  `MarkerWithoutWorkloads`, `WorkloadsInstalled`). The coordinator
  picks between prompt, breadcrumb, and silent in one read.
- `IFirstRunCoordinator.EnsureFirstRunPromptedAsync` now returns
  `Task<int?>`. `null` means "continue with the user's command"; a
  value means the coordinator handled it (currently only used for the
  post-setup re-run hint on `init` / `new`, which returns `0`).
- `FileFirstRunStateStore` takes an `IWorkloadStore` and writes the
  marker on every terminal setup outcome that did not fail mid-install
  (success, opt-out, partial install with `--check`). Real install
  failures leave the marker unwritten so the user is re-prompted next
  time.

## Notes on the deferred workload loader reload

`IWorkloadStore.GetWorkloadsAsync` reads `workloads.json` fresh on every
call, but templates / bundles adapters snapshot workload metadata at
host build time. Auditing every consumer is out of scope here; the
documented fallback (re-run hint for `init` / `new`) is the path taken.
A follow-up issue should be filed for the in-process reload work.

## Tests

- Unit coverage in `FirstRunCoordinatorTests`, `FileFirstRunStateStoreTests`,
  and `SetupRunnerTests` for every behavior listed above (prompt,
  breadcrumb, Ctrl+C marker, init/new short-circuit, installed-stack
  filtering, skip list, non-interactive suppression).
- Local: 990 passed, 0 failed. Release build with
  `TreatWarningsAsErrors`: 0 warnings.

## Docs

`docs/first-run-experience.md` is the source of truth for the behavior
above. The "Implementation status" table at the bottom tracks which
items shipped here and which are deferred (loader reload).

### DIFF
--- a/src/Func/Console/SpectreInteractionService.cs
+++ b/src/Func/Console/SpectreInteractionService.cs
@@ -319,7 +319,10 @@ internal class SpectreInteractionService : IInteractionService
             .ShowAsync(_stderr, cancellationToken);
     }
 
-    public async Task<IReadOnlyList<string>> PromptForMultiSelectionAsync(string title, IEnumerable<string> choices, CancellationToken cancellationToken = default)
+    public Task<IReadOnlyList<string>> PromptForMultiSelectionAsync(string title, IEnumerable<string> choices, CancellationToken cancellationToken = default)
+        => PromptForMultiSelectionAsync(title, choices.Select(static c => new MultiSelectionChoice(c)), cancellationToken);
+
+    public async Task<IReadOnlyList<string>> PromptForMultiSelectionAsync(string title, IEnumerable<MultiSelectionChoice> choices, CancellationToken cancellationToken = default)
     {
         var choiceList = choices.ToList();
         if (!IsInteractive || choiceList.Count == 0)
@@ -327,13 +330,19 @@ internal class SpectreInteractionService : IInteractionService
             return [];
         }
 
-        List<string> selected = await new MultiSelectionPrompt<string>()
+        // NotRequired() lets the user press ENTER with nothing selected to exit
+        // the prompt cleanly. Callers treat an empty result as "no selection",
+        // which (for `func setup`) is the documented escape hatch from the
+        // stack picker.
+        List<MultiSelectionChoice> selected = await new MultiSelectionPrompt<MultiSelectionChoice>()
             .Title(title)
-            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm)[/]")
+            .NotRequired()
+            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm; ENTER with no selection exits)[/]")
+            .UseConverter(static choice => choice.Label)
             .AddChoices(choiceList)
             .ShowAsync(_stderr, cancellationToken);
 
-        return selected;
+        return [.. selected.Select(static choice => choice.Value)];
     }
 
     public async Task<string> PromptForInputAsync(string prompt, string? defaultValue = null, CancellationToken cancellationToken = default)
--- a/src/Func/Console/SpectreInteractionService.cs
+++ b/src/Func/Console/SpectreInteractionService.cs
@@ -337,7 +337,7 @@ internal class SpectreInteractionService : IInteractionService
         List<MultiSelectionChoice> selected = await new MultiSelectionPrompt<MultiSelectionChoice>()
             .Title(title)
             .NotRequired()
-            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm; ENTER with no selection exits)[/]")
+            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm; [green]<enter>[/] with no selection exits)[/]")
             .UseConverter(static choice => choice.Label)
             .AddChoices(choiceList)
             .ShowAsync(_stderr, cancellationToken);
--- a/src/Func/Console/SpectreInteractionService.cs
+++ b/src/Func/Console/SpectreInteractionService.cs
@@ -334,28 +334,15 @@ internal class SpectreInteractionService : IInteractionService
         // the prompt cleanly. Callers treat an empty result as "no selection",
         // which (for `func setup`) is the documented escape hatch from the
         // stack picker.
-        MultiSelectionPrompt<MultiSelectionChoice> prompt = new MultiSelectionPrompt<MultiSelectionChoice>()
+        List<MultiSelectionChoice> selected = await new MultiSelectionPrompt<MultiSelectionChoice>()
             .Title(title)
             .NotRequired()
             .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm; ENTER with no selection exits)[/]")
             .UseConverter(static choice => choice.Label)
-            .AddChoices(choiceList);
+            .AddChoices(choiceList)
+            .ShowAsync(_stderr, cancellationToken);
 
-        foreach (MultiSelectionChoice choice in choiceList)
-        {
-            if (choice.IsPreselected)
-            {
-                prompt.Select(choice);
-            }
-        }
-
-        List<MultiSelectionChoice> selected = await prompt.ShowAsync(_stderr, cancellationToken);
-
-        // Disabled entries are decorative: they render so the user can see them
-        // in context, but toggling them has no effect. Strip them from the
-        // result so callers never act on values they didn't intend to expose
-        // as selectable.
-        return [.. selected.Where(static choice => !choice.IsDisabled).Select(static choice => choice.Value)];
+        return [.. selected.Select(static choice => choice.Value)];
     }
 
     public async Task<string> PromptForInputAsync(string prompt, string? defaultValue = null, CancellationToken cancellationToken = default)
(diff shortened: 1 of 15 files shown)


src/Func/Hosting/FirstRun/FirstRunCoordinator.cs
```diff
@@ -10,92 +10,172 @@ namespace Azure.Functions.Cli.Hosting.FirstRun;
...
+        bool runSetup;
+        try
+        {
+            runSetup = await _interaction.ConfirmAsync(PromptMessage, defaultValue: true, cancellationToken);
+        }
+        catch (OperationCanceledException)
+        {
+            // Ctrl+C at the prompt = "stop pestering me". Write the marker
+            // so we don't ask again next time, then propagate the cancel.
+            await TryMarkCompleteAsync(cancellationToken);
```
satvu: AI review:

 🐛 Bug: Marker never written on Ctrl+C

When the user presses Ctrl+C at the confirm prompt, the code catches OperationCanceledException and calls TryMarkCompleteAsync(cancellationToken) — but that token is already cancelled. Inside TryMarkCompleteAsync, MarkCompleteAsync(cancellationToken) calls File.WriteAllTextAsync(..., cancellationToken), which immediately throws another OperationCanceledException that gets swallowed. The marker is never written, so the user gets re-prompted every time.

Fix: Pass CancellationToken.None to TryMarkCompleteAsync (and through to MarkCompleteAsync) on the cancellation path, since the intent is to persist despite cancellation:

```
   catch (OperationCanceledException)
   {
       await TryMarkCompleteAsync(CancellationToken.None);
       throw;
   }
```

liliankasem: Good catch, fixed in 01f68897. Passing `CancellationToken.None` on the cancel path and tightened the regression test to assert the token isn't cancelled (the mock-based test was hiding the bug).

liliankasem:
UX

```
> f5

Looks like this is your first time running func.

The new CLI uses installable workloads to bring in language stacks (Node.js, Python, .NET, Go) and the host runtime. Running `func setup` now will install the host and the dev environment / dependencies for the
stack(s) you pick.

You can always add more later with `func setup --features <node|dotnet|go|python>`.

Run `func setup` now? [y/n] (y):
```

on init

```
> f5 init

Looks like this is your first time running func.

The new CLI uses installable workloads to bring in language stacks (Node.js, Python, .NET, Go) and the host runtime. Running `func setup` now will
install the host and the dev environment / dependencies for the stack(s) you pick.

You can always add more later with `func setup --features <node|dotnet|go|python>`.

Run `func setup` now? [y/n] (y): y
Including prerelease workload versions.
Setting up Azure Functions
✓ Installed host 4.1048.200-local.20260529232203.
✓ Installed go worker 1.0.0-local.20260530004646.
✓ Installed go stack 1.0.0-local.20260530004646.
✓ Installed extension bundle 4.35.0-local.20260529232203.
✓ Azure Functions setup is complete.

Setup complete. Re-run `func init` to use the new stacks.
```

on already installed items:

```bash
>f5 setup
Already installed (use `func workload uninstall <name>` to remove):
   [✓] go
   [✓] node
   
Select stacks to install (SPACE to toggle, ENTER to confirm; ENTER with no selection exits):

> [ ] dotnet
  [ ] python
  (press <space> to toggle, <enter> to confirm; ENTER with no selection exits)
```

on no selection:

`No stacks selected. Nothing to install. Run `func setup` again, or `func setup --features <name>`, when you're ready.
`

src/Func/Console/SpectreInteractionService.cs
```diff
@@ -318,21 +318,30 @@ public async Task<string> PromptForSelectionAsync(string title, IEnumerable<stri
...
-        List<string> selected = await new MultiSelectionPrompt<string>()
+        // NotRequired() lets the user press ENTER with nothing selected to exit
+        // the prompt cleanly. Callers treat an empty result as "no selection",
+        // which (for `func setup`) is the documented escape hatch from the
+        // stack picker.
+        List<MultiSelectionChoice> selected = await new MultiSelectionPrompt<MultiSelectionChoice>()
             .Title(title)
-            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm)[/]")
+            .NotRequired()
+            .InstructionsText("[grey](press [blue]<space>[/] to toggle, [green]<enter>[/] to confirm; ENTER with no selection exits)[/]")
```
fabiocav: `ENTER` here is inconsistent. It shows as `[green]<enter>[/]` and `ENTER` here on the same line.

src/Func/Hosting/FirstRun/FirstRunCoordinator.cs
```diff
@@ -40,62 +56,130 @@ internal sealed class FirstRunCoordinator(
...
+        // command. The workload loader snapshots templates/bundles/host metadata
+        // at host build time, so anything the user typed (init, new, run,
+        // start, quickstart, ...) is operating against the pre-install view
+        // and would either fail outright ("no project here") or silently use
+        // stale metadata. Asking them to re-run is the only safe answer until
+        // we add an in-process reload.
+        if (setupRan && !string.IsNullOrEmpty(commandName))
+        {
+            _interaction.WriteBlankLine();
+            _interaction.WriteHint($"Setup complete. Re-run `func {commandName.ToLowerInvariant()}` to use the new stacks.");
```
fabiocav: Will this display `unknown` for bare `func` calls?

src/Func/Hosting/FirstRun/FileFirstRunStateStore.cs
```diff
@@ -2,22 +2,64 @@
...
 
     private string MarkerPath => Path.Combine(_paths.Home, MarkerFileName);
 
-    public bool IsFirstRun() => !File.Exists(MarkerPath);
+    public async Task<bool> IsFirstRunAsync(CancellationToken cancellationToken = default)
+        => await GetStateAsync(cancellationToken) == FirstRunState.NeverPrompted;
+
+    public async Task<FirstRunState> GetStateAsync(CancellationToken cancellationToken = default)
+    {
+        bool markerExists = File.Exists(MarkerPath);
```
fabiocav: It would be good to check the performance impact of this change. This is on the hot path and will impact every command. We want to avoid anything that would cause the initial run to lag.

Prior to the change, we only checked for the marker file. Now we're loading workloads, which will incur additional disk IO, read the workload JSON and deserialize it with very call.

src/Func/Console/SpectreInteractionService.cs
liliankasem: Good catch, fixed in ce8e185 — both halves of the instruction string now use `[green]<enter>[/]`.

src/Func/Hosting/FirstRun/FirstRunCoordinator.cs
liliankasem: Yes, that's possible — `CommandNameResolver` returns `help` for bare `func`, `version` for `func --verbose`, and `unknown` on parse errors. Fixed in ce8e185: for those sentinel names we now print a generic "Setup complete. Run \`func <command>\` to use the new stacks." instead of "Re-run \`func help\`".

src/Func/Hosting/FirstRun/FileFirstRunStateStore.cs
liliankasem: Agreed — fixed in ce8e185. `FileFirstRunStateStore` no longer reads/deserializes `workloads.json`; it just probes `File.Exists` on the registry path as the "workloads installed" signal. The hot path is now two file stats and zero JSON. The corner case (installed → uninstalled all) leaves the empty registry file present so we'd treat them as workloads-installed and skip the breadcrumb — fine tradeoff for the perf win.

jviau (approved):

PR #5223, opened 2026-05-29
Implement first-run experience spec

liliankasem:
Stacked on top of #5220. Targeted at `liliankasem/feat/setup-ux` so the diff
shows only this PR's changes; merge into `vnext` after #5220 lands.

Implements the pending items from
[`docs/first-run-experience.md`](docs/first-run-experience.md).

## Items

1. **Bare `func` triggers the prompt.** Tightened the bare-invocation
   check in `FirstRunCoordinator.IsSkipped` so any invocation with
   `Tokens.Count == 0` reaches the prompt, regardless of whether the
   parser reports a "Required command was not provided" error.

2. **New prompt copy.** Replaced the short explanation with the
   multi-paragraph copy from the spec (workloads, language stacks, host
   runtime, `--features <node|dotnet|go|python>` hint). Routed through
   `IInteractionService.WriteLine` / `WriteHint` (no raw `Console`).

3. **Ctrl+C writes the marker.** The confirm prompt is wrapped so
   `OperationCanceledException` writes the first-run marker before
   rethrowing. Best-effort: a failure to write the marker is swallowed
   so the original cancellation surfaces unchanged.

4. **Installed stacks are filtered out, not labelled.** Spectre's
   `MultiSelectionPrompt<T>` does not support per-row disabling, so I
   went with the preferred option from the brief: filter installed
   stacks out of the choice list and surface them as a single muted
   `Already installed: node, python` hint above the prompt. If every
   supported stack is already installed, the prompt is skipped and the
   run treats it as a clean opt-out.

5. **`func init` / `func new` prompt + post-setup hint.** Neither was
   in the skip list, so the prompt already fires (verified with new
   tests). After the in-process setup succeeds, the coordinator
   short-circuits with `Setup complete. Re-run \`func init\` to use the
   new stacks.` and returns exit code 0. **Spike result on the workload
   loader reload:** `IWorkloadStore.GetWorkloadsAsync` reads
   `workloads.json` fresh on every call, but the templates / bundles
   adapter stack snapshots metadata at host build time. Auditing every
   consumer is out of scope for this PR; the spec's documented fallback
   (re-run hint) is the path taken here. Follow-up issue will be filed
   for the in-process reload work.

6. **Subsequent-run breadcrumb.** Added a `FirstRunState` enum
   (`NeverPrompted`, `MarkerWithoutWorkloads`, `WorkloadsInstalled`)
   exposed via `IFirstRunStateStore.GetStateAsync`. The coordinator
   uses the new state in one read to pick between prompt / breadcrumb
   / silent. The muted `Tip: run \`func setup\` to install your dev
   environment and language stack dependencies.` line is suppressed
   for the same skip list as the prompt and in non-interactive
   contexts.

## Type / API changes

- `IFirstRunCoordinator.EnsureFirstRunPromptedAsync` now returns
  `Task<int?>`. `null` means "continue with the user's command";
  a value means the coordinator handled it (currently only the
  post-setup re-run hint for `init` / `new`, which returns `0`).
- `IFirstRunStateStore` gains `GetStateAsync`. `IsFirstRunAsync` stays
  for backward compatibility and is implemented in terms of the new
  state.

## Tests

All new behavior covered in `FirstRunCoordinatorTests` (+10 cases) and
`FileFirstRunStateStoreTests` (+3 cases). `SetupRunnerTests` updated to
assert the new "Already installed" hint. Local run: 990 passed,
release build (CI warnings-as-errors): 0 warnings.

## Docs

`docs/first-run-experience.md` "Implementation status" table updated and
the "interactive" and "init/new" sections rewritten to match the
shipped behavior. Note added about the deferred loader reload.

### DIFF
--- a/docs/first-run-experience.md
+++ b/docs/first-run-experience.md
@@ -66,9 +66,11 @@ Behaviour:
 
 ### Interactive
 
-- Multi-select shows all stacks (`node`, `python`, `dotnet`, `go`).
-- Stacks already installed are **disabled and grayed out**, with an
-  `(installed)` suffix. They cannot be re-toggled.
+- Multi-select shows the stacks that are not already installed.
+- Stacks already installed are filtered out of the choice list. When any
+  are filtered, a single muted line above the prompt records them:
+  `Already installed: node, python`. If every supported stack is already
+  installed, the prompt is skipped and the run exits cleanly.
 - ENTER with nothing selected is a clean opt-out: exit 0, marker written,
   hint shown ("No stacks selected").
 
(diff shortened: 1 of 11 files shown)


liliankasem:
Folded into #5220.

PR #5229, opened 2026-05-29
Adopt existing projects in `func init`

### LINKED ISSUE #5228: [v5] func init should initialize an existing project
### Version

bug bash

### OS / shell

osx

### Stack

_No response_

### Description

`func init` will setup a new function app project for you. But with this new CLI we now have a new configuration via .func/config.json an `func init` should be able to be used to initalize an existing project by creating the `.func/config.json` if missing , and filling in the stack and language

### Steps to reproduce

n/a

### Expected behavior

_No response_

### Actual behavior

_No response_

### Stack trace / logs

```shell

```

liliankasem:
Fixes #5228.

Today `func init` refuses to run inside a folder that already has a Functions project. v5 considers a project initialized when `.func/config.json` exists, but plenty of real projects predate that marker (they only have `host.json`). Asking those users to either delete files or pass `--force` (which scaffolds on top of their code) is a bad first experience.

## Behavior

- **Empty folder**: scaffold (unchanged).
- **`host.json` present, no `.func/config.json`**: **adopt** — write `.func/config.json` based on the project's stack and skip scaffolding so user source is untouched.
- **`.func/config.json` present**: refuse unless `--force` (unchanged).
- **`--force`**: always scaffolds, bypassing adopt mode.

## Adopt-mode stack resolution

**Step 1 — pick the project signal:**

`project_runtime` = `local.settings.json`'s `FUNCTIONS_WORKER_RUNTIME` ?? null

The process env var is intentionally ignored. Adoption is about the on-disk project shape; an env var set in the user's shell shouldn't change what we write into `.func/config.json`.

**Step 2 — apply this 3×3:**

| `project_runtime` → | `--stack` omitted | `--stack X` (installed) agrees | `--stack X` conflicts |
|---|---|---|---|
| **null** | interactive: prompt "Adopting an existing project. Which stack is it?". Non-interactive: refuse with "Couldn't detect the project's stack". | adopt with `X` | n/a |
| **installed** | adopt + snap | adopt with `X` | refuse (use `--force` to override) |
| **uninstalled** | refuse with `func setup --features <stack>` hint | refuse (same hint) | refuse |

Any candidate stack (from `--stack` or `local.settings.json`) that doesn't match an installed initializer is refused. The hint uses a literal `<stack>` placeholder rather than the raw runtime value, because the value itself may not be a valid feature id (e.g. typos, or `native` for Go).

## Tests

35 `InitCommand` tests cover every cell in the matrix, including: adoption when only `host.json` is present, snap from `local.settings.json`, refuse-on-conflict, refuse-on-uninstalled (both signal sources), refuse-when-no-signal in non-interactive mode, prompt-when-no-signal in interactive mode, `--force` override, and the pre-existing fresh-init flows.

Full suite: 981 passing.

### DIFF
--- a/docs/cli-architecture.md
+++ b/docs/cli-architecture.md
@@ -362,11 +362,24 @@ If a newer version is found, a notice is printed after the command completes.
    └── No matching initializer    → "No installed workload supports stack '<x>'." (exit 1)
 3. Detect existing project state:
    ├── .func/config.json present  → fully initialized; refuse without --force
-   ├── host.json only             → adopt: write .func/config.json, skip scaffolding
+   ├── host.json only             → adopt mode (see below); skip scaffolding
    └── empty                      → scaffold via IProjectInitializer.InitializeAsync
 4. Build InitContext (WorkingDirectory, ProjectName, Language, Force)
 5. Delegate to IProjectInitializer.InitializeAsync(context, parseResult) (skipped in adopt mode)
 ```
+
+**Adopt mode** (host.json present, no `.func/config.json`):
+
+1. Resolve `project_runtime` = `FUNCTIONS_WORKER_RUNTIME` env (preferred, matches the host) ?? `local.settings.json` ?? null. Warn if env and `local.settings.json` disagree.
+2. Apply:
+
+   | `project_runtime` → | `--stack` omitted | `--stack X` agrees | `--stack X` conflicts |
+   |---|---|---|---|
+   | null | current behavior (prompt / auto-select / error) | adopt with `X` | n/a |
+   | installed | adopt + snap | adopt with `X` | refuse (use `--force`) |
+   | uninstalled | adopt + hint `func setup --features <stack>` | adopt with `X` | refuse (use `--force`) |
+
+3. Write `.func/config.json` with the resolved stack. Skip scaffolding so user source is untouched. `--force` always bypasses adopt mode and takes the full scaffold path.
 
 Each registered initializer also contributes options to `func init` via `GetInitOptions(IInitOptionRegistry)`; values are read back inside `InitializeAsync` via the `ParseResult`. The registry collapses same-named contributions across workloads so shared options (e.g. `--no-bundles`) show up once in `--help` and every contributing workload reads the canonical instance.
 
--- a/docs/cli-architecture.md
+++ b/docs/cli-architecture.md
@@ -379,7 +379,7 @@ If a newer version is found, a notice is printed after the command completes.
    | installed | adopt + snap | adopt with `X` | refuse (use `--force`) |
    | uninstalled | refuse with `func setup --features <stack>` hint | refuse (same hint) | refuse |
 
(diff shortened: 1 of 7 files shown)


fabiocav (commented):
Do we need to ensure we're aligned with `local.settings.json` if one is present?

Should `func init` allow a stack to be specified/selected? Should we snap to what's in local.settings.json? Should we enable selection, but fail if `local.settings.json` has something different?

liliankasem:
> Do we need to ensure we're aligned with `local.settings.json` if one is present?
> 
> Should `func init` allow a stack to be specified/selected? Should we snap to what's in local.settings.json? Should we enable selection, but fail if `local.settings.json` has something different?

**Step 1 — pick the project signal:**

```
project_runtime = process env FUNCTIONS_WORKER_RUNTIME
                ?? local.settings.json's FUNCTIONS_WORKER_RUNTIME
                ?? null
```

(Env wins over `local.settings.json`, matching the host. If they conflict, warn but keep going.)

|  | `--stack` omitted | `--stack X` agrees with `project_runtime` | `--stack X` conflicts with `project_runtime` |
|---|---|---|---|
| **`project_runtime` = null** | current behavior (prompt / auto-select / error) | adopt with `X` | n/a |
| **`project_runtime` → installed stack** | adopt + snap | adopt with `X` | refuse (use `--force` to override) |
| **`project_runtime` → uninstalled stack** | adopt + hint `func setup --features <stack>` | adopt with `X` | refuse (use `--force`) |

liliankasem:
Tested scenarios.

Trying to init when the stack is not installed:

```
> f5 init -s dotnet
Error: The 'dotnet' stack is not installed. Run `func setup --features <stack>` to setup your dev environment, then re-run `func init`.
```

Trying to init with -s <stack> and the stack is installed

```
> f5 init -s go
· Wrote .func/config.json.

✓ Existing project adopted for go.
```

Trying to init when the project already has a func/config.json

```
> f5 init
Error: This directory already contains a Functions project ('.func/config.json' is present). Pass --force to re-initialize.
```

Trying to init when multiple stacks are available, and the stack is not know (no local settings json worker runtime)

```
> f5 init
Select a stack:

> Go
  Node.js
  .NET
  Python
```

Trying to init and the worker runtime is set in local settings json

```
> f5 init
· Wrote .func/config.json.

✓ Existing project adopted for dotnet.
```

Trying to init and the worker runtime is set in local settings json but the stack is not installed

```
> f5 init
Error: The 'dotnet-isolated' stack is not installed. Run `func setup --features <stack>` to setup your dev environment, then re-run `func init`.
```

---

init in an empty folder with 1 stack installed:

```
> f5 init
Auto-selecting 'go' (the only installed workload).
Install more with func workload install or run func workload search to discover available stacks.
· Wrote .func/config.json.

✓ Project initialized for go.
```

init in an empty folder with >1 stack installed:

```
> f5 init
Select a stack:

> Go
  Node.js

(Type to search)
```

init in an empty folder with 0 stacks installed:

```
Warning: No stacks installed.

Set up a stack to scaffold a quickstart project:

 func setup --features dotnet  C#, F#
 func setup --features node    JavaScript, TypeScript
 func setup --features python  Python
 func setup --features go      Go

Run func workload search to discover available stacks.
```

satvu (approved):

PR #5230, opened 2026-05-29
Bypass proxy on Azurite probe HttpClient

### LINKED ISSUE #5200: [v5] `func run` fails if azurite is already running
### Version

```
╭──────────────┬─────────────────────────────────────────────────────────────────────╮
│ Property     │ Value                                                               │
├──────────────┼─────────────────────────────────────────────────────────────────────┤
│ Version      │ 5.0.0-preview.1.ci.26279.0                                          │
│ Build        │ 5.0.0-preview.1.ci.26279.0+3a2339c01cbb94ca4d4468c5e296d51c6c2b9380 │
│ Runtime      │ .NET 10.0.8                                                         │
│ OS           │ Microsoft Windows 10.0.26200                                        │
│ Architecture │ X64                                                                 │
╰──────────────┴─────────────────────────────────────────────────────────────────────╯
```

### OS / shell

PowerShell 7.6.2

### Stack

dotnet

### Description

`func run` times out starting azurite

### Steps to reproduce

1. Run azurite yourself with default ports
2. Run `func run`

### Expected behavior

Func start succeeds, either using existing azurite or a new azurite using free ports.

### Actual behavior

```
❯ func run
Starting Azurite (native) for AzureWebJobsStorage...Azure Functions CLI
  Executable: C:\Program Files\nodejs\azurite.cmdAzure Functions CLI
  Data: C:\Users\javia\.azure-functions\azurite\dataAzure Functions CLI
  Logs: C:\Users\javia\.azure-functions\azurite\azurite.logAzure Functions CLI
  Endpoints: blob 127.0.0.1:10000, queue 127.0.0.1:10001, table 127.0.0.1:10002Azure Functions CLI
Error: Azurite did not become ready within 30 seconds.

Expected endpoints:
  Blob:  http://127.0.0.1:10000/devstoreaccount1
  Queue: http://127.0.0.1:10001/devstoreaccount1
  Table: http://127.0.0.1:10002/devstoreaccount1

Log file: C:\Users\javia\.azure-functions\azurite\azurite.log
```

### Stack trace / logs

```shell

```

liliankasem:
Fixes #5200.

## Problem

When the user already has Azurite running on default ports and has an HTTP proxy configured (`HTTP_PROXY`, system proxy, or WPAD on Windows), `func run` fails with:

```
Starting Azurite (native) for AzureWebJobsStorage...
  Endpoints: blob 127.0.0.1:10000, queue 127.0.0.1:10001, table 127.0.0.1:10002
Error: Azurite did not become ready within 30 seconds.
```

## Root cause

`AzuriteProbe` uses a named `HttpClient` whose primary handler was the default. The default handler honours the system/env proxy even for `127.0.0.1`. With a proxy in the path, every probe request fails (connection refused or timeout), the probe returns `NotListening`, and `ManagedAzuriteOrchestrator` launches a second Azurite on top of the already-running one. The new process cannot bind the occupied ports and the readiness poll times out.

Confirmed locally: with `HTTP_PROXY=http://127.0.0.1:39999` set against a live Azurite, the probe reports `NotListening` for all three endpoints. Without the env var it reports `Ready`.

## Fix

Configure the probe's primary handler as a `SocketsHttpHandler` with `UseProxy = false`. Azurite always runs on the loopback interface, so a proxy should never be in the path.

## Tests

- New `AddAzuriteProbe_PrimaryHandler_DisablesProxy` regression test walks the configured handler chain and asserts `UseProxy = false` on the inner `SocketsHttpHandler`.
- All 104 `Azurite*` tests pass.
- Manually verified end-to-end: a probe against a live Azurite with `HTTP_PROXY` pointing at an unreachable address now correctly returns `Ready`.

### DIFF
--- a/src/Func/Commands/Start/Azurite/AzuriteServiceCollectionExtensions.cs
+++ b/src/Func/Commands/Start/Azurite/AzuriteServiceCollectionExtensions.cs
@@ -27,6 +27,17 @@ internal static class AzuriteServiceCollectionExtensions
             // probe itself also enforces a per-request timeout via a linked
             // CancellationTokenSource so callers can preempt sooner.
             client.Timeout = AzuriteProbe.PerRequestTimeout;
+        })
+        .ConfigurePrimaryHttpMessageHandler(static () => new SocketsHttpHandler
+        {
+            // Azurite always runs on the loopback interface, so a system or
+            // environment proxy (HTTP_PROXY, WinHTTP, WPAD) must never be in
+            // the path. Without this, a configured proxy intercepts the probe
+            // and returns Connection refused / timeout, the probe reports
+            // NotListening, and the orchestrator launches a second Azurite
+            // that fails to bind the already-occupied ports (issue #5200).
+            UseProxy = false,
+            AllowAutoRedirect = false,
         });
 
         services.AddSingleton<IAzuriteProbe, AzuriteProbe>();
--- a/test/Func.Tests/Commands/Start/Azurite/AzuriteProbeTests.cs
+++ b/test/Func.Tests/Commands/Start/Azurite/AzuriteProbeTests.cs
@@ -46,6 +46,26 @@ public class AzuriteProbeTests : IAsyncLifetime
     }
 
     [Fact]
+    public void AddAzuriteProbe_PrimaryHandler_DisablesProxy()
+    {
+        // Regression for #5200: a configured HTTP_PROXY / system proxy must
+        // not be in the path of the probe, since Azurite always runs on the
+        // loopback interface. If the proxy is used, probes fail with
+        // Connection refused and the orchestrator launches a second Azurite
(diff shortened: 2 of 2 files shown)


jviau (approved):
Approved. But we should maybe have further discussion about the expected azurite behavior here. Do we want to use an existing azurite? What if customer wants to use azurite on non-standard ports? In memory persistence for azurite? Do we want to add support for any of that into func? Or is the expectation they start azurite themselves / update local settings.

liliankasem:
> Approved. But we should maybe have further discussion about the expected azurite behavior here. Do we want to use an existing azurite? What if customer wants to use azurite on non-standard ports? In memory persistence for azurite? Do we want to add support for any of that into func? Or is the expectation they start azurite themselves / update local settings.

Yeah these are all items I want to address post-build

PR #5235, opened 2026-05-29
Dotnet project Output Path Resolution

### LINKED ISSUE #5207: [v5] `func run` for dotnet projects is unreliable with output path resolution
### Version

```
╭──────────────┬─────────────────────────────────────────────────────────────────────╮
│ Property     │ Value                                                               │
├──────────────┼─────────────────────────────────────────────────────────────────────┤
│ Version      │ 5.0.0-preview.1.ci.26279.0                                          │
│ Build        │ 5.0.0-preview.1.ci.26279.0+3a2339c01cbb94ca4d4468c5e296d51c6c2b9380 │
│ Runtime      │ .NET 10.0.8                                                         │
│ OS           │ Microsoft Windows 10.0.26200                                        │
│ Architecture │ X64                                                                 │
╰──────────────┴─────────────────────────────────────────────────────────────────────╯
```

### OS / shell

PowerShell 7.6.2

### Stack

dotnet

### Description

A project can build to a different directory than what `func run` believes is the output path, leading to no worker actually started and no functions loaded.

The issue is we call `dotnet build -getproperty:OutputPath` to discover the apps output and _then_ build the app. Getting the property in a separate step does not guarantee it will accurately reflect the apps true output. As output behavior can vary between the property fetch and a true build.

While this will be "fine" for vast majority of apps, I think it is not a good stance to take to make assumptions about how a project builds/outputs itself.

Recommendation:

We can perform the build & output resolution in a one-shot: `dotnet build -getTargetResult:Build`. We will get a JSON output with exactly what was built:

``` json
{
  "TargetResults": {
    "Build": {
      "Result": "Success",
      "Items": [
        {
          "Identity": "C:\\repos\\bugbash\\dotnet1\\bin\\Debug\\net10.0\\dotnet1.dll",
          "TargetFrameworkIdentifier": ".NETCoreApp",
          "TargetPlatformMoniker": "",
          "CopyUpToDateMarker": "C:\\repos\\bugbash\\dotnet1\\obj\\Debug\\net10.0\\dotnet1.csproj.Up2Date",
          "TargetPlatformIdentifier": "",
          "TargetFrameworkVersion": "10.0",
          "ReferenceAssembly": "C:\\repos\\bugbash\\dotnet1\\obj\\Debug\\net10.0\\ref\\dotnet1.dll",
          "FullPath": "C:\\repos\\bugbash\\dotnet1\\bin\\Debug\\net10.0\\dotnet1.dll",
          "RootDir": "C:\\",
          "Filename": "dotnet1",
          "Extension": ".dll",
          "RelativeDir": "C:\\repos\\bugbash\\dotnet1\\bin\\Debug\\net10.0\\",
          "Directory": "repos\\bugbash\\dotnet1\\bin\\Debug\\net10.0\\",
          "RecursiveDir": "",
          "ModifiedTime": "2026-05-29 12:13:09.4655988",
          "CreatedTime": "2026-05-29 12:13:12.8131613",
          "AccessedTime": "2026-05-29 12:13:12.8141559",
          "DefiningProjectFullPath": "C:\\Program Files\\dotnet\\sdk\\10.0.300\\Microsoft.Common.CurrentVersion.targets",
          "DefiningProjectDirectory": "C:\\Program Files\\dotnet\\sdk\\10.0.300\\",
          "DefiningProjectName": "Microsoft.Common.CurrentVersion",
          "DefiningProjectExtension": ".targets"
        }
      ]
    }
  }
}
```

### Steps to reproduce

1. create a new dotnet app via `func init`
2. add some trigger to your app to demonstrate this wont be loaded
3. setup your csproj with the following:

``` xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <AzureFunctionsVersion>V4</AzureFunctionsVersion>
    <OutputType>Exe</OutputType>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <TargetDir>out\bin\$(Configuration)\net10.0\</TargetDir>
  </PropertyGroup>

  <ItemGroup>
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.52.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.OpenTelemetry" Version="1.2.0" />
    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.15.3" />
    <PackageReference Include="Azure.Monitor.OpenTelemetry.Exporter" Version="1.7.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="2.1.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.7" />
  </ItemGroup>

  <Target Name="_BreakFuncRun" BeforeTargets="BeforeBuild">
    <PropertyGroup>
      <OutputPath>out\bin\$(Configuration)\net10.0\</OutputPath>
      <OutDir>out\bin\$(Configuration)\net10.0\</OutDir>
    </PropertyGroup>
  </Target>

</Project>
```

4. `dotnet build`
5. `func run`

### Expected behavior

App is launched, functions discovered

### Actual behavior

App launches, no functions discovered.

### Stack trace / logs

```shell

```

satvu:

resolves #5207 

## Summary

Fixes #5207

When `func run` builds a .NET project, the output directory is now resolved from the actual build result rather than a separate MSBuild property query. This fixes cases where targets dynamically modify OutputPath/OutDir during the build, causing the CLI to look in the wrong directory for the built app.

## Changes

- **DotNetSourceProject.cs**: Replaced the two-step build-then-query approach with a single dotnet build --getTargetResult:Build invocation that both builds the project and returns the actual output paths as JSON. When --no-build is used, the existing --getProperty:OutputPath fallback is preserved.
- **DotNetSourceProjectTests.cs**: Updated tests to validate the new one-shot build+resolve behavior, added tests for JSON parsing edge cases (missing fields, invalid JSON, empty items).

## How it works

Previously:
1. `dotnet build <project>` (build only)
2. `dotnet msbuild <project> --getProperty:OutputPath` (query output path separately)

Now:
1. `dotnet build <project> --getTargetResult:Build` (build + get actual output in one shot)

The JSON response from --getTargetResult:Build contains the RelativeDir and FullPath of the built assembly, which accurately reflects where the build actually placed outputs — even when MSBuild targets modify the output directory dynamically.

## Testing

- All 48 existing + new tests in Workloads.DotNet.Tests pass.
- Covers: normal build path, FullPath fallback, skip-build path, build failures, empty/invalid JSON, null context, cancellation.

### DIFF
--- a/src/Abstractions/Common/GracefulException.cs
+++ b/src/Abstractions/Common/GracefulException.cs
@@ -8,29 +8,21 @@ namespace Azure.Functions.Cli.Common;
 /// </summary>
 public class GracefulException : Exception
 {
-    public bool IsUserError { get; }
-    public string? VerboseMessage { get; }
-
     public GracefulException(string message, bool isUserError = false, string? verboseMessage = null)
-        : base(message)
+        : this(message, null, isUserError, verboseMessage)
     {
-        IsUserError = isUserError;
-        VerboseMessage = verboseMessage;
     }
 
-    public GracefulException(string message, Exception innerException, bool isUserError = false, string? verboseMessage = null)
+    public GracefulException(string message, Exception? innerException, bool isUserError = false, string? verboseMessage = null)
         : base(message, innerException)
     {
         IsUserError = isUserError;
         VerboseMessage = verboseMessage;
     }
 
-    public GracefulException(string message, string verboseMessage)
-        : base(message)
-    {
-        IsUserError = true;
-        VerboseMessage = verboseMessage;
-    }
+    public bool IsUserError { get; }
+
+    public string? VerboseMessage { get; }
 }
 
 /// <summary>
--- a/src/Abstractions/Common/GracefulException.cs
+++ b/src/Abstractions/Common/GracefulException.cs
@@ -18,6 +18,13 @@ public class GracefulException : Exception
         VerboseMessage = verboseMessage;
     }
 
+    public GracefulException(string message, Exception innerException, bool isUserError = false, string? verboseMessage = null)
+        : base(message, innerException)
+    {
+        IsUserError = isUserError;
+        VerboseMessage = verboseMessage;
+    }
+
     public GracefulException(string message, string verboseMessage)
         : base(message)
     {
--- a/src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
+++ b/src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Text.Json;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Projects;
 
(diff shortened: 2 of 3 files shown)


src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
```diff
@@ -26,15 +27,101 @@ public override async Task PrepareForHostRunAsync(FunctionsProjectHostRunContext
...
         {
-            await BuildAsync(projectDir, cancellationToken);
+            // Build and resolve output in one shot so the output path reflects the actual build result,
+            // even when targets modify OutputPath/OutDir dynamically during the build.
+            outputDirectory = await BuildAndGetOutputDirectoryAsync(projectDir, cancellationToken);
+        }
+        else
+        {
+            // Assume customer has already built, best effort resolution.
+            outputDirectory = await GetOutputDirectoryAsync(projectDir, cancellationToken);
```
jviau: Lets also update this to call `dotnet build {project} -getTargetResult:GetTargetPath`

With that refactor, the skip-build path can follow the exact same code-path, all that changes is the value passed to `getTargetResult`. Their output JSON will basically be the same.

src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
```diff
@@ -26,15 +27,101 @@ public override async Task PrepareForHostRunAsync(FunctionsProjectHostRunContext
...
+                    {
+                        string dir = Path.GetDirectoryName(assemblyPath)!;
+                        return new DirectoryInfo(dir);
+                    }
+                }
+            }
+        }
+        catch (JsonException)
+        {
+            // Fall through to the error below.
```
jviau: Throw a different `GracefulException` here, the message of the one below is not exactly applicable. Also include the JsonException as the inner exception.

jviau (changes requested):

src/Abstractions/Common/GracefulException.cs
```diff
@@ -8,22 +8,21 @@ namespace Azure.Functions.Cli.Common;
 /// </summary>
 public class GracefulException : Exception
 {
-    public bool IsUserError { get; }
-    public string? VerboseMessage { get; }
-
     public GracefulException(string message, bool isUserError = false, string? verboseMessage = null)
-        : base(message)
+        : this(message, null, isUserError, verboseMessage)
```
fabiocav, line 12: While here, I simplified the type a bit: constructors now chain through a single initializer instead of repeating assignments, the inner-exception param is nullable, and I dropped the redundant (string, string) ctor. Removing the ctor is technically a public-API change, but since we're in unpublished preview with no external consumers it's a safe cleanup.

src/Workloads/Stacks/DotNet/DotNetSourceProject.cs
```diff
@@ -26,53 +28,85 @@ public override async Task PrepareForHostRunAsync(FunctionsProjectHostRunContext
...
+                    firstItem["FullPath"]?.GetValue<string>()
+                    ?? firstItem["Identity"]?.GetValue<string>()
+                    ?? string.Empty;
+
+                if (!string.IsNullOrWhiteSpace(assemblyPath))
+                {
+                    // With --no-build we only resolve the target path; nothing compiled it. If the
+                    // assembly isn't actually present, fail here with an actionable message instead of
+                    // letting the host fail later with a more cryptic "worker/app not found" error.
+                    if (requireAssemblyExists && !File.Exists(assemblyPath))
```
fabiocav, line 89: When --no-build is used, we now verify the resolved assembly actually exists on disk and fail with an actionable message, rather than letting the host fail later with a cryptic error.

fabiocav (approved):

fabiocav:
Picked this up to address @jviau's feedback and make some additional changes to get it ready to land in vnext. Need to unblock the project status updates and live logs and want to make sure I bring this in and don't end up introducing a conflict for you.

PR #5242, opened 2026-05-30
Show per-install progress bars in func setup

liliankasem:
## Summary

`func setup` now renders a progress bar for each workload it installs, so users see what's currently being installed and that it's making progress. Particularly useful for the Python worker, which is ~273 MB compressed / 849 MB / 13.6k files unpacked and feels indistinguishable from a hang.

### Before

```
Setting up Azure Functions
✓ Installed host 4.1048.200-local.20260530021444.
✓ Installed python worker 4.43.0-local.20260530021444.
...
```

(...with a multi-minute silent pause on the python worker line.)

### After

```
Setting up Azure Functions
⠋ Installing python worker 4.43.0-local.20260530021444  ━━━━━━━━━━ 23%
```

…with the label cycling through the installer's existing phases (`Resolving…` → `Downloading…` → `Extracting…` → `Registering…`) and then collapsing to:

```
✓ Installed python worker 4.43.0-local.20260530021444.
```

## What changed

- `SetupRunner.EnsureDependencyAsync` wraps the `InstallFromCatalogAsync` call in `IInteractionService.RunWithProgressAsync`, with the initial label `Installing {DisplayName} {targetVersion}`. The installer already reports phases via `IProgress<WorkloadInstallProgress>`; setup was passing `null` and discarding them.
- JSON output mode (`--output json`) skips the progress wrapper so the structured event stream stays clean.
- Extracted the duplicated `ProgressAdapter` from `WorkloadInstallCommand` and `WorkloadUpdateCommand` into a shared `WorkloadInstallProgressAdapter`.

## Behavior in non-TTY (CI logs)

Spectre's `RunWithProgressAsync` falls back to its `LoggingProgressContext`, which prints one line per phase transition. CI logs gain visibility too, no garbled escape codes.

## Validation

- Clean release build, 0 warnings.
- Full test suite: 1004/1004 pass. Existing setup/workload tests already mocked `Arg.Any<IProgress<WorkloadInstallProgress>?>()`, so no test changes were required.

## Follow-up (not in this PR)

The python worker is slow because the upstream package ships every (Python version × OS × arch) combo (~849 MB / 13.6k files). A proper fix is splitting `Workloads.Workers.Python` per RID so installs only fetch the slice they need. This PR is the UX mitigation in the meantime.

### DIFF
--- a/src/Func/Commands/Setup/SetupRunner.cs
+++ b/src/Func/Commands/Setup/SetupRunner.cs
@@ -444,15 +444,28 @@ internal sealed class SetupRunner(
 
         try
         {
-            WorkloadInstallResult installResult = await _workloadInstaller.InstallFromCatalogAsync(
-                package.PackageId,
-                package.Version,
-                options.Source,
-                includePrerelease: options.IncludePrerelease,
-                exact: true,
-                force: false,
-                progress: null,
-                cancellationToken);
+            WorkloadInstallResult installResult = options.OutputMode == SetupOutputMode.Json
+                ? await _workloadInstaller.InstallFromCatalogAsync(
+                    package.PackageId,
+                    package.Version,
+                    options.Source,
+                    includePrerelease: options.IncludePrerelease,
+                    exact: true,
+                    force: false,
+                    progress: null,
+                    cancellationToken)
+                : await _interaction.RunWithProgressAsync(
+                    $"Installing {dependency.DisplayName} {targetVersion}",
+                    async (ctx, ct) => await _workloadInstaller.InstallFromCatalogAsync(
+                        package.PackageId,
+                        package.Version,
+                        options.Source,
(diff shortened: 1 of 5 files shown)


liliankasem:
progress bar shows when waiting on a workload install:

Screenshot 2026-05-29 at 8 15 34 PM

then after is collapses:

```bash
Setting up Azure Functions
✓ host 4.1048.200 is already installed.
✓ Installed dotnet stack 1.0.0.
✓ Installed dotnet templates 1.0.0.
✓ Installed python worker 4.43.0.
✓ Installed python stack 1.0.0.
✓ Installed python templates 1.0.0.
✓ extension bundle 4.35.0 is already installed.
✓ Azure Functions setup is complete.
```

satvu (approved):

PR #5254, opened 2026-06-01
build(templates/dotnet): restore upstream nupkg via PackageDownload

soninaren:
Move the upstream Functions item-templates package download out of pack and into restore.

## Problem

The DotNet templates workload hydrate step ran `dotnet new install <id>@<version>` at pack time, which made an HTTP call to NuGet.org every build to fetch `Microsoft.Azure.Functions.Worker.ItemTemplates`. That broke any build environment without network access at pack time and made local iteration brittle.

## Fix

- Declare `<PackageDownload Include="$(SourceItemTemplatesPackageId)" Version="[$(SourceItemTemplatesVersion)]" />` in `Workloads.Templates.DotNet.csproj` so restore pulls the upstream nupkg into the global packages folder ([PackageDownload docs](https://learn.microsoft.com/en-us/nuget/consume-packages/packagedownload-functionality)).
- `HydrateDotnetTemplates` resolves the restored nupkg path via `[MSBuild]::NormalizePath($(NuGetPackageRoot), …)` and passes it directly to the hydrate script.
- `hydrate-dotnet-templates.ps1` now takes a required `-LocalNupkgPath` (replacing the old `-NuGetSource`). `dotnet new install <local-path>` resolves from the on-disk file without consulting a feed; the same path is reused to extract per-template host descriptions, replacing the `$HiveRoot/.templateengine/packages/` fallback search.
- Drop the `DotnetTemplatesNuGetSource` MSBuild knob — configure alternate feeds via `NuGet.Config` (the standard restore mechanism).

## Verification

Local `dotnet pack` against a clean GPF:

1. `dotnet restore` downloaded `microsoft.azure.functions.worker.itemtemplates/4.0.5584` into the GPF.
2. `dotnet pack --no-restore` ran `dotnet new install C:\Nuget\…\.nupkg` (resolved locally, no feed access in the log).
3. 36 templates filtered, projected, and written to `tools/any/content/dotnet-templates.json`; `source.json` written.
4. Re-running `dotnet pack` is a no-op (incremental check honoured); deleting the `.hydrated` marker triggers a clean rehydrate.

Built nupkg layout unchanged.

### DIFF
--- a/eng/build/Workloads.Templates.DotNet.targets
+++ b/eng/build/Workloads.Templates.DotNet.targets
@@ -5,8 +5,10 @@
 
     Hydrates the workload payload (tools/any/content/dotnet-templates.json
     + tools/any/content/source.json) from the upstream
-    Microsoft.Azure.Functions.Worker.ItemTemplates NuGet package via
-    `dotnet new install` against a build-scoped template hive. See
+    Microsoft.Azure.Functions.Worker.ItemTemplates NuGet package. The
+    package is restored into the global packages folder by the
+    <PackageDownload> item in the csproj; this target consumes the
+    already-downloaded nupkg locally and does not reach the network. See
     proposed/templates-workload-spec.md §5.3 / §5.3.1 / §6.3 for the
     contract and the projection rules.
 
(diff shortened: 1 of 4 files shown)


jviau (approved):
https://dev.azure.com/azfunc/internal/_build/results?buildId=282526&view=results

Pack succeeded.

PR #5257, opened 2026-06-01
fix: auto-detect workload prerelease for prerelease CLI builds

liliankasem:
Builds on top of #5184; supersedes its customer-facing UX: the env var is now an override, not a requirement.

## The bug

Running `f5 run` on a prerelease bug-bash install hit:

```
Resolve worker...: No worker workload package 'Azure.Functions.Cli.Workloads.Workers.node' satisfies profile range '[3.13.0]'.
```

Two layered problems:

1. The built-in `flex` profile pins `node "[3.13.0]"` exact. The published package is `3.13.0-preview.1`. Per SemVer, `3.13.0-preview.1 < 3.13.0`. The previous customer-facing fix required `FUNC_CLI_WORKLOADS_PRERELEASE=true` to opt in, which customers should not have to set on a prerelease CLI.
2. Even with prerelease allowed, NuGet's `VersionRange.Satisfies` rejects prerelease candidates whose numeric version matches a stable range bound, so `[3.13.0]` still rejected `3.13.0-preview.1`.

## The fix

1. **Auto-enable prerelease when the CLI itself is prerelease.** `WorkloadCatalogOptionsSetup` now defaults `IncludePrerelease = true` when the running CLI's `InformationalVersion` contains a `-` (SemVer prerelease marker). The `FUNC_CLI_WORKLOADS_PRERELEASE` env var remains an explicit override in **both** directions (`true|1|yes|on` forces on; `false|0|no|off` forces off; anything else falls through to auto-detect).
2. **Prerelease-aware range satisfaction.** `WorkloadCatalog.SatisfiesRange` re-checks a prerelease candidate's numeric portion against the range when prerelease is allowed, so `[3.13.0]` now accepts `3.13.0-preview.1`.
3. **Tri-state `--prerelease`.** `setup`, `workload install`, `workload update`, and `workload search` now use `Option<bool?>`. Unset means "defer to the catalog options" (which itself is auto-detected). Explicit `--prerelease` or `--prerelease false` wins in both directions. Internal callers (start initialization, host workload resolver, worker installer) pass `null` so the catalog default applies.

## Tests

* New `WorkloadCatalogTests` coverage: range `[3.13.0]` resolves `3.13.0-preview.1` when prerelease is on; auto-detect picks up prerelease CLI version; env var overrides off; stable CLI defaults to off.
* Existing prerelease tests updated to reflect tri-state semantics.
* Full `Func.Tests` suite: 1140/1140 green.

## Docs

* README bug-bash section updated: env var is now an override of auto-detect, not a required setting.

## Notes

* The bug-bash installers still export `FUNC_CLI_WORKLOADS_PRERELEASE=true` for clarity, but it's now redundant on a prerelease build (the env var keeps working as an explicit override).
* The cherry-picked commit from #5184 keeps its original `Co-authored-by: Copilot` trailer (original author's commit); the follow-up commit does not add one.


src/Func/Workloads/Catalog/WorkloadCatalogOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+                return false;
+            }
+        }
+
+        // Auto-detect: a prerelease CLI build resolves prerelease workload
+        // packages by default. SemVer prerelease suffixes are delimited with
+        // '-' (e.g. "5.0.0-preview.1"), so the presence of a dash in the
+        // informational version is a reliable, build-agnostic signal.
+        string informational = _cliVersionProvider.InformationalVersion;
+        return !string.IsNullOrEmpty(informational) && informational.Contains('-');
```
jviau, line 70: Going to make a note to follow up on this. I think we can make this a compile time constant via an `#if` def

jviau: https://github.com/Azure/azure-functions-core-tools/pull/5290

PR #5259, opened 2026-06-01
Add `import logging` to Python function_app.py template

liliankasem:
The scaffolded `function_app.py` for the Python stack was missing `import logging`, so users adding a sample HTTP handler with `logging.info(...)` had to add the import themselves.

## Changes

- `src/Workloads/Stacks/Python/Templates/function_app.py`: add `import logging`.
- `src/Workloads/Stacks/Python/release_notes.md`: note the fix.

### DIFF
--- a/src/Workloads/Stacks/Python/Templates/function_app.py
+++ b/src/Workloads/Stacks/Python/Templates/function_app.py
@@ -1,3 +1,4 @@
 import azure.functions as func
+import logging
 
 app = func.FunctionApp(http_auth_level=func.AuthLevel.ANONYMOUS)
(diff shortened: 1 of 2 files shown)


soninaren (approved):

src/Workloads/Stacks/Python/Templates/function_app.py
```diff
@@ -1,3 +1,4 @@
 import azure.functions as func
+import logging
 
 app = func.FunctionApp(http_auth_level=func.AuthLevel.ANONYMOUS)
```
manvkaur, line 4:
```suggestion
app = func.FunctionApp()
```

liliankasem: @manvkaur can you open a PR for the format you expect python to be in please

PR #5260, opened 2026-06-01
fix(run): always tear down host process on every exit path

liliankasem:
## Problem

After `func run` exits, the spawned `Azure.Functions.Cli.Workloads.Host` process can sometimes survive and stay visible in Activity Monitor. It's hard to reproduce because the happy paths (clean source completion, Ctrl+C handled in the OCE catch) do clean up correctly; the host also has a stdin-EOF fallback that catches abrupt CLI exits.

The real leaks are on the CLI side, in the ownership chain between starting the host PID and the dashboard pipeline taking over.

## Leak paths fixed

1. **`HostProcessEventStream` was never deterministically disposed.** It implemented `IHostEventStream` (no `IAsyncDisposable`) and only disposed the wrapped `Process` from the `finally` of `CompleteWhenProcessExitsAsync`, which only runs after the process actually exits. So a stuck host kept itself alive.
2. **`StartCommand` had a window after init returned the started host stream and before the pipeline began.** Anything that threw in that window (renderer construction, log-file sink open, recording-renderer disposal, summary emit) skipped all teardown.  `await using ManagedAzuriteHandle? azurite` covered Azurite; nothing covered the host.
3. **`DashboardPipeline.RunAsync` only called `RequestShutdownAsync` from the `OperationCanceledException` catch.** A non-OCE exception (renderer crashing inside `OnEventAsync`, event sink IO error, etc.) propagated without ever signalling the host. The `finally` only disposed the renderer.

## What this PR does

- **`IHostEventStream` now extends `IAsyncDisposable`.** All five implementations (`HostProcessEventStream`, `CompositeHostEventStream`, `InMemoryHostEventStream`, `DemoEventSource`, `StartInitializationLogEventStream`) get the method.
- **`HostProcessEventStream.DisposeAsync`** is idempotent and routes through `RequestShutdownAsync` so disposal closes stdin, waits the shutdown timeout, then `KillTree`s on timeout. Best-effort: a thrown `RequestShutdownAsync` falls through to a direct `KillTree` + `Process.Dispose` so the OS handle is released no matter what.
- **`CompositeHostEventStream.DisposeAsync`** forwards to every child; failures from one child don't stop the rest, then surface as an `AggregateException`.
- **`StartCommand.ExecuteAsync`** `await using`s the host event stream the moment the init runner returns it, before any renderer / sink / pipeline construction. Now every throw path between init and pipeline completion tears the host PID down.
- **`DashboardPipeline.RunAsync`** always signals shutdown via `RequestShutdownAsync` in `finally`. Belt-and-braces over the StartCommand-level `await using`; idempotent, so the existing OCE path is unchanged behaviorally.
- **`DemoStartInitializationRunner.RunAsync`** disposes `state.EventStream` if any step (or `ToResult`) throws after the host has been started. Today `StartHostInitializationStep` is last, but this defends against future step additions.

## Tests

- New `HostProcessEventStreamTests`: `DisposeAsync_KillsBlockingProcessTree`, `DisposeAsync_IsIdempotent`.
- New `DashboardPipelineTests.RunAsync_WhenRendererThrows_StillShutsDownHostAndRethrows` — proves the non-OCE leak path is closed.
- New `CompositeHostEventStreamTests` covering forward-to-children and exception aggregation.
- All 1,175 `Func.Tests` pass. Release build clean (0 warnings, `CI=true`).

## Out of scope

- Initialization-runner Azurite leak when a step after `EnsureAzuriteInitializationStep` fails. Not currently triggerable (host start is last) and orthogonal to this fix; will be a follow-up.
- Host workload changes: stdin-EOF monitor and Ctrl+C handling already cover their side of the contract.

### DIFF
--- a/src/Func/Commands/Start/Host/HostProcessEventStream.cs
+++ b/src/Func/Commands/Start/Host/HostProcessEventStream.cs
@@ -19,6 +19,7 @@ internal sealed class HostProcessEventStream : IHostEventStream, IHostEventStrea
     private readonly Task _stdoutTask;
     private readonly Task _stderrTask;
     private int _shutdownRequested;
+    private int _disposed;
 
     public HostProcessEventStream(
         IHostProcess process,
@@ -80,6 +81,41 @@ internal sealed class HostProcessEventStream : IHostEventStream, IHostEventStrea
 
     public Task<int> WaitForExitAsync(CancellationToken cancellationToken)
         => _exitTask.WaitAsync(cancellationToken);
+
+    /// <summary>
+    /// Ensures the host process is shut down (gracefully if it cooperates,
+    /// killed otherwise) before returning. Idempotent and best-effort, so
+    /// it's safe to call from <c>await using</c> on every code path,
+    /// including ones that already drove the pipeline to completion.
+    /// </summary>
+    public async ValueTask DisposeAsync()
+    {
+        if (Interlocked.Exchange(ref _disposed, 1) != 0)
+        {
+            return;
+        }
+
+        try
+        {
+            await RequestShutdownAsync(CancellationToken.None);
+        }
+        catch
+        {
+            // RequestShutdownAsync is best-effort during disposal; the
+            // process is killed below regardless. Swallowing here keeps the
+            // outer failure (which is why we're disposing) primary.
+            try
+            {
+                _process.KillTree();
+            }
+            catch
+            {
+                // Process may already be gone; nothing more we can do.
+            }
+
+            await _process.DisposeAsync();
+        }
+    }
 
     private void WriteProcessStarted(HostProcessLaunchInfo launchInfo)
     {
--- a/src/Func/Commands/Start/Host/HostProcessEventStream.cs
+++ b/src/Func/Commands/Start/Host/HostProcessEventStream.cs
@@ -83,10 +83,7 @@ internal sealed class HostProcessEventStream : IHostEventStream, IHostEventStrea
         => _exitTask.WaitAsync(cancellationToken);
 
     /// <summary>
-    /// Ensures the host process is shut down (gracefully if it cooperates,
-    /// killed otherwise) before returning. Idempotent and best-effort, so
-    /// it's safe to call from <c>await using</c> on every code path,
-    /// including ones that already drove the pipeline to completion.
+    /// Idempotent, best-effort shutdown safe to call from <c>await using</c> on any exit path.
     /// </summary>
     public async ValueTask DisposeAsync()
     {
@@ -101,16 +98,14 @@ internal sealed class HostProcessEventStream : IHostEventStream, IHostEventStrea
         }
         catch
         {
-            // RequestShutdownAsync is best-effort during disposal; the
-            // process is killed below regardless. Swallowing here keeps the
-            // outer failure (which is why we're disposing) primary.
+            // Graceful shutdown failed; force-kill below.
             try
             {
                 _process.KillTree();
             }
             catch
             {
-                // Process may already be gone; nothing more we can do.
+                // Process already gone.
             }
 
             await _process.DisposeAsync();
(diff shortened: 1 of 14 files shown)


src/Func/Commands/Start/Host/HostProcessEventStream.cs
```diff
@@ -81,6 +82,36 @@ public async Task RequestShutdownAsync(CancellationToken cancellationToken)
...
+    /// </summary>
+    public async ValueTask DisposeAsync()
+    {
+        if (Interlocked.Exchange(ref _disposed, 1) != 0)
+        {
+            return;
+        }
+
+        try
+        {
```
satvu, line 96: Do we also want to do any cleanup/checks with `_exitTask` or `_process` here?

src/Func/Commands/Start/Host/HostProcessEventStream.cs
```diff
@@ -81,6 +82,36 @@ public async Task RequestShutdownAsync(CancellationToken cancellationToken)
...
+            await RequestShutdownAsync(CancellationToken.None);
```
satvu, line 97: Was suggested by an agent and did a little digging - some ASP.NET implementations particularly those related to process cleanup will have timeouts: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.hostoptions.shutdowntimeout?view=net-11.0-pp.

src/Func/Commands/Start/Host/HostProcessEventStream.cs
liliankasem: Good catch — the happy path is fine because `RequestShutdownAsync` already awaits `_exitTask` (which drains stdout/stderr and calls `_process.DisposeAsync()` in its finally). But the catch path skips `_exitTask`, leaving it as a fire-and-forget task that could throw unobserved after `KillTree`.

I'll add an `_exitTask` await (with a swallowed exception) in the catch block so we don't leak unobserved tasks.

src/Func/Commands/Start/Host/HostProcessEventStream.cs
liliankasem: `RequestShutdownAsync` already has a `_shutdownTimeout` + `KillTree` fallback built in (lines 72–76), which is the same pattern as ASP.NET's `HostOptions.ShutdownTimeout`. So `DisposeAsync` gets timeout behavior for free on the happy path by delegating to `RequestShutdownAsync`. No additional timeout needed here.

jviau (approved):

satvu (approved):
This looks solid on second review. The ownership chain is now clear and deterministic:

1. **Interface change** — IHostEventStream : IAsyncDisposable is the right forcing function to ensure every implementation has a cleanup path.

2. **HostProcessEventStream.DisposeAsync** — Good use of the interlocked guard for idempotency. The catch-path correctly falls through to KillTree + awaiting _exitTask so no tasks go unobserved and the OS handle is always released.

3. **StartCommand await-using** — Taking ownership immediately after init returns is the correct fix for the window between init and pipeline setup. The dashboardEventStream factory just wraps the same reference, so the lifetime is properly scoped.

4. **DashboardPipeline finally-block** — Belt-and-braces shutdown signal on all exit paths (not just OCE) closes the non-cancellation leak path. The best-effort catch is appropriate since StartCommand's dispose is the backstop.

5. **DemoStartInitializationRunner** — Defensive dispose on step failure protects against future step ordering changes without over-engineering.

6. **Tests** — Good coverage of the new paths: idempotency, force-kill on blocking process, renderer-throw still triggers shutdown, composite aggregation. The NeverEndingLifecycleEventStream refactor to use InMemoryHostEventStream is a nice simplification.

One minor observation (non-blocking): In HostProcessEventStream.DisposeAsync, if RequestShutdownAsync succeeds (graceful exit), the process resources are cleaned up via _exitTask's finally block. But if it throws, the catch path also awaits _exitTask which handles disposal. This asymmetry is fine since both paths converge, just noting it for future readers.

Ship it. 🚢

PR #5292, opened 2026-06-02
feat(update): scaffold `func update` command

liliankasem:
## Summary

Draft PR scaffolding a new `func update` top-level command that updates the
installed `func` CLI in place.

The full spec lives on the `docs` orphan branch:
**[`proposed/func-update.spec.md`](https://github.com/Azure/azure-functions-core-tools/blob/docs/proposed/func-update.spec.md)**.
Leave review comments on that file.

## What's in this PR

- `src/Func/Commands/Update/UpdateCommand.cs` — `FuncCliCommand + IBuiltInCommand` with `--prerelease`, `--version`, `-y` / `--yes`. `ExecuteAsync` throws a `GracefulException` pointing users at the install script until the pipeline lands.
- Registered in `src/Func/Hosting/BuiltInCommands.cs`.

## Follow-up commits

- [ ] `IReleaseFeed` + `GitHubReleaseFeed` + tests (SemVer-based stable/prerelease classification)
- [ ] `RidResolver` (shared with install scripts where possible)
- [ ] `IInstallMethodDetector` + tests (defer to npm/brew/choco/winget; in-place only for install-script deploys)
- [ ] `ICliInstaller` (extract / backup / swap / verify / rollback) + tests
- [ ] Wire pipeline into `UpdateCommand.ExecuteAsync` + integration tests
- [ ] `ICliUpdateNotifier` (background "update available" hint, 24h cache, package-manager-aware)
- [ ] README + release notes


src/Func/Commands/Update/UpdateCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System.CommandLine;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Hosting;
+
+namespace Azure.Functions.Cli.Commands.Update;
+
+/// <summary>
+/// Updates the installed func CLI in place. Modeled on
+/// <c>aspire update --self</c>: resolves the requested channel, downloads the
```
liliankasem: No need to mention aspire here, keep the comment simple

docs/func-update-spec.md
```diff
@@ -0,0 +1,152 @@
...
+
+- `UpdateCommandTests` — option parsing, channel selection precedence, confirm prompt, `--yes` enforcement under non-interactive.
+- `InstallMethodDetectorTests` — npm/brew/choco/winget/script paths, env-var overrides, unknown location.
+- `CliInstallerTests` — fake filesystem + fake archive: happy path, version-probe failure → rollback, write-permission failure → rollback, partial extract failure → rollback, backup cleanup.
+- `GitHubReleaseFeedTests` — `HttpMessageHandler` stub returning fixture payloads; stable vs prerelease filtering, asset matching by RID, hash field plumbing.
+
+## Out of scope (for this spec)
+
+- Background "update available" notifier on every command (Aspire's `ICliUpdateNotifier`). Track separately.
+- Auto-update on launch. Always explicit.
```
liliankasem: Wonder if this should be in scope for first iteration

jviau:
I am curious how this will even work? If the CLI is running, won't the file be locked on disk?

liliankasem:
> I am curious how this will even work? If the CLI is running, won't the file be locked on disk?

this is really only a Windows issue. On Linux/macOS the file isn't locked while it's executing, so we can replace it directly.

The flow handles this safely on both: download the archive to a temp dir, validate (SHA-256 + extract), and only then touch the install location. On Windows a running .exe can't be overwritten but can be renamed, so we'll follow the same pattern Aspire uses in aspire update: rename the current func.exe to func.exe.old.<timestamp>, then copy the validated binary from temp into place. The running process keeps executing from the renamed file, and the next func launch picks up the new one. Stale .old.* files get swept on the next update; if anything goes wrong post-rename we restore from the backup.

liliankasem:
Superseded by #5297 (branch was renamed to drop the duplicated `liliankasem/` prefix; closing this and reopening as a new PR was the only path since GitHub branch rename invalidates PR head refs).

PR #5294, opened 2026-06-02
Refactor install scripts

liliankasem:
Rewrites `install.sh` and `install.ps1` to mirror the structure and polish of the aspire.dev install scripts, and updates the README invocation examples to match.

## Changes

**Both scripts**
- Flag-based argument parsing (`--prerelease`, `--version`, `--install-path`, `--skip-path`, `--force`, `--dry-run`, `--verbose`, `--help`).
- Structured logging helpers (`say_*` in bash, `Write-Message -Level …` in pwsh).
- Hardened downloads: TLS 1.2 pinning, timeouts, retries, max-redirect caps, HEAD content-type probe to surface HTML error pages.
- Prefers `gh` CLI when available + authenticated to dodge the anonymous GitHub API 60/hr rate limit.
- GitHub Actions support via `$GITHUB_PATH`.
- Bash adds fish/sh profile detection alongside bash/zsh.
- Removed bug bash flag and persistent env-var blocks (no longer needed).

**Bugfix: `iex` no longer crashes the terminal**
On vnext, running `iex "& { $(irm …/install.ps1) } -Prerelease"` when func is already installed (or any other `exit` path) closes the whole terminal window, because `iex` runs the script in the host shell and a bare `exit` kills the host. This PR routes every exit through an `Exit-Script` helper that checks `$PSCommandPath`; when invoked via `iex` it `return`s instead of calling `exit`, leaving your shell alive.

**Post-install output**
Restyled to match aspire's format with section headers and dashed underlines:

```
func CLI v5.0.0-preview.2 successfully installed to: /Users/…/.azure-functions-cli/func
Successfully added func to $PATH in /Users/…/.zshrc

Telemetry
---------

The Azure Functions CLI collects usage data. …

Side-by-side notice
-------------------

Detected an existing 'func' at /opt/homebrew/bin/func, leaving it as the default.
Use 'func5' to invoke v5.

Reload shell
------------

To use the func CLI in new terminal sessions, restart your terminal or run:
  source /Users/…/.zshrc
```

Side-by-side and reload sections only render when relevant.

**README**
Install one-liners now use the cleaner `bash -s -- --flag` and `iex "& { $(irm …) } -Flag"` invocation patterns so flags pass cleanly through pipes. Adds a note that v5 is currently pre-release only and a callout about the `func5` side-by-side wrapper.


satvu (approved):

install.ps1
```diff
@@ -2,112 +2,262 @@
...
+    [string] $Source = "Azure/azure-functions-core-tools",
     [switch] $Prerelease,
     [switch] $Force,
-    [switch] $BugBash,
-    [string] $Source = 'Azure/azure-functions-core-tools',
-    [string] $InstallDir = (Join-Path $HOME '.azure-functions')
+    [switch] $SkipPath,
+    [switch] $KeepArchive,
+    [switch] $DryRun,
+    [switch] $Help
```
jviau, line 55: Powershell has built in `-?` semantics, do we need our own?

install.ps1
```diff
@@ -2,112 +2,262 @@
...
+    [string] $Version = "",
+    [string] $Source = "Azure/azure-functions-core-tools",
     [switch] $Prerelease,
     [switch] $Force,
-    [switch] $BugBash,
-    [string] $Source = 'Azure/azure-functions-core-tools',
-    [string] $InstallDir = (Join-Path $HOME '.azure-functions')
+    [switch] $SkipPath,
+    [switch] $KeepArchive,
+    [switch] $DryRun,
```
jviau, line 54: Powershell has built in `-WhatIf` mechanism

install.ps1
```diff
@@ -2,112 +2,262 @@
...
 
 .DESCRIPTION
-    Downloads and installs the func CLI from GitHub Releases.
-    Usage: irm https://aka.ms/func-cli/install.ps1 | iex
+    Downloads the func CLI from GitHub Releases, extracts it to the install dir,
+    drops a func5 wrapper for side-by-side use with Core Tools v4, and updates
+    PATH so 'func' / 'func5' are available in new terminals.
+
+    Piped usage:
+        iex "& { $(irm https://aka.ms/func-cli/install.ps1) }"
```
jviau, line 13: why the change in usage examples?

We originally had `irm https://aka.ms/func-cli/install.ps1 | iex`

install.ps1
```diff
@@ -2,112 +2,262 @@
...
+$Script:HeadRequestTimeoutSec = 60
+$Script:DefaultInstallDir = Join-Path $HOME '.azure-functions'
+
+# True when invoked as a file (pwsh -File ... or .\install.ps1), false when piped
+# into iex. We use this so a piped 'exit' doesn't kill the user's pwsh session.
+$Script:InvokedFromFile = -not [string]::IsNullOrEmpty($PSCommandPath)
+
+function Exit-Script {
+    param([int] $Code = 0)
+    if ($Script:InvokedFromFile) { exit $Code } else { return }
```
jviau, line 73: Should return the error code

```suggestion
    if ($Script:InvokedFromFile) { exit $Code } else { return $Code }
```

install.ps1
liliankasem: fixed~

install.ps1
liliankasem: Get-Help install.ps1 and install.ps1 -? both need the script to exist as a file PowerShell can resolve by path, so neither works through the documented irm | iex flow without the user saving it locally first

install.ps1
liliankasem: dropping -DryRun

PR #5298, opened 2026-06-02
feat(update): add IReleaseFeed + CdnReleaseFeed

liliankasem:
Stacked on #5297. PR #2 in the `func update` chain.

Adds the CDN-backed release feed the update pipeline will consume. No RID matching, no download, no command wiring; those land in subsequent PRs.

## What changed

Pivoted from GitHub Releases API to the **Azure Functions CDN** for version discovery and artifact downloads:

- **`CdnReleaseFeed`** fetches a lightweight version manifest (`version.json`) from `public/cli/v5/version.json` on the CDN.
- Manifest shape: `{"stable": "5.1.0", "preview": "5.2.0-preview.1"}`.
- Download URLs are constructed deterministically: `public/{version}/Azure.Functions.Cli.{rid}.{version}.zip`.
- `GetVersionAsync` verifies artifact existence with a `HEAD` request before returning a `Release`.

## Why the pivot

- **No rate limits** — CDN is unauthenticated and globally cached; GitHub API caps at 60 req/hr unauthenticated.
- **No auth required** — removes the `GITHUB_TOKEN` dependency and associated error paths.
- **Predictable URLs** — matches the existing v4 distribution pattern.
- **Single lightweight fetch** — one small JSON file vs paginating 100+ releases.

## Design notes

- **`IReleaseFeed`** interface is preserved — downstream consumers (`func update` command) are unchanged.
- **Classification** still follows the SemVer prerelease label: a version is stable iff it has no prerelease identifiers.
- **Error mapping**: 404 on manifest → `GracefulException` mentioning feed not configured; 5xx → retry-later message; malformed JSON → format-changed message.
- **`VersionManifest`** DTO uses source-generated `JsonSerializerContext` (AOT/trim-friendly).
- **`Release`** record simplified: removed `TagName`/`Assets`, added `DownloadUrl` (constructed from version + RID).

## Pipeline requirements (follow-up)

The release pipeline will need to:
1. Upload versioned artifacts to `public/{version}/Azure.Functions.Cli.{rid}.{version}.zip`
2. Update `public/cli/v5/version.json` **after** artifacts are uploaded (two-phase publish for atomicity)

part of #5333


src/Func/Update/GitHubReleaseFeed.cs
```diff
@@ -0,0 +1,177 @@
...
+        try
+        {
+            raw = await JsonSerializer.DeserializeAsync(
+                stream,
+                UpdateJsonContext.Default.GitHubReleaseArray,
+                cancellationToken);
+        }
+        catch (JsonException ex)
+        {
+            throw new GracefulException(
```
liliankasem: Does this need to be a GracefulException? is this something a user can act on? I don't think it is so we should rethink how we handle the exceptions here

src/Func/Update/GitHubReleaseFeed.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (response.IsSuccessStatusCode)
+        {
+            return;
+        }
+
+        if (response.StatusCode == HttpStatusCode.Forbidden && IsRateLimited(response))
+        {
+            throw new GracefulException(
+                "GitHub API rate limit reached. Set the GITHUB_TOKEN environment variable "
```
liliankasem: Do we really want to support this? Not sure how I feel about us asking users to set the GITHUB_TOKEN

src/Func/Update/GitHubReleaseFeed.cs
```diff
@@ -0,0 +1,183 @@
...
+            .FirstOrDefault();
+    }
+
+    public async Task<Release?> GetVersionAsync(SemVersion version, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(version);
+
+        IReadOnlyList<Release> releases = await FetchReleasesAsync(cancellationToken);
+
+        return releases.FirstOrDefault(r => SemVersion.PrecedenceComparer.Equals(r.Version, version));
```
satvu: Can we just look directly for a specific link instead of going through `FetchReleasesAsync`? That seems to rely on `ReleasesEndpoint` which only pulls the latest 100.

liliankasem: Yeah this is a open q/consideration: `Pagination. v1 fetches a single page of 100. The repo has ~50 releases historically. Open follow-up if we ever cross 100 and want GetVersionAsync to stay exhaustive.`

But I guess there is no reason we can't go through to the release link directly esp if we know the exact version - let me consider :)

src/Func/Update/Release.cs
```diff
@@ -0,0 +1,17 @@
...
+
+/// <summary>
+/// A published func CLI release as seen by the update pipeline. The version
+/// drives both channel classification (stable vs preview) and "newer than
+/// current" comparisons. The download URL is constructed from the CDN base
+/// and the version/RID.
+/// </summary>
+internal sealed record Release(
+    SemVersion Version,
+    bool IsPrerelease,
```
jviau: Is an explicit `IsPrerelease` field needed, or can it be computed from `SemVersion`? I think `SemVersion` has an `IsPrerelease` directly on it.

``` CSharp
internal sealed record Release(SemVersion Version, Uri DownloadUri)
{
    public bool IsPrerelease => Version.IsPrerelease;
}
```

src/Func/Update/Release.cs
```diff
@@ -0,0 +1,17 @@
...
+    string DownloadUrl);
```
jviau: Use `Uri` when representing urls when possible.
```suggestion
    Uri DownloadUrl);
```

src/Func/Update/VersionManifest.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Azure.Functions.Cli.Update;
+
+/// <summary>
+/// JSON shape of the CDN version manifest at
+/// <c>https://cdn.functions.azure.com/public/cli/v5/version.json</c>.
+/// Contains the latest stable and preview versions published to the feed.
+/// </summary>
+internal sealed record VersionManifest(
+    [property: JsonPropertyName("stable")] string? Stable,
```
jviau, line 14: I am kind of torn here. Part of me wants these to be typed directly as `SemVersion` and use a custom JsonConverter for this. But will that make identifying and exposing parse issues to the user more complicated?

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        return new Release(
+            Version: version,
+            IsPrerelease: version.IsPrerelease,
+            DownloadUrl: downloadUrl);
+    }
+
+    private async Task<VersionManifest> FetchManifestAsync(CancellationToken cancellationToken)
+    {
+        using HttpClient client = _httpClientFactory.CreateClient(HttpClientName);
```
jviau: It looks like you are already going for a named HTTP client with the base URI configured. You can inject `HttpClient` directly into the ctor in that case and use DI to configure it:

``` CSharp
services.AddHttpClient<IReleaseFeed, CdnReleaseFeed>(
    client =>
    {
        // can even get the base CDN uri from configuration
        client.BaseAddress = new Uri(<base-cdn-uri>);
    })
    .AddPolicyHandler(...); // add HTTP retry polices as needed
```

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            manifest = await JsonSerializer.DeserializeAsync(
+                stream,
+                UpdateJsonContext.Default.VersionManifest,
+                cancellationToken);
+        }
+        catch (JsonException ex)
+        {
+            throw new GracefulException(
+                "Could not parse the CDN version manifest. The format may have changed.",
```
jviau: I don't think we should mention "The format may have changed". I am also not sure `GracefulException` is the right choice. Just let the json exception bubble up. Maybe `InvalidOperationException` with a message of `"Error reading version manifest from '<uri>': invalid format"`

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -0,0 +1,148 @@
...
+                ex);
+        }
+
+        if (manifest is null)
+        {
+            throw new GracefulException(
```
jviau: Is this a user error though? This can also be the same exception message as above.

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -0,0 +1,148 @@
...
+    private static void EnsureSuccessOrThrowGraceful(HttpResponseMessage response)
+    {
+        if (response.IsSuccessStatusCode)
+        {
+            return;
+        }
+
+        if (response.StatusCode == HttpStatusCode.NotFound)
+        {
+            throw new GracefulException(
```
jviau: This doesn't seem like the correct place for `GracefulException` and these do not seem like user errors. These error messages are also not helpful for users. 

1. use `InvalidOperationException`, or some other more specific exception to this scenario
2. Throw/catch the HTTP exception (if any) and include it in the inner exception
3. Keep error messages all consistent for this flow: `"Error reading version manifest from '<uri>': <http status short name>"`

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -0,0 +1,148 @@
...
+    internal const string ManifestPath = "public/cli/v5/version.json";
+
+    private readonly IHttpClientFactory _httpClientFactory = httpClientFactory ?? throw new ArgumentNullException(nameof(httpClientFactory));
+    private readonly ILogger<CdnReleaseFeed> _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+    public async Task<Release?> GetLatestAsync(bool includePrerelease, CancellationToken cancellationToken)
+    {
+        VersionManifest manifest = await FetchManifestAsync(cancellationToken);
+
+        string? versionString = includePrerelease ? manifest.Preview : manifest.Stable;
```
jviau: I think we should only use pre-release iff the prerelease version is higher than stable.

src/Func/Update/IReleaseFeed.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Returns the latest release on the requested channel, or <c>null</c>
+    /// when no version is published for that channel.
+    /// </summary>
+    /// <param name="includePrerelease">
+    /// <c>false</c> to return the latest stable release;
+    /// <c>true</c> to return the latest preview release.
+    /// </param>
+    public Task<Release?> GetLatestAsync(bool includePrerelease, CancellationToken cancellationToken);
```
jviau: Do we want nullable here? Or should this be non-null and always throw if it cannot resolve the latest version.

src/Func/Update/IReleaseFeed.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Returns a <see cref="Release"/> for the given <paramref name="version"/>
+    /// if it exists on the CDN (i.e. the zip artifact is downloadable), or
+    /// <c>null</c> when the version is not available.
+    /// Used by <c>func update --version X.Y.Z</c>.
+    /// </summary>
+    public Task<Release?> GetVersionAsync(SemVersion version, CancellationToken cancellationToken);
```
jviau: Same here: consider using non-null and always throw if the request version is not found.

jviau (changes requested):
We should make sure that even when using specifies pre-release, we will only actually use the pre-release version if it is newer than stable. Otherwise, we still use stable.

I am also contemplating if `version.json` should just be an ordered array of published versions? We can identify the newest and stable vs preview by SemVer conventions. But if that is too much more complexity than the 2-field version.json in this PR today no worries.

src/Func/Update/VersionManifest.cs
liliankasem: Yeah let's just leave as is

liliankasem:
@jviau ready for second look!

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -17,75 +16,91 @@ namespace Azure.Functions.Cli.Update;
...
+        if (includePrerelease)
         {
-            return null;
+            // Pick whichever is higher: stable or preview.
+            SemVersion? stable = TryParseVersion(manifest.Stable, "stable");
+            SemVersion? preview = TryParseVersion(manifest.Preview, "preview");
+
+            SemVersion? best = (stable, preview) switch
+            {
+                (not null, not null) => SemVersion.PrecedenceComparer.Compare(preview, stable) > 0 ? preview : stable,
```
jviau: I think `PrecedenceComparer.Compare` can take in null, so the switch statement may not be needed. It will already correctly choose the non-null version when one is not null. And then the `best is null` check below will handle when both are null.

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -17,75 +16,91 @@ namespace Azure.Functions.Cli.Update;
...
-
-        if (!SemVersion.TryParse(versionString, SemVersionStyles.Strict, out SemVersion? version))
+        else
         {
-            _logger.LogWarning("CDN manifest contains unparseable version '{Version}' for channel {Channel}.", versionString, includePrerelease ? "preview" : "stable");
-            return null;
+            if (string.IsNullOrEmpty(manifest.Stable))
+            {
+                throw new InvalidOperationException(
+                    $"Error reading version manifest from '{ManifestPath}': stable version is missing");
```
jviau: nit: remove `stable` from message

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -17,75 +16,91 @@ namespace Azure.Functions.Cli.Update;
...
+            }
+
+            if (!SemVersion.TryParse(manifest.Stable, SemVersionStyles.Strict, out SemVersion? version))
+            {
+                throw new InvalidOperationException(
+                    $"Error reading version manifest from '{ManifestPath}': invalid stable version '{manifest.Stable}'");
```
jviau: nit: remove `stable` from message

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -17,75 +16,91 @@ namespace Azure.Functions.Cli.Update;
...
 
-        // Verify the artifact exists on CDN with a HEAD request.
-        string downloadUrl = BuildDownloadUrl(version);
+        Uri downloadUri = BuildDownloadUri(version);
 
-        using HttpClient client = _httpClientFactory.CreateClient(HttpClientName);
-        using var request = new HttpRequestMessage(HttpMethod.Head, downloadUrl);
-
-        using HttpResponseMessage response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+        using var request = new HttpRequestMessage(HttpMethod.Head, downloadUri);
```
jviau, line 73: The comment `// Verify the artifact exists on CDN with a HEAD request.` from your previous changes was nice. Can you add it back in? It helped me notice this isn't the actual download.

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -17,75 +16,91 @@ namespace Azure.Functions.Cli.Update;
...
     }
 
     private async Task<VersionManifest> FetchManifestAsync(CancellationToken cancellationToken)
     {
-        using HttpClient client = _httpClientFactory.CreateClient(HttpClientName);
-        using HttpResponseMessage response = await client.GetAsync(ManifestPath, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+        using HttpResponseMessage response = await _httpClient.GetAsync(ManifestPath, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
 
-        EnsureSuccessOrThrowGraceful(response);
+        if (!response.IsSuccessStatusCode)
```
jviau, line 99: Would it be helpful to throw the HTTP exception via `response.EnsureSuccessStatusCode()` and then wrap it in a catch block?

jviau (approved):

PR #5300, opened 2026-06-03
Detect language when adopting an existing project in func init

### LINKED ISSUE #5301: func init adopt mode doesn't persist stack.language for multi-language stacks
## Repro

Run `func init` in an existing pre-v5 project (has `host.json`, no `.func/config.json`):

```
> func init
· Wrote .func/config.json.

✓ Existing project adopted for dotnet.

> func new --list
Error: Cannot determine language for stack 'dotnet' in '/Users/.../bugbash/dotnetqs/http'.
`stack.language` is missing from .func/config.json — run func init to scaffold the project (which writes the language) before adding functions.
```

The written `.func/config.json` contains:

```json
{
  "stack": {
    "runtime": "dotnet"
  }
}
```

`stack.language` (e.g. `"c#"`) is missing.

## Expected

For multi-language stacks (dotnet C#/F#, node JS/TS), adoption should infer the language from the on-disk project shape (`.csproj`/`.fsproj`, `tsconfig.json`, etc.) and persist `stack.language` so subsequent commands like `func new --list` work without the user repeating `--language`.

## Actual

`InitCommand`'s adopt path skips language resolution entirely when `--language` isn't supplied, so the language is never written.

## Scope

Affects every multi-language stack:

- `dotnet` (C#, F#)
- `node` (JavaScript, TypeScript)

Single-language stacks (Python, Go) are unaffected since the stack runtime already implies the language.

liliankasem:
Resolves #5301

## Problem

When `func init` adopts a pre-existing project (host.json present, no `.func/config.json`), it wrote `stack.runtime` but not `stack.language`. For multi-language stacks (dotnet C#/F#, node JS/TS) this left the config incomplete, so subsequent commands errored:

```
> func init
✓ Existing project adopted for dotnet.

> func new --list
Error: Cannot determine language for stack 'dotnet' in '...'.
`stack.language` is missing from .func/config.json — run func init to scaffold the project (which writes the language) before adding functions.
```

## Fix

Reuse the existing `IFunctionsProjectResolver` (the same one `func start` / `func new` use) to fingerprint the directory during adoption, instead of standing up parallel init-only detection rules.

- Added a `Language` property on `FunctionsProject` (default `null`).
- `DotNetSourceProject.Language` is derived from the project file extension (`.csproj` → C#, `.fsproj` → F#).
- `NodeFunctionsProject.Language` looks at `tsconfig.json` / top-level `*.ts` for TypeScript and otherwise defaults to JavaScript (the factory has already vouched that it's a Node project).
- Single-language stacks (Python, Go) and `DotNetOutputProject` leave `Language` as `null`; adoption then writes the stack runtime without a language entry, matching the prior behavior.
- `InitCommand` injects `IFunctionsProjectResolver`. In adopt mode with no `--language`, it asks the resolver for the project, reads `Language`, and feeds it through the existing language-resolution path. `--language` still wins when supplied.

## Tests

- 3 new `InitCommand` adopt tests covering: resolver-supplied language is persisted, resolver returns null language → omitted, explicit `--language` wins.
- New `DotNetSourceProject.Language` theory (`.csproj` / `.fsproj` / case-insensitive).
- 4 new `NodeFunctionsProject.Language` tests covering tsconfig, top-level `.ts`, package.json-only, and the bare default.
- All 1289 tests pass; clean build, 0 warnings.

### DIFF
--- a/src/Abstractions/Projects/FunctionsProject.cs
+++ b/src/Abstractions/Projects/FunctionsProject.cs
@@ -17,6 +17,15 @@ public abstract class FunctionsProject
 
     public abstract string StackDisplayName { get; }
 
+    /// <summary>
+    /// The canonical display name of the language used by this project (e.g.
+    /// "C#", "F#", "JavaScript", "TypeScript"), or <c>null</c> when the
+    /// project layout doesn't carry enough information to tell. Single-language
+    /// stacks (Python, Go) leave this <c>null</c> because the stack already
+    /// implies the language.
+    /// </summary>
+    public virtual string? Language => null;
+
     public abstract bool SupportsExtensionBundles { get; }
 
     public abstract FunctionsWorkerReference WorkerReference { get; }
--- a/src/Abstractions/Projects/FunctionsProject.cs
+++ b/src/Abstractions/Projects/FunctionsProject.cs
@@ -18,11 +18,8 @@ public abstract class FunctionsProject
     public abstract string StackDisplayName { get; }
 
     /// <summary>
-    /// The canonical display name of the language used by this project (e.g.
-    /// "C#", "F#", "JavaScript", "TypeScript"), or <c>null</c> when the
-    /// project layout doesn't carry enough information to tell. Single-language
-    /// stacks (Python, Go) leave this <c>null</c> because the stack already
-    /// implies the language.
+    /// Display name of the project's language (e.g. "C#", "TypeScript"), or
+    /// <c>null</c> for single-language stacks where the stack implies it.
     /// </summary>
     public virtual string? Language => null;
 
--- a/src/Workloads/Stacks/Node/NodeProjectFactory.cs
+++ b/src/Workloads/Stacks/Node/NodeProjectFactory.cs
@@ -40,11 +40,8 @@ internal sealed class NodeProjectFactory : IFunctionsProjectFactory
         return Task.FromResult(Created(project, reason));
     }
 
-    // Returns the human-readable detection reason (null when nothing matched)
-    // alongside the language ("TypeScript" if a tsconfig.json or top-level
-    // *.ts file is present, otherwise "JavaScript"). The language is
-    // unconditional once we've decided it's a Node project: even a bare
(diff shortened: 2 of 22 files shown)


src/Abstractions/Projects/FunctionsProject.cs
```diff
@@ -17,6 +17,15 @@ public abstract class FunctionsProject
 
     public abstract string StackDisplayName { get; }
 
+    /// <summary>
+    /// The canonical display name of the language used by this project (e.g.
+    /// "C#", "F#", "JavaScript", "TypeScript"), or <c>null</c> when the
+    /// project layout doesn't carry enough information to tell. Single-language
+    /// stacks (Python, Go) leave this <c>null</c> because the stack already
+    /// implies the language.
+    /// </summary>
```
liliankasem, line 23: comment is too verbose, not needed

src/Workloads/Stacks/Node/NodeProjectFactory.cs
```diff
@@ -30,40 +30,48 @@ public Task<ProjectCreationResult> TryCreateProjectAsync(ProjectCreationContext
...
-    private static string? TryGetReason(DirectoryInfo workingDirectory)
+    // Language defaults to JavaScript so a Node directory with only package.json
+    // (no .ts files) still classifies cleanly.
+    private static (string? Reason, string Language) Fingerprint(DirectoryInfo workingDirectory)
     {
-        string packageJsonPath = Path.Combine(workingDirectory.FullName, "package.json");
+        string root = workingDirectory.FullName;
+        bool hasTsConfig = File.Exists(Path.Combine(root, "tsconfig.json"));
+        bool hasTopLevelTs = Directory.EnumerateFiles(root, "*.ts", SearchOption.TopDirectoryOnly).Any();
+        string language = hasTsConfig || hasTopLevelTs ? "TypeScript" : "JavaScript";
```
satvu: can we short-circuit running `hasTopLevelTs` if `hasTsConfig` is already true to avoid searching the directory?

satvu (approved):

PR #5306, opened 2026-06-03
fix(func new): render resolution failures exactly once

### LINKED ISSUE #5304: func new prints "Cannot determine language" error 3 times (v5)
## Summary

When a v5 (`vnext`) project's `.func/config.json` is missing `stack.language` on a multi-language stack (e.g. `dotnet`), running `func5 new -t Http` prints the **same** "Cannot determine language" error message **three times** in a single invocation, instead of once.

## Repro

1. In a Functions project where `.func/config.json` has `stack.runtime` but no `stack.language` (for example, an existing dotnet project initialized with today's `vnext` `func init`, where the missing-language adoption gap is the subject of #5300).
2. Run:

   ```
   func5 new -t Http
   ```

3. Observe the same error printed three times:

   ```
   Error: Cannot determine language for stack 'dotnet' in '<path>'.
   `stack.language` is missing from .func/config.json — run func init to scaffold the project (which writes the language) before adding functions.
   Error: Cannot determine language for stack 'dotnet' in '<path>'.
   `stack.language` is missing from .func/config.json — …
   Error: Cannot determine language for stack 'dotnet' in '<path>'.
   `stack.language` is missing from .func/config.json — …
   ```

## Expected

The error should be rendered **exactly once** per invocation, then the command should exit 1.

## Root cause notes

- The error is emitted by `NewCommandRenderer.RenderMissingLanguage` (`src/Func/Templates/NewCommandRenderer.cs`), called from inside `NewCommandRunner.ResolveContextAsync` (`src/Func/Templates/NewCommandRunner.cs`).
- `ResolveContextAsync` mixes context resolution and user-facing error rendering: every code path that returns `null` also calls a `_renderer.RenderXxx(...)` first. Any caller that runs the resolver more than once (directly or indirectly through pre-parse / option-hydration paths) therefore renders the same error multiple times.
- The fix is to split resolution from rendering: `ResolveContextAsync` returns a result that describes the failure (kind + metadata), and the single execution entry-point (`ExecuteAsync` / `ListAsync`) renders it once.

## Related

- PR #5300 fixes the upstream cause (`func init` not persisting `stack.language` on adoption) but does not fix the duplicate-print bug surfaced when the language is genuinely missing.

liliankasem:
Resolves #5304

## Problem

When `.func/config.json` is missing `stack.language` on a multi-language stack (e.g. `dotnet`), running `func5 new -t Http` printed the same "Cannot determine language" error three times in a single invocation.

The root cause is that `NewCommandRunner.ResolveContextAsync` mixed resolution with rendering: every failure branch called `_renderer.RenderXxx(...)` (or `_interaction.WriteError`) inline and then returned `null`. Any caller that runs the resolver more than once (directly, or through a pre-parse / hydration pass) therefore prints the same error N times.

## Fix

Split resolution from rendering:

- `ResolveContextAsync` now returns a `ResolutionOutcome` (success carrying `ResolvedContext`, or a `ResolutionFailure` record carrying the failure kind + minimum metadata).
- All inline `_renderer` / `_interaction` calls inside the resolver are removed.
- `ExecuteAsync` and `ListAsync` each render the failure exactly once via a new `RenderResolutionFailure(failure)` helper.

This makes the duplication structurally impossible from a single entry-point call. Future pre-parse / hydration callers can now run the resolver silently and let the actual execute/list call render once.

## Tests

- New `NewCommandRunnerTests` exercises the missing-language path via both `ExecuteAsync` and `ListAsync` and asserts the error is rendered exactly once.
- Full suite: 1151 tests pass, clean build (0 warnings).

## Out of scope

- The upstream `func init` adoption gap (where the language was never persisted to begin with) is being fixed in #5300.

### DIFF
--- a/src/Func/Templates/NewCommandRunner.cs
+++ b/src/Func/Templates/NewCommandRunner.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.CommandLine;
+using System.Diagnostics;
 using Azure.Functions.Cli.Bundles;
 using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Configuration;
@@ -411,6 +412,14 @@ internal sealed class NewCommandRunner
             case ResolutionFailureKind.MissingLanguage:
                 _renderer.RenderMissingLanguage(failure.Stack!, failure.ProjectPath!);
                 break;
+
+            default:
+                // Guard against silently swallowing a future ResolutionFailureKind.
+                // Without this, a new enum value would return null from the caller
+                // with no user-visible error, reintroducing the class of bug this
+                // refactor exists to prevent.
+                throw new UnreachableException(
+                    $"Unhandled {nameof(ResolutionFailureKind)}: {failure.Kind}.");
         }
     }
 
--- a/src/Func/Templates/NewCommandRunner.cs
+++ b/src/Func/Templates/NewCommandRunner.cs
@@ -78,11 +78,14 @@ internal sealed class NewCommandRunner
     {
         ArgumentNullException.ThrowIfNull(invocation);
 
-        ResolvedContext? resolved = await ResolveContextAsync(invocation, cancellationToken);
-        if (resolved is null)
+        ResolutionOutcome outcome = await ResolveContextAsync(invocation, cancellationToken);
+        if (outcome.Failure is { } executeFailure)
         {
+            RenderResolutionFailure(executeFailure);
             return 1;
         }
+
+        ResolvedContext resolved = outcome.Context!;
 
         // Steps 11a / 11b: extension-bundle presence + min-bundle gate.
         // DotNet doesn't ship a templates-workload.json, so the gate is a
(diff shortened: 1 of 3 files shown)


src/Func/Templates/NewCommandRunner.cs
```diff
@@ -344,17 +333,85 @@ await _profileResolver.ResolveAsync(
...
+
+    /// <summary>
+    /// Renders a single <see cref="ResolutionFailure"/>. Centralizing the
+    /// render here keeps <see cref="ResolveContextAsync"/> side-effect free:
+    /// each <see cref="ExecuteAsync"/> / <see cref="ListAsync"/> entry-point
+    /// renders the failure at most once, regardless of how many internal
+    /// resolution passes a caller adds (pre-parse / hydration paths run the
+    /// same resolver but consume the outcome silently).
+    /// </summary>
+    private void RenderResolutionFailure(ResolutionFailure failure)
```
satvu, line 360: no default case?

liliankasem: Good catch. A statement switch over an enum doesn't get exhaustiveness checking, so adding a new `ResolutionFailureKind` later would silently render nothing (caller still returns null, user sees no error, which is exactly the class of bug this refactor exists to prevent). Added a `default` arm that throws `UnreachableException` in 968477f so future additions fail loudly.

satvu (approved):

PR #5314, opened 2026-06-04
Add preview-only func5 alias nudge

liliankasem:
While v5 is in preview, side-by-side installs with v4 leave `func` pointing at v4, so the docs/output hints that say "run func ..." can be misleading. Rather than rewrite copy that will be correct at GA, this adds a small preview-only banner the CLI prints next to those hints:

> Preview: if you also have Core Tools (v4) installed, you may need to invoke the new CLI via `func5` instead.

### Behavior
- Gated on `IsPrerelease && IsInteractive`.
- Shown when `exitCode != 0` or `commandName == "help"` (covers bare `func`, explicit `func help`, and any failed invocation).
- Silent on successful commands and in non-interactive/CI contexts.
- Best-effort: render failures are swallowed so the nudge never affects exit code.

### Implementation
- New `FuncAliasNudge` in `src/Func/Hosting/`, registered as a singleton in `CliHostFactory`.
- `Program.cs` captures it from DI inside the host scope and calls `TryPrint(exitCode, commandName)` after `PrintVersionNotice`.
- 9 unit tests covering each gate and the success/failure/help paths.

resolves #


src/Func/Hosting/FuncAliasNudge.cs
```diff
@@ -0,0 +1,111 @@
...
+    // the version string and nothing else.
+    private static readonly HashSet<string> _allowedRootCommands =
+        new(StringComparer.OrdinalIgnoreCase)
+        {
+            "help",
+            "init",
+            "new",
+            "profile",
+            "setup",
+            "start",
```
liliankasem: run

liliankasem:
Example

```bash
> ./../out/bin/Func/debug/func new -t Http
Error: Template 'Http' was not found for this project's stack.
Run func new --list to see available templates.

Preview: if you also have Core Tools (v4) installed, you may need to invoke the new CLI via 'func5' instead.
```

satvu (approved):

PR #5317, opened 2026-06-04
Convert release notes to Keep a Changelog format

liliankasem:
Migrates every `release_notes.md` in the repo to the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) model so each package has a consistent, scannable history with an `## [Unreleased]` section ready for new entries.

### Changes

- All 14 `release_notes.md` files (`src/Func`, `src/Abstractions`, and every workload under `src/Workloads/**`) updated with:
  - Standard header explaining the format + SemVer link.
  - `## [Unreleased]` section.
  - Existing entries preserved, grouped under `### Added` / `### Changed` / `### Fixed`.
  - Seed entry added to previously empty/stub files.
- `docs/building-a-workload.md` and `.github/skills/create-workload/SKILL.md` examples updated to match.

MSBuild integration is unchanged: `eng/build/Release.props` still points `ReleaseNotesFile` at `release_notes.md` and `Release.targets` inlines the file verbatim into `PackageReleaseNotes`. Verified by packing `Abstractions` successfully.

### DIFF
--- a/.github/skills/create-workload/SKILL.md
+++ b/.github/skills/create-workload/SKILL.md
@@ -77,7 +77,16 @@ Create the following files:
   ```markdown
   # Azure.Functions.Cli.Workloads.<Name>
 
-  ## 1.0.0-preview.1
+  All notable changes to this package are documented here.
+
+  The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+  and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+  ## [Unreleased]
+
+  ## [1.0.0-preview.1]
+
+  ### Added
 
   - Initial scaffold of the <Name> workload (entry point + stub project initializer).
   ```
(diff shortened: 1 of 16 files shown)


jviau:
Right now, release_notes.md get incorporated into the nupkg. Having the entire changelog there won't be ideal. How do we have both? It seems like we would need separate files: release_notes.md for the current changes, and `CHANGELOG.md` for this change. Or do we have build-time parse `CHANGELOG.md` and pull in only the unreleased changes.

liliankasem:
> Right now, release_notes.md get incorporated into the nupkg. Having the entire changelog there won't be ideal. How do we have both? It seems like we would need separate files: release_notes.md for the current changes, and `CHANGELOG.md` for this change. Or do we have build-time parse `CHANGELOG.md` and pull in only the unreleased changes.

That's a good call out, let me think on this a bit

fabiocav (changes requested):
Can we discuss this to keep things consistent across repos? Not opposed to the change, but we need to discuss impact to dependencies on the current format and how we roll this across all repos if changing.

liliankasem:
I agree that this is worth a team conversation, I am going to close this for now and create an issue to track if this is something we want to do.

@fabiocav - I created an issue here #5411

PR #5365, opened 2026-06-10
build: infer bundle channel for templates workloads

liliankasem:
Applies the [#5364](https://github.com/Azure/azure-functions-core-tools/pull/5364) pattern (channel inferred from properties, per-channel defaults, hash-based invalidation) to the Node and Python templates workloads.

Drops the separate `$(TemplatesChannel)` MSBuild property and reuses `$(BundleChannel)`, so a single channel concept drives both the bundles workload and any workload that snapshots from a bundle. The runtime `TemplatesChannelMapper` (CLI-side) is a different concept and is intentionally left alone.

### Changes

- **`src/Workloads/Templates/Directory.Version.props`** — infers `$(BundleChannel)` from `BUILD_SOURCEBRANCHNAME` when `PublicReleaseTag` is set (same regex as bundles); defaults `$(SourceBundleVersion)` to the channel-resolved `$(BundleVersion)` from the shared `Workloads.Templates.SourceBundleVersion.props`. `$(VersionPrefix)` keeps the existing `1.0.1` / `1.0.1-preview` / `1.0.1-experimental` shape.
- **`eng/build/Workloads.Templates.targets`** — drops the duplicate `_TplCdnId` block and reuses `$(_BundleCdnId)` from the shared props. Splits staging setup into `_TemplatesProperties` (always runs, computes per-channel staging path) and `_DownloadTemplatesProperties` (CDN-only, computes URL and writes hash files capturing URL + `IncludeV1/V2Templates` flags + `TemplatesStackLanguageFilter`). All staging moves under `obj/<config>/<tfm>/templates/$(BundleChannel)/` so channel-switching builds don't collide.
- Doc/comment updates in the Node csproj and both `README.DEV.md` files.

### Verification

- `dotnet pack` for Node templates (static mode) stable and preview channels — both produce correctly versioned nupkgs and filter against the right bundle.
- `dotnet pack` for Python templates (cdn mode) stable channel — downloads from the stable bundle and packs cleanly.
- Existing test suite: 1192/1192 passing.

resolves no open issue.

### DIFF
--- a/src/Workloads/Templates/Directory.Version.props
+++ b/src/Workloads/Templates/Directory.Version.props
@@ -6,7 +6,7 @@
     Directory.Version.props because it does not snapshot from an
     extension bundle (see proposed/templates-workload-spec.md §4.4.3).
 
-      $(TemplatesVersion)         workload pkg version. Bare three-part SemVer.
+      $(VersionPrefix)            workload pkg version. Bare three-part SemVer.
       $(BundleChannel)            stable | preview | experimental. Single
                                   channel concept shared with the bundles
                                   workload. Selects the source bundle id at
@@ -29,10 +29,10 @@
                                   so the three storage locations cannot
                                   drift (Templates Workload Spec §4.4.4).
 
-    Derived $(VersionPrefix):
-      stable        -> 1.0.0
-      preview       -> 1.0.0-preview
-      experimental  -> 1.0.0-experimental
+    Derived workload pkg version:
+      stable        -> 1.0.1
+      preview       -> 1.0.1-preview
+      experimental  -> 1.0.1-experimental
   -->
 
   <PropertyGroup Condition="'$(BundleChannel)' == '' AND '$(PublicReleaseTag)' != ''">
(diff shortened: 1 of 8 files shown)


src/Workloads/Templates/Directory.Version.props
```diff
@@ -31,21 +35,21 @@
...
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(TemplatesChannel)' == 'stable'">
-    <VersionPrefix>$(TemplatesVersion)</VersionPrefix>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(TemplatesChannel)' != 'stable'">
-    <VersionPrefix>$(TemplatesVersion)-$(TemplatesChannel)</VersionPrefix>
+  <PropertyGroup>
+    <VersionPrefix Condition="'$(BundleChannel)' == 'stable'">$(TemplatesVersion)</VersionPrefix>
```
jviau: This looks wrong from before these changes. VersionPrefix is being overloaded with the hyphen. Use `VersionSuffix` to incorporate `BundleChannel`, just like the bundles PR I made


``` xml
<!--

Do we even need TemplatesVersion defined separately here?
Or can we just set the `1.0.1` directly here and delete TemplatesVersion

-->
<VersionPrefix>$(TemplatesVersion)</VersionPrefix>
<VersionSuffix Condition="'$(BundleChannel)' != 'stable'">$(BundleChannel)</VersionSuffix>
```

liliankasem: Good call on both. Switched to `VersionSuffix=$(BundleChannel)` and inlined the pkg version directly into `VersionPrefix` here and in the DotNet templates props (it had the same indirection). Final shape:

```xml
<VersionPrefix>1.0.1</VersionPrefix>
<VersionSuffix Condition="'$(BundleChannel)' != 'stable'">$(BundleChannel)</VersionSuffix>
```

Release tags still produce `1.0.1` / `1.0.1-preview` / `1.0.1-experimental`. Pushed in 3b38c6756.

liliankasem:
@jviau @soninaren thoughts on latest?

jviau (approved):
LGTM -- I guess templates targets already had all the inputs/outputs and per-channel folder structures considered in its targets.

PR #5378, opened 2026-06-11
Fall back to stable templates when project bundle channel workload is missing

### LINKED ISSUE #5369: `func new` should warn and fall back to stable templates when bundle channel workload is missing
## Problem

When running `func new`, if the project's bundle channel (e.g. `preview`, `experimental`) doesn't have a matching installed templates workload, the command hard-fails:

```
>> func5 new
Error: No installed templates workload matches this project's bundle channel (Microsoft.Azure.Functions.ExtensionBundle.Preview -> channel 'preview').
Install one with: func workload install Azure.Functions.Cli.Workloads.Templates.Python --version <version>-preview.1.
```

This is too strict. A stable templates workload is often perfectly usable, and breaking `func new` entirely blocks the user from making progress.

## Expected behaviour

Channel resolution should degrade gracefully:

1. **Preferred:** use the templates workload that matches the project's bundle channel (e.g. `preview`, `experimental`).
2. **If no channel-matched workload is installed but a stable one is:** fall back to the stable templates workload and emit a **warning** explaining:
   - Their bundle channel is `<channel>` but templates are from the `stable` channel.
   - Behaviour/templates may differ from what their bundles ship.
   - Suggest installing the matching workload: `func workload install Azure.Functions.Cli.Workloads.Templates.<Lang> --version <version>-<channel>.1`.
3. **If no templates workload is installed at all:** then hard-fail with the current install instruction.

## Why

- Channel-mismatched templates are usually still functional; the user gets a usable scaffold plus an informational nudge.
- Matches the spirit of the CLI: tell the user what to do next, don't block them when a reasonable fallback exists.
- Avoids forcing every preview/experimental bundle user to install a separate workload just to run `func new`.

## Repro

1. Create/open a Functions project whose bundle is `Microsoft.Azure.Functions.ExtensionBundle.Preview` (channel `preview`).
2. Install only the stable Python templates workload.
3. Run `func new`. Today: hard error. Expected: warning + scaffold using stable templates.

liliankasem:
Resolves #5369.

When `func new` runs in a project whose extension bundle is `Preview` or `Experimental` and no channel-matched templates workload is installed, the command used to hard-fail. This change makes the resolution degrade gracefully:

1. Prefer the channel-matched templates workload.
2. If none is installed and a stable one is, fall back to it and emit a warning explaining the mismatch plus the `func workload install ... --version <version>-<channel>.1` hint.
3. If nothing is installed at all, surface the existing hard-fail message unchanged.

The warning fires from `Execute` / `List` only, so pre-parse hydration paths stay silent and we don't double-render.

### Tests
- `ListAsync_ProjectChannelMissingWorkload_FallsBackToStableAndWarns`
- `ListAsync_ProjectChannelMissingAndNoStable_StillHardFails`
- Existing template tests still green (77/77).

### DIFF
--- a/src/Func/Commands/Workload/WorkloadSearchCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadSearchCommand.cs
@@ -165,7 +165,7 @@ internal sealed class WorkloadSearchCommand : FuncCliCommand
 
         _interaction.WriteHint(countLine);
         _interaction.WriteHint("Run 'func workload install <alias | package id>' to install one (stable by default).");
-        _interaction.WriteHint("When several versions or channels exist, pass --version (-v) to install a specific version.");
+        _interaction.WriteHint("When several versions or channels exist, pass --version (-v) to install a specific one.");
     }
 
     private static string DisplayNameOrPackageId(CatalogSearchResult result)
--- a/src/Func/Commands/Workload/WorkloadSearchCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadSearchCommand.cs
@@ -164,7 +164,8 @@ internal sealed class WorkloadSearchCommand : FuncCliCommand
         }
 
         _interaction.WriteHint(countLine);
-        _interaction.WriteHint("Run 'func workload install <alias>' to install one.");
+        _interaction.WriteHint("Run 'func workload install <alias | package id>' to install one (stable by default).");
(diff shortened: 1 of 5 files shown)


liliankasem:
Screenshot 2026-06-12 at 12 49 27 PM

satvu (approved):

PR #5379, opened 2026-06-11
fix(workload): show one row per channel in `workload search` --prerelease

### LINKED ISSUE #5370: `func workload search` only shows highest version per package, hides other channels
## Problem

`func workload search` collapses results to one row per package (the highest version), so users only see the latest channel's build. Channels like `stable` are hidden behind newer `preview`/`experimental` versions.

## Repro

```
> func5 workload search bundle
Including prerelease workload versions.
Extension Bundles
Version: 4.42.0-preview.1
Package ID: azure.functions.cli.workloads.extensionbundles
Alias: bundles
...
Showing 3 results.
```

Only the `4.42.0-preview.1` bundle row is shown. The stable `4.35.0` bundle is not listed at all.

## Expected

Search should surface the latest version **per channel** for each package, so users can see and install the channel they want. For example:

```
Extension Bundles
Version: 4.42.0-preview.1
Package ID: azure.functions.cli.workloads.extensionbundles
Alias: bundles
...

Extension Bundles
Version: 4.35.0
Package ID: azure.functions.cli.workloads.extensionbundles
Alias: bundles
...
```

Same package, one row per channel (e.g. `stable`, `preview`, `experimental`), each showing that channel's latest version. This way users can discover and `func workload install bundles --version 4.35.0` if they want the stable channel.

## Notes

- Tightly related to #5369 (channel-mismatch fallback for `func new`). Users need a way to discover the channel-matched workload to install; today search doesn't show it if a newer prerelease exists.
- Should still respect `--prerelease` / "Including prerelease workload versions." filtering: when prerelease is off, only stable-channel rows appear.

liliankasem:
resolves #5370

## Problem
`func workload search` collapses results to one row per package (the highest version returned by the V3 search endpoint), so users only ever see the latest channel's build. With `--prerelease`, the stable row was hidden behind a newer `-preview` / `-experimental` version, and there was no way to discover the right `--version` for a non-default channel.

## Fix
When `--prerelease` is on, expand each search hit into one row per release channel by listing all versions of the package and picking the latest per channel. A `Channel:` field is rendered on each card (and surfaced in `--json` as `channel`) so it's obvious which row maps to which install target.

Channels are derived from the prerelease label: no label → `stable`, otherwise the first dot-separated label segment (so `4.42.0-preview.1` → `preview`, `5.0.0-experimental.2` → `experimental`, `1.0.0-rc.1` → `rc`, etc.). Per the issue's follow-up note, the command isn't restricted to a hard-coded `preview`/`experimental` allow-list; any post-dash label is its own channel.

Behaviour without `--prerelease` is unchanged: one stable row per package, no extra version listing call.

Other notes:
- `IWorkloadCatalog.ListVersionsAsync` is a new primitive that delegates to the existing source-client implementation.
- The "more may be available, refine your query" hint is now anchored to the package count returned by the catalog (the actual paging unit), not the post-expansion row count.
- Version listing per package runs in parallel (`Task.WhenAll`). A listing failure for one package falls back to the original search row instead of dropping it.

## Tests
- New unit coverage for channel expansion (latest-per-channel selection, no expansion without `--prerelease`, fallback when version listing fails, channel parsing).
- All 1201 existing tests pass.

### DIFF
--- a/src/Func/Commands/Workload/WorkloadSearchCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadSearchCommand.cs
@@ -6,6 +6,7 @@ using Azure.Functions.Cli.Common;
 using Azure.Functions.Cli.Console;
 using Azure.Functions.Cli.Workloads.Catalog;
 using Microsoft.Extensions.Options;
+using NuGet.Versioning;
 
 namespace Azure.Functions.Cli.Commands.Workload;
 
@@ -24,6 +25,11 @@ internal sealed class WorkloadSearchCommand : FuncCliCommand
     // workload (the only kind `func init --stack` can target). Mirrors the
     // `kind:workload` PackageTag stack csprojs emit.
     private const string StackKind = "workload";
+
+    // Label used for non-prerelease versions when grouping search results by
+    // release channel. Versions without a prerelease label all collapse to
+    // this single bucket so the stable channel renders as one row.
+    private const string StableChannel = "stable";
 
     private readonly IInteractionService _interaction;
     private readonly IWorkloadCatalog _catalog;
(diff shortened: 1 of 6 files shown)


liliankasem:
Screenshot 2026-06-12 at 12 24 26 PM
Screenshot 2026-06-12 at 12 24 45 PM

satvu (approved):

PR #5385, opened 2026-06-16
Default func setup to stable bundle; honor host.json channel

### LINKED ISSUE #5374: `func setup` should install the latest stable bundle by default, only pick preview/experimental when host.json opts in
## Summary

`func setup` currently doesn't differentiate between stable, preview, and experimental extension bundle channels. It should default to the **latest stable** bundle, and only install a preview/experimental bundle when the project's `host.json` explicitly points at that channel.

### `func setup`

- **No `host.json` present** (or `host.json` doesn't specify an extension bundle channel): install the **latest stable** bundle.
- **`host.json` present and pointing to a preview/experimental channel**: install the matching preview/experimental bundle.
- **`host.json` present and pointing to stable** (or no bundle section): install the latest stable bundle.

### `func init -s <stack> -c preview` (or `-c experimental`)

If the customer later runs `func init` with a preview/experimental channel and the corresponding preview/experimental bundle is **not installed**, the CLI should:

- Detect the missing bundle for that channel.
- Suggest the exact command to install it (e.g. `func setup --bundle-channel preview`).

## Why

Today a fresh `func setup` can leave customers on a preview/experimental bundle when they only wanted the stable developer experience, and there's no guided path to add a preview bundle on demand when a project actually needs one.

## Acceptance criteria

- [ ] `func setup` installs the latest **stable** bundle by default.
- [ ] `func setup` honors a preview/experimental channel declared in `host.json` and installs the matching bundle instead.
- [ ] `func init -s <stack> -c preview|experimental` detects when the required bundle channel isn't installed and prints a clear suggestion with the install command.
- [ ] Behavior is covered by tests.

liliankasem:
resolves #5374

## What

`func setup` now installs the latest **stable** extension bundle by default. Preview/experimental bundles (and matching node/python templates) are only installed when the project's `host.json` opts into that channel via `extensionBundle.id`.

## Behavior

- **Setup, stable by default:** a prerelease CLI (or `--prerelease`) no longer pulls a preview/experimental bundle unless `host.json` asks for it.
- **Channel applies to templates too:** node/python template workloads track the bundle channel (dotnet templates stay channel-less, since they don't use bundles).
- **Graceful fallback:** when `host.json` pins preview/experimental but that channel has nothing published, setup falls back to stable and warns instead of failing.
- **Init discovery hint:** `func init -s <stack> -c preview|experimental` warns when no matching-channel bundle workload is installed and points at `func workload search bundles --prerelease`.

## Tests

- 1217 Func.Tests pass; added coverage for channel resolution, stable fallback + warning, the `--check` fallback-already-installed case, and init detection.
- Release build clean under `TreatWarningsAsErrors=true`.

## Docs

- Added a channel-selection section to `src/Workloads/Tools/ExtensionBundles/README.md`.

### DIFF
--- a/src/Func/Bundles/BundleHelpers.cs
+++ b/src/Func/Bundles/BundleHelpers.cs
@@ -51,6 +51,37 @@ internal static class BundleHelpers
         };
 
     /// <summary>
+    /// Returns the NuGet prerelease label that identifies <paramref name="channel"/>,
+    /// or <c>null</c> for the stable channel (whose versions carry no prerelease label).
+    /// </summary>
+    /// <param name="channel">The channel to convert.</param>
+    /// <returns>The prerelease label, or <c>null</c> for stable.</returns>
+    public static string? ToPrereleaseLabel(this BundleChannel channel) =>
+        channel switch
+        {
+            BundleChannel.Stable => null,
+            BundleChannel.Preview => PreviewLabel,
+            BundleChannel.Experimental => ExperimentalLabel,
+            _ => throw new ArgumentOutOfRangeException(nameof(channel), channel, null)
+        };
+
+    /// <summary>
+    /// Determines whether <paramref name="version"/> belongs to <paramref name="channel"/>.
+    /// Stable matches only released versions; preview and experimental match prerelease
+    /// versions whose label identifies that channel.
+    /// </summary>
+    /// <param name="version">The nuget package version.</param>
+    /// <param name="channel">The channel to test against.</param>
+    /// <returns>True when the version belongs to the channel.</returns>
+    public static bool MatchesChannel(NuGetVersion version, BundleChannel channel)
+    {
+        ArgumentNullException.ThrowIfNull(version);
+        return channel == BundleChannel.Stable
+            ? !version.IsPrerelease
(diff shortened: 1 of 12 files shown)


src/Func/Bundles/BundleHelpers.cs
```diff
@@ -50,6 +50,37 @@ public static string ToDisplayString(this BundleChannel channel) =>
...
+    /// or <c>null</c> for the stable channel (whose versions carry no prerelease label).
+    /// </summary>
+    /// <param name="channel">The channel to convert.</param>
+    /// <returns>The prerelease label, or <c>null</c> for stable.</returns>
+    public static string? ToPrereleaseLabel(this BundleChannel channel) =>
+        channel switch
+        {
+            BundleChannel.Stable => null,
+            BundleChannel.Preview => PreviewLabel,
+            BundleChannel.Experimental => ExperimentalLabel,
```
satvu, line 64: Just curious - what is the difference between preview and experimental bundles?

satvu (approved):

src/Func/Bundles/BundleHelpers.cs
liliankasem: That's a question for @soninaren or @manvkaur  :D

PR #5388, opened 2026-06-18
build: introduce Workload SDK

jviau:
This PR introduces a Workload SDK and onboards the Workloads.DotNet.csproj` and `Workloads.Host.csproj` projects onto it (will onboard more in later PRs). The SDK today is a local project holding tasks and direct targets, in the future this will be packed to a nupkg for re-use when we move workloads to other repositories.

## Experience

A csproj now defines the following properties:

``` xml
<Project>
  <PropertyGroup>
    <!-- specifying RuntimeIdentifiers turns on the meta and per-RID nupkg feature. Leaving empty will produce a single nupkg. -->
    <RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>

    <!-- Setting to 'true' hooks in the SDK. -->
    <PackAsWorkload>true</PackAsWorkload>

    <!-- workload|content|meta, will be validated. The kind will influence how the package is built -->
    <WorkloadKind>workload</WorkloadKind>

    <!-- Controls the 'alias:{alias}' tag -->
    <WorkloadAlias>dotnet</WorkloadAlias>
  </PropertyGroup>
</Project>
```

## Changes:

1. Automatic generation of `workload.json`, including entry point discovery.
    - Adds a `CliWorkloadAttribute<T>` to allow for the entry point to be defined in code.
    - MSBuild task will discover that attribute and emit the workload.json
 2. Automatic nupkg defaults - setting of PackageType and PackageTags, including workload.json
 3. Workload-kind packaging will include the assembly plus any non-unified dependencies. Content and Meta will leave it out.
 4. Per-RID meta & sub packaging
     - When a workload has multiple RIDs (uses `RuntimeIdentifiers`), the targets will automatically build per RID and then package a meta RID-less nupkg, then a per-RID nupkg.
     - The meta package is currently empty. In the future it will need to include the RID-discovery mechanism when that spec is ready.

## For Future PRs

1. Onboard other workload projects
2. Improve support for content & meta workload kinds. Content should automatically package the build output under `/tools/any`. Meta should translate `PackageReference`'s (or some other item type) into the `workload.json`


shahariyaictseba-247 (approved):
https://docs.github.com/api/article/body?pathname=/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints

src/Func/Func.csproj
```diff
@@ -5,15 +5,25 @@
...
-    <SelfContained>true</SelfContained>
-    <PublishSingleFile>true</PublishSingleFile>
-    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Pack settings, for installing as framework dependent dotnet tool. -->
+    <PackageId>Azure.Functions.Cli</PackageId>
+    <!-- <CreateRidSpecificToolPackages>false</CreateRidSpecificToolPackages> -->
+    <PackAsTool>true</PackAsTool>
```
jviau: need to undo these changes, didn't intend to include them.

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    [Required]
+    public string AssemblyPath { get; set; } = string.Empty;
+
+    [Output]
+    public string WorkloadType { get; private set; } = string.Empty;
+
+    public override bool Execute()
+    {
+        static bool IsCliWorkloadAttribute(CustomAttributeData attribute)
```
liliankasem: Do we want to extract this out to an ext class (or helper) - if we think this will be a commonly used check?

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+        string[] runtimeAssemblies = Directory.GetFiles(runtimeDir, "*.dll");
+        IEnumerable<string> paths = runtimeAssemblies.Concat([ AssemblyPath, _abstractionsAssembly ]);
+        var resolver = new PathAssemblyResolver(paths);
+
+        using MetadataLoadContext mlc = new(resolver);
+        Assembly asm = mlc.LoadFromAssemblyPath(AssemblyPath);
+
+        foreach (CustomAttributeData? attribute in asm.GetCustomAttributesData())
+        {
+            Log.LogMessage(MessageImportance.High, $"Found custom attribute: {attribute.AttributeType.FullName}");
```
liliankasem: This could be quite noisy; do we want to drop the importance to Low? Or is there a justification for this to be high?

src/Workloads/Sdk/Tasks/ResolveWorkloadCopyLocal.cs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Build.Framework;
+
+namespace Azure.Functions.Cli.Workloads.Sdk.Tasks;
+
+public sealed class ResolveWorkloadCopyLocal : Microsoft.Build.Utilities.Task
```
liliankasem, line 8: Missing test?

src/Workloads/Sdk/Tasks/WriteWorkloadJson.cs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Build.Framework;
+
+namespace Azure.Functions.Cli.Workloads.Sdk.Tasks;
+
+public sealed class WriteWorkloadJson : Microsoft.Build.Utilities.Task
```
liliankasem, line 10: Missing test?

src/Workloads/Sdk/Tasks/ResolveWorkloadCopyLocal.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    [Required]
+    public ITaskItem[] UnifiedItems { get; set; } = [];
+
+    [Output]
+    public ITaskItem[] WorkloadCopyLocal { get; set; } = [];
+
+    public override bool Execute()
+    {
+        GetUnifiedItems(out HashSet<string> runtimeProjects, out HashSet<string> runtimePackages);
```
liliankasem: nit: the class defines these outputs as unifiedProjects/Packages - so it's odd to see them referred to differently here

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System.Reflection;
+using System.Runtime.InteropServices;
+using Microsoft.Build.Framework;
+
+namespace Azure.Functions.Cli.Workloads.Sdk.Tasks;
+
+public sealed class ResolveWorkloadEntry : Microsoft.Build.Utilities.Task
+{
+    private static readonly string _abstractionsAssembly = Path.Combine(
```
liliankasem: Could this throw if `Assembly.Location` is null/empty?

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+    public string WorkloadType { get; private set; } = string.Empty;
+
+    public override bool Execute()
+    {
+        static bool IsCliWorkloadAttribute(CustomAttributeData attribute)
+        {
+            Type type = attribute.AttributeType;
+            return type.IsGenericType
+                && type.GetGenericTypeDefinition().FullName
+                    == "Azure.Functions.Cli.Workloads.CliWorkloadAttribute`1";
```
liliankasem: nit: Can we use a const for this instead of hardcoding inline

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
```diff
@@ -0,0 +1,53 @@
...
+        var resolver = new PathAssemblyResolver(paths);
+
+        using MetadataLoadContext mlc = new(resolver);
+        Assembly asm = mlc.LoadFromAssemblyPath(AssemblyPath);
+
+        foreach (CustomAttributeData? attribute in asm.GetCustomAttributesData())
+        {
+            Log.LogMessage(MessageImportance.High, $"Found custom attribute: {attribute.AttributeType.FullName}");
+
+            if (IsCliWorkloadAttribute(attribute))
```
liliankasem: Do we want to wrap this in a try-catch to avoid a potential FileNotFoundException? Agent is flagging this concern so what to check in with you on it:

> a workload with an assembly-level attribute from a package dependency would throw here and break discovery before we ever reach the CliWorkload attribute

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
jviau: Meant to remove that! Was debugging the target.

src/Workloads/Sdk/Tasks/ResolveWorkloadCopyLocal.cs
jviau: What do you mean by referred to differently?

liliankasem: One place calls them "runtimeProjects" (here) the other "unifiedProjects" (GetUnifiedItems method sig)

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
jviau: This should never be null or empty for an MSBuild task.

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
jviau: done

src/Workloads/Sdk/Tasks/WriteWorkloadJson.cs
jviau: added

src/Workloads/Sdk/Tasks/ResolveWorkloadCopyLocal.cs
jviau: updated the names

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
jviau: made constants.

src/Workloads/Sdk/Tasks/ResolveWorkloadEntry.cs
jviau: done

src/Workloads/Sdk/Tasks/ResolveWorkloadCopyLocal.cs
jviau: added.

liliankasem (approved):

PR #5392, opened 2026-06-22
Replace deprecated url.parse() in npm install script

### LINKED ISSUE #5366: npm publish-tools: replace deprecated `url.parse()` in install.js with WHATWG URL API
## Summary

`eng/tools/publish-tools/npm/lib/install.js` uses the legacy `url.parse()` API, which emits a `DEP0169` deprecation warning on Node.js 20+:

```
(node:XXXXX) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
```

The package declares `"engines": { "node": ">=20" }`, so every `npm install azure-functions-core-tools` on a supported Node version surfaces this warning during `postinstall`.

## Repro

```bash
cd eng/tools/publish-tools/npm
npm ci --ignore-scripts
node lib/install.js
```

Warning prints right after the `attempting to GET ...` line.

## Source

[`eng/tools/publish-tools/npm/lib/install.js`](https://github.com/Azure/azure-functions-core-tools/blob/main/eng/tools/publish-tools/npm/lib/install.js):

```js
// line 4
const url = require('url');
...
// line 52
const options = url.parse(endpoint);
```

## Suggested fix

Replace with the WHATWG `URL` class (built-in, available since Node 10) and pass the components directly to `https.get`:

```js
const parsed = new URL(endpoint);
const options = {
    protocol: parsed.protocol,
    hostname: parsed.hostname,
    port: parsed.port,
    path: parsed.pathname + parsed.search,
    agent: proxy ? new HttpsProxyAgent(proxy) : undefined,
};
```

Or pass the `URL` instance directly to `https.get(parsed, ...)`, which Node supports natively.

The `const url = require('url');` import can then be removed (only `url.parse` is used).

## Impact

- Removes deprecation noise from every user's `npm install`.
- Future-proofs against eventual removal of `url.parse()` in a future Node major.
- No behavior change expected; the CDN endpoint is always a well-formed HTTPS URL.

## Notes

Discovered while validating https-proxy-agent bump in #4850. Pre-existing issue, not caused by that PR.

vribeirocmp:
Fixes #5366.

The npm install script builds the download request options with url.parse().
That call is deprecated (DEP0169) and prints a warning on every npm install
of the package.

This change uses the WHATWG URL API instead. The endpoint is parsed with
new URL() and passed as the first argument to https.get(), and the options
object (which carries the proxy agent when one is configured) is passed as
the second argument. The behavior is the same, including the proxy code path.

Tested locally: the script passes node --check, the download request reaches
the server, and the proxy agent is still used when set.

### DIFF
--- a/eng/tools/publish-tools/npm/lib/install.js
+++ b/eng/tools/publish-tools/npm/lib/install.js
@@ -1,7 +1,6 @@
 #! /usr/bin/env node
 
 const extract = require('extract-zip');
-const url = require('url');
 const { HttpsProxyAgent } = require('https-proxy-agent');
 const https = require('https');
 const version = require('../package.json').version;
@@ -49,7 +48,8 @@ const fileName = 'Azure.Functions.Cli.' + platform + '.' + version + '.zip';
 const endpoint = 'https://cdn.functions.azure.com/public/' + consolidatedBuildId + '/' + fileName;
 
 console.log('attempting to GET %j', endpoint);
-const options = url.parse(endpoint);
+const requestUrl = new URL(endpoint);
+const options = {};
 
 // npm config preceed system environment
 // https://github.com/npm/npm/blob/19397ad523434656af3d3765e80e22d7e6305f48/lib/config/reg-client.js#L7-L8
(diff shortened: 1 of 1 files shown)


liliankasem:
Thanks for the contribution @vribeirocmp, but looks like we already have a PR for this here: #5371

Closing as duplicate.

PR #5405, opened 2026-06-24
build: migrate templates to workloads sdk

jviau:
migrate the following projects to workloads sdk

```
Workloads.Templates.Dotnet.csproj
Workloads.Templates.Node.csproj
Workloads.Templates.Python.csproj
```


liliankasem (approved):
Just double checking that the `func-workload` tag is applied to all workloads via the sdk?

jviau:
> Just double checking that the `func-workload` tag is applied to all workloads via the sdk?

Yes: https://github.com/Azure/azure-functions-core-tools/blob/vnext/eng/build/Workloads/Workload.Pack.targets#L7

Issue #5333, opened 2026-06-09
Implement `func update` command


liliankasem:
Spec lives at `docs/proposed/func-update.spec.md` (on the `docs` branch). Implement the command
so users can update an installed v5 CLI to the latest version without
re-running the installer script.

## Definition of Done

- [ ] `func update` command implemented per the existing spec.
- [ ] Works across the installer's supported channels (script-based
      install for Windows, macOS, Linux).
- [ ] Surfaces "no update available" cleanly when already current.
- [ ] Honours `--yes` to skip confirmation (required when non-interactive).
- [ ] Tests cover happy path + no-update path + failure path.
- [ ] Help text + docs updated.

## PRs

| PR | Status | What it delivers |
|----|--------|-----------------|
| #5297 | ✅ Merged | `UpdateCommand` scaffold — options (`--prerelease`, `--version`, `--yes`), help text, Parser registration |
| #5298 | ✅ Merged | `IReleaseFeed` + `CdnReleaseFeed` — fetches version manifest from CDN, resolves latest stable/preview |
| #5428 | Draft | `ICliUpdater` + `CliUpdater` — download → extract → backup → swap → verify → rollback pipeline |
| #5458 | Draft | Wire `UpdateCommand` handler — install-method detection (npm/brew/choco/winget), version compare, confirmation prompt, progress reporting, DI registration. Completes the feature. |

PR #5428, opened 2026-06-25
feat(update): add ICliUpdater download/swap/rollback pipeline

liliankasem:
Stacked on #5298. PR #3 in the `func update` chain.

Adds the download → extract → backup → swap → verify → rollback pipeline that takes a `Release` from `IReleaseFeed` and installs it in place.

## What's in

- **`ICliUpdater` + `CliUpdater`** (`src/Func/Update/`) — orchestrates the full update flow
- **`IUpdateFileSystem` + `UpdateFileSystem`** — file-system seam (temp dir, move/delete dir, stream-to-file, extract zip)
- **`ICliEnvironment` + `CliEnvironment`** (`src/Func/Common/`) — abstracts `Environment.ProcessPath`
- **`CliUpdaterTests`** — 4 tests covering happy path, download failure, verify mismatch + rollback, swap failure + rollback

## Update flow

1. **Download** — stream release zip from CDN to temp file
2. **Extract** — unzip to temp directory
3. **Backup** — rename current install directory to `.backup`
4. **Swap** — move extracted files into place
5. **Verify** — spawn new `func --version`, confirm expected version
6. **Cleanup** — delete backup on success; rollback (restore backup) on any failure

## Design notes

- All I/O behind interfaces (`IUpdateFileSystem`, `ICliEnvironment`, `IProcessRunner`) for full testability
- Rollback is always attempted on failure before rethrowing
- Errors surface as `GracefulException` with actionable messages
- DI wiring left for the command-integration PR

part of #5333


src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,192 @@
...
+                _fileSystem.ExtractZip(zipPath, extractDir);
+            }
+            catch (InvalidDataException ex)
+            {
+                throw new GracefulException(
+                    $"Failed to extract the downloaded package for func {release.Version}. The archive may be corrupt. Try running 'func update' again.",
+                    ex);
+            }
+
+            _fileSystem.MoveDirectory(installDir, backupDir);
```
ahmedmuhsin: This blocks self-update on Windows. A running process can't rename the folder holding its own loaded exe and dlls, so this throws IOException access denied right here, before any swap happens. It's also unguarded, so it escapes as a raw stack trace instead of a GracefulException. I tested it on Win11 to be sure. What does work is renaming the files in place (func.exe to func.exe.old), copying the new ones in, then deleting the .old files on next launch. You can rename a loaded file, you just can't delete it.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            // Swap extracted files into place, then verify. Any failure after
+            // the backup triggers a rollback so the previous version is restored.
+            bool installSucceeded = false;
+            try
+            {
+                _fileSystem.MoveDirectory(extractDir, installDir);
```
ahmedmuhsin: extractDir comes from Path.GetTempPath, so if func is installed on a different volume than TEMP this throws about identical roots. I hit it going C: to Q:. Not Windows only either, Linux and mac fail the same way with EXDEV. Stage onto the same volume as the install dir, or copy plus delete when the roots differ.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,192 @@
...
+                    $"CDN returned {(int)response.StatusCode} while downloading func {release.Version}. Try again later.",
+                    isUserError: true);
+            }
+
+            await using Stream content = await response.Content.ReadAsStreamAsync(cancellationToken);
+            await _fileSystem.SaveStreamToFileAsync(zipPath, content, cancellationToken);
+        }
+    }
+
+    private async Task VerifyAsync(Release release, string installDir, CancellationToken cancellationToken)
```
ahmedmuhsin, line 223: The only check here is func --version. A truncated or tampered zip that still runs and prints the right version would sail through this. Can we publish a checksum in the release metadata and verify the zip before extracting? This thing pulls executable code onto the machine so it feels worth it.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,192 @@
...
+        string tempWorkDir = _fileSystem.CreateTempDirectory();
+        string zipPath = Path.Combine(tempWorkDir, "func-update.zip");
+        string extractDir = _fileSystem.CreateTempDirectory();
+
+        string installDir = GetInstallDirectory();
+        string backupDir = installDir + ".backup";
+
+        if (_fileSystem.DirectoryExists(backupDir))
+        {
+            throw new GracefulException(
```
ahmedmuhsin: The download throws set isUserError true but this one, plus the verify and GetInstallDirectory throws, leave it off. Let's make it consistent.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,192 @@
...
+
+using System.Runtime.InteropServices;
+using Azure.Functions.Cli.Common.Processes;
+using Azure.Functions.Cli.Common;
+using Microsoft.Extensions.Logging;
+
+namespace Azure.Functions.Cli.Update;
+
+/// <inheritdoc cref="ICliUpdater" />
+internal sealed class CliUpdater(
```
ahmedmuhsin: I don't see ICliUpdater, IUpdateFileSystem, ICliEnvironment or the HttpClient registered anywhere. Is the command wiring landing in a follow-up? If so let's call it out so the two go in together.

test/Func.Tests/Common/Processes/ProcessRunnerTests.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using Azure.Functions.Cli.Commands.Start.Azurite.Processes;
+using Azure.Functions.Cli.Common.Processes;
```
ahmedmuhsin, line 4: The runner moved to Common/Processes but its test stayed under Azurite. Move it over to match.

ahmedmuhsin (commented):
Took a proper look and ran a few things locally to check my assumptions. Structure is solid and the seams are nice and testable. Two blockers before this ships. The directory move can't self-update on Windows, and there's no integrity check on the download. Rest is inline.

src/Func/Update/CliUpdater.cs
satvu: Follow up PR in draft here: https://github.com/Azure/azure-functions-core-tools/pull/5458

src/Func/Update/CliUpdater.cs
satvu: Fixed in 23ec5df4. Replaced the directory-move strategy with rename-in-place: existing files get renamed to \.old\, new files are copied in, then \.old\ files are cleaned up after verification. If the running exe can't be deleted (locked on Windows), it's deferred to next launch. Verified e2e on Windows — rename of a locked exe succeeds, new files write in place.

src/Func/Update/CliUpdater.cs
satvu: Fixed in 23ec5df4. Added \CreateTempDirectory(siblingPath)\ which stages temp dirs on the same volume as the install directory (parent dir + \.func-update-<random>\). This avoids cross-volume \Directory.Move\ failures (EXDEV on Unix, IOException on Windows).

src/Func/Update/CliUpdater.cs
satvu: Fixed in 23ec5df4. Added \isUserError: true\ to all \GracefulException\ throws — extract failure, verification failure, both \GetInstallDirectory\ guards.

test/Func.Tests/Common/Processes/ProcessRunnerTests.cs
satvu: Fixed in 23ec5df4. Moved \ProcessRunnerTests.cs\ to \	est/Func.Tests/Common/Processes/\ and updated the namespace to \Azure.Functions.Cli.Tests.Common.Processes\.

src/Func/Update/CliUpdater.cs
satvu: Good call — agreed this would be valuable. It requires changes to the release metadata/feed to publish checksums alongside artifacts, so it's out of scope for this service-layer PR. Will track as a follow-up.

satvu: Actually, the issue already is being tracked here: https://github.com/Azure/azure-functions-core-tools/issues/5445

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,269 @@
...
+        IReadOnlyList<string> existingFiles = _fileSystem.GetFiles(installDir);
+        foreach (string file in existingFiles)
+        {
+            // Skip any leftover .old files from a previous failed update
+            if (file.EndsWith(OldFileSuffix, StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
+
+            _fileSystem.RenameFile(file, file + OldFileSuffix);
```
soninaren: **(Blocking) Running `func update` a second time collides on a stale `.old`.** `RenameFile` is two-arg `File.Move`, which throws if the destination already exists. After a first `func update`, the running `func.exe.old` stays locked and `CleanupOldFiles` can't delete it ("cleaned up on next launch" — but there's no launch-time sweep in the tree). The next run's `RenameFile(func.exe, func.exe.old)` hits that stale `.old` and throws `IOException`; rollback then also collides trying to restore the stale `.old` over the still-present `func.exe`, landing in the "inconsistent state" path. Suggest deleting/overwriting any pre-existing `.old` before renaming, plus adding the startup `*.old` sweep this design relies on (or dropping the reliance on it).

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,269 @@
...
+                    _fileSystem.RenameFile(file, originalPath);
+                }
+            }
+
+            _logger.LogInformation("Previous version restored.");
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Rollback failed. The installation at {InstallDir} may be in an inconsistent state.", installDir);
+        }
```
soninaren: A few smaller items on this rollback path:

1. **Rollback aborts on the first per-file failure.** The whole loop is wrapped in a single `try`, so if one `DeleteFile`/`RenameFile` throws (locked file / permissions), the remaining `.old` files are never restored — a worse state than all-or-nothing. Consider a per-file try/continue.
2. **Two error contracts in one feature.** `CliUpdater` throws `GracefulException` straight from the service layer, while sibling `CdnReleaseFeed` throws `InvalidOperationException` (specific-exception-per-convention). Worth settling on one contract so `UpdateCommand` faces a single failure model.
3. **Misleading "restored" messaging.** `VerifyAsync` hardcodes "The previous version has been restored." before rollback actually runs (and it may itself fail), and this method logs "Previous version restored." unconditionally — including download-stage failures where nothing was swapped. Worth making both conditional on what actually happened.

src/Func/Update/IUpdateFileSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Renames <paramref name="sourcePath"/> to <paramref name="destinationPath"/>.
+    /// </summary>
+    public void RenameFile(string sourcePath, string destinationPath);
+
+    /// <summary>
+    /// Copies a single file from <paramref name="sourcePath"/> to
+    /// <paramref name="destinationPath"/>, overwriting if it exists.
+    /// </summary>
+    public void CopyFile(string sourcePath, string destinationPath);
```
soninaren: Nits: `MoveDirectory` and `CopyFile` are declared on this seam but never called by `CliUpdater` — dead surface that still has to be mocked/maintained. Also in `CliUpdaterTests`: the `.old` suffix is hardcoded as string literals instead of reusing `CliUpdater.OldFileSuffix` (tests drift silently if it changes), and the tests use raw `Assert.*` and assert exact collaborator call sequences rather than the repo-standard AwesomeAssertions.

soninaren (commented):
A few notes on the download/swap/rollback pipeline. Nice cleanup on the copied-files rollback and the exact-match version check since the last push. Remaining items inline.

src/Func/Update/CliUpdater.cs
satvu: Addressed in cbffdd6e:

1. **Per-file rollback** — Fixed. Each `DeleteFile`/`RenameFile` in `TryRollbackInPlace` now has its own `try/catch`, so one locked file doesn't abort the remaining restorations.

2. **Two error contracts** — Acknowledged as a valid design concern but out of scope for this PR. `CliUpdater` throws `GracefulException` because it's the command-boundary wrapper per repo conventions. Happy to align `CdnReleaseFeed` in a follow-up.

3. **Misleading 'restored' messaging** — Fixed. Removed the premature claim from `VerifyAsync`'s exception message. `TryRollbackInPlace` now tracks whether any files were actually restored and only logs 'Previous version restored.' when that's true.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,238 @@
...
+                    $"Failed to extract the downloaded package for func {release.Version}. The archive may be corrupt. Try running 'func update' again.",
+                    ex,
+                    isUserError: true);
+            }
+
+            // On Windows a running process holds its own exe and dlls locked,
+            // so we can't move or delete the install directory. Instead, rename
+            // existing files in-place (func.exe → func.exe.old) and copy the
+            // new files into the same directory.
+            SwapInPlace(installDir, extractDir);
```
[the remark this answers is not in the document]
satvu: This concern is incorrect. Windows **does** allow renaming a running executable.

The NT kernel's PE loader opens executables with `FILE_SHARE_DELETE` sharing mode, which explicitly permits rename and delete operations on the file while it's loaded. The process continues executing from the same physical pages in memory regardless of what the file is named on disk — renaming only changes the directory entry, not the file data.

**Evidence:**

1. [Stack Overflow — "Rename a running executable (exe) file"](https://stackoverflow.com/questions/6695496/rename-a-running-executable-exe-file) — community-confirmed that `File.Move` on a running `.exe` succeeds on NTFS.

2. [Visual Studio Magazine — "Replace a Running Application with a New Version"](https://visualstudiomagazine.com/articles/2017/12/15/replace-running-app.aspx) — describes this exact pattern: *"Though you can't replace a running application, you **can rename it**."*

3. This is the same approach used by Squirrel.Windows (the update framework behind VS Code, GitHub Desktop, Slack, and Teams on Windows).

What Windows does **not** allow is overwriting or truncating a running exe in place (writing to the same NTFS stream). But `File.Move` (rename) is always safe because it operates on the directory entry, not the file content.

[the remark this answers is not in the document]
satvu: Also had an agent test this locally and it seems to be ok?

src/Func/Common/ICliEnvironment.cs
jviau, line 1: This feels like it should just be an `Options<>` pattern, not its own interface.

src/Func/Update/UpdateFileSystem.cs
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.IO.Compression;
+
+namespace Azure.Functions.Cli.Update;
+
+/// <inheritdoc cref="IUpdateFileSystem" />
+internal sealed class UpdateFileSystem : IUpdateFileSystem
```
jviau: We already have an `IProfileFileSystem`. We should consolidate these file system types into one abstraction. Also consider using https://www.nuget.org/packages/TestableIO.System.IO.Abstractions, as it has great test mocks via another package.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+    public async Task UpdateAsync(Release release, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(release);
+
+        string installDir = GetInstallDirectory();
+
+        // Stage the extraction on the same volume as the install directory so
+        // Directory.Move / File.Move never crosses volume boundaries.
+        string tempWorkDir = _fileSystem.CreateTempDirectory(installDir);
+        string zipPath = Path.Combine(tempWorkDir, "func-update.zip");
```
jviau: The file extension is OS dependent. On windows it is a zip, but it is a `.tar.gz` on linux and mac.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+            // On Windows a running process holds its own exe and dlls locked,
+            // so we can't move or delete the install directory. Instead, rename
+            // existing files in-place (func.exe → func.exe.old) and copy the
+            // new files into the same directory.
+            copiedFiles = SwapInPlace(installDir, extractDir);
+
+            await VerifyAsync(release, installDir, cancellationToken);
+
+            CleanupOldFiles(installDir);
+            _logger.LogInformation("func {Version} installed successfully.", release.Version);
```
jviau: Should we make a habit of using logger source gen?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+                    $"Failed to extract the downloaded package for func {release.Version}. The archive may be corrupt. Try running 'func update' again.",
+                    ex,
+                    isUserError: true);
+            }
+
+            // On Windows a running process holds its own exe and dlls locked,
+            // so we can't move or delete the install directory. Instead, rename
+            // existing files in-place (func.exe → func.exe.old) and copy the
+            // new files into the same directory.
+            copiedFiles = SwapInPlace(installDir, extractDir);
```
jviau: I don't think it is a good idea to blindly rename all files to `.old` in the directory. What if there are unrelated user-placed files here? Or what if we eventually start writing config that is meant to persist between versions. Plus, we currently ship as a single file app. So there shouldn't even be any files to rename beyond `func` / `func.exe`.

What I recommend: extract the new directory to the temp location. Look at what files it all has, then look for the same files in the existing install. Rename only those matching file names to `.old`.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            await VerifyAsync(release, installDir, cancellationToken);
+
+            CleanupOldFiles(installDir);
+            _logger.LogInformation("func {Version} installed successfully.", release.Version);
+        }
+        catch (Exception) when (TryRollbackInPlace(installDir, copiedFiles))
```
jviau: Why use exception filter as a side-effect? This seems more convoluted then just:

``` csharp
catch (Exception)
{
    RollBack(..);
    throw;
}
```

What is the filter-side-effect approach doing that the above doesn't?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+            await _fileSystem.SaveStreamToFileAsync(zipPath, content, cancellationToken);
+        }
+    }
+
+    private async Task VerifyAsync(Release release, string installDir, CancellationToken cancellationToken)
+    {
+        string funcBinary = GetFuncBinaryPath(installDir);
+        string expectedVersion = release.Version.ToString();
+
+        ProcessOutcome outcome = await _processRunner.RunAsync(
```
jviau, line 227: I think this verification should be done via a checksum approach. But we also need to ensure the checksum is published to the CDN.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        try
+        {
+            _logger.LogInformation("Downloading func {Version}.", release.Version);
+            await DownloadAsync(release, zipPath, cancellationToken);
+
+            _logger.LogInformation("Extracting update package.");
+            try
+            {
+                _fileSystem.ExtractZip(zipPath, extractDir);
```
jviau: As mentioned, this won't always be a zip.

Also, do we want to unzip to a temp directly? Could we load the archive metadata in memory to find its files, perform the renames, then extract the archive in place?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+        string? dir = Path.GetDirectoryName(processPath);
+        if (string.IsNullOrEmpty(dir))
+        {
+            throw new GracefulException("Could not determine the func installation directory.", isUserError: true);
+        }
+
+        return dir;
+    }
+
+    private static string GetFuncBinaryPath(string installDir) =>
```
jviau: We already have `ProcessPath` DI'd into this type. Is that not what this tries to compute already?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+            if (line.Equals(expectedVersion, StringComparison.Ordinal))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private void TryDeleteDirectory(string path)
```
jviau: I have a suggestion to avoid the temp directory at all. But if you do opt to keep it, I recommend moving it into its own struct so we can follow the `using` pattern here. If you move to file system abstractions, or whatever file system abstraction layer you get, aim to make something along the lines of:

``` csharp
await using TempDirectory temp = _fileSystem.CreateTempDirectory();

private readonly struct TempDirectory(string path) : IAsyncDisposable
{
    public string Path { get; } = path;

    public ValueTask DisposeAsync()
    {
        // clean up Path.
    }
}
```

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            if (!response.IsSuccessStatusCode)
+            {
+                throw new GracefulException(
+                    $"CDN returned {(int)response.StatusCode} while downloading func {release.Version}. Try again later.",
+                    isUserError: true);
+            }
+
+            await using Stream content = await response.Content.ReadAsStreamAsync(cancellationToken);
+            await _fileSystem.SaveStreamToFileAsync(zipPath, content, cancellationToken);
```
jviau, line 196: We should make sure the CDN release work includes publishing a checksum, and this download logic should verify against the checksum. This will replace the `VerifyAsync` logic here.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,295 @@
...
+        // and track which files were introduced for rollback purposes.
+        IReadOnlyList<string> newFiles = _fileSystem.GetFiles(extractDir);
+        List<string> copiedFiles = new(newFiles.Count);
+        foreach (string newFile in newFiles)
+        {
+            string relativePath = Path.GetRelativePath(extractDir, newFile);
+            copiedFiles.Add(Path.Combine(installDir, relativePath));
+        }
+
+        _fileSystem.CopyDirectory(extractDir, installDir);
```
jviau: If this throws we may have a partially copied upgrade, but the return value of this method will no longer be available and we won't know the progress, and thus be unable to perform a rollback.

Do we need to track the files explicitly? Or can rollback independently look at what `.old` files exist and determine what to remove from there?

jviau (changes requested):
Primary asks:

1. Lets consolidate the file system abstractions
2. Use a checksum verify approach
3. Do not rename all files in the install dir, only rename ones we know the new version will need to overwrite
4. Partial copy failure will prevent a rollback
5. Handle .zip vs .tar.gz

The choice of using a temp dir for archive staging vs .NET APIs to examine contents of the archive and extract in place is up to you.

src/Func/Update/CliUpdater.cs
satvu: There's a separate issue for this: https://github.com/Azure/azure-functions-core-tools/issues/5445. A stubbed approach was added with a TODO.

src/Func/Update/CliUpdater.cs
satvu: That is tracked here: https://github.com/Azure/azure-functions-core-tools/issues/5445

src/Func/Update/UpdateFileSystem.cs
satvu: Consolidated, but keeping with this slimmer file system abstraction since it wasn't too difficult to mock and TestableIO has a lot more functionality than we need.

src/Func/Common/ICliEnvironment.cs
satvu: Done — replaced `ICliEnvironment` + `CliEnvironment` with a `CliEnvironmentOptions` POCO injected via `IOptions<CliEnvironmentOptions>`. Tests use `Options.Create()` directly instead of NSubstitute.

Worth noting the trade-off: `Environment.ProcessPath` is a runtime intrinsic rather than a configuration value, so the options pattern is a slightly unconventional fit. But it's simpler (no interface + impl pair for a single property) and equally testable, so net win. Committed in f47770bf.

src/Func/Update/CliUpdater.cs
satvu: Simplified to rename one file - if we run into this later, we can tackle it then. This should be sufficient for preview.

src/Func/Update/CliUpdater.cs
satvu: No longer relevant w/ change to focus on single binary/file (`func`) .

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                throw new GracefulException(
+                    $"The downloaded archive for func {release.Version} does not contain '{binaryName}'.",
+                    isUserError: true);
+            }
+
+            // Rename the running binary out of the way, then copy the new one in.
+            // On Windows, renaming a running executable is allowed by the NT kernel.
+            _fileSystem.MoveFile(binaryPath, backupPath, overwrite: true);
+            _fileSystem.CopyFile(extractedBinary, binaryPath);
```
brettsam: if an exception is thrown here, `swapped` will be `false` and we won't roll back...

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,301 @@
...
+            swapped = true;
+
+            await VerifyAsync(release, installDir, cancellationToken);
+
+            // Best-effort removal of the backup; on Windows the running exe may
+            // still be locked, in which case it gets cleaned up on next launch.
+            TryDeleteFile(backupPath);
```
brettsam: where is the code that cleans up the backup on next launch?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,301 @@
...
+        string binaryPath = GetBinaryPath();
+        string installDir = Path.GetDirectoryName(binaryPath)!;
+        string backupPath = binaryPath + OldFileSuffix;
+
+        // Stage the download on the same volume as the install directory so
+        // File.Move never crosses volume boundaries.
+        string tempWorkDir = _fileSystem.CreateTempDirectory(installDir);
+        string archiveExtension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "zip" : "tar.gz";
+        string archivePath = Path.Combine(tempWorkDir, $"func-update.{archiveExtension}");
+        string extractDir = _fileSystem.CreateTempDirectory(installDir);
```
brettsam: should these `CreateTempDirectory()` calls be inside the `try`? that way if they fail somehow (or the second one fails), we clean up still?

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,301 @@
...
+{
+    private const string OldFileSuffix = ".old";
+
+    private readonly HttpClient _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+    private readonly IFileSystem _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+    private readonly CliEnvironmentOptions _environment = (environmentOptions ?? throw new ArgumentNullException(nameof(environmentOptions))).Value;
+    private readonly IProcessRunner _processRunner = processRunner ?? throw new ArgumentNullException(nameof(processRunner));
+    private readonly ILogger<CliUpdater> _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+    public async Task UpdateAsync(Release release, CancellationToken cancellationToken)
```
brettsam, line 27: in general, we have no interprocess protection here for two updates running simultaneously... we likely want to drop a "lock" file to prevent another process from updating simultaneously on the same machine.

src/Func/Update/CdnReleaseFeed.cs
```diff
@@ -149,6 +146,13 @@ private async Task<VersionManifest> FetchManifestAsync(CancellationToken cancell
     private static Uri BuildDownloadUri(SemVersion version)
     {
         string rid = RuntimeInformation.RuntimeIdentifier;
-        return new Uri($"public/cli/v5/{version}/Azure.Functions.Cli.{rid}.{version}.zip", UriKind.Relative);
+        string extension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "zip" : "tar.gz";
```
brettsam: this is done in a couple places... use a helper instead?

src/Func/Common/Processes/IProcessRunner.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace Azure.Functions.Cli.Commands.Start.Azurite.Processes;
+namespace Azure.Functions.Cli.Common.Processes;
 
 /// <summary>
 /// Launches short-lived child processes for the CLI. Wraps
```
brettsam, line 7: note... comment below still mentions Azurite but this is more general-purpose now.

src/Func/Update/CliUpdater.cs
satvu: Fixed in https://github.com/Azure/azure-functions-core-tools/pull/5428/commits/41c2406d784ad29c617ab1fa6184656e5e30fef7

src/Func/Update/CliUpdater.cs
satvu: comment updated - overwritten on next update.

src/Func/Update/CliUpdater.cs
satvu: Concurrent invocations aren't very likely as `func update` requires an explicit call/action but noted. I opened this issue for trackign: https://github.com/Azure/azure-functions-core-tools/issues/5524

src/Func/Common/CliEnvironmentOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+/// Runtime environment options for the CLI process. Bind at startup and
+/// inject via <see cref="Microsoft.Extensions.Options.IOptions{TOptions}"/>.
+/// </summary>
+internal sealed class CliEnvironmentOptions
+{
+    /// <summary>
+    /// Full path of the running executable, or <c>null</c> when the platform
+    /// cannot determine it.
+    /// </summary>
+    public string? ProcessPath { get; set; }
```
jviau: Consider ensuring this is non-nullable. Maybe default it to `Environment.ProcessPath`

src/Func/Common/IFileSystem.cs
jviau, line 1: I still recommend bringing in TestableIO file system abstractions. 

1. We use it in other projects, so it is easier for dev knowledge transfer to standardize on one abstraction
2. We don't need to test our own implementation
3. We can leverage their test mocks

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,333 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(release);
+
+        string binaryPath = GetBinaryPath();
+        string installDir = Path.GetDirectoryName(binaryPath)!;
+
+        // Stage the download on the same volume as the install directory so
+        // File.Move never crosses volume boundaries.
+        using TempDirectory tempWorkDir = new(_fileSystem.CreateTempDirectory(installDir), _fileSystem);
```
jviau: Having this be a method off the file system itself will give a cleaner API:

``` csharp
using TempDirectory temp = _fileSystem.CreateTempDirectory(....);
```

src/Func/Common/PhysicalFileSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+    public void CreateDirectory(string path) => Directory.CreateDirectory(path);
+
+    public string CreateTempDirectory()
+    {
+        string path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        Directory.CreateDirectory(path);
+        return path;
+    }
+
+    public string CreateTempDirectory(string siblingPath)
```
jviau: Some concerns issues with this API:

1. I would expect this to still use `Path.GetTempPath()`. Nothing about this is "temp" if you pass in a path outside of the temp dir.
2. Hardcoding the `.func-update-` logic into here is leaking business logic into this abstraction
3. The "sibling" path logic is a bit confusing to me. I wouldn't expect a file system abstraction to have such business logic in it.

src/Func/Update/CliUpdater.cs
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        string actual = await _fileSystem.ComputeSha256Async(filePath, cancellationToken);
+
+        if (!string.Equals(actual, release.Sha256Checksum, StringComparison.OrdinalIgnoreCase))
+        {
+            throw new GracefulException(
+                $"Checksum mismatch for func {release.Version}. " +
+                $"Expected '{release.Sha256Checksum}' but got '{actual}'. " +
+                "The download may be corrupt or tampered with. Try running 'func update' again.",
```
jviau: Let's remove the corrupt/tampered line, it is speculative and kind of an alarming message to give to customers. Just leave it as checksum verification failed, expected/actual. Try again.

jviau (approved):
Approving, but I'd still encourage using `TestableIO` abstractions for the file system

src/Func/Common/IFileSystem.cs
satvu: Filed an issue for this b/c the scope is rather large (https://github.com/Azure/azure-functions-core-tools/issues/5561)

jviau (approved):

Issue #5329, opened 2026-06-09
Design: profile updates as part of release pipelines


liliankasem:
Design the integration between SKU/host release pipelines and the CLI
profiles. Today profile updates are manual; we want a release of (e.g.)
the Flex host to automatically refresh the Flex profile to the new host
version. Output is a written design plus follow-up issues for each
pipeline change.

## Definition of Done

- [ ] Design doc landed in `docs/proposed/` covering: triggers (what
      release events update which profiles), publishing flow (who builds
      the profile, who signs, who pushes to CDN), versioning, rollback,
      and validation gates.
- [ ] Per-SKU contract documented (Flex, Consumption, App Service, etc.)
      including which fields update and which stay pinned.
- [ ] Follow-up issues opened for each pipeline change the design
      surfaces.
- [ ] Doc reviewed by host team + release pipeline owners.

liliankasem:
## Profiles CDN - Initial Setup

### What has been done

1. **Registry uploaded** to blob storage (**staging**):
   ```
   public/cli/v5/profiles/v1/registry.json
   public/cli/v5/profiles/v1/registry.json.sha256
   ```
   Accessible via: `https://cdn-staging.functions.azure.com/public/cli/v5/profiles/v1/registry.json`

- [ ] **TODO: This will still need to go to prod CDN**

2. **CLI wired up** — `RemoteProfileSource` added on `vnext` branch (PR incoming) that fetches from the CDN with SHA-256 checksum validation and local caching.

### CDN Caching — Still Needs Configuration

Per the [design doc](proposed/cli-profiles.md §6.1), the CDN caching strategy should be:

- **1-hour TTL** — the CLI uses a local cache with 1h TTL before re-fetching
- **ETag-based cache** — CDN should serve `ETag` and respect `If-None-Match` for conditional GETs (not yet configured on the CDN/blob level)
- **`generatedAt` staleness check** — if cached registry is >7 days old and remote fails, CLI warns users

**Action needed for this issue:**
- Configure blob storage / CDN cache-control headers (e.g., `Cache-Control: public, max-age=3600`)
- Ensure ETag headers are served (Azure Blob Storage does this by default, but verify CDN pass-through)
- Set up the publishing pipeline to update `registry.json` + `registry.json.sha256` atomically as part of SKU deployment

[...]

PR #5434, opened 2026-06-26
Add RemoteProfileSource for CDN-backed profile registry

### LINKED ISSUE #5332: Stand up profiles CDN
Profiles are currently served from GitHub raw, which couples profile
lookups to GitHub availability and gives us no versioning or cache
controls. Pick a CDN, define the path scheme and versioning strategy,
build the publishing pipeline, and switch `func setup` / profile
commands over.

## Definition of Done

- [x] CDN + path scheme picked and documented.
    - `cli/v5/profiles/v1/registry.json`
- [x] Versioning + cache-bust strategy documented.
- [x] Fallback / failure behaviour documented (what happens if the CDN
      is unreachable).
- [x] `func setup` and profile commands read from the CDN instead of
      GitHub raw.

liliankasem:
## Summary

Adds `RemoteProfileSource` that fetches the built-in profile registry from the CDN instead of relying solely on the bundled embedded resource.

## CDN Configuration

The CLI defaults to the **prod CDN** base URL (`https://cdn.functions.azure.com/public/`). For testing, override via environment variable:

```bash
export FUNC_CLI_PROFILES_CDN_BASE_URL="https://cdn-staging.functions.azure.com/public/"
```

This follows the same pattern as `FUNC_CLI_QUICKSTART_MANIFEST_URL`.

> **⚠️ Note:** The profile registry currently **only exists on staging CDN** (`cdn-staging.functions.azure.com`). A follow-up task will publish it to prod CDN before GA.

## Fetch Strategy (per design doc §14.1)

1. **Check local cache** — if within 1-hour TTL, use it directly
2. **Remote fetch** — download registry + SHA-256 checksum in parallel (10s timeout)
3. **Validate checksum** — discard on mismatch
4. **Stale cache fallback** — if remote fails but cache exists (warns if >7 days old)
5. **Empty fallback** — defers to `BuiltInProfileSource` (bundled)

## Changes

- `src/Func/Profiles/RemoteProfileSource.cs` — new profile source with DI-injected `HttpClient`
- `src/Func/Profiles/Sha256Extensions.cs` — `HashDataLowerHex` helper for checksum validation
- `src/Func/Profiles/IProfileFileSystem.cs` / `ProfileFileSystem.cs` — write methods ensure parent directories; atomic writes via temp+rename
- `src/Func/Hosting/ProfileRegistration.cs` — named HttpClient with `BaseAddress` configured at DI time, env var override support
- `test/Func.Tests/Profiles/RemoteProfileSourceTests.cs` — 7 test cases

## Notes

- CDN caching headers (Cache-Control, ETag pass-through) still need configuration — tracked in #5329.

Closes #5332


src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,178 @@
...
+
+namespace Azure.Functions.Cli.Profiles;
+
+/// <summary>
+/// Fetches the built-in profile registry from the CDN, falling back to a local cache.
+/// </summary>
+internal sealed class RemoteProfileSource : IProfileSource
+{
+    internal const string RegistryUrl = "https://cdn-staging.functions.azure.com/public/cli/v5/profiles/v1/registry.json";
+    internal const string ChecksumUrl = "https://cdn-staging.functions.azure.com/public/cli/v5/profiles/v1/registry.json.sha256";
```
liliankasem: We can create a work item to track that we switch to prod before GA. Or do something like we are already doing for the catalog/nuget feed (if the CLI is preview, prerelease is true for all catalog interactions) e.g.

```csharp
// If CLI version has a prerelease suffix (e.g., 5.0.0-preview.1), use staging CDN
private static string GetBaseUrl()
{
    string? version = typeof(RemoteProfileSource).Assembly
        .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
        ?.InformationalVersion;
    
    bool isPrerelease = version?.Contains('-') == true;
    return isPrerelease 
        ? "https://cdn-staging.functions.azure.com" 
        : "https://cdn.functions.azure.com";
}
```

satvu (approved):

satvu (commented):
**Potential race condition on cache writes** (RemoteProfileSource.cs, lines 232–235)

The three cache files are written sequentially:
`csharp
await _fileSystem.WriteAllTextAsync(cachePath, registryJson, cancellationToken);
await _fileSystem.WriteAllTextAsync(cacheChecksumPath, expectedChecksum, cancellationToken);
await _fileSystem.WriteAllTextAsync(cacheMetaPath, DateTimeOffset.UtcNow.ToString("O"), cancellationToken);
`

If the process crashes (or is killed) between these writes, the on-disk cache can end up in an inconsistent state:

1. **Registry written, checksum missing** → next load discards the cache ("no checksum file") and refetches. This is safe but wasteful.
2. **Registry + checksum written, meta missing** → next load finds no timestamp, skips the fresh-cache path, tries remote, fails, then hits the stale-cache path which *also* has no timestamp — so it still works but logs a spurious warning.
3. **Old checksum + new registry** (concurrent CLI invocations writing simultaneously) → checksum mismatch, cache discarded as corrupt.

Case 3 is the most concerning in practice (two \unc\ processes racing on the same home directory).

**Suggestion:** Write to temporary files and atomically rename, or acquire a simple file lock around the cache-write block. Alternatively, if you want to keep it simple, write the meta file *first* (so a partial write looks like an expired cache rather than a corrupt one) and document that concurrent writes may cause a single extra refetch.

src/Func/Profiles/IProfileFileSystem.cs
```diff
@@ -18,4 +18,14 @@ internal interface IProfileFileSystem
     /// </summary>
     public Task WriteAllTextAsync(string path, string contents, CancellationToken cancellationToken);
 
+    /// <summary>
+    /// Writes a text file atomically via temp-file-and-rename.
+    /// </summary>
+    public Task WriteAllTextAtomicAsync(string path, string contents, CancellationToken cancellationToken);
```
jviau: Does this need to be a new method? Or can we have `WriteAllTextAsync` just do this by default?

src/Func/Profiles/IProfileFileSystem.cs
jviau, line 1: Did we evaluate using [System.IO.Abstractions](https://www.nuget.org/packages/TestableIO.System.IO.Abstractions)?

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,230 @@
...
+        ILogger<RemoteProfileSource> logger)
+    {
+        _httpClientFactory = httpClientFactory ?? throw new ArgumentNullException(nameof(httpClientFactory));
+        _parser = parser ?? throw new ArgumentNullException(nameof(parser));
+        _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+        _configurationPaths = configurationPaths ?? throw new ArgumentNullException(nameof(configurationPaths));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    internal static string RegistryUrl => GetBaseUrl() + RegistryPath;
```
jviau: nit: use `Uri` if these represent urls.

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,230 @@
...
+        string cacheChecksumPath,
+        string cacheMetaPath,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(_fetchTimeout);
+
+            using HttpClient client = _httpClientFactory.CreateClient(HttpClientName);
```
jviau: consider just DI'ing with HttpClient preconfigured. Can make the prod vs staging CDN decision at configuration time then.

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private static string GetBaseUrl()
+    {
+        string? version = typeof(RemoteProfileSource).Assembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+            ?.InformationalVersion;
+
+        bool isPrerelease = version?.Contains('-') == true;
+        return isPrerelease ? StagingBaseUrl : ProdBaseUrl;
```
jviau: I believe I added a different way to resolving this for bundles I think? Can you look at that approach? I can't remember where it is, will take a look and get back.

I also unsure about forcing staging CDN when a preview build. This seems to be conflating channels with release quality.

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private static string ComputeSha256(string content)
+    {
+        byte[] hash = SHA256.HashData(System.Text.Encoding.UTF8.GetBytes(content));
```
jviau: Consider adding a static extension method to SHA256:

``` CSharp
extension (SHA256)
{
    public static string HashDataLowerCase(string contents)
    {
        byte[] hash = SHA256.HashData(System.Text.Encoding.UTF8.GetBytes(content));
        return Convert.ToHexStringLower(hash);
    }
}
```

Can now be used as `SHA256.HashDataLowerCase(contents);`

src/Func/Profiles/RemoteProfileSource.cs
liliankasem: Yeah I think you're right on this. I just wanted to use staging for now as that is where I have the profiles uploaded - so it was less about it being preview and more about validation. I don't think we will ever realistically expose the staging profiles to customers/ the GA CLI.

src/Func/Profiles/IProfileFileSystem.cs
```diff
@@ -14,8 +14,13 @@ internal interface IProfileFileSystem
...
     /// <summary>
-    /// Writes a text file.
+    /// Writes a text file, creating parent directories if needed.
     /// </summary>
     public Task WriteAllTextAsync(string path, string contents, CancellationToken cancellationToken);
 
+    /// <summary>
+    /// Writes a text file atomically via temp-file-and-rename, creating parent directories if needed.
+    /// </summary>
+    public Task WriteAllTextAtomicAsync(string path, string contents, CancellationToken cancellationToken);
```
jviau: I see this method is still here after the latest iteration. Is the goal to keep it separate from `WriteAllTextAsync`?

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,196 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Azure.Functions.Cli.Profiles;
+
+/// <summary>
+/// Fetches the profile registry from the CDN, falling back to a local cache.
+/// </summary>
+internal sealed class RemoteProfileSource(
+    IHttpClientFactory httpClientFactory,
```
jviau: We can DI in `HttpClient` directly. Either with keyed services or with the `.AddHttpClient<RemoteProfileSource>`

Keyed services: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientbuilderextensions.addaskeyed?view=net-11.0-pp

Then you use an attribute on your ctor to tell it what key to use.

src/Func/Profiles/RemoteProfileOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Configuration for the remote profile registry source.
+/// </summary>
+internal sealed class RemoteProfileOptions
+{
+    /// <summary>
+    /// CDN base URL from which the profile registry is fetched.
+    /// </summary>
+    public Uri CdnBaseUrl { get; set; } = new("https://cdn.functions.azure.com/public/");
```
jviau: Can we extract this to a constant (or `static readonly` so it can be `Uri`).

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,196 @@
...
+        string cacheDir,
+        string cachePath,
+        string cacheChecksumPath,
+        string cacheMetaPath,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(_options.HttpTimeout);
```
jviau: Timeout can be configured on the HTTP client itself. This way we can separate the HTTP client behavior from this implementation for better testability.

jviau (approved):

src/Func/Profiles/IProfileFileSystem.cs
satvu: Couldn't understand the reason for it and there was an acknowledgement on the last comment/iteration, so I cleaned this up into one.

soninaren (commented):
### Checksum-valid-but-malformed registry poisons the cache and breaks _all_ profile resolution

**File:** `src/Func/Profiles/RemoteProfileSource.cs` (`LoadAsync` — the three unguarded `_parser.ParseBuiltInRegistry(...)` call sites) &nbsp;·&nbsp; **Priority:** Critical &nbsp;·&nbsp; **Blocking**

#### What happens

Remote content is **persisted to the cache and returned _before_ it is parsed**, and the parse is outside any `try/catch` — only network exceptions are caught, and only inside `TryFetchRemoteAsync`. If the CDN publishes a `registry.json` that hashes correctly against its `registry.json.sha256` but is **semantically invalid** (empty body, truncated-then-rehashed, malformed JSON, a forward-incompatible schema, or a bad version range), `ParseBuiltInRegistry` throws _past_ the entire fallback ladder.

Trace:

1. `TryFetchRemoteAsync` validates the checksum (passes), writes `registry.json` + `.sha256` + a fresh `.meta` to the cache, and returns the bad JSON.
2. Back in `LoadAsync`, `_parser.ParseBuiltInRegistry(remoteJson, ...)` throws.
3. The throw escapes `LoadAsync` → escapes `ProfileCatalog.LoadAsync` → fails whatever command needed profiles.
4. The `"return empty so BuiltInProfileSource (bundled fallback) handles it"` safety net is **never reached**.
5. On the next run, `TryLoadValidatedCacheAsync` finds fresh meta (within the 1h TTL), the checksum still matches the poisoned bytes, and `ParseBuiltInRegistry` throws **again** — before remote is ever retried. The stale-cache path parses the same poison and throws too.

The cache never heals until TTL expiry, and if the CDN still serves the same authentic-but-bad content, TTL expiry just re-fetches and **re-poisons**. The one mechanism designed to guarantee graceful degradation instead guarantees a hard failure on every user's machine.

#### Why this is Critical, not just High (the severity-deciding dependency)

The blast radius depends on whether `ProfileCatalog` isolates per-source exceptions. It does **not**:

```csharp
// src/Func/Profiles/ProfileCatalog.cs — LoadAsync
foreach (IProfileSource source in _sources)
{
    cancellationToken.ThrowIfCancellationRequested();
    snapshots.Add(await source.LoadAsync(context, cancellationToken)); // no try/catch
}
```

Because the loop awaits each source with no `try/catch`, an exception thrown by `RemoteProfileSource.LoadAsync` propagates out of the catalog and takes down **all** profile resolution — including the `Project` and `User` sources and the bundled `BuiltInProfileSource` — not just the remote path. A single bad CDN publish therefore breaks profile resolution for every user, which is why this escalates from High to **Critical**.

#### Suggested fix

- **Validate by parse _before_ persisting.** In `TryFetchRemoteAsync`, parse into a throwaway model after the checksum passes and _only then_ write to the cache, so an unparseable registry is never persisted.
- **Wrap all three `ParseBuiltInRegistry` call sites** (cached, remote, stale) in `try/catch`. On parse failure: log, **invalidate/delete the offending cache files**, and continue down the ladder (remote → stale → empty → bundled).
- **Isolate per-source exceptions in `ProfileCatalog.LoadAsync`** so one misbehaving source can never take down the others — defense in depth regardless of the fix above.

#### Test to add

Feed a **checksum-valid but malformed/empty** registry and assert `LoadAsync` returns an empty snapshot (falls through to bundled) instead of throwing — and that the poisoned cache is not re-used on a second call. No current test exercises this path: the suite only feeds valid registries or simulates network failure.

src/Func/Hosting/ProfileRegistration.cs
```diff
@@ -23,7 +24,40 @@ public static IServiceCollection AddProfiles(this IServiceCollection services)
...
+            .Configure(opts =>
+            {
+                string? overrideUrl = Environment.GetEnvironmentVariable(Constants.ProfilesCdnBaseUrlEnvironmentVariable);
+                if (!string.IsNullOrWhiteSpace(overrideUrl))
+                {
+                    if (!Uri.TryCreate(overrideUrl, UriKind.Absolute, out Uri? parsed) ||
+                        (parsed.Scheme != Uri.UriSchemeHttp && parsed.Scheme != Uri.UriSchemeHttps))
+                    {
+                        throw new InvalidOperationException(
+                            $"Environment variable '{Constants.ProfilesCdnBaseUrlEnvironmentVariable}' must be an absolute HTTP(S) URL. Got: '{overrideUrl}'.");
```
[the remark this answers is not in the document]
satvu: This runs inside `PostConfigure` during DI options setup — it's not at the command boundary. Per repo conventions, `GracefulException` is thrown by commands that catch specific failures from their dependencies. An `InvalidOperationException` here is appropriate because it surfaces as a startup/configuration failure, which is the correct signal for a misconfigured environment variable.

src/Func/Profiles/RemoteProfileSource.cs
```diff
@@ -0,0 +1,189 @@
...
+                _logger.LogWarning("Profile registry checksum mismatch. Discarding remote fetch.");
+                return null;
+            }
+
+            // Persist to cache atomically: write to temp files then rename so
+            // concurrent CLI processes never see a half-written cache.
+            await _fileSystem.WriteAllTextAsync(Path.Combine(cacheDir, CacheFileName), registryJson, cancellationToken);
+            await _fileSystem.WriteAllTextAsync(Path.Combine(cacheDir, CacheChecksumFileName), expectedChecksum, cancellationToken);
+            // Meta written last: a partial write looks like an expired cache (safe).
+            await _fileSystem.WriteAllTextAsync(Path.Combine(cacheDir, CacheMetaFileName), DateTimeOffset.UtcNow.ToString("O"), cancellationToken);
```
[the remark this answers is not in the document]
satvu: Fixed — removed the unused `cacheDir` parameter and now using the passed `cachePath`/`cacheChecksumPath`/`cacheMetaPath` directly.

src/Func/Profiles/Sha256Extensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Azure.Functions.Cli.Profiles;
+
+/// <summary>
+/// Convenience helpers for SHA256 hashing.
+/// </summary>
+internal static class Sha256Extensions
```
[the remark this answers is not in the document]
satvu: Acknowledged — the name is slightly misleading since it's a static helper, not an extension method. However this is pre-existing code outside the scope of this change. Will address in a separate cleanup if needed.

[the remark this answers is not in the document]
satvu: Fixed — renamed to `Sha256Helpers` since it's a static helper, not an extension method holder.

soninaren (approved):

PR #5462, opened 2026-07-06
Use default host name for flex health check

### LINKED ISSUE #5465: Use default host name for flex deployment health check
The health check uses `enabledHostNames` which can list multiple names when we only really need/want the `defaultHostName` for a proper check.

satvu:

Resolves #5465 

Leverage [defaultHostName](https://learn.microsoft.com/en-us/dotnet/api/azure.provisioning.appservice.website.defaulthostname?view=azure-dotnet) instead of `.FirstOrDefault()` of `enabledHostNames` property to make sure check goes to the correct site.

### Pull request checklist

  * [] Otherwise: I've added my notes to `release_notes.md`

### Additional Information

Went back through the history w/ an agent, the use of `enabledHostNames` goes all the way to "day one" and has no documented reasoning for why. One potential reason that `enabledHostNames` is used is because it only lists hostnames that are currently enabled/accessible, because `defaultHostName` exists even if the site is stopped or hostname is disabled. 

That still seems incorrect to me. Using `defaultHostName` should be more reliable because it returns the canonical `*.azurewebsites.ne`t hostname, avoiding issues with custom domains in `enabledHostNames` that might have different TLS/routing configurations.

Example of what is returned from each:

{
  "defaultHostName": "some-app-name.azurewebsites.net",
  "enabledHostNames": [
    "some-custom-domain-name.net",
    "some-app-name.azurewebsites.net",
    "some-app-name.scm.azurewebsites.net"
  ]
}

### Testing

- Tested locally e2e, deployed flex app and checked that health check logs were written and check succeeded.

### DIFF
--- a/src/Cli/func/Helpers/AzureHelper.cs
+++ b/src/Cli/func/Helpers/AzureHelper.cs
@@ -443,7 +443,7 @@ namespace Azure.Functions.Cli.Helpers
             var url = new Uri($"{managementURL}{site.SiteId}?api-version={ArmUriTemplates.WebsitesApiVersion}");
             var armSite = await ArmHttpAsync<ArmWrapper<ArmWebsite>>(HttpMethod.Get, url, accessToken);
 
-            site.HostName = armSite.Properties.EnabledHostNames?.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) == -1);
+            site.HostName = armSite.Properties.DefaultHostName;
             site.ScmUri = armSite.Properties.EnabledHostNames?.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) != -1);
             site.Location = armSite.Location;
             site.Kind = armSite.Kind;
--- a/eng/ci/templates/official/steps/prebuild-common.yml
+++ b/eng/ci/templates/official/steps/prebuild-common.yml
@@ -21,9 +21,21 @@ steps:
 
 - template: /eng/ci/templates/steps/restore-nuget.yml@self
 
+- pwsh: |
+    $diff = git diff --name-only origin/main...HEAD
+    if ($diff -match 'eng/build/Packages\.props') {
+      Write-Output "Packages.props changed, running worker version validation..."
+      Write-Output "##vso[task.setvariable variable=runWorkerValidation]true"
+    } else {
+      Write-Output "Packages.props not changed, skipping worker version validation."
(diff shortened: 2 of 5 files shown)


src/Cli/func/Helpers/AzureHelper.cs
```diff
@@ -443,7 +443,7 @@ public static async Task<Site> LoadSiteObjectAsync(Site site, string accessToken
             var url = new Uri($"{managementURL}{site.SiteId}?api-version={ArmUriTemplates.WebsitesApiVersion}");
             var armSite = await ArmHttpAsync<ArmWrapper<ArmWebsite>>(HttpMethod.Get, url, accessToken);
 
-            site.HostName = armSite.Properties.EnabledHostNames?.FirstOrDefault(s => s.IndexOf(".scm.", StringComparison.OrdinalIgnoreCase) == -1);
+            site.HostName = armSite.Properties.DefaultHostName;
```
kshyju, line 446: Could `DefaultHostName` be `null`? If so, should we fallback (to an entry from EnabledHostNames) ?

kshyju, line 446: Is there any downstream impact of chaning this value? I see this property (HostName) is being used in other places (Ex: `CheckFunctionHostStatusForFlex`) . Would they continue to work or this would cause any behavior changes? Good to confirm this is not causing any unexpected behaviors.

kshyju: Can you share a sample value(s) of both DefaultHostName and EnabledHostNames ,in the PR description or linked GH issue?

kshyju (approved):

src/Cli/func/Helpers/AzureHelper.cs
satvu: The same `null` risk was present before so that failure would go the same way (UriFormatException). We do not want to fall back on `enabledHostNames` any longer for security reasons, so if `defaultHostName` doesn't work, we should fail.

satvu: Before opening the PR I had an agent check where else it's used, and it's only used for `CheckFunctionHostStatusForFlex` which is what we are trying to fix. Updated the PR description. `enabledHostNames` includes the defaultHostName and scm site + custom domains. If defaultHostName is missing, it should also be missing from enabledHostNames.

kshyju: Makes sense.

jviau (approved):

PR #5471, opened 2026-07-14
feat(workloads): add Java stack workload (#5320)

### LINKED ISSUE #5320: Implement Java stack workload
Add the Java stack workload that wires the Java worker into `func init`,
`func new`, and `func start` flows (language detection, project scaffolding
hooks, runtime resolution). Worker package itself is tracked separately.

## Definition of Done

- [ ] Stack workload nupkg builds and publishes via the workloads pipeline.
- [ ] CI pipeline (public + official/signed) wired up for the workload.
- [ ] `func init --stack java` produces a valid project.
- [ ] Stack metadata surfaces in `func templates list` and `func --help`.
- [ ] Stack manifest validated by the schema in CI.

ahmedmuhsin:
## What this does

This adds the Java stack workload with alias `java`. It wires Java into `func init`, `func new`, and `func start`. It pairs with the Java worker in #5470. The two are separate packages so they can merge in either order, though worker first is the natural order since the stack pulls in the worker on the first `func start`.

## What is in it

The new code lives under `src/Workloads/Stacks/Java`.

- `JavaWorkload` is the entry point and registers the initializer and the project factory
- `JavaProjectInitializer` handles `func init --stack java`. It scaffolds a Maven project with a `pom.xml`, a sample HTTP trigger at `src/main/java/com/function/Function.java`, a `host.json` with the extension bundle, a `local.settings.json`, a `.gitignore`, and a `.funcignore`. It also contributes the shared `--no-bundles` and `--bundles-channel` options
- `JavaProjectFactory` claims a directory as Java when it sees a `pom.xml`, a Gradle build file, or a `src/main/java` tree
- `JavaFunctionsProject` runs `mvn clean package` before the host starts, then points the host at the staged `target/azure-functions/<app>` output. `func start --no-build` reuses an existing build

I also added the test project with 33 unit tests, registered both projects in `Azure.Functions.Cli.slnx`, and added the release pipeline at `eng/ci/release/official-release.workload.java.yml`.

## Scaffold defaults

The generated `pom.xml` targets Java 21, uses the latest `azure-functions-maven-plugin` 1.42.0, and defaults to Flex Consumption on Linux for deployment. It keeps `azure-functions-java-library` at 3.1.0.

## How I tested it

The build is clean with no warnings and all 33 tests pass. `dotnet pack` produces a valid nupkg with a `workload.json` whose kind is workload and an entry point that points at `JavaWorkload`.

I ran the whole thing end to end with a real `func` build.

- `func init --stack java` scaffolds the project with the expected `pom.xml` (Java 21, plugin 1.42.0, Flex Consumption on Linux), the sample function, and the settings files
- `mvn clean package` on JDK 21 builds and stages `target/azure-functions/<app>` the way `JavaFunctionsProject` expects
- `func start` resolves the project as Java, pulls in the worker, host, and bundles, and the HTTP trigger returns a 200 locally
- I deployed to a real Flex Consumption Linux app in Azure with `mvn azure-functions:deploy`. The app came up as Flex Consumption running Java 21 and the endpoint returned a 200 with the expected greeting. I deleted the app afterward

## Notes

`func start` needs JDK 21 with `JAVA_HOME` set, and Maven on the path or a `mvnw` wrapper in the project. If Maven is missing or the build fails the stack surfaces a clear error.

`func new` template listing goes through the repo wide `ITemplateProvider` contribution point, which isn't implemented for any stack yet. This workload registers the same way the Go and Python stacks do.

## Definition of done

- [x] `func init --stack java` produces a valid Maven project, verified end to end and it builds with `mvn package`
- [x] The stack registers through the workload model and shows up in `func --help` and the workload list once installed
- [x] The workload manifest matches the schema with kind workload and an entry point
- [x] The stack workload nupkg builds and the release pipeline is wired

Closes #5320

### DIFF
--- a/src/Workloads/Stacks/Java/JavaFunctionsProject.cs
+++ b/src/Workloads/Stacks/Java/JavaFunctionsProject.cs
@@ -1,0 +1,280 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Diagnostics;
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Projects;
+using Azure.Functions.Cli.Workers;
+
+namespace Azure.Functions.Cli.Workloads.Java;
+
+/// <summary>
+/// Java Functions project. Before the host runs, builds the user's Maven module
+/// with <c>mvn clean package</c> so the Azure Functions Maven plugin stages the
+/// app under <c>target/azure-functions/&lt;appName&gt;</c>, then points the host
+/// startup directory at that staged output.
+/// </summary>
+internal sealed class JavaFunctionsProject : FunctionsProject
+{
+    private const string StagedAppRootRelativePath = "target/azure-functions";
+    private const string HostJsonFileName = "host.json";
+
+    private readonly WorkingDirectory _workingDirectory;
+    private readonly FunctionsWorkerReference _workerReference;
+
+    public JavaFunctionsProject(WorkingDirectory workingDirectory)
+    {
+        _workingDirectory = workingDirectory ?? throw new ArgumentNullException(nameof(workingDirectory));
+
+        // The Java worker's worker.config.json declares language "java", which matches the
+        // workload id, so no runtime override is needed (unlike the Go worker's "native").
+        _workerReference = FunctionsWorkerReference.FromWorkload("java");
+    }
+
+    // Internal seams so tests can stub the Maven discovery/build without spawning real processes.
+    internal Func<string, CancellationToken, Task<string?>> ResolveMavenCommand { get; set; } = DefaultResolveMavenCommandAsync;
+
+    internal Func<string, string, Action<string>, Action<string>, CancellationToken, Task<int>> RunMavenPackage { get; set; } = DefaultRunMavenPackageAsync;
+
+    public override WorkingDirectory WorkingDirectory => _workingDirectory;
+
+    public override string StackName => "java";
+
+    public override string StackDisplayName => "Java";
+
+    public override bool SupportsExtensionBundles => true;
+
+    public override FunctionsWorkerReference WorkerReference => _workerReference;
+
+    public override async Task PrepareForHostRunAsync(FunctionsProjectHostRunContext context, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+        cancellationToken.ThrowIfCancellationRequested();
+
+        string root = _workingDirectory.Info.FullName;
+
+        if (!context.SkipBuild)
+        {
+            string? mavenCommand = await ResolveMavenCommand(root, cancellationToken).ConfigureAwait(false);
+            if (mavenCommand is null)
+            {
+                throw new GracefulException(
+                    "Could not find Apache Maven. Install Maven 3.6 or later (https://maven.apache.org/install.html) "
+                    + "and ensure 'mvn' is on your PATH, or add a Maven wrapper ('mvnw') to the project.",
+                    isUserError: true);
+            }
+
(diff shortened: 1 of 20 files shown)


src/Workloads/Stacks/Java/JavaFunctionsProject.cs
```diff
@@ -0,0 +1,212 @@
...
+            {
+                onErrorLine(e.Data);
+            }
+        };
+
+        process.Start();
+        process.BeginOutputReadLine();
+        process.BeginErrorReadLine();
+
+        await process.WaitForExitAsync(cancellationToken).ConfigureAwait(false);
```
satvu: Do we need to also kill the `mvn` process (or entire process tree) explicitly?

src/Workloads/Stacks/Java/JavaFunctionsProject.cs
```diff
@@ -0,0 +1,212 @@
...
+        return process.ExitCode;
```
satvu: Before exiting, we can wait for output to fully drain:

await stdoutComplete.Task.ConfigureAwait(false);
await stderrComplete.Task.ConfigureAwait(false);

src/Workloads/Stacks/Java/JavaFunctionsProject.cs
ahmedmuhsin: Good call. Wrapped the wait in a try/finally that calls `process.Kill(entireProcessTree: true)` guarded by `HasExited`, so a Ctrl+C or a failed build no longer leaves `mvn` and its child JVM running. Mirrors the Go stack runner.

src/Workloads/Stacks/Java/JavaFunctionsProject.cs
ahmedmuhsin: Done. Added stdout and stderr `TaskCompletionSource` sentinels and await both after `WaitForExitAsync`, so the trailing build output is not dropped. Mirrors the Go runner. Verified end to end with a real `func start` build, which also surfaced that `mvn.cmd` has to run through `cmd.exe` on Windows, fixed here too.

src/Workloads/Stacks/Java/JavaProjectFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Azure.Functions.Cli.Workloads.Java;
+
+/// <summary>
+/// Creates Java Functions projects from Java-specific fingerprints
+/// (Maven <c>pom.xml</c>, Gradle build scripts, or a <c>src/main/java</c> tree).
+/// </summary>
+internal sealed class JavaProjectFactory : IFunctionsProjectFactory
+{
+    private static readonly FunctionsWorkerId _workerId = new("java");
+
```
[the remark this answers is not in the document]
ahmedmuhsin: Removed the unused `_workerId` field and its now-unused `Workers` using. It was not actually breaking the build (the same field is in the Go, Node, and Python factories and CI is green), but it is cleaner gone.

src/Workloads/Stacks/Java/JavaProjectInitializer.cs
```diff
@@ -0,0 +1,134 @@
...
+                Path.Combine(root, "host.json"),
+                host => EnsureExtensionBundle(host, channel));
+        }
+
+        return Task.CompletedTask;
+    }
+
+    private static void EnsureExtensionBundle(JsonObject host, BundleChannel channel)
+    {
+        // Only fill in when missing so a user-customised bundle survives `--force`.
```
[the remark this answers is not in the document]
ahmedmuhsin: Good catch. Reworded the comment to match the behavior. The guard preserves an existing bundle on a re-init without `--force`, and a forced re-init recreates host.json first so the default bundle is written.

satvu (approved):

PR #5483, opened 2026-07-17
Wire `func start`/`func run` for PowerShell Stack

### LINKED ISSUE #5480: PowerShell `func start`
Tracking completion of e2e implementation of `func start` for PowerShell stack.

satvu:

resolves #5480

**Wire PowerShell workload with `func run`/`func start`**

Adds project factory and project class so the PowerShell workload participates in project resolution during `func start`/`func run`.

### Changes

- **`PowerShellProjectFactory`** — fingerprints PowerShell projects (`profile.ps1`, `requirements.psd1`, `*.ps1`, `Modules/` folder)
- **`PowerShellFunctionsProject`** — `FunctionsProject` implementation (interpreted, no build step needed)
- **`PowerShellWorkload.Configure`** — registers the factory via `AddProjectFactory`
- Tests for all new classes

### Notes

PowerShell is an interpreted language so `PrepareForHostRunAsync` is a no-op — the host worker handles module resolution at runtime via `requirements.psd1` and the `Modules` folder.

### Testing
Local E2E testing was performed. Full end-to-end verified:

✅ resolve_project complete: PowerShell — factory fingerprints profile.ps1
✅ resolve_worker complete: powershell 4.0.4778 — worker resolved from installed workload
✅ resolve_bundle complete: Bundle ... 4.35.0 — extension bundles loaded
✅ HttpTrigger loaded at http://localhost:7071/api/HttpTrigger
✅ curl returned Hello from PowerShell Functions!
The PowerShell workload is fully wired with func start.

### DIFF
--- a/src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
+++ b/src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
@@ -1,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Azure.Functions.Cli.Common;
+using Azure.Functions.Cli.Projects;
+using Azure.Functions.Cli.Workers;
+
+namespace Azure.Functions.Cli.Workloads.PowerShell;
+
+/// <summary>
+/// PowerShell Functions project. PowerShell is an interpreted language with no
+/// compile step, so <see cref="PrepareForHostRunAsync"/> is a no-op. The host
+/// worker handles module resolution at runtime via <c>requirements.psd1</c> and
+/// the <c>Modules</c> folder.
+/// </summary>
+internal sealed class PowerShellFunctionsProject : FunctionsProject
+{
+    private readonly WorkingDirectory _workingDirectory;
+    private readonly FunctionsWorkerReference _workerReference;
+
+    public PowerShellFunctionsProject(WorkingDirectory workingDirectory)
+    {
+        _workingDirectory = workingDirectory ?? throw new ArgumentNullException(nameof(workingDirectory));
+        _workerReference = FunctionsWorkerReference.FromWorkload("powershell");
+    }
+
+    public override WorkingDirectory WorkingDirectory => _workingDirectory;
+
+    public override string StackName => "powershell";
+
+    public override string StackDisplayName => "PowerShell";
+
+    public override bool SupportsExtensionBundles => true;
+
+    public override FunctionsWorkerReference WorkerReference => _workerReference;
+}
--- a/src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
+++ b/src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
@@ -15,18 +15,20 @@ namespace Azure.Functions.Cli.Workloads.PowerShell;
 /// </summary>
 internal sealed class PowerShellFunctionsProject : FunctionsProject
 {
+    internal const string WorkloadId = "powershell";
+
     private readonly WorkingDirectory _workingDirectory;
     private readonly FunctionsWorkerReference _workerReference;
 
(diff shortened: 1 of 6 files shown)


src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
```diff
@@ -0,0 +1,36 @@
...
+    public override WorkingDirectory WorkingDirectory => _workingDirectory;
+
+    public override string StackName => "powershell";
+
+    public override string StackDisplayName => "PowerShell";
+
+    public override bool SupportsExtensionBundles => true;
+
+    public override FunctionsWorkerReference WorkerReference => _workerReference;
+}
```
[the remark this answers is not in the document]
satvu: The style as-is matches the other stacks (https://github.com/Azure/azure-functions-core-tools/blob/vnext/src/Workloads/Stacks/Node/NodeFunctionsProject.cs).

src/Workloads/Stacks/PowerShell/PowerShellFunctionsProject.cs
```diff
@@ -0,0 +1,36 @@
...
+/// </summary>
+internal sealed class PowerShellFunctionsProject : FunctionsProject
+{
+    private readonly WorkingDirectory _workingDirectory;
+    private readonly FunctionsWorkerReference _workerReference;
+
+    public PowerShellFunctionsProject(WorkingDirectory workingDirectory)
+    {
+        _workingDirectory = workingDirectory ?? throw new ArgumentNullException(nameof(workingDirectory));
+        _workerReference = FunctionsWorkerReference.FromWorkload("powershell");
```
soninaren: Wondering if we should store this as constant string somewhere.

src/Workloads/Stacks/PowerShell/PowerShellProjectFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        // requirements.psd1 declares managed dependencies for PowerShell Functions.
+        if (File.Exists(Path.Combine(root, "requirements.psd1")))
+        {
+            return "found requirements.psd1";
+        }
+
+        // Fallback: any *.ps1 at the project root.
+        if (Directory.EnumerateFiles(root, "*.ps1", SearchOption.TopDirectoryOnly).Any())
```
soninaren: From Co-pilot:

### `*.ps1` fallback matches more than `.ps1` files on Windows

**File:** `src/Workloads/Stacks/PowerShell/PowerShellProjectFactory.cs` &nbsp;·&nbsp; **Priority:** Medium

`Directory.EnumerateFiles(root, "*.ps1", ...)` isn't a strict `.ps1`-only match on Windows. When the search-pattern extension is exactly three characters, the Win32 matcher also matches files whose extension *starts with* those characters (a legacy 8.3 behavior). So a directory containing only, say, `MyModule.Format.ps1xml` (a common, non-runnable PowerShell format file) — or `x.ps1backup` — gets fingerprinted as a PowerShell Functions project.

**Two consequences:**

1. **False positives** — a directory with no actual `.ps1` script matches the fallback.
2. **Platform divergence** — on Linux/macOS the same call matches *none* of these (extension is `ps1xml`, not `ps1`, and matching is case-sensitive), so the same directory resolves differently on Windows vs. CI/WSL — the kind of "works on my machine" report that's painful to diagnose.

**Suggested fix** — filter by exact extension:

```csharp
if (Directory.EnumerateFiles(root, "*.ps1", SearchOption.TopDirectoryOnly)
        .Any(f => Path.GetExtension(f).Equals(".ps1", StringComparison.OrdinalIgnoreCase)))
{
    return "found *.ps1 file";
}
```

Worth adding a test with a `.ps1xml` file present (and no real `.ps1`) asserting `NotCreated`, since the current `*.ps1` case is only covered by the positive theory.

soninaren (approved):
Minor comments. Looks good otherwise.

PR #5484, opened 2026-07-17
fix: correct func startup binding

jviau:
## Summary

Updates `func start` local host binding and makes the bind address configurable.

Changes in `src/Cli/func/Actions/HostActions/StartHostAction.cs` (and `src/Cli/func/Common/HostStartSettings.cs`):
- **Default binding**: both the HTTP path (`Setup()`) and the HTTPS path (Kestrel `Listen`) bind to the IPv4 loopback address (`127.0.0.1` / `IPAddress.Loopback`).
- **New opt-in `--address` flag**: lets you specify the local IP address to bind to. When unset it defaults to `127.0.0.1`.
- **New `local.settings.json` setting**: `Host.LocalHttpAddress` provides the same override via config.
- **Precedence**: `--address` flag → `Host.LocalHttpAddress` → `127.0.0.1`.

The address is resolved once via a shared `ResolveBindAddress()` helper that validates the value and throws a clear `CliException` on invalid input. It feeds both binding paths (the listen URI is built through an `IPEndPoint` so IPv6 values are bracketed correctly). `BaseUri` and displayed function URLs remain `localhost`, so normal local development output is unchanged.

## Notes

- Default behavior is localhost-only; broadening the bind address is strictly opt-in.
- No change to IPv6 defaults.
- Existing localhost-based workflows and tests (which connect via `http://localhost:{port}`) are unaffected.
- `NetworkHelpers` and unrelated build assets are untouched.

## Validation

- `dotnet build` of `src/Cli/func` succeeds with 0 warnings / 0 errors.
- 4 new unit tests in `StartHostActionTests.cs` cover the default, the `--address` override, the config-supplied value, and invalid-input handling — all pass.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -19,3 +19,5 @@
 - Bumped https-proxy-agent dependency (#5335)
 - Updated target framework to .NET 10 (#4850)
 - Fix Flex Health Check to use `defaultHostName` instead of `enabledHostNames` (#5462)
+- Changed `func start` to bind to the IPv4 loopback address (`127.0.0.1`) by default instead of `0.0.0.0`, and added an opt-in `--address` flag (and `Host.LocalHttpAddress` setting in `local.settings.json`) to override the bind address (#5484)
+  - **Potential breaking change:** the host now binds to `127.0.0.1` by default, so it is only reachable from the local machine. Setups that relied on binding to `0.0.0.0` — for example reaching the host from outside a Docker container via a published port — will no longer connect. Workaround: start the host with `func start --address 0.0.0.0` (or set `"LocalHttpAddress": "0.0.0.0"` under `Host` in `local.settings.json`) to restore the previous behavior.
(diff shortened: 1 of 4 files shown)


satvu (approved):

satvu (approved):

kshyju (approved):
1) Release notes entry missing?

2) My agent review flagged this. Is this something we should mention in the "breaking change" docs?:

>The 0.0.0.0 → 127.0.0.1 default change breaks a common workflow: running func start in a container and reaching it from the host via a mapped port (docker run -p 7071:7071, Compose, etc.). Docker forwards host traffic to the container's eth0, not loopback, so a loopback-only bind refuses those connections — a confusing "connection refused" on a port that's clearly mapped. Same for container-to-container and WSL2 → Windows host. This worked before with no flags; now it requires --address 0.0.0.0.

kshyju (approved):

PR #5488, opened 2026-07-21
Make host.json optional across func publish, and pack

gavin-aguiar:
## Summary

Aligns Core Tools with the fact that the Azure Functions runtime does not require a
`host.json` file. Previously `func azure functionapp publish` and `func pack` hard-failed
when a project had no `host.json` — root detection used `host.json` as the sole project
marker, so publish/pack aborted before packaging. This PR makes `host.json` optional for
`func pack` and `func azure functionapp publish` across all runtimes, while leaving
projects that *do* have a `host.json` completely unchanged.

Note: this PR does **not** change `func start`. The host runtime still synthesizes and
writes a default `host.json` during `func start` when one is missing; suppressing that
write requires a dedicated host-side option and will be handled separately (reusing the
host's `IsFileSystemReadOnly` flag was rejected because it also disables jobhost file
watching and the in-memory extension bundle).

Verified with a Python v2 (`function_app.py`) single HTTP-trigger project with no
`host.json`: `func pack` produces a deployable zip containing the app files and no
`host.json`.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Cli/func/Actions/HostActions/Startup.cs
+++ b/src/Cli/func/Actions/HostActions/Startup.cs
@@ -94,6 +94,14 @@ namespace Azure.Functions.Cli.Actions.HostActions
                 o.LogPath = _hostOptions.LogPath;
                 o.IsSelfHost = _hostOptions.IsSelfHost;
                 o.SecretsPath = _hostOptions.SecretsPath;
+
+                // Only force read-only (never clear it), so a real read-only deployment
+                // detected by the host is preserved. This prevents the host from writing a
+                // default host.json to the project when the project doesn't have one.
+                if (_hostOptions.IsFileSystemReadOnly)
+                {
+                    o.IsFileSystemReadOnly = true;
+                }
             });
 
             services.AddSingleton<IConfigureBuilder<IConfigurationBuilder>>(_ => new ExtensionBundleConfigurationBuilder(_hostOptions));
--- a/src/Cli/func/Actions/HostActions/Startup.cs
+++ b/src/Cli/func/Actions/HostActions/Startup.cs
@@ -94,14 +94,6 @@ namespace Azure.Functions.Cli.Actions.HostActions
                 o.LogPath = _hostOptions.LogPath;
                 o.IsSelfHost = _hostOptions.IsSelfHost;
                 o.SecretsPath = _hostOptions.SecretsPath;
-
-                // Only force read-only (never clear it), so a real read-only deployment
-                // detected by the host is preserved. This prevents the host from writing a
-                // default host.json to the project when the project doesn't have one.
-                if (_hostOptions.IsFileSystemReadOnly)
-                {
-                    o.IsFileSystemReadOnly = true;
-                }
             });
 
             services.AddSingleton<IConfigureBuilder<IConfigurationBuilder>>(_ => new ExtensionBundleConfigurationBuilder(_hostOptions));
--- a/src/Cli/func/Actions/LocalActions/PackAction/CustomPackSubcommandAction.cs
+++ b/src/Cli/func/Actions/LocalActions/PackAction/CustomPackSubcommandAction.cs
@@ -34,14 +34,25 @@ namespace Azure.Functions.Cli.Actions.LocalActions.PackAction
             {
                 dir =>
                 {
+                    var validateCustomHandlerTitle = "Validate Custom Handler Configuration";
+                    var hostJsonPath = Path.Combine(dir, Constants.HostJsonFileName);
+
+                    // host.json is optional. Without it there is no custom handler configuration
+                    // to validate, so surface a non-blocking warning instead of failing the pack.
+                    if (!FileSystemHelpers.FileExists(hostJsonPath))
+                    {
+                        PackValidationHelper.DisplayValidationWarning(
+                            validateCustomHandlerTitle,
+                            $"No {Constants.HostJsonFileName} found. Skipping custom handler configuration validation.");
+                        return;
+                    }
+
                     // Validate custom handler configuration and executable
                     try
                     {
-                        var hostJsonPath = Path.Combine(dir, Constants.HostJsonFileName);
                         var hostJsonContent = FileSystemHelpers.ReadAllTextFromFileAsync(hostJsonPath).Result;
                         var hostConfig = JObject.Parse(hostJsonContent);
                         var customHandler = hostConfig["customHandler"];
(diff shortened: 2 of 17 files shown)


src/Cli/func/Actions/HostActions/Startup.cs
```diff
@@ -94,6 +94,14 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
                 o.LogPath = _hostOptions.LogPath;
                 o.IsSelfHost = _hostOptions.IsSelfHost;
                 o.SecretsPath = _hostOptions.SecretsPath;
+
+                // Only force read-only (never clear it), so a real read-only deployment
+                // detected by the host is preserved. This prevents the host from writing a
+                // default host.json to the project when the project doesn't have one.
+                if (_hostOptions.IsFileSystemReadOnly)
+                {
+                    o.IsFileSystemReadOnly = true;
```
satvu: This will also affect file watching since `options.FileWatchingEnabled` leverages this value (see [here](https://github.com/Azure/azure-functions-host/blob/22cc6f25cb0df09b311321d15696cd309600f8a2/src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs#L72)).

gavin-aguiar: Removed func start changes from this PR

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
```diff
@@ -41,9 +41,11 @@ public static string ResolveOutputPath(string functionAppRoot, string outputPath
 
         public static void ValidateFunctionAppRoot(string functionAppRoot)
         {
+            // host.json is optional and Core Tools no longer adds one, so its absence is not a
```
satvu: can you update the comments so that they are accurate? the host still adds one so I would just remove "and Core tools ... adds one" and leave the rest. There's a few comments like this to be updated.

test/Cli/Func.UnitTests/ActionsTests/PackAction/GoPackSubcommandActionTests.cs
```diff
@@ -39,15 +39,15 @@ public void ValidateFunctionApp_ValidStructure_PassesValidation()
         }
 
         [Fact]
-        public void ValidateFunctionApp_MissingHostJson_ThrowsCliException()
+        public void ValidateFunctionApp_MissingHostJson_DoesNotThrow()
         {
+            // host.json is optional. Go pack should still validate on go.mod alone.
```
satvu, line 44: is `go.mod` enough of a marker to determine a function app? Even if `host.json` isn't required, we may need another marker to use to make sure we can fail fast if we're not in a function app.

satvu (commented):
Please update the description and title as well to match the new scope of this PR

test/Cli/Func.UnitTests/ActionsTests/PackAction/GoPackSubcommandActionTests.cs
gavin-aguiar: To even reach Go pack , the worker runtime must first resolve to  WorkerRuntime.Go . That only happens with explicit Functions configuration  `WorkerRuntimeLanguageHelper.GetCurrentWorkerRuntimeLanguage`

•  FUNCTIONS_CLI_NATIVE_LANGUAGE=go  or
•  FUNCTIONS_WORKER_RUNTIME=native  +  go.mod  present.

A plain Go project with just a  go.mod  and no Functions config resolves to  None , and  PackAction.RunAsync  throws "Unable to determine the worker runtime" before any Go logic runs. So what we have here should be enough for a go function app.

src/Cli/func/Actions/LocalActions/PackAction/PackHelpers.cs
gavin-aguiar: Updated

src/Cli/func/Helpers/ScriptHostHelpers.cs
```diff
@@ -17,7 +17,19 @@ public static string GetFunctionAppRootDirectory(string startingDirectory, IEnum
                 return startingDirectory;
             }
 
-            searchFiles = searchFiles ?? new List<string> { ScriptConstants.HostMetadataFileName };
+            // host.json is no longer strictly required in a Functions project (the runtime
+            // synthesizes a default when it is missing). Detect the project root using any of
+            // the common Functions project markers (across languages) so publishing/packing
+            // works without host.json.
+            searchFiles = searchFiles ?? new List<string>
```
satvu: Did a check - this is used by more than just publish (it's also used as part of `func start` and `func init` for `dotnet`). 

I think to keep this scoped for `func publish` and `func pack` only, it should be refactored. Those two commands pass in this list in `searchFiles` parameter instead. 

You can also make the passed-in list more "runtime aware" since publish/pack would have resolved the worker runtime, so if you know the runtime is node just pass in a check for `package.json`.

src/Cli/func/Actions/LocalActions/PackAction/CustomPackSubcommandAction.cs
```diff
@@ -34,14 +34,25 @@ protected internal override void ValidateFunctionApp(string functionAppRoot, Pac
...
+                    var validateCustomHandlerTitle = "Validate Custom Handler Configuration";
+                    var hostJsonPath = Path.Combine(dir, Constants.HostJsonFileName);
+
+                    // host.json is optional. Without it there is no custom handler configuration
+                    // to validate, so surface a non-blocking warning instead of failing the pack.
+                    if (!FileSystemHelpers.FileExists(hostJsonPath))
+                    {
+                        PackValidationHelper.DisplayValidationWarning(
+                            validateCustomHandlerTitle,
+                            $"No {Constants.HostJsonFileName} found. Skipping custom handler configuration validation.");
```
satvu: CustomHandler apps won't work w/o the following setting that is typically in `host.json`:

`Ensure the custom handler executable is configured using the  AzureFunctionsJobHost__customHandler__description__defaultExecutablePath  application setting after deployment.`

The official docs only document the `host.json` use: https://learn.microsoft.com/en-us/azure/azure-functions/functions-custom-handlers#hostjson

I think this is worth raising to customers in the warning message.

satvu (approved):

satvu (approved):
I think this change should be added to the release notes.

vrdmr (approved):

PR #5494, opened 2026-07-24
Fix worker resolution failures by setting WEBSITE_SKU in host process environment

### LINKED ISSUE #5218: [v5] Worker resolution sometimes fails in the func run experience even when the worker is present
### Version

5.0.0-preview.1.ci.26279.0+3a2339c01cbb94ca4d4468c5e296d51c6c2b9380

### OS / shell

Windows, PowerShell

### Stack

TBD

### Description

Worker resolution seems to sometimes fail in the `func run` experience, even when the worker is present. This needs further investigation to identify the exact conditions and root cause.

### Steps to reproduce

Details pending. Observed intermittently in the `func run` experience when the expected worker is present but worker resolution still fails.

### Expected behavior

`func run` should resolve and use the available worker when the worker is present.

### Actual behavior

`func run` sometimes fails worker resolution despite the worker being present.

### Stack trace / logs

TBD

satvu:
## Summary

Fixes #5218

Sets `WEBSITE_SKU=Dynamic` in the host child process environment so that worker profile conditions referencing the SKU host property evaluate correctly during local development with `func run`.

## Root Cause

When `func run` launches the host as a child process, the host internally uses `WorkerProfileManager` to evaluate `worker.config.json` profiles. These profiles can include `HostPropertyCondition` entries (e.g., for `Sku`, `platform`) that resolve via `ScriptSettingsManager.GetSetting()` — which reads environment variables.

In the local dev environment, `WEBSITE_SKU` is unset, so:

1. `HostPropertyCondition.Evaluate()` returns `false` for any SKU-gated condition
2. Unresolvable conditions are replaced with `FalseCondition()` (Info-level log only)
3. No profiles match → host falls back to the base `description` (often a non-runnable placeholder)
4. `ThrowIfFileNotExists` throws in `BuildWorkerConfig`, caught by `catch when (!ex.IsFatal())`, logged generically as "Failed to initialize worker provider", returns `null`
5. Worker isn't added → "no worker for runtime"

The failure is **silent** because every swallow point only emits Info/Debug-level logs.

## Fix

Added `WEBSITE_SKU=Dynamic` via `TryAdd` in `HostProcessStartInfoFactory.Create()` — the single point where the host child process environment is built. `TryAdd` ensures user-provided values (e.g., from `local.settings.json`) take precedence.

### Why `Dynamic`?

`WEBSITE_SKU` maps to the Azure Functions hosting plan:

| Value | Plan |
|-------|------|
| `Dynamic` | Consumption (serverless, scale-to-zero) |
| `ElasticPremium` | Premium (pre-warmed, VNET) |
| `FlexConsumption` | Flex Consumption |
| *(unset)* | Dedicated / App Service |

`Dynamic` (Consumption) was chosen because:

1. **Most common local dev scenario** — the majority of Azure Functions projects target the Consumption plan, so local behavior should match that by default.
2. **Least restrictive profile evaluation** — Consumption is the baseline SKU that worker profiles are designed to support. Premium/Flex profiles add capabilities on top; Consumption profiles represent the minimal runnable configuration.
3. **Matches legacy Core Tools behavior** — the v4 `func start` experience implicitly operated under Consumption semantics locally.
4. **`TryAdd` preserves user intent** — if a developer sets `WEBSITE_SKU` in `local.settings.json` (e.g., to test Premium-specific behavior), that value is respected and not overwritten.

## Testing

- Added assertion in existing test to verify `WEBSITE_SKU` is set by default
- Added new test `Create_WhenWebsiteSkuAlreadySet_PreservesUserValue` to verify user-provided SKU is not overwritten
- All `HostProcessStartInfoFactory` tests pass

### DIFF
--- a/src/Func/Commands/Start/Host/HostProcessStartInfoFactory.cs
+++ b/src/Func/Commands/Start/Host/HostProcessStartInfoFactory.cs
@@ -14,6 +14,8 @@ internal sealed class HostProcessStartInfoFactory
     public const string ScriptRootEnvironmentVariable = "AzureWebJobsScriptRoot";
     public const string AzureFunctionsEnvironmentVariable = "AZURE_FUNCTIONS_ENVIRONMENT";
     public const string WebsiteHostnameEnvironmentVariable = "WEBSITE_HOSTNAME";
+    public const string WebsiteSkuEnvironmentVariable = "WEBSITE_SKU";
+    public const string DefaultLocalWebsiteSku = "Dynamic";
     public const string AspNetCoreSuppressStatusMessagesEnvironmentVariable = "ASPNETCORE_SUPPRESSSTATUSMESSAGES";
     public const string HostingLifetimeLogLevelEnvironmentVariable = "Logging__LogLevel__Microsoft.Hosting.Lifetime";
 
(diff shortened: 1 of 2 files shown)


satvu:
Closing as we are not 100% confident on RCA

PR #5500, opened 2026-07-29
Support .NET Isolated Debug and JSON file output options

### LINKED ISSUE #5219: [v5] No support for .NET isolated debug and JSON output file options in the start experience
### Version

5.0.0-preview.1.ci.26279.0+3a2339c01cbb94ca4d4468c5e296d51c6c2b9380

### OS / shell

Windows, PowerShell

### Stack

.NET isolated

### Description

The .NET stack/project start experience does not appear to support the .NET debug flags and environment variable injection flow that existed in previous Core Tools versions.

Previous Core Tools start support includes `--dotnet-isolated-debug`, which pauses the .NET Worker process until a debugger is attached, plus `--enable-json-output` and `--json-output-file`, which enable JSON line output and optionally write that output to a file.

The implementation also flows environment variables/startup hooks such as `FUNCTIONS_ENABLE_DEBUGGER_WAIT=true`, `FUNCTIONS_ENABLE_JSON_OUTPUT=true`, and `DOTNET_STARTUP_HOOKS=Microsoft.Azure.Functions.Worker.Core` when the relevant options are enabled.

Reference implementation: Azure/azure-functions-core-tools `src/Cli/func/Actions/HostActions/StartHostAction.cs`, options `dotnet-isolated-debug`, `enable-json-output`, and `json-output-file`.

### Steps to reproduce

1. Use a .NET isolated Functions project.
2. Start the project through the new start/run experience.
3. Try to enable the previous Core Tools behavior equivalent to `func start --dotnet-isolated-debug --enable-json-output --json-output-file <path>`.
4. Observe that the .NET isolated debug and JSON output file options are not available in the experience.

### Expected behavior

The .NET stack/project should expose options equivalent to previous Core Tools support for `--dotnet-isolated-debug`, `--enable-json-output`, and `--json-output-file`, including the required environment variable/startup hook flow and JSON output file piping.

### Actual behavior

The start experience does not expose support for the .NET isolated debug flow or JSON output file flow.

### Stack trace / logs

N/A

satvu:

resolves #5219 

### Pull request checklist

### Additional information

Tested locally e2e w/ Visual Studio (attached debugger, hit breakpoint for an HTTP function from CLI run).

- v4 has a known gap w/ VS Code for this feature (json output support, logging output) so just testing VS parity.

### DIFF
--- a/src/Abstractions/Projects/StartHostConfiguration.cs
+++ b/src/Abstractions/Projects/StartHostConfiguration.cs
@@ -1,0 +1,30 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Projects;
+
+/// <summary>
+/// Stack-specific host run adjustments a workload derives from the parsed
+/// <c>func start</c>/<c>func run</c> options (see <see cref="IStartHostOptionContributor.Configure"/>).
+/// Applied by the CLI only when the resolved project's stack matches the contributing workload.
+/// </summary>
+public sealed record StartHostConfiguration
+{
+    /// <summary>
+    /// A configuration that changes nothing about the host run.
+    /// </summary>
+    public static StartHostConfiguration Empty { get; } = new();
+
+    /// <summary>
+    /// Environment variables to inject into the host process. Merged over other sources, so a
+    /// value here wins. Empty when the workload's options are not enabled.
+    /// </summary>
+    public IReadOnlyDictionary<string, string> EnvironmentVariables { get; init; }
+        = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// When set, the CLI mirrors the host process's raw (JSON) standard output to this file path.
+    /// <c>null</c> disables file capture.
+    /// </summary>
+    public string? JsonOutputFilePath { get; init; }
+}
--- a/src/Abstractions/Projects/StartHostConfiguration.cs
+++ b/src/Abstractions/Projects/StartHostConfiguration.cs
@@ -27,4 +27,11 @@ public sealed record StartHostConfiguration
     /// <c>null</c> disables file capture.
     /// </summary>
     public string? JsonOutputFilePath { get; init; }
+
+    /// <summary>
+    /// An optional message the CLI shows the user before the host starts (e.g. a note that the
+    /// worker will pause until a debugger attaches). The workload owns the wording; the CLI just
+    /// renders it. <c>null</c> shows nothing.
+    /// </summary>
+    public string? StartupNotice { get; init; }
 }
--- a/src/Abstractions/Projects/StartHostConfiguration.cs
+++ b/src/Abstractions/Projects/StartHostConfiguration.cs
@@ -31,10 +31,11 @@ public sealed record StartHostConfiguration
         = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
     /// <summary>
-    /// When set, the CLI mirrors the host process's raw (JSON) standard output to this file path.
-    /// <c>null</c> disables file capture.
+    /// An optional interceptor that inspects raw stdout lines from the host process before they
+    /// enter the log pipeline. Lines the interceptor consumes are suppressed from rendering.
+    /// The CLI disposes this when the host event stream shuts down.
     /// </summary>
-    public string? JsonOutputFilePath { get; init; }
+    public IHostOutputInterceptor? OutputInterceptor { get; init; }
 
     /// <summary>
     /// An optional message the CLI shows the user before the host starts (e.g. a note that the
--- a/src/Abstractions/Projects/StartHostConfiguration.cs
+++ b/src/Abstractions/Projects/StartHostConfiguration.cs
@@ -11,6 +11,14 @@ namespace Azure.Functions.Cli.Projects;
 public sealed record StartHostConfiguration
 {
     /// <summary>
+    /// Environment variable prefix the CLI uses to defer variables that must not be set on the host
+    /// process itself (e.g. DOTNET_STARTUP_HOOKS) but should be inherited by worker child processes.
+    /// The host strips this prefix post-boot and re-emits the target variable. Both the CLI workload
+    /// and the host must agree on this value.
+    /// </summary>
+    public const string DeferredWorkerEnvironmentPrefix = "FUNCTIONS_CORETOOLS_DEFER_ENV__";
+
+    /// <summary>
     /// A configuration that changes nothing about the host run.
     /// </summary>
     public static StartHostConfiguration Empty { get; } = new();
--- a/src/Abstractions/Projects/StartOptionRegistry.cs
+++ b/src/Abstractions/Projects/StartOptionRegistry.cs
@@ -25,6 +25,17 @@ public sealed class StartOptionRegistry
     public StartOptionRegistry(Command command)
     {
         _command = command ?? throw new ArgumentNullException(nameof(command));
+
+        // Seed from the command's existing options so GetOrAdd detects collisions with built-ins.
+        foreach (Option existing in _command.Options)
+        {
+            var entry = new Entry(existing, "built-in");
+            _byName[existing.Name] = entry;
+            foreach (string alias in existing.Aliases)
+            {
+                _byName[alias] = entry;
+            }
+        }
     }
 
     /// <summary>
--- a/src/Abstractions/Projects/StartOptionRegistry.cs
+++ b/src/Abstractions/Projects/StartOptionRegistry.cs
@@ -1,0 +1,87 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.CommandLine;
+
+namespace Azure.Functions.Cli.Projects;
+
+/// <summary>
+/// Coordinates the options that <see cref="IStartHostOptionContributor"/> implementations
+/// contribute to <c>func start</c>/<c>func run</c>. Adds each unique option to the supplied
+/// <see cref="Command"/> exactly once and returns a shared canonical instance across workloads
+/// that contribute the same name, so a shared option shows once in <c>--help</c> and every
+/// contributor reads the same parsed value.
+/// </summary>
+/// <remarks>
+/// Public so workload test projects can drive <see cref="IStartHostOptionContributor.GetStartOptions"/>
+/// without taking a dependency on the host's internals.
+/// </remarks>
+public sealed class StartOptionRegistry
+{
+    private readonly Command _command;
+    private readonly Dictionary<string, Entry> _byName = new(StringComparer.Ordinal);
+    private string _activeStack = "?";
+
+    public StartOptionRegistry(Command command)
+    {
+        _command = command ?? throw new ArgumentNullException(nameof(command));
+    }
+
+    /// <summary>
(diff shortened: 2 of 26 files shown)


src/Abstractions/Projects/StartOptionRegistry.cs
```diff
@@ -0,0 +1,87 @@
...
+public sealed class StartOptionRegistry
+{
+    private readonly Command _command;
+    private readonly Dictionary<string, Entry> _byName = new(StringComparer.Ordinal);
+    private string _activeStack = "?";
+
+    public StartOptionRegistry(Command command)
+    {
+        _command = command ?? throw new ArgumentNullException(nameof(command));
+    }
```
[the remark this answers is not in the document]
satvu: Fixed — the constructor now seeds \_byName\ from the command's existing options (labeled \uilt-in\), so \GetOrAdd\ detects collisions with built-in options. Added tests for both the type-mismatch throw and same-type de-dupe paths.

src/Workloads/Stacks/DotNet/DotNetStartOptionContributor.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        Description = "Pauses the .NET Worker process on startup until a debugger is attached."
+    };
+
+    public Option<bool> EnableJsonOutputOption { get; } = new("--enable-json-output")
+    {
+        Description = "Emit JSON line output from the .NET worker when applicable."
+    };
+
+public Option<string?> JsonOutputFileOption { get; } = new("--json-output-file")
```
brettsam: nit: indent this?

src/Abstractions/Projects/StartHostConfiguration.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// value here wins. Empty when the workload's options are not enabled.
+    /// </summary>
+    public IReadOnlyDictionary<string, string> EnvironmentVariables { get; init; }
+        = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// When set, the CLI mirrors the host process's raw (JSON) standard output to this file path.
+    /// <c>null</c> disables file capture.
+    /// </summary>
+    public string? JsonOutputFilePath { get; init; }
```
brettsam: i don't love this... I'd almost rather have a way for a workload to register a LogInterceptor or something that allows it to: 
- read a log and do something unique with it (like write to a file)
- instruct the host to filter it (so we don't write this system log to the console)

as this sits, i imagine we're writing *all* logs to this file, which can get huge. And is a very dotnet-isolated thing that other stacks likely will never need...

brettsam (changes requested):

src/Func/Commands/Start/LegacyStartArgumentNormalizer.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Azure.Functions.Cli.Commands.Start;
+
+internal static class LegacyStartArgumentNormalizer
```
satvu, line 6: VS calls `func host start` still - this is needed for compatibility. May need to evaluate whether we can change this or not for VS (is this required for v4? does this need to be kept to make sure both v4 + v5 work with VS)?

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
```diff
@@ -0,0 +1,115 @@
...
+    private static partial Regex WorkerPidRegex();
+
+    private readonly TextWriter? _writer;
+    private bool _pidCaptured;
+
+    internal DotNetHostOutputInterceptor(string? outputFilePath)
+    {
+        if (!string.IsNullOrWhiteSpace(outputFilePath))
+        {
+            var stream = new FileStream(outputFilePath, FileMode.Create, FileAccess.Write, FileShare.ReadWrite, bufferSize: 4096, useAsync: true);
```
brettsam: let's not iniitalize any of this unless we know we need to write something... my agent says that this could create the file even if dotnet-isolated isn't being used

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public bool TryIntercept(string line)
+    {
+        // v4 path: raw azfuncjsonlog: prefix (if the host passes it through verbatim).
+        if (line.StartsWith(JsonLogPrefix, StringComparison.OrdinalIgnoreCase))
+        {
+            string payload = line[JsonLogPrefix.Length..];
+            _writer?.WriteLine(payload);
+            return true;
```
brettsam: --enable-json-output` without a file discards the output. With no path, `_writer` is null, but both protocol branches return `true` and suppress the line (`DotNetHostOutputInterceptor.cs:45-50,81-86`). V4 stripped the prefix and emitted the payload to the console.

src/Func/Commands/Start/Initialization/Steps/PrepareProjectHostRunInitializationStep.cs
```diff
@@ -101,6 +101,21 @@ private Dictionary<string, string> BuildEnvironmentVariables(StartInitialization
...
+        if (context.Options.StackHostConfigurations.TryGetValue(project.StackName, out StartHostConfiguration? stackConfiguration))
+        {
+            foreach ((string name, string value) in stackConfiguration.EnvironmentVariables)
+            {
+                environmentVariables[name] = value;
+            }
+
+            if (!string.IsNullOrWhiteSpace(stackConfiguration.StartupNotice))
+            {
+                _interaction.WriteHint(stackConfiguration.StartupNotice);
```
brettsam: `StartupNotice` still corrupts `--output json`. `PrepareProjectHostRunInitializationStep.cs:115` calls `WriteHint`, writing plain text to stdout in the middle of NDJSON. It should flow through `StartInitializationStepContext.ReportLogAsync`.

src/Func/Commands/Start/Host/HostProcessEventStream.cs
```diff
@@ -141,9 +145,17 @@ private void WriteProcessStarted(HostProcessLaunchInfo launchInfo)
 
     private async Task ReadLinesAsync(TextReader reader, string streamName)
     {
+        bool canIntercept = _outputInterceptor is not null
+            && string.Equals(streamName, HostProcessStreamNames.StandardOutput, StringComparison.Ordinal);
+
         string? line;
         while ((line = await reader.ReadLineAsync()) is not null)
         {
+            if (canIntercept && _outputInterceptor!.TryIntercept(line))
```
brettsam: an interceptor could throw, causing failures... maybe that's okay? another alternative is to catch/log those exceptions?

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
```diff
@@ -0,0 +1,115 @@
...
+        // v5 path: structured JSON record from the host.
+        if (!_pidCaptured && line.Length > 0 && line[0] == '{')
+        {
+            return TryInterceptStructuredRecord(line);
+        }
+
+        return false;
+    }
+
+    private bool TryInterceptStructuredRecord(string line)
```
brettsam, line 61: agent had some issues with this behavior compared to v4:

-----

The tested Visual Studio happy path is **not broken**. The problem is duplicate output and fragile fallback behavior around it.

Current flow:

```text
Worker emits azfuncjsonlog with PID
→ host wraps it in structured JSON
→ interceptor writes PID record to output file
→ _pidCaptured remains false

Worker emits “.NET Worker (PID: 1234) initialized...”
→ broad regex matches
→ interceptor writes the same PID record again
→ _pidCaptured becomes true
```

The resulting file can contain duplicates:

```json
{ "name":"dotnet-worker-startup", "workerProcessId":1234 }
{ "name":"dotnet-worker-startup", "workerProcessId":1234 }
```

After `_pidCaptured` becomes true, the interceptor stops examining **all structured host records**. Therefore later `azfuncjsonlog:` records are no longer extracted into the file or suppressed from the console.

The PID fallback is also too broad. It matches `(PID: n)` in any structured message without validating its category or complete text. For example, with JSON output but no debug wait:

```csharp
logger.LogInformation("Started helper process (PID: 4242)");
```

could append this false event:

```json
{ "name":"dotnet-worker-startup", "workerProcessId":4242 }
```

A safer implementation would:

- Always process genuine `azfuncjsonlog:` records.
- Match them only from `Host.Function.ToolingConsoleLog`.
- Use the human-readable PID regex only as a fallback when no genuine PID event was received.
- Match the complete known worker message, not any `(PID: n)`.
- Deduplicate only the fallback PID—not disable structured interception entirely.

brettsam (changes requested):
some more comments...

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
satvu: Addressed in d1650230:

- **Deduplication**: \_pidCaptured\ is now set on both the \azfuncjsonlog:\ path and the regex path, so we never write the PID twice.
- **Narrow regex**: Changed from \\(PID:\s*(\d+)\)\ to the full known message: \Azure Functions \.NET Worker \(PID:\s*(\d+)\) initialized in debug mode\ — won't match unrelated log lines.
- **Structured interception continues**: The \_pidCaptured\ guard only gates the regex fallback branch. Genuine \zfuncjsonlog:\ records embedded in structured JSON are always processed regardless of \_pidCaptured\ state.
- **No suppression without a file**: When \--json-output-file\ isn't provided, intercepted lines return \alse\ so they still render in the console (matching v4 behavior).

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
```diff
@@ -0,0 +1,147 @@
...
+            string payload = line[JsonLogPrefix.Length..];
+            WriteToDisk(payload);
+            _pidCaptured = true;
+
+            // When there is no output file, don't suppress — let the line flow to console.
+            return _writer is not null;
+        }
+
+        // v5 path: structured JSON record from the host.
+        if (!_pidCaptured && line.Length > 0 && line[0] == '{')
```
brettsam: after the first pid is captured, we won't check anymore... is that the desired behavior? should we not gate on that?

src/Func/Commands/Start/Host/HostProcessEventStream.cs
```diff
@@ -141,9 +145,27 @@ private void WriteProcessStarted(HostProcessLaunchInfo launchInfo)
...
+                try
+                {
+                    if (_outputInterceptor!.TryIntercept(line))
+                    {
+                        continue;
+                    }
+                }
+                catch (Exception)
+                {
+                    // Interceptor failure must not crash the event loop — let the line through.
```
brettsam, line 168: can we emit a warming, maybe? may make it clearer if VS hangs or something due to a file permission issue...

src/Workloads/Stacks/DotNet/DotNetHostOutputInterceptor.cs
satvu: V4 doesn't gate on this, but it was added as a performance check (do we really need any other logs after?) but had an agent look and it doesn't think that this would have large impact. Removing the gate.

brettsam (approved):

PR #5505, opened 2026-08-03
Add diagnostic logging to worker resolution pipeline

satvu:
## Summary

Adds structured `ILogger` diagnostics to the worker resolution path to help investigate [#5218](https://github.com/Azure/azure-functions-core-tools/issues/5218) — intermittent worker resolution failures where `func run` fails to find a worker that is present on disk.

**This PR adds diagnostics only.** The confidence on the root cause is low, and no behavioral fix is included. The logging will help capture the resolution state when the failure reproduces.

## Changes

Logging is added to three components in the worker resolution pipeline:

| Component | What's logged | Level |
|---|---|---|
| `DefaultFunctionsWorkerContentResolver` | Installed workloads found, per-candidate version constraint filtering (with skip reasons), selected candidate, missing `worker.config.json` | Debug / Warning |
| `DefaultFunctionsWorkerResolver` | Active constraint for the worker ID, workload search results (package ID + alias lookup) | Debug |
| `ResolveFunctionsWorkerInitializationStep` | Active profile name, worker version ranges, install-fallback attempts, final failure type and message | Debug / Warning |

All log calls use structured templates (`{WorkerId}`, `{PackageId}`, etc.). No user-facing behavior changes.

## Test validation

- All 26 `DefaultFunctionsWorker*` tests pass
- All 23 `StartInitialization` tests pass
- Zero build warnings

### DIFF
--- a/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
+++ b/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
@@ -7,6 +7,7 @@ using Azure.Functions.Cli.Workers;
 using Azure.Functions.Cli.Workloads.Catalog;
 using Azure.Functions.Cli.Workloads.Discovery;
 using Azure.Functions.Cli.Workloads.Install;
+using Microsoft.Extensions.Logging;
 using NuGet.Versioning;
 
 namespace Azure.Functions.Cli.Commands.Start.Initialization;
@@ -16,7 +17,8 @@ namespace Azure.Functions.Cli.Commands.Start.Initialization;
 /// </summary>
 internal sealed class ResolveFunctionsWorkerInitializationStep(
     IFunctionsWorkerResolverFactory workerResolverFactory,
-    IFunctionsWorkerInstaller workerInstaller) : FuncStartInitializationStep
+    IFunctionsWorkerInstaller workerInstaller,
+    ILogger<ResolveFunctionsWorkerInitializationStep> logger) : FuncStartInitializationStep
 {
     public const string StepId = "resolve_worker";
 
@@ -24,6 +26,9 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
         ?? throw new ArgumentNullException(nameof(workerResolverFactory));
 
     private readonly IFunctionsWorkerInstaller _workerInstaller = workerInstaller ?? throw new ArgumentNullException(nameof(workerInstaller));
+
+    private readonly ILogger<ResolveFunctionsWorkerInitializationStep> _logger = logger
+        ?? throw new ArgumentNullException(nameof(logger));
 
     public override string Id => StepId;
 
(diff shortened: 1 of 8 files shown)


RohitRanjanMS:
Are these customer-visible logs? What's the sink?

src/Func/Workers/DefaultFunctionsWorkerResolver.cs
```diff
@@ -28,8 +31,15 @@ public Task<FunctionsWorkerResolutionResult> ResolveWorkerAsync(FunctionsWorkerI
         cancellationToken.ThrowIfCancellationRequested();
 
         _activeWorkerConstraints.TryGetValue(workerId.Value, out VersionRange? constraint);
+        _logger.LogDebug(
+            "[worker-resolve] Resolving worker '{WorkerId}' with constraint '{Constraint}'.",
+            workerId.Value, constraint?.ToString() ?? "(none)");
```
ahmedmuhsin: Answering Rohit's question about the sink, there isn't one today, so none of this output is reachable.

`CliHostFactory.CreateBuilder` uses `Host.CreateEmptyApplicationBuilder(null)` and never registers a logging provider. `AddOpenTelemetry()` wires `.WithTracing` and `.WithMetrics` but not `.WithLogging`. Probed the fully built host on this branch, workloads registered:

```
factory = Microsoft.Extensions.Logging.LoggerFactory
ILoggerProvider count = 0 []
WorkerResolver  : trace=False dbg=False warn=False crit=False
ContentResolver : trace=False dbg=False warn=False crit=False
InitStep        : trace=False dbg=False warn=False crit=False
```

Disabled at every level including Critical, so every `_logger.Log*` call added here is a no-op. `--verbose` does not change that either. It only gates `GracefulException.VerboseMessage` in `Program.cs` and switches output mode for `func version` and `func workload list`.

This is not something the PR introduced. The existing `[bundle-resolve]` logs in `ExtensionBundleResolver` have the same problem. But it does mean the PR cannot do what the description says it will, which is capture resolution state when #5218 reproduces. A sink is the prerequisite, either a `--verbose` gated console provider or a rolling file under the CLI home, and it probably belongs here since that is the whole point of the change.

Unrelated but while I'm here, the description says `Closes #5218` and also says there is no fix. That will auto close an open bug on merge. `Refs #5218` is probably what you want.

src/Func/Workers/DefaultFunctionsWorkerResolver.cs
```diff
@@ -28,8 +31,15 @@ public Task<FunctionsWorkerResolutionResult> ResolveWorkerAsync(FunctionsWorkerI
...
 
         IReadOnlyList<ContentWorkloadInfo> installedWorkers = GetWorkerWorkloads(workerId);
+        _logger.LogDebug(
+            "[worker-resolve] Workload search for '{WorkerId}' found {Count} candidate(s).",
+            workerId.Value, installedWorkers.Count);
```
ahmedmuhsin: For the failure in #5218, worker present on disk but resolution fails, the count is the least useful thing to record at this point.

`GetWorkerWorkloads` searches by two keys it computes internally, `FunctionsWorkerWorkloadPackages.GetPackageId(workerId)` and the alias `workerId.Value + "-worker"`, and neither is logged. When the count comes back 0, what you want in the log is which package id and alias were searched for, and what `_workloadProvider.GetContentWorkloads()` actually returned (id, version, aliases). That is where the mismatch would be visible.

Also the description says this logs "workload search results (package ID + alias lookup)". Right now it is only the count.

src/Func/Workers/DefaultFunctionsWorkerContentResolver.cs
```diff
@@ -35,6 +40,7 @@ public FunctionsWorkerResolutionResult ResolveWorker(
 
         if (installedWorkers.Count == 0)
         {
+            _logger.LogDebug("[worker-resolve] No installed workloads found for worker '{WorkerId}'.", workerId.Value);
```
ahmedmuhsin: This is the branch that matches the symptom reported in #5218, and it is the one carrying the least detail. Debug level with no inventory, while the `candidates.Count == 0` branch further down logs the full installed list at Warning.

Worth flipping that around, since "the worker is on disk but we found zero workloads" is exactly the case you are trying to catch.

src/Func/Workers/DefaultFunctionsWorkerResolver.cs
satvu: Addressing in #5522 as this is a broader issue.

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
```diff
@@ -25,6 +27,9 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
 
     private readonly IFunctionsWorkerInstaller _workerInstaller = workerInstaller ?? throw new ArgumentNullException(nameof(workerInstaller));
 
+    private readonly ILogger<ResolveFunctionsWorkerInitializationStep> _logger = logger
```
jviau: I often type the field as just `ILogger`, since the generic interface is purely a DI-convenience wrapper

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
```diff
@@ -46,12 +51,21 @@ public override async Task<StartInitializationStepResult> ExecuteAsync(StartInit
             && TryGetInstallableWorker(notResolved.Failure, out FunctionsWorkerId? workerId)
             && workerId is not null)
         {
+            _logger.LogDebug(
```
jviau: Consider adopting logger source generation

jviau (approved):

PR #5507, opened 2026-08-04
Remove unsupported runtimes from SKU built-in profiles

### LINKED ISSUE #5293: [v5] Warn if running a SKU profile with a stack that it doesn't support
### Version

5.0.0-preview.2

### OS / shell

Windows Powershell

### Stack

python

### Description

Python apps run just fine with Windows profiles (there's not much difference between SKU profiles currently), but in the future, if SKU profiles diverge more, customers may wrongly test with a windows profile and lose time when they should only be using linux profiles (to simulate prod env).

Not a huge priority since customers will be able to fix themselves, but this case can also be expanded to accommodate preview/new languages that are only on specific SKUs.

### Steps to reproduce

- Set up a python app
- Set up the profile to windows (which does not support python based on [official docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-scale#operating-systemruntime))
- `func run` runs normally

### Expected behavior

Warning for the customer that python is not supported on Windows in prod and that they should switch to one of the available Linux profiles.

### Actual behavior

No issues, runs normally w/o warnings or errors.

satvu:
## Summary

Fixes #5293

Aligns built-in profiles with Azure runtime support and enforces `supportedRuntimes` during `func start`. Unsupported stacks are rejected before worker resolution or auto-install, with a second check against the resolved worker runtime.

### Built-in profiles

| Profile | Removed from `supportedRuntimes` | Removed from `workers` |
|---------|----------------------------------|------------------------|
| `windows-consumption` | `python` | `python`, `go` |
| `windows-dedicated` | `python` | `python`, `go` |
| `linux-premium` | — | `go` |
| `linux-consumption` | — | `go` |

The profiles continue to use `dotnet-isolated`, matching the .NET worker runtime identity consumed by `func setup` and the Functions host.

### Runtime validation

- Validates profile support before resolving or installing a worker, so missing or invalid worker installations do not mask unsupported-stack errors.
- Retains validation against the actual runtime after worker resolution.
- Accepts either the project stack name or worker runtime identity: `.NET` uses stack `dotnet` and runtime `dotnet-isolated`; Go uses stack/workload `go` and runtime `native`.
- Reads the expected runtime from `FunctionsWorkerReference` and normalizes newly installed workers to that identity, preserving the worker ID, config path, and version. First-run Go workers therefore resolve as `native`, not `go`.
- Reports unsupported stacks as user errors (exit code 1), including the profile's supported runtime list.

### Tests and release notes

- Covers rejection before worker resolution/installation with missing and invalid worker installations, asserting against `IFunctionsWorkerInstaller`.
- Loads the shipped registry and asserts supported runtimes and worker keys for all five built-in profiles.
- Covers `.NET` runtime matching and both `go` and `native` profile values during Go auto-install. The fake installer returns `go`; the final worker must be `native` with its metadata preserved.
- Covers worker-reference runtime identities and updates `src/Func/release_notes.md` with the user-facing behavior change.

## Supporting documentation

**Python is Linux-only.** The [Azure Functions supported languages](https://learn.microsoft.com/en-us/azure/azure-functions/supported-languages) documentation marks Python as supported on Linux and unsupported on Windows. The [Azure Functions scale and hosting](https://learn.microsoft.com/en-us/azure/azure-functions/functions-scale) documentation also states:

> Linux is the only supported operating system for the Python runtime stack.

**Go is Flex Consumption-only.** The [Go developer reference](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-go) states:

> Go support is currently available only for function apps hosted in the Flex Consumption plan.

### Automated

- `dotnet test test\Func.Tests\Func.Tests.csproj --no-restore --filter "FullyQualifiedName~StartInitializationTests|FullyQualifiedName~FunctionsWorkerReferenceTests|FullyQualifiedName~BuiltInProfileSourceTests" -v minimal` — 47 passed.
- Release build with `CI=true`: `dotnet build -c Release --no-restore -v q` — succeeded with 0 warnings and 0 errors.

### Manual — real workload packages

Built a Windows x64 CLI and real workload `.nupkg`s, published packages to a local NuGet v3 feed (`http://localhost:5555/v3/index.json`), and used an isolated CLI/workload home under `artifacts\manual-go-test`. Created a Go Functions project with `func init --stack go`. Runs used an interactive terminal so worker auto-install was enabled.

| Scenario | Procedure | Observed result |
|----------|-----------|-----------------|
| Supported Go/Flex, worker missing | Install real host and Go stack, remove Go worker, then run `func start <app> --profile flex --output plain --no-azurite` | Auto-installed Go worker `1.0.0-local.20260916232418`; reported `Resolved worker native` and `resolve_worker complete: native`. |
| Unsupported Go/Linux Premium, worker missing | Republish the latest CLI at `c046eded`, remove Go worker, then run `func start <app> --profile linux-premium --output plain --no-azurite` | Rejected with `Profile 'linux-premium' does not support the 'Go' stack`, listed supported runtimes, and exited with code **1**. The workload list afterward still contained only the host and Go stack: **no worker was installed**. |

**Manual-test scope:** The Flex run verified real-package auto-install and runtime normalization, not a complete host startup or function invocation. It stopped afterward at extension-bundle resolution because no compatible bundle workload was installed. Go was also unavailable on the test machine. The Linux Premium rejection occurred before those downstream dependencies were needed.

### DIFF
--- a/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
+++ b/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
@@ -47,7 +47,8 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
             && workerId is not null)
         {
             // Block auto-install if neither the stack name nor the workload id match any supported runtime.
-            ValidateSupportedRuntime(context, project.StackName, workerId.Value, project.StackDisplayName);
+            // workerId.Value is the workload identifier (e.g. "python", "dotnet") which may differ from WorkerRuntime.
+            ValidateSupportedRuntime(context, project.StackName, runtimeIdentifier: workerId.Value, project.StackDisplayName);
             result = await TryInstallAndResolveWorkerAsync(context, workerId, workerVersionRanges, notResolved.Failure, cancellationToken);
         }
 
@@ -57,7 +58,7 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
             throw CreateWorkerResolutionException(failedResult.Failure, context);
         }
 
-        ValidateSupportedRuntime(context, project.StackName, resolved.Worker.WorkerRuntime, project.StackDisplayName);
+        ValidateSupportedRuntime(context, project.StackName, runtimeIdentifier: resolved.Worker.WorkerRuntime, project.StackDisplayName);
         context.State.Worker = resolved.Worker;
 
         string completionMessage = string.IsNullOrWhiteSpace(resolved.Worker.Version)
@@ -205,14 +206,14 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
 
     /// <summary>
     /// Checks whether the profile supports the runtime by matching against both the stack name
-    /// and the worker runtime name. Some stacks differ between the two (e.g. Go stack = "go"
-    /// but worker runtime = "native"; DotNet stack = "dotnet" but worker runtime = "dotnet-isolated").
-    /// A match on either is sufficient.
+    /// and the runtime identifier (workload id or worker runtime name). Some stacks differ between
+    /// the two (e.g. Go stack = "go" but worker runtime = "native"; DotNet stack = "dotnet" but
+    /// worker runtime = "dotnet-isolated"). A match on either is sufficient.
     /// </summary>
     private static void ValidateSupportedRuntime(
         StartInitializationStepContext context,
         string stackName,
-        string workerRuntimeName,
+        string runtimeIdentifier,
         string stackDisplayName)
     {
         if (context.State.ResolvedProfile is not { SupportedRuntimes: { } supportedRuntimes } profile)
@@ -222,7 +223,7 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
 
         if (supportedRuntimes.Any(runtime =>
             string.Equals(runtime, stackName, StringComparison.OrdinalIgnoreCase)
-            || string.Equals(runtime, workerRuntimeName, StringComparison.OrdinalIgnoreCase)))
+            || string.Equals(runtime, runtimeIdentifier, StringComparison.OrdinalIgnoreCase)))
         {
             return;
         }
--- a/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
+++ b/src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
@@ -46,6 +46,8 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
             && TryGetInstallableWorker(notResolved.Failure, out FunctionsWorkerId? workerId)
             && workerId is not null)
         {
+            // Block auto-install if neither the stack name nor the workload id match any supported runtime.
+            ValidateSupportedRuntime(context, project.StackName, workerId.Value, project.StackDisplayName);
             result = await TryInstallAndResolveWorkerAsync(context, workerId, workerVersionRanges, notResolved.Failure, cancellationToken);
         }
 
@@ -55,7 +57,7 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
             throw CreateWorkerResolutionException(failedResult.Failure, context);
         }
 
-        ValidateSupportedRuntime(context, resolved.Worker);
+        ValidateSupportedRuntime(context, project.StackName, resolved.Worker.WorkerRuntime, project.StackDisplayName);
         context.State.Worker = resolved.Worker;
 
         string completionMessage = string.IsNullOrWhiteSpace(resolved.Worker.Version)
@@ -201,19 +203,31 @@ internal sealed class ResolveFunctionsWorkerInitializationStep(
             + $"{failure.Message} Run '{repairCommand}' to repair the install.";
     }
 
-    private static void ValidateSupportedRuntime(StartInitializationStepContext context, IFunctionsWorker worker)
+    /// <summary>
+    /// Checks whether the profile supports the runtime by matching against both the stack name
+    /// and the worker runtime name. Some stacks differ between the two (e.g. Go stack = "go"
+    /// but worker runtime = "native"; DotNet stack = "dotnet" but worker runtime = "dotnet-isolated").
+    /// A match on either is sufficient.
+    /// </summary>
+    private static void ValidateSupportedRuntime(
+        StartInitializationStepContext context,
+        string stackName,
+        string workerRuntimeName,
+        string stackDisplayName)
     {
         if (context.State.ResolvedProfile is not { SupportedRuntimes: { } supportedRuntimes } profile)
         {
             return;
         }
 
-        if (supportedRuntimes.Any(runtime => string.Equals(runtime, worker.WorkerRuntime, StringComparison.OrdinalIgnoreCase)))
+        if (supportedRuntimes.Any(runtime =>
+            string.Equals(runtime, stackName, StringComparison.OrdinalIgnoreCase)
+            || string.Equals(runtime, workerRuntimeName, StringComparison.OrdinalIgnoreCase)))
         {
             return;
         }
 
-        string message = $"Profile '{profile.Name}' does not support the detected runtime '{worker.WorkerRuntime}'. "
+        string message = $"Profile '{profile.Name}' does not support the '{stackDisplayName}' stack. "
             + $"Supported runtimes: {string.Join(", ", supportedRuntimes)}.";
         throw new GracefulException(message, isUserError: true);
     }
--- a/src/Func/Profiles/BuiltInProfiles.json
+++ b/src/Func/Profiles/BuiltInProfiles.json
@@ -47,9 +47,6 @@
         },
         "python": {
           "version": "[4.43.0]"
-        },
-        "go": {
-          "version": "[1.0.0]"
         }
       },
       "supportedRuntimes": [
@@ -73,17 +70,10 @@
       "workers": {
         "node": {
           "version": "[3.13.0]"
-        },
-        "python": {
-          "version": "[4.43.0]"
-        },
-        "go": {
-          "version": "[1.0.0]"
         }
       },
       "supportedRuntimes": [
         "node",
-        "python",
         "java",
         "powershell",
         "dotnet-isolated",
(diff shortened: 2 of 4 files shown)


src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
```diff
@@ -201,19 +203,23 @@ private static string CreateInvalidInstallationMessage(FunctionsWorkerResolution
...
+    /// worker resolution or installation. This prevents unnecessary downloads for unsupported runtimes.
+    /// </summary>
+    private static void ValidateSupportedStack(StartInitializationStepContext context, FunctionsProject project)
     {
         if (context.State.ResolvedProfile is not { SupportedRuntimes: { } supportedRuntimes } profile)
         {
             return;
         }
 
-        if (supportedRuntimes.Any(runtime => string.Equals(runtime, worker.WorkerRuntime, StringComparison.OrdinalIgnoreCase)))
```
RohitRanjanMS, line 211: Is the worker.WorkerRuntime and project.StackName same? What would be stack and runtime for isolated?

satvu: For one language they differ:

Go -> native

To handle both cases we'll have to use both WorkerRuntime and StackName to find a match. 

For vnext, all references to dotnet refer to dotnet-isolated in the workloads, args, and commands since dotnet inproc is not supported in vnext.

satvu: Actually, we should just use "dotnet" in the profile - we're moving towards "dotnet" meaning "dotnet-isolated" anyways and in-proc is not supported at all in vnext.

src/Func/Profiles/BuiltInProfiles.json
```diff
@@ -28,7 +28,7 @@
         "python",
         "java",
         "powershell",
-        "dotnet-isolated",
+        "dotnet",
```
ahmedmuhsin: This rename breaks `func setup --features dotnet` on every built-in profile.

`SetupRunner.DotNetProfileRuntime` is the literal `"dotnet-isolated"`, and `BuildDependencyPlanAsync` compares `runtimeFeature.ProfileRuntime` against `profile.SupportedRuntimes`, so with `dotnet` in the JSON that lookup misses. Matrix against this branch:

```
no profile    / dotnet -> passed
flex          / dotnet -> Error: Profile 'flex' does not support runtime 'dotnet'. Supported runtimes: go, node, python, java, powershell, dotnet, custom.
flex          / node   -> passed
flex          / go     -> passed
win-consump   / python -> Error: ... does not support runtime 'python'   (intended by this PR)
win-consump   / dotnet -> Error: ... does not support runtime 'dotnet'
linux-premium / dotnet -> Error: ... does not support runtime 'dotnet'
```

The same commands pass the gate on vnext, so it is a regression. Note the message lists `dotnet` as supported while rejecting it. It only bites when a profile is active (explicit `--profile`, profiles in `.func/config.json`, or a user default), since `SetupProfileScope.Unconstrained` skips the check.

I tried reverting just this rename back to `dotnet-isolated` and keeping everything else, including the new OR match. The matrix goes green apart from the intended python-on-windows rejection, and all 1323 tests still pass.

Worth calling out that the rename also makes the new OR match dead weight for the bundled profiles. After it, every value in every `supportedRuntimes` equals a stack name, so the `workerRuntimeName` arm never fires for a built-in profile. Keeping `dotnet-isolated` is what makes that arm earn its place and keeps setup working.

Nothing caught this because every profile test builds `ResolvedProfile` inline and no test reads this file. A test that walks the shipped JSON and checks each `supportedRuntimes` value against both consumers would be worth adding.

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
```diff
@@ -46,6 +46,8 @@ public override async Task<StartInitializationStepResult> ExecuteAsync(StartInit
             && TryGetInstallableWorker(notResolved.Failure, out FunctionsWorkerId? workerId)
             && workerId is not null)
         {
+            // Block auto-install if neither the stack name nor the workload id match any supported runtime.
+            ValidateSupportedRuntime(context, project.StackName, workerId.Value, project.StackDisplayName);
```
ahmedmuhsin: Minor, but `workerId.Value` is the workload id, not the worker runtime, and it lands in a parameter named `workerRuntimeName`. Go is exactly the case where those differ (`go` vs `native`), which is the distinction the doc comment on the method is about. Your comment on this line already says "workload id".

No behavior impact since the stack name arm covers it, just makes the method harder to read than it needs to be.

test/Func.Tests/Commands/StartInitializationTests.cs
```diff
@@ -489,7 +489,107 @@ public async Task DemoRunner_ProfileRejectsUnsupportedDetectedRuntime()
...
+            Arg.Any<bool>(),
+            Arg.Any<IProgress<WorkloadInstallProgress>?>(),
+            Arg.Any<CancellationToken>());
+    }
+
+    [Fact]
+    public async Task DemoRunner_ProfileAllowsStackWhenWorkerRuntimeMatches()
+    {
+        // DotNet has StackName "dotnet" but WorkerRuntime "dotnet-isolated".
+        // The profile lists "dotnet-isolated" so it should pass validation via WorkerRuntime match.
```
ahmedmuhsin, line 553: This profile shape no longer exists in any shipped profile, since all five now list `dotnet` rather than `dotnet-isolated`. Fine as unit coverage of the OR branch, but the comment reads like it mirrors production and it does not.

If the rename gets reverted (see my comment on BuiltInProfiles.json) then it does mirror production and this becomes the right test to have.

test/Func.Tests/Commands/StartInitializationTests.cs
```diff
@@ -1347,26 +1448,39 @@ public void Dispose()
...
+            string stackDisplayName,
+            bool supportsExtensionBundles,
+            FunctionsWorkerReference? workerReference = null)
+        {
+            _workingDirectory = workingDirectory;
+            _stackName = stackName;
+            _stackDisplayName = stackDisplayName;
+            _supportsExtensionBundles = supportsExtensionBundles;
+            _workerReference = workerReference ?? FunctionsWorkerReference.FromWorkload(stackName);
+        }
```
ahmedmuhsin: Primary constructors take optional parameters, so this could have stayed as it was:

```csharp
private sealed class TestFunctionsProject(
    WorkingDirectory workingDirectory,
    string stackName,
    string stackDisplayName,
    bool supportsExtensionBundles,
    FunctionsWorkerReference? workerReference = null) : FunctionsProject
{
    private readonly FunctionsWorkerReference _workerReference =
        workerReference ?? FunctionsWorkerReference.FromWorkload(stackName);
```

Saves the four extra backing fields, and AGENTS.md asks for primary constructors.

src/Func/Profiles/BuiltInProfiles.json
satvu: Fixed — reverted all profiles back to \dotnet-isolated\ in \supportedRuntimes\. This keeps \SetupRunner.DotNetProfileRuntime\ working and makes the OR-match arm earn its place (DotNet projects pass validation via \WorkerRuntime = \dotnet-isolated\\ matching the profile). The test \DemoRunner_ProfileAllowsStackWhenWorkerRuntimeMatches\ now correctly mirrors a shipped profile shape.

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
satvu: Fixed — renamed the parameter from \workerRuntimeName\ to \
untimeIdentifier\ and added an inline comment at the pre-install call site explaining that \workerId.Value\ is the workload identifier.

test/Func.Tests/Commands/StartInitializationTests.cs
satvu: Fixed — with the revert back to \dotnet-isolated\ in profiles, this test now correctly mirrors the shipped profile shape. The comment is accurate again.

test/Func.Tests/Commands/StartInitializationTests.cs
satvu: Done — converted \TestFunctionsProject\ to a primary constructor with optional \workerReference\ parameter.

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
satvu: Reverted since this is to match host runtime easily, but we keep "dotnet" for all customer-facing commands. This will be internal implementation.

satvu:
@copilot resolve the merge conflicts in this pull request

src/Func/Commands/Start/Initialization/Steps/ResolveFunctionsWorkerInitializationStep.cs
```diff
@@ -66,7 +68,7 @@ public override async Task<StartInitializationStepResult> ExecuteAsync(StartInit
             throw CreateWorkerResolutionException(failedResult.Failure, context);
         }
 
-        ValidateSupportedRuntime(context, resolved.Worker);
+        ValidateSupportedRuntime(context, project.StackName, runtimeIdentifier: resolved.Worker.WorkerRuntime, project.StackDisplayName);
         context.State.Worker = resolved.Worker;
```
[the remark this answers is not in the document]
satvu: Fixed in `f2ecf5d5`. The auto-install path now normalizes the installed worker to `project.WorkerReference.WorkerRuntime` before storing it, preserving the worker ID, config path, and version. I also changed the regression test so the fake installer returns `WorkerRuntime = "go"`; the test verifies the final resolved worker is `"native"`, matching first-run production behavior. Targeted tests pass (39/39), and the strict Release build succeeds with 0 warnings.

satvu:
Addressed the suppressed findings from the latest Copilot review in c046eded:

- Validate the project stack/reference runtime before calling the worker resolver, retaining the post-resolution check. Unsupported profiles now take precedence over missing or invalid worker installations.
- Correct the rejection test to configure NotInstalled/InvalidInstallation results and assert that neither the worker resolver nor IFunctionsWorkerInstaller.InstallAsync is called.
- Load the actual bundled registry and assert supportedRuntimes and worker keys for all five profiles, including Python only on Linux, Go only on Flex, and dotnet-isolated retained.
- Cover both go and native profile values for first-run Go installation, verifying normalization from the installer runtime go to native while preserving worker metadata.
- Update release notes to describe the user-facing rejection before worker resolution/installation.

Validation: 47 targeted tests passed; strict Release build completed with zero warnings/errors.

ahmedmuhsin (approved):

RohitRanjanMS (approved):

PR #5508, opened 2026-08-04
Improve 'No installed stack matches' error with install guidance

### LINKED ISSUE #5210: [v5] Improve "No installed stack matches" error details to include information about how to install or aka.ms link?
### Version

_No response_

### OS / shell

_No response_

### Stack

_No response_

### Description

Not an error. But an improvement. It would be nice to give more details or an aka.ms link explaining how to install the stack.

Image

### Steps to reproduce

Image

### Expected behavior

_No response_

### Actual behavior

_No response_

### Stack trace / logs

```shell

```

satvu:

resolves #5210

### Pull request checklist

  * [] Otherwise: I've added my notes to `release_notes.md`

### DIFF
--- a/src/Func/Workloads/WorkloadHintRenderer.cs
+++ b/src/Func/Workloads/WorkloadHintRenderer.cs
@@ -60,6 +60,32 @@ internal sealed class WorkloadHintRenderer(IInteractionService interaction) : IW
     {
         _interaction.WriteError($"No installed stack matches '{hint.RequestedStack}'.");
         WriteInstalledStacks(hint.InstalledStacks);
+
+        _interaction.WriteBlankLine();
+
+        if (hint.RequestedStack is not null
+            && KnownInstallableWorkloads.LanguageMap.ContainsKey(hint.RequestedStack))
+        {
+            _interaction.WriteHint($"Install it with:");
+            _interaction.WriteLine(l => l
+                .Muted("  ")
+                .Command($"func setup --features {hint.RequestedStack}"));
+        }
+        else
+        {
+            _interaction.WriteHint($"Set up a stack to {hint.ActionDescription}:");
+            _interaction.WriteBlankLine();
+
+            IEnumerable<DefinitionItem> items = KnownInstallableWorkloads.LanguageMap.Select(static kvp =>
+                new DefinitionItem($"func setup --features {kvp.Key}", string.Join(", ", kvp.Value)));
+            _interaction.WriteDefinitionList(items);
+        }
+
+        _interaction.WriteBlankLine();
+        _interaction.WriteLine(l => l
+            .Muted("Run ")
+            .Command("func workload search")
+            .Muted(" to discover available stacks."));
     }
 
     private void RenderAmbiguousStackChoice(WorkloadHint hint)
--- a/src/Func/Workloads/WorkloadHintRenderer.cs
+++ b/src/Func/Workloads/WorkloadHintRenderer.cs
@@ -42,18 +42,7 @@ internal sealed class WorkloadHintRenderer(IInteractionService interaction) : IW
     {
         _interaction.WriteWarning("No stacks installed.");
         _interaction.WriteBlankLine();
-        _interaction.WriteHint($"Set up a stack to {hint.ActionDescription}:");
-        _interaction.WriteBlankLine();
-
-        IEnumerable<DefinitionItem> items = KnownInstallableWorkloads.LanguageMap.Select(static kvp =>
-            new DefinitionItem($"func setup --features {kvp.Key}", string.Join(", ", kvp.Value)));
-        _interaction.WriteDefinitionList(items);
-
-        _interaction.WriteBlankLine();
-        _interaction.WriteLine(l => l
-            .Muted("Run ")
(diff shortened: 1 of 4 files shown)


src/Func/Workloads/WorkloadHintRenderer.cs
```diff
@@ -60,6 +60,32 @@ private void RenderNoMatchingStack(WorkloadHint hint)
     {
         _interaction.WriteError($"No installed stack matches '{hint.RequestedStack}'.");
         WriteInstalledStacks(hint.InstalledStacks);
+
+        _interaction.WriteBlankLine();
+
+        if (hint.RequestedStack is not null
+            && KnownInstallableWorkloads.LanguageMap.ContainsKey(hint.RequestedStack))
```
ahmedmuhsin, line 56: This guard is what makes echoing the stack name safe, and there is a second place in `func init` that could use it.

When a directory has a `host.json` but no `.func/config.json`, `func init` takes the adopt path in `InitCommand` and prints this instead:

```
The 'python' stack is not installed. Run `func setup --features <stack>` to setup your dev environment, then re-run `func init`.
```

The comment there says the `<stack>` placeholder is deliberate, because echoing an unknown candidate "would promise a `func setup` command that won't necessarily work". This check removes that objection, so that path could render `NoMatchingStack` and get the same guidance. Cloning an existing repo and running `func init` without the stack installed feels like the more common way to hit this.

Fine as a follow-up if you would rather keep this PR tight. Just want it to be a decision rather than an oversight.

src/Func/Workloads/WorkloadHintRenderer.cs
```diff
@@ -60,6 +60,32 @@ private void RenderNoMatchingStack(WorkloadHint hint)
...
+        }
+        else
+        {
+            _interaction.WriteHint($"Set up a stack to {hint.ActionDescription}:");
+            _interaction.WriteBlankLine();
+
+            IEnumerable<DefinitionItem> items = KnownInstallableWorkloads.LanguageMap.Select(static kvp =>
+                new DefinitionItem($"func setup --features {kvp.Key}", string.Join(", ", kvp.Value)));
+            _interaction.WriteDefinitionList(items);
+        }
```
ahmedmuhsin, line 66: This branch can contradict the line right above it.

`SelectInitializerAsync` matches only on `IProjectInitializer.Stack`, not `WorkerRuntimeAliases`, unlike `FindInitializerForCandidate` a few lines up. So `func init --stack dotnet-isolated` with dotnet installed lands in `NoMatchingStack` and renders this:

```
No installed stack matches 'dotnet-isolated'.

Installed stacks:
  dotnet

Set up a stack to initialize a project:

  func setup --features dotnet    C#, F#
  func setup --features node      JavaScript, TypeScript
  func setup --features python    Python
  func setup --features go        Go

Run func workload search to discover available stacks.
```

We tell the user dotnet is installed and then offer to install it. Same thing with `--stack native` when go is installed. Before this change the message stopped after the installed list, so this is new.

Smallest fix is filtering `hint.InstalledStacks` out of this menu. The deeper fix is making `SelectInitializerAsync` alias-aware like `FindInitializerForCandidate`, which would keep aliases out of `NoMatchingStack` entirely. Either works for me, though the filter seems worth having regardless.

src/Func/Workloads/WorkloadHintRenderer.cs
```diff
@@ -60,6 +60,32 @@ private void RenderNoMatchingStack(WorkloadHint hint)
...
+
+        _interaction.WriteBlankLine();
+        _interaction.WriteLine(l => l
+            .Muted("Run ")
+            .Command("func workload search")
+            .Muted(" to discover available stacks."));
```
ahmedmuhsin, line 72: This trailer plus the `else` branch above is line for line the body of `RenderNoWorkloadsInstalled`. Worth pulling out a `WriteSetupMenu(string actionDescription)` and calling it from both, otherwise the two copies will drift the next time someone tweaks the wording.

src/Func/Workloads/WorkloadHintRenderer.cs
satvu: Went with alias-aware fix

src/Func/Workloads/WorkloadHintRenderer.cs
satvu: Good call — keeping it as a follow-up. The adopt path (line 224-236) already uses \FindInitializerForCandidate\ so aliases resolve there. The issue was only in \SelectInitializerAsync\ (the fresh-scaffold path), which is now fixed in this PR. The adopt path's hardcoded error message at line 233 could use the same \WorkloadHint\ renderer for consistency, but that's a separate change.

src/Func/Workloads/WorkloadHintRenderer.cs
satvu: Done — extracted \WriteSetupMenu(string actionDescription, IReadOnlyList<string> installedStacks)\ and both paths call it now. The installed-stacks filter is baked in so neither copy can drift.

jviau (approved):

PR #5512, opened 2026-08-05
Add RID pointer package support

fabiocav:

N/A

### Pull request checklist

  * [x] Otherwise: Documentation is updated in this PR

### Additional information

Adds RID-pointer workload packages while preserving schema v1 and legacy installed registry rows.

* Authors payload-free `FuncCliWorkload` pointer packages with exact RID-to-package maps.
* Authors `FuncCliWorkloadRidPackage` implementations with required `runtimeIdentifier` metadata.
* Resolves implementations by exact package ID and pointer version from the pointer source, including deterministic local sibling resolution.
* Tracks physical payload ownership across explicit installs, logical pointers, and meta workloads for safe install, update, uninstall, list, and prune behavior.
* Prevents incompatible-RID activation and integrates logical identities with host, setup, and Python worker flows.

User-visible behavior: installing a multi-RID workload by its logical package ID now selects the implementation for the current RID while lifecycle commands continue to present the logical identity.

Installed packages are keyed on package ID and version only, matching the NuGet global packages folder. A payload already installed from one feed is reused when a pointer resolved from another feed maps to the same ID and version, rather than failing. Source is still recorded and still pins pointer implementation resolution to the feed that supplied the pointer.

Also fixes a bug where the payload content root was hardcoded to `tools/any`, so RID-specific host packs never received the executable bit on Linux and macOS.

Validation:

* `dotnet restore`
* `CI=true dotnet build -c Release --no-restore` (0 warnings, 0 errors)
* Release test suite across Func and workload test projects
* Real Host workload pack inspection: one payload-free pointer plus six exact-version RID implementation packages
* Local NuGet v3 feed end-to-end install/list/update/uninstall validation

### DIFF
--- a/docs/building-a-workload.md
+++ b/docs/building-a-workload.md
@@ -2,9 +2,7 @@
 
 This guide walks through building a workload for the Azure Functions Core Tools v5 CLI. A workload is a NuGet package that the CLI loads at runtime to extend its behavior, most commonly to provide `func init` / `func new` support for a specific language stack (e.g. Node.js, Python, Java), but a workload can also contribute brand-new subcommands.
 
-> **Status**: the abstractions, DI host, and `func workload install` / `uninstall` commands described below are in the tree as of this PR. Workloads are installed from a local `.nupkg` on disk; NuGet feed acquisition lands in a follow-up.
->
-> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
+> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta` / `rid-pointer`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
 
 ## Architecture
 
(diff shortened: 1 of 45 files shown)


fabiocav:
Superseded by stack #5521, which splits this change into six reviewable PRs.

This PR was complete and green, but at 45 files / +4165−740 it was impractical to review as a unit. The same change now lands bottom-to-top as:

1. #5515 — Workload SDK: author RID pointer and implementation packages
2. #5516 — Workloads: add rid-pointer manifest kind and runtime identifier
3. #5517 — Workloads: model physical and logical package ownership in the registry
4. #5518 — Workloads: resolve and install RID pointer packages
5. #5519 — Workloads: use RID-aware installer APIs in workload commands
6. #5520 — Workloads: activate RID pointer workloads and fix Python worker content root

Every one of the 45 files here is carried by exactly one layer, so the stack reproduces this change in full. Each layer builds with `CI=true dotnet build -c Release` at 0 warnings and passes the suite on its own, so each is independently reviewable and revertible.

**Three intentional differences from this PR:**

1. **Test assertions.** The new tests here used raw `Assert.*`, which `.github/instructions/tests.instructions.md` forbids in favour of AwesomeAssertions. All layers converted them (`Assert.Throws<T>` → `ThrowExactly<T>()`, since the xUnit form is exact-type-matching). Behaviour is unchanged.

2. **A stale doc comment** on `WorkloadJsonContext` was corrected in #5516; it explained a camel-casing rationale that no longer applied once the converter became a plain `JsonConverter<WorkloadKind>`.

3. **An added fix not present here.** `Workloads.Workers.Python.csproj` hardcoded `PackagePath="tools/any/"` while this PR changes the CLI to derive the payload root from the entry's runtime identifier (`WorkloadPackageLayout.GetContentRoot`). `DefaultFunctionsWorkerInstaller` has no `tools/any` fallback, so as written this PR would have broken the Python worker's payload lookup. It only reproduces at pack time, which is why builds and the host workload install both passed. Fixed in #5520 and verified by packing all five RIDs: the payload now lands under `tools/<rid>/` with zero entries under `tools/any/`.

PR #5515, opened 2026-08-06
Workload SDK: author RID pointer and implementation packages

fabiocav:
This is **layer 1 of 6** in a stack decomposing #5512 ("Add RID pointer package support") into small dependent PRs. It targets `vnext` and is the bottom of the stack.

This layer is the MSBuild/SDK authoring half of the feature. It has no dependency on the `func` CLI code — `WriteWorkloadJson` only references `System.Text.Json` and `Microsoft.Build.Framework` — so it builds and tests standalone on `vnext`.

## What the SDK now emits

When a workload sets `RuntimeIdentifiers`, the Workload SDK produces two package shapes:

- **Outer pointer package** — `kind: rid-pointer`, with a generated non-empty RID-to-package map in `packages`. It carries no payload and no entry point.
- **Inner RID implementation packages** — keep their authored `workload` or `content` kind and emit a required, lowercase `runtimeIdentifier`.

`WriteWorkloadJson` gained a validation pass that fails the task (rather than writing a bad manifest) when:

- a `rid-pointer` declares an entry point, a `runtimeIdentifier`, or no inner packages;
- a non-pointer kind declares inner packages;
- a `runtimeIdentifier` is empty, padded, or not lowercase;
- inner package runtime identifiers are duplicated, or an inner package is not named `<pointer-package-id>.<rid>`;
- a `workload` kind is missing either the entry-point assembly path or type.

`Workload.targets` flows `PackageId` and the resolved per-RID identifier into the task, and includes the identifier in the incremental hash so the manifest regenerates when the RID changes.

`docs/building-a-workload.md` documents the same authoring contract, including that implementation payloads ship under `tools/<rid>/`.

## Validation

- `CI=true dotnet build -c Release` — **0 warnings, 0 errors**.
- `dotnet test test/Workloads/Sdk.Tests/Workloads.Sdk.Tests.csproj` — **35/35 passed**.
- End-to-end pack check on the multi-RID Python worker workload: the pointer package emits `kind: rid-pointer` with all five RIDs mapped and no `tools/` payload; each implementation package emits `kind: content` with its `runtimeIdentifier`.

## Notes

- No `src/Func/` code, version props, or release notes are touched — this is a decomposition of already-reviewed work, not a release.
- The eight new tests are written with AwesomeAssertions to match the rest of the file and `.github/instructions/tests.instructions.md`; behavior is identical to the source diff.

### DIFF
--- a/docs/building-a-workload.md
+++ b/docs/building-a-workload.md
@@ -2,9 +2,7 @@
 
 This guide walks through building a workload for the Azure Functions Core Tools v5 CLI. A workload is a NuGet package that the CLI loads at runtime to extend its behavior, most commonly to provide `func init` / `func new` support for a specific language stack (e.g. Node.js, Python, Java), but a workload can also contribute brand-new subcommands.
 
-> **Status**: the abstractions, DI host, and `func workload install` / `uninstall` commands described below are in the tree as of this PR. Workloads are installed from a local `.nupkg` on disk; NuGet feed acquisition lands in a follow-up.
->
-> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
+> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta` / `rid-pointer`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
 
 ## Architecture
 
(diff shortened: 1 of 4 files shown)


eng/build/Workloads/Workload.targets
```diff
@@ -104,8 +106,10 @@
       OutputPath="$(_WorkloadJsonFile)"
       Schema="$(_FuncWorkloadSchema)"
       Kind="$(WorkloadKind)"
+      PackageId="$(PackageId)"
```
fabiocav, line 109: This enables enforcement of the `{PackageId}.{rid}` naming contract. The CLI resolves implementations by *exact id* computed from the pointer's id plus the RID. A misnamed inner package would fail at `func workload install` on a user's machine, but with this,  it's a build error.

eng/build/Workloads/Workload.targets
```diff
@@ -104,8 +106,10 @@
...
       DisplayName="$(Title)"
       Description="$(Description)"
+      RuntimeIdentifier="$(_WorkloadManifestRuntimeIdentifier)"
```
fabiocav, line 112: Emit RID into `workload.json`, which is what the CLI reads both to verify the package matches the RID it resolved, and to derive `tools/<rid>`

jviau (approved):

eng/build/Workloads/Workload.targets
```diff
@@ -47,6 +47,7 @@
       <_WorkloadRidPath Condition="'$(RuntimeIdentifier)' == ''">any</_WorkloadRidPath>
       <_WorkloadRidPath Condition="'$(RuntimeIdentifier)' != ''">$(RuntimeIdentifier)</_WorkloadRidPath>
       <_WorkloadPackageContentRoot>tools/$(_WorkloadRidPath)/</_WorkloadPackageContentRoot>
+      <_WorkloadManifestRuntimeIdentifier Condition="'$(RuntimeIdentifier)' != '' AND $(CreateRidSpecificWorkloadPackages)">$(RuntimeIdentifier)</_WorkloadManifestRuntimeIdentifier>
```
fabiocav, line 50: Gate on CreateRidSpecificWorkloadPackages, not RuntimeIdentifier alone: a project can be
RID-specific for ordinary build reasons without being an inner package of a pointer set.
This is the same condition that sets the id suffix, rid: tag, and FuncCliWorkloadRidPackage
type, so the manifest can never claim a RID the package type denies, which the CLI would reject.

satvu (approved):

docs/building-a-workload.md
```diff
@@ -2,9 +2,7 @@
...
 
-> **Status**: the abstractions, DI host, and `func workload install` / `uninstall` commands described below are in the tree as of this PR. Workloads are installed from a local `.nupkg` on disk; NuGet feed acquisition lands in a follow-up.
->
-> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
+> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta` / `rid-pointer`), and the install pipeline are specified in [`docs/proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
```
satvu:
```suggestion
> **Spec**: this guide is the authoring view. The on-disk and on-feed layout, the `workload.json` schema, the `kind` discriminator (`workload` / `content` / `meta` / `rid-pointer`), and the install pipeline are specified in [`./proposed/workload-package-layout.md`](./proposed/workload-package-layout.md). Consult that doc for the contract; this guide stays focused on the happy-path authoring experience for `kind: workload`.
```

PR #5516, opened 2026-08-06
Workloads: add rid-pointer manifest kind and runtime identifier

fabiocav:
Layer 2 of 6 in a stack decomposing #5512 ("Add RID pointer package support") into small dependent PRs. Stacks on #5515 (layer 1) and targets `fabiocav/workload-sdk-rid-packages`.

## What this changes

This layer is the `workload.json` manifest contract that the CLI reads. It teaches the contract to describe RID pointer packages and their RID-specific implementations.

- **`WorkloadKind.RidPointer`** — a fourth kind whose JSON wire value is the hyphenated `rid-pointer`. Camel casing can't produce that, so `WorkloadKindJsonConverter` is rewritten from a `JsonStringEnumConverter<WorkloadKind>` into an explicit `JsonConverter<WorkloadKind>` with a declared read/write mapping that throws `JsonException` on unknown values.
- **`WorkloadMetadata.Packages`** — the RID-to-implementation package id map, required only for `rid-pointer`.
- **`WorkloadMetadata.RuntimeIdentifier`** — the RID an implementation package declares for itself.
- **`WorkloadMetadataReader` validation** for both shapes:
  - `rid-pointer` must declare a non-empty `packages` map, with lowercase and unique keys and non-empty package ids, and must not declare an `entryPoint` or a `runtimeIdentifier`.
  - `workload` and `content` must not declare `packages`, and any `runtimeIdentifier` they declare must be normalized (lowercase, untrimmed-free).
  - `meta` stays payload-free: no `entryPoint`, no `packages`, no `runtimeIdentifier`.

Validation ordering is preserved exactly as written in the source PR — it's load-bearing, since a test asserts which message wins when a manifest is invalid in more than one way.

## Testing

- `dotnet build -c Release` with `CI=true`: **0 warnings, 0 errors**. Adding the enum member produced no non-exhaustive `switch` fallout — every existing `switch` over `WorkloadKind` already had a default arm, so nothing outside the slice needed to change.
- `Func.Tests` filtered to `WorkloadMetadataReader`: 28 passed.
- Full `Func.Tests`: 1328 passed, 1 failed — `ProcessRunnerTests.RunAsync_Timeout_IsReported`, a pre-existing environment-dependent failure unrelated to this change.

No version or release-note changes: this is a decomposition of already-reviewed work, not a release.

### DIFF
--- a/src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
+++ b/src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
@@ -95,19 +95,13 @@ internal sealed class WorkloadMetadataReader : IWorkloadMetadataReader
                 return metadata;
 
             case WorkloadKind.Content:
-                if (metadata.EntryPoint is not null)
-                {
-                    throw new InvalidWorkloadException(
-                        $"'{metadataPath}' declares kind 'content' " +
-                        "but also defines an entryPoint. Remove the entryPoint or change the kind to 'workload'.");
-                }
-
+                ValidateNoEntryPoint(metadata, metadataPath);
                 ValidateNoPackages(metadata, metadataPath);
                 ValidateRuntimeIdentifier(metadata, metadataPath);
                 return metadata;
 
             case WorkloadKind.Meta:
-                ValidateNoEntryPoint(metadata, metadataPath, "meta");
+                ValidateNoEntryPoint(metadata, metadataPath);
                 ValidateNoPackages(metadata, metadataPath);
                 if (metadata.RuntimeIdentifier is not null)
                 {
@@ -118,7 +112,7 @@ internal sealed class WorkloadMetadataReader : IWorkloadMetadataReader
                 return metadata;
 
             case WorkloadKind.RidPointer:
-                ValidateNoEntryPoint(metadata, metadataPath, "rid-pointer");
+                ValidateNoEntryPoint(metadata, metadataPath);
                 if (metadata.RuntimeIdentifier is not null)
                 {
                     throw new InvalidWorkloadException(
@@ -134,12 +128,13 @@ internal sealed class WorkloadMetadataReader : IWorkloadMetadataReader
         }
     }
 
-    private static void ValidateNoEntryPoint(WorkloadMetadata metadata, string metadataPath, string kind)
+    private static void ValidateNoEntryPoint(WorkloadMetadata metadata, string metadataPath)
     {
         if (metadata.EntryPoint is not null)
         {
             throw new InvalidWorkloadException(
-                $"'{metadataPath}' declares kind '{kind}' but also defines an entryPoint.");
+                $"'{metadataPath}' declares kind '{metadata.Kind.ToWireValue()}' " +
+                "but also defines an entryPoint. Remove the entryPoint or change the kind to 'workload'.");
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
```diff
@@ -90,19 +90,42 @@ public WorkloadMetadata Read(string packageDirectory)
...
+                ValidateRuntimeIdentifier(metadata, metadataPath);
                 return metadata;
 
             case WorkloadKind.Content:
-            case WorkloadKind.Meta:
-                // entryPoint is meaningless for content/meta packages — reject
-                // it at author time instead of silently ignoring the field.
                 if (metadata.EntryPoint is not null)
                 {
                     throw new InvalidWorkloadException(
```
jviau: Should this use `ValidateNoEntryPoint()`?

src/Func/Workloads/Discovery/WorkloadMetadataReader.cs
```diff
@@ -111,6 +134,72 @@ public WorkloadMetadata Read(string packageDirectory)
...
+            }
+        }
+    }
+
+    private static bool IsNormalizedRuntimeIdentifier(string value)
+        => !string.IsNullOrWhiteSpace(value)
+            && string.Equals(value, value.Trim(), StringComparison.Ordinal)
+            && string.Equals(value, value.ToLowerInvariant(), StringComparison.Ordinal);
+
+    private static string GetWireKind(WorkloadKind kind)
```
jviau: Add an extension method to `WorkloadKind` for resolving the display string.

src/Func/Workloads/Storage/WorkloadJsonContext.cs
```diff
@@ -25,9 +23,34 @@ namespace Azure.Functions.Cli.Workloads.Storage;
...
+            "content" => WorkloadKind.Content,
+            "meta" => WorkloadKind.Meta,
+            "rid-pointer" => WorkloadKind.RidPointer,
+            _ => throw new JsonException($"Unknown workload kind '{value}'."),
+        };
+    }
 
+    public override void Write(Utf8JsonWriter writer, WorkloadKind value, JsonSerializerOptions options)
+    {
+        string wireValue = value switch
```
jviau: this can use the same extension method I mentioned in the other comment.

src/Func/Workloads/Storage/WorkloadJsonContext.cs
```diff
@@ -25,9 +23,34 @@ namespace Azure.Functions.Cli.Workloads.Storage;
...
-/// <see cref="JsonStringEnumConverter{TEnum}"/> would emit PascalCase.
+/// Serializes <see cref="WorkloadKind"/> using the workload manifest's wire values.
 /// </summary>
-internal sealed class WorkloadKindJsonConverter() : JsonStringEnumConverter<WorkloadKind>(JsonNamingPolicy.CamelCase);
+internal sealed class WorkloadKindJsonConverter : JsonConverter<WorkloadKind>
+{
+    public override WorkloadKind Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        string? value = reader.GetString();
+        return value switch
```
jviau: consider a static extension method `WorkloadKind.Parse(...)`

jviau (approved):

satvu (approved):

PR #5517, opened 2026-08-06
Workloads: model physical and logical package ownership in the registry

fabiocav:
Layer 3 of 6 in a stack decomposing #5512 ("Add RID pointer package support") into small dependent PRs. Stacks on #5516 (layer 2) — this PR targets `fabiocav/stack-2-rid-pointer-manifest-schema`, not `vnext`.

## What this layer does

Teaches the on-disk workload registry the difference between the **physical** package installed on disk (a RID-specific implementation) and the **logical** package the user asked for (the RID pointer). This is the storage model everything above it in the stack depends on.

- `WorkloadEntry` gains `RuntimeIdentifier` (which RID the row was validated for) and `LogicalPackage` (which pointer selected it).
- `WorkloadEntry` gains `IsExplicitlyInstalled`, backed by a nullable persisted property. Rows written by older CLIs have no such field, and deserializing a missing `bool` as `false` would silently strip explicit ownership from every previously user-installed workload, making them eligible for `func workload prune`. The public property defaults a missing value to `true` instead. This is deliberate backward compatibility for the first release that ships this change.
- `IWorkloadStore` gains ownership APIs: `AddOwnershipAsync`, `RemoveOwnershipAsync`, `MoveLogicalOwnershipAsync`, `MoveExplicitOwnershipAsync`. A physical row can be owned explicitly by the user, logically by a pointer, or both. `InstallRefCount` now counts distinct owners, so repeating an install by the same owner is idempotent and removing one owner leaves the other intact. A row accepts at most one logical owner; attaching a different one throws.
- `IWorkloadRuntimeIdentifierProvider` makes the current RID substitutable in tests, registered in `AddWorkloadStorage`.

The persisted JSON stays a flat array; a test asserts that shape.

## Expected loose ends at this layer

`IWorkloadRuntimeIdentifierProvider` is referenced only by its DI registration here. Its consumers are `WorkloadInstaller` (layer 4) and `WorkloadLoader` (layer 6); it lives at this layer because it must sit below both. Likewise the new store ownership APIs are exercised only by tests here — the installer starts calling them in layer 4.

## Validation

- `CI=true dotnet build -c Release` — **0 warnings, 0 errors**.
- `dotnet test test/Func.Tests/Func.Tests.csproj -c Release --filter "FullyQualifiedName~WorkloadStore"` — 19/19 passed.
- `dotnet test test/Func.Tests/Func.Tests.csproj -c Release` — 1335 passed, 1 failed: `ProcessRunnerTests.RunAsync_Timeout_IsReported`, a pre-existing timing flake unrelated to this change.

No version or release-note changes in this layer.

### DIFF
--- a/src/Func/Workloads/Storage/WorkloadRegistry.cs
+++ b/src/Func/Workloads/Storage/WorkloadRegistry.cs
@@ -37,6 +37,8 @@ internal sealed class WorkloadRegistry
 /// </summary>
 internal sealed class WorkloadEntry
 {
+    private bool? _isExplicitlyInstalled;
+
     /// <summary>
     /// NuGet package id (e.g. <c>"Azure.Functions.Cli.Workloads.Dotnet"</c>).
     /// Matched case-insensitively.
@@ -78,11 +80,38 @@ internal sealed class WorkloadEntry
     public string Source { get; init; } = string.Empty;
 
     /// <summary>
-    /// Number of installs that brought this package in. An explicit
-    /// <c>func workload install &lt;id&gt;</c> bumps the count; each meta package
-    /// that lists this package as a member also bumps the count at meta install.
-    /// Explicit uninstall removes the entry regardless of the count; meta
-    /// uninstall decrements and removes only when the count hits zero.
+    /// Runtime identifier validated for a RID-specific implementation package.
+    /// Null for ordinary and legacy entries.
+    /// </summary>
+    public string? RuntimeIdentifier { get; init; }
+
+    /// <summary>
+    /// Whether this physical package has an explicit install owner. Defaults
+    /// to true so registry rows written by older CLIs retain their ownership.
+    /// </summary>
+    [JsonIgnore]
+    public bool IsExplicitlyInstalled
+    {
+        get => _isExplicitlyInstalled ?? true;
+        init => _isExplicitlyInstalled = value;
+    }
+
+    [JsonPropertyName("isExplicitlyInstalled")]
+    public bool? PersistedIsExplicitlyInstalled
+    {
+        get => _isExplicitlyInstalled;
+        init => _isExplicitlyInstalled = value;
+    }
+
+    /// <summary>
+    /// Logical pointer package that selected this implementation, or null when
+    /// no pointer owner is attached.
+    /// </summary>
+    public LogicalPackage? LogicalPackage { get; init; }
+
+    /// <summary>
+    /// Number of distinct owners: explicit ownership, logical pointer ownership,
+    /// and meta owners. Repeating an install by the same owner is idempotent.
     /// Defaults to 1 for legacy entries that pre-date this field.
     /// </summary>
     public int InstallRefCount { get; init; } = 1;
@@ -94,6 +123,24 @@ internal sealed class WorkloadEntry
     /// for <see cref="WorkloadKind.Content"/>.
     /// </summary>
     public EntryPointSpec? EntryPoint { get; init; }
+}
+
+/// <summary>
+/// User-facing pointer package metadata persisted with its selected physical implementation.
+/// </summary>
+internal sealed class LogicalPackage
+{
+    public required string PackageId { get; init; }
+
+    public required string PackageVersion { get; init; }
+
+    public IReadOnlyList<string> Aliases { get; init; } = [];
+
+    public string DisplayName { get; init; } = string.Empty;
+
+    public string Description { get; init; } = string.Empty;
+
+    public string Source { get; init; } = string.Empty;
 }
 
 /// <summary>
--- a/src/Func/Workloads/Storage/WorkloadRegistry.cs
+++ b/src/Func/Workloads/Storage/WorkloadRegistry.cs
@@ -37,8 +37,6 @@ internal sealed class WorkloadRegistry
 /// </summary>
 internal sealed class WorkloadEntry
 {
-    private bool? _isExplicitlyInstalled;
-
     /// <summary>
     /// NuGet package id (e.g. <c>"Azure.Functions.Cli.Workloads.Dotnet"</c>).
     /// Matched case-insensitively.
@@ -86,22 +84,11 @@ internal sealed class WorkloadEntry
     public string? RuntimeIdentifier { get; init; }
 
     /// <summary>
-    /// Whether this physical package has an explicit install owner. Defaults
-    /// to true so registry rows written by older CLIs retain their ownership.
+    /// Whether this physical package lacks an explicit install owner. Defaults
+    /// to false so registry rows written by older CLIs retain their ownership.
     /// </summary>
-    [JsonIgnore]
-    public bool IsExplicitlyInstalled
-    {
-        get => _isExplicitlyInstalled ?? true;
-        init => _isExplicitlyInstalled = value;
-    }
-
-    [JsonPropertyName("isExplicitlyInstalled")]
-    public bool? PersistedIsExplicitlyInstalled
-    {
-        get => _isExplicitlyInstalled;
-        init => _isExplicitlyInstalled = value;
-    }
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+    public bool IsImplicitlyInstalled { get; init; }
 
     /// <summary>
     /// Logical pointer package that selected this implementation, or null when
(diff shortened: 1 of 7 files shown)


src/Func/Workloads/Storage/WorkloadStore.cs
```diff
@@ -93,6 +93,153 @@ public async Task ReplaceWorkloadAsync(
...
+        => await MoveOwnershipAsync(oldPackageId, oldVersion, newEntry, WorkloadOwnershipKind.Logical, cancellationToken);
+
+    public async Task<WorkloadOwnershipMoveResult> MoveExplicitOwnershipAsync(
+        string oldPackageId,
+        string oldVersion,
+        WorkloadEntry newEntry,
+        CancellationToken cancellationToken = default)
+        => await MoveOwnershipAsync(oldPackageId, oldVersion, newEntry, WorkloadOwnershipKind.Explicit, cancellationToken);
+
+    private async Task<WorkloadOwnershipMoveResult> MoveOwnershipAsync(
```
satvu: Can this leverage "remove" + "add"? There's one more I/O read this way (since both methods do one) but less code duplication.

fabiocav: Good suggestion, and I went and measured it rather than eyeballing it -- but I'd like to keep the single-pass version, for two reasons.

**The cost is a bit higher than one read.** `AddOwnershipAsync` and `RemoveOwnershipAsync` each do their *own* read and their *own* write (`ReadRegistryAsync` + `WriteRegistryAsync`), whereas `MoveOwnershipAsync` does one read and one write total. So composing it costs an extra read *and* an extra write -- and the write is the atomic temp-file-plus-rename, which is the expensive half.

**The bigger issue is atomicity.** Composing from remove + add turns one registry mutation into two separately persisted states. In the common case the old row's only remaining owner *is* the one being moved, so `RemoveOwnershipAsync` takes the `remainingReferences <= 0` branch, deletes the row, and commits that to disk. If anything fails between that write and the subsequent add -- cancellation, a crash, or the `UnauthorizedAccessException` we already see intermittently on Windows during the rename -- the workload is gone from the registry while its payload is still on disk, and nothing owns it.

There's also a signalling wrinkle: `MoveOwnershipAsync` returns `PreviousEntryRemoved`, which the caller uses to decide whether to delete the old install directory. That would have to be reconstructed across the two calls.

Happy to revisit if you'd still prefer the smaller surface -- the duplication is real, and an internal helper taking the already-loaded registry might get most of the dedup without giving up the single write.

src/Func/Workloads/Storage/WorkloadStore.cs
```diff
@@ -108,6 +255,75 @@ private static int FindIndex(WorkloadRegistry registry, string packageId, string
...
+        }
+
+        // Payload metadata comes from the incoming entry: the caller may have just replaced the payload
+        // on disk, and a legacy row would otherwise keep stale values such as a missing runtime identifier.
+        return CopyWithOwnership(
+            incoming,
+            current.IsExplicitlyInstalled || ownership == WorkloadOwnershipKind.Explicit,
+            current.LogicalPackage ?? incoming.LogicalPackage,
+            current.InstallRefCount + (attachExplicit ? 1 : 0) + (attachLogical ? 1 : 0));
+    }
```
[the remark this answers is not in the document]
fabiocav: Confirmed -- this is a real invariant break, and thanks for catching it. I reproduced it with a throwaway test before responding:

- `AddOwnershipAsync(entry, Explicit)` where the stored row has no logical owner and `entry.LogicalPackage` is non-null attaches the logical owner but leaves `InstallRefCount` unchanged.
- A subsequent `RemoveOwnershipAsync(..., Logical)` then returns `EntryRemoved: true` -- it decrements a count that never included that owner, hits zero, and deletes a row whose *explicit* ownership should have survived.

Corroborating evidence that it's an oversight rather than intent: `NormalizeNewEntry` handles the equivalent new-row case and counts `entry.LogicalPackage is not null` **regardless** of ownership kind. So for the same owner set the new-row path yields a refcount of 2 and the merge path yields 1 -- an asymmetry inside the same file.

**Scope, for whoever picks this up:** it is not reachable from product code today. The only caller is `WorkloadInstaller.AttachOwnershipAsync`, which is guarded by `HasOwnership(existing, ownership)`; getting past that guard with `Explicit` requires `IsExplicitlyInstalled == false`, and combined with a null stored `LogicalPackage` that describes a zero-owner row, which the "every row has at least one owner" invariant excludes. So this is a latent contract defect reachable through `IWorkloadStore` directly, not shipping data loss. Worth fixing regardless, since the store's own invariant shouldn't depend on an accident of its caller.

Taking the second option you offered (allow the attach, make the count follow it), the fix is one line:

```csharp
bool attachLogical = current.LogicalPackage is null && incoming.LogicalPackage is not null;
```

That counts the logical owner whenever one is actually attached rather than only when the caller labelled the operation `Logical`, which also makes `MergeOwnership` symmetric with `NormalizeNewEntry`. I checked it against every installer-reachable state and it is behaviour-identical in all of them, so it carries no risk for the layers built on top.

One coordination note: this PR is one layer of a stack carved from #5512, and the defect originates there, so the layers above inherit it from the same snapshot. I'm confirming where the fix should land before pushing it.

test/Func.Tests/Workloads/Storage/WorkloadStoreTests.cs
```diff
@@ -184,6 +199,172 @@ public async Task SaveWorkloadAsync_PreservesPreviousRegistry_WhenSerializationF
...
+            Kind = WorkloadKind.Content,
+            RuntimeIdentifier = "win-x64",
+            IsExplicitlyInstalled = false,
+            LogicalPackage = new LogicalPackage { PackageId = "other.pkg", PackageVersion = "1.0.0" },
+            InstallRefCount = 1,
+        };
+
+        await FluentActions.Awaiting(() => _store.AddOwnershipAsync(conflicting, WorkloadOwnershipKind.Logical))
+            .Should().ThrowExactlyAsync<InvalidOperationException>();
+    }
```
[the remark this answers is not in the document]
fabiocav: Agreed, and the test is written -- it fails against the current behaviour exactly as you describe. It asserts both halves of the invariant: that an `Explicit` add doesn't silently attach a logical owner, and that `RemoveOwnershipAsync(..., Logical)` afterwards doesn't delete a row that is still explicitly installed. It'll land together with the `MergeOwnership` fix discussed on the other thread.

src/Func/Workloads/Storage/WorkloadStore.cs
fabiocav: Marking this as unresolved. Removed some rogue agent comments :)

src/Func/Workloads/Storage/WorkloadRegistry.cs
```diff
@@ -78,11 +80,38 @@ internal sealed class WorkloadEntry
...
+    /// Null for ordinary and legacy entries.
+    /// </summary>
+    public string? RuntimeIdentifier { get; init; }
+
+    /// <summary>
+    /// Whether this physical package has an explicit install owner. Defaults
+    /// to true so registry rows written by older CLIs retain their ownership.
+    /// </summary>
+    [JsonIgnore]
+    public bool IsExplicitlyInstalled
```
jviau: Can we simplify this serialization by flipping the meaning of this property?

``` csharp
[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
public bool IsImplicitlyInstalled { get; init; }
```

satvu (approved):

src/Func/Workloads/Storage/WorkloadStore.cs
jviau: The implementation could also be decomposed into steps, and then the public methods would orchestrate those steps for their goal:

Remove -> Read + RemoveCore + Write
Add -> Read + AddCore + Write
Move -> Read + RemoveCore + AddCore + Write

jviau (approved):

src/Func/Workloads/Storage/WorkloadStore.cs
fabiocav: Composing the public remove and add operations would split the move across two registry writes and break its atomicity guarantee. We can reduce duplication by extracting mutation-only `RemoveOwnershipCore` and `AddOwnershipCore` helpers.

jviau (approved):

PR #5518, opened 2026-08-06
Workloads: resolve and install RID pointer packages

fabiocav:
Layer 4 of 6 decomposing #5512 ("Add RID pointer package support"). Stacked on #5517 (`fabiocav/stack-3-registry-ownership-model`) — review and merge that one first.

This is the core layer: the installer learns to resolve a RID pointer package to the correct RID-specific implementation, install it, and track ownership.

## Resolution

`InstallFromCatalogAsync` / `InstallFromPackageAsync` now detect a rid-pointer package, select the implementation mapped to the current runtime identifier, and install that implementation instead of the pointer. Selection is **exact-id and exact-version** — a pointer names its implementations explicitly, so there is no fuzzy or latest-wins matching inside the pointer map.

Failure modes are explicit:

- No entry for the current RID → `WorkloadPackageNotFoundException` naming the current RID and the RIDs the pointer does cover.
- Entry exists but the implementation isn't on the feed → `WorkloadPackageNotFoundException` calling out a likely partial publish, with the implementation id and source.
- Local pointer with no sibling `.nupkg` alongside it → `FileNotFoundException`; the catalog is deliberately not consulted for local installs.
- A RID implementation carrying the ordinary `FuncCliWorkload` package type, or one built for a different RID, is rejected with `InvalidWorkloadException`.

## Ownership

Installing through a pointer records the physical implementation as the registry row and attaches the pointer as the *logical* package. Reuse is keyed on **package id + version** — not the source feed. A package with the same id served from two different feeds is the same package, matching NuGet's own behavior, so an implementation already on disk is reused even when it was originally installed from another source.

When a pointer resolves to a different implementation than the one it currently owns (a RID pivot, or a version bump), logical ownership **moves** off the previous row rather than being duplicated, and the orphaned install directory is cleaned up. `force: true` never reuses an installed implementation; it reinstalls.

## Layout

`WorkloadPackageLayout` replaces the hardcoded `tools/any` payload path: RID-specific packages resolve to `tools/<rid>`, RID-agnostic ones keep `tools/any`. Only the installer consumes it at this layer; the remaining consumers land in layer 6.

## Interface

`IWorkloadInstaller` gains `UpdateAsync` and `UninstallAsync` overloads that take a `WorkloadOwnershipKind`. The existing signatures are unchanged so current command call sites keep compiling — the commands move to the new overloads in layer 5.

## Validation

- `CI=true dotnet build -c Release` — 0 warnings, 0 errors.
- `dotnet test test/Func.Tests/Func.Tests.csproj -c Release` — 1351 passed, 1 failed.
- Full solution test run — all other projects green.

The single failure is `ProcessRunnerTests.RunAsync_Timeout_IsReported`, which is pre-existing and unrelated (verified failing on the base branch with these changes stashed).

No command production or test files needed changes at this layer. Test assertions carved from the source PR were converted from raw `Assert.*` to AwesomeAssertions per `tests.instructions.md`; the production code is byte-identical to the reviewed source.


src/Func/Workloads/Install/WorkloadInstaller.cs
```diff
@@ -12,293 +12,1009 @@
...
             await _store.SaveWorkloadAsync(entry, cancellationToken);
+            return new WorkloadInstallResult(entry, false);
         }
         catch
         {
-            TryDeleteDirectory(installPath);
+            TryDeleteDirectory(stagingPath);
+            if (movedToFinal && existing is null)
+            {
+                TryDeleteDirectory(installPath);
```
satvu: Is there anything actionable that can be surfaced to the user from here? Should they try installation again (maybe with `--force`?

satvu (commented):
`WorkloadInstaller` is massive - can this be decomposed into something smaller (it's 1570 lines)?

src/Func/Workloads/Install/WorkloadInstaller.cs
jviau, line 1: Can we break up this file? Partial class and splitting across separate files is okay. Or decomposing into different types if that works better

fabiocav: Already working on that. @satvu  also mentioned. Planned on having this broken up before sending the original PR, but will have this updated soon.

src/Func/Workloads/Install/WorkloadPackageInspection.cs
```diff
@@ -0,0 +1,18 @@
...
+
+internal enum WorkloadPackageRole
+{
+    Ordinary,
+    Pointer,
+    RuntimeIdentifierImplementation,
+}
+
+internal sealed record WorkloadPackageIdentity(string PackageId, string Version, IReadOnlyList<string> PackageTypes,
+    IReadOnlyList<string> Aliases, IReadOnlyList<string> RuntimeIdentifierTags, string? Title, string? Description);
```
jviau: nit: remove `Tags`, have just `RuntimeIdentifiers`

src/Func/Workloads/Install/WorkloadPackageInspector.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        if (metadata.Kind == WorkloadKind.RidPointer)
+        {
+            RequirePackageType(identity, standardType, WorkloadPackageTypes.Workload);
+            if (identity.RuntimeIdentifierTags.Count > 0)
+            {
+                throw new InvalidWorkloadException($"RID pointer package '{identity.PackageId}' cannot declare a rid: tag.");
+            }
+
+            if (files.Any(IsPayloadFile))
```
jviau, line 105: Why do we care about validating this? Would it not be sufficient to simply ignore any content they bundle in the nupkg (which we already will ignore when package is a rid-pointer)? They could package with files under any other random folder and we won't care. This feels arbitrary to complain about this specific folder.

jviau (approved):

src/Func/Workloads/Install/WorkloadDeploymentService.cs
```diff
@@ -0,0 +1,629 @@
...
+            EntryPoint = package.Metadata.EntryPoint,
+            Kind = package.Metadata.Kind,
+            Source = source,
+            RuntimeIdentifier = package.Metadata.RuntimeIdentifier,
+            IsExplicitlyInstalled = ownership == WorkloadOwnershipKind.Explicit,
+            LogicalPackage = logicalPackage,
+            InstallRefCount = 1,
+        };
+
+    private static WorkloadEntry CopyForOwnership(WorkloadEntry entry, WorkloadOwnershipKind ownership, LogicalPackage? logicalPackage)
```
satvu, line 376: CopyForOwnership and MergeForReinstall both manually copy all 12 WorkloadEntry properties (the same pattern as WorkloadStore.CopyWithOwnership). If a property is added to WorkloadEntry, all three methods need updating independently.

The store layer has RoundTripsAllFields to catch this, but the deployment service doesn't have an equivalent. Should we add a similar test?

src/Func/Workloads/Install/WorkloadPackageInspector.cs
fabiocav: `tools/` is the defined workload payload root, while a RID pointer is intended to be metadata-only.

This is strictly following the spec. I agree that having a folder here won't cause problems, but if we don't have good scenarios to support allowing this, the validation catches an accidentally bundled payload instead of silently ignoring it. Files in unrelated folders are not interpreted as workload payload.

We can relax this eventually if needed.

src/Func/Workloads/Install/WorkloadDeploymentService.cs
fabiocav: This would be a good validation.  Let me add something to cover the scenario.

satvu (approved):

PR #5519, opened 2026-08-06
Workloads: use RID-aware installer APIs in workload commands

fabiocav:
Layer 5 of a 6-layer stack decomposing #5512 (RID-pointer workload support) into small, independently reviewable PRs.

**Targets `fabiocav/stack-4-installer-rid-resolution` (layer 4), not `vnext`.** Review the base first.

## What this layer does

Layer 4 added the ownership-aware `IWorkloadInstaller` overloads *additively*, deliberately preserving the pre-existing ones so the commands kept compiling unchanged. This layer is where the commands move onto the new overloads, so a RID pointer install is presented and managed as its **logical** identity rather than the physical RID-specific package that backs it.

- **`install`** — resolve an already-installed match through the logical package first, falling back to explicitly-installed physical packages. The non-interactive "already installed" hint now names the pointer id, so the suggested `func workload update <id>` is a command the user can actually run. Success/prompt text prefers the logical id, version, and aliases. The `--force to repair` suffix is dropped because the installer's own message now carries the actionable guidance.
- **`list`** — projects both logical and physical package identity, the resolved runtime identifier, and an ownership summary (`explicit`, `logical:<id>`, `meta:<n>`). Falls back to the loaded-workload projection when the registry is empty.
- **`uninstall`** / **`prune`** — release *logical* ownership through the logical overload, so a physical package still referenced by another owner isn't removed out from under it. Matching prefers logical candidates and only falls back to explicitly-installed physical ones.
- **`update`** — resolves the target to an explicit or logical identity and updates through the matching overload. `--all` skips physical packages already covered by a pointer, since updating them separately would re-resolve and replace the payload the logical update just staged. A RID pivot that lands on the same version now renders as a runtime-identifier change instead of a confusing "from X to X".

## Slice

Exactly 10 files, +612/-105:

**Production** (`src/Func/Commands/Workload/`): `WorkloadInstallCommand.cs`, `WorkloadListCommand.cs`, `WorkloadPruneCommand.cs`, `WorkloadUninstallCommand.cs`, `WorkloadUpdateCommand.cs`

**Tests** (`test/Func.Tests/Commands/Workload/`): the five matching `*CommandTests.cs`

No files outside this slice were touched.

## Deviations from the source PR

- **Assertion style.** The source wrote its 5 new tests with raw `Assert.Equal` / `Assert.Contains`. `.github/instructions/tests.instructions.md` requires AwesomeAssertions, and every prior layer in this stack converted them, so **11 assertions** were converted to `.Should()` form for consistency. Pre-existing assertions were left alone — no drive-by normalization.
- Fixed a stray-whitespace formatting defect in `WorkloadUpdateCommand.cs` (`WorkloadUpdateResult             result`) carried over from the source.

## Validation

- `CI=true dotnet build -c Release` — **0 Warning(s), 0 Error(s)**.
- Targeted: `--filter "FullyQualifiedName~WorkloadInstallCommand|...ListCommand|...PruneCommand|...UninstallCommand|...UpdateCommand"` — **87/87 passed**.
- Full `Func.Tests`: **1357 passed, 1 failed**. The only failure is `ProcessRunnerTests.RunAsync_Timeout_IsReported`, a machine-timing-dependent test that fails identically on `vnext` and on every layer of this stack, unrelated to this change.

### DIFF
--- a/src/Func/Commands/Workload/WorkloadInstallCommand.cs
+++ b/src/Func/Commands/Workload/WorkloadInstallCommand.cs
@@ -199,19 +199,38 @@ internal sealed class WorkloadInstallCommand : FuncCliCommand
             return null;
         }
 
-        // Same matching rules as uninstall/update: by package id always,
-        // by alias unless --exact. We don't need to enforce uniqueness
-        // here; if any version matches we prompt.
-        WorkloadEntry? match = installed.FirstOrDefault(e =>
+        IReadOnlyList<WorkloadEntry> logicalMatches = [.. installed.Where(e =>
             e.LogicalPackage is not null
             && (string.Equals(e.LogicalPackage.PackageId, identifier, StringComparison.OrdinalIgnoreCase)
                 || (!exact && e.LogicalPackage.Aliases.Any(a =>
-                    string.Equals(a, identifier, StringComparison.OrdinalIgnoreCase)))));
-        match ??= installed.FirstOrDefault(e =>
-            e.IsExplicitlyInstalled
-            && (string.Equals(e.PackageId, identifier, StringComparison.OrdinalIgnoreCase)
-                || (!exact && e.Aliases.Any(a =>
-                    string.Equals(a, identifier, StringComparison.OrdinalIgnoreCase)))));
+                    string.Equals(a, identifier, StringComparison.OrdinalIgnoreCase)))))];
+        string[] logicalIds = [.. logicalMatches
+            .Select(e => e.LogicalPackage!.PackageId)
+            .Distinct(StringComparer.OrdinalIgnoreCase)];
+        if (logicalIds.Length > 1)
+        {
+            throw CreateAmbiguousInstalledAlias(identifier, logicalIds);
+        }
+
+        WorkloadEntry? match = logicalMatches.FirstOrDefault();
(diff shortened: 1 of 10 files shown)


src/Func/Commands/Workload/WorkloadInstallCommand.cs
```diff
@@ -167,9 +167,7 @@ protected override async Task<int> ExecuteAsync(ParseResult parseResult, Cancell
         }
         catch (InvalidOperationException ex)
         {
-            throw new GracefulException(
-                $"{ex.Message} Pass --force to repair the install.",
-                isUserError: true);
+            throw new GracefulException(ex.Message, isUserError: true);
```
satvu, line 170: Is the guidance being removed invalid now?

fabiocav: Yep, the removal here is intentional. The installer now repairs an orphaned or invalid install during a normal install, so `--force` isn’t needed for that case anymore. Other `InvalidOperationException` cases aren’t fixed by `--force`, so keeping the blanket guidance would be misleading.

satvu (approved):

jviau (approved):

jviau (approved):

PR #5531, opened 2026-08-07
feat(templates): add TemplateEngine integration foundation

jviau:
## Summary

- add the `template-engine-integration` OpenSpec proposal, design, specification, and task breakdown from the stabilized source snapshot
- add Microsoft.TemplateEngine dependency wiring and the func-owned settings hive, host, bootstrapper, context parameters, and command-facing `Templater` foundation
- add focused coverage for host creation, context propagation, per-invocation environments, component registration, and `Templater` construction

## Scope

This PR establishes the existing integration foundation only. The OpenSpec tasks remain unchecked because the full context-bound catalog, constraint evaluation, group resolution, resolved-template invocation, strict argument parsing, and command integration are not implemented here.

Func-specific constraints are intentionally excluded: `FuncTemplateComponents` remains an empty integration-owned registration scaffold. Command-owned selection is also excluded, along with `TemplateGroupResolver`, `LanguageTiebreaker`, template-package lifecycle behavior, and `func init`/`func new` execution changes. The host parameters use the new `func:bundle-id` and `func:bundle-version` names.

## Validation

- `openspec validate template-engine-integration --strict`
- `dotnet test test\Func.Tests\Func.Tests.csproj -c Release --no-restore --filter "FullyQualifiedName~Templates.Engine" /p:TreatWarningsAsErrors=true` — 9 passed
- clean `dotnet build src\Func\Func.csproj -c Release --no-restore /p:TreatWarningsAsErrors=true` — 0 warnings, 0 errors


src/Func/Templates/Engine/Templater.cs
jviau, line 1: This implementation is only partially on the way towards what the spec asks. It is meant to be a starting point.

src/Abstractions/Bundles/ExtensionBundleResolution.cs
```diff
@@ -17,6 +17,9 @@ private ExtensionBundleResolution()
     public sealed record Resolved(string BundleId, string Version, string Path, ExtensionBundleSupportedRuntimeWarning? RuntimeWarning)
         : ExtensionBundleResolution;
 
+    public sealed record NotResolved(string Message, ExtensionBundleSupportedRuntimeWarning? RuntimeWarning = null)
```
satvu, line 20: There are other cases that seem like they may overlap and are more specific (EmptyIntersection, WorkloadMissing) what is the intended use of this? A catch-all for everything else? Or a specific scenario?

jviau: It is meant for a catch all of ~bundle resolution was never ran.~

correction: it means bundle resolution ran, but there is nothing to resolve (ie: no initialized project).

ahmedmuhsin: Backing Sarah on this. Every other case (`Resolved`, `WorkloadMissing`, `EmptyIntersection`, `NoCompatibleInstall`) means resolution ran and produced an outcome, so `NotResolved` reads as a fifth failure mode rather than "never attempted". `NotAttempted` or `Unknown` would say it.

Also worth asking whether it needs a case at all. `Templater.Create` only manufactures one when handed null, which is inventing a value to represent nothing. A nullable `ExtensionBundleResolution` might be the better model.

docs/template-engine-integration/design.md
```diff
@@ -0,0 +1,318 @@
...
+
+`ListAsync` loads installed templates and evaluates their constraints in the command context. It projects each engine template into a func-owned `TemplateCatalogEntry`, including identity, aliases, group identity, language, precedence, package origin when known, visibility, parameter metadata, and eligibility diagnostics. The same projected candidate model is used by template groups so listing and execution cannot disagree about a symbol.
+
+Restricted entries remain in the catalog so `func new --list` and diagnostic flows can explain why an installed template cannot run. The command decides whether ordinary presentation hides host-hidden templates; exact identity lookup still has access to them.
+
+The integration uses a common `ConstraintEvaluation` model for both catalog entries and resolution:
+
+```text
+Eligible
+Restricted(message, call-to-action)
```
ahmedmuhsin, line 114: `func init` doesn't require a bundle, so `func init` then `func new` on a clean machine is a normal path. At that point resolution returns `WorkloadMissing`, `Templater.Create` sets no bundle host params, and a bundle constraint has nothing to read. It can't evaluate, so it lands in `NotEvaluated`, which the design says points to a template or host configuration defect.

Nothing is broken though. The user just hasn't installed a bundle yet. `NewCommandBundleValidator` handles this today by pointing at `func workload install` or `func setup`, and I'd rather not lose that.

Can "required host context wasn't available" be its own state, or route to `Restricted`? Still fail-closed, it just needs a message that fits and somewhere to put the next step.

docs/template-engine-integration/tasks.md
```diff
@@ -0,0 +1,51 @@
...
+- [ ] 2.2 Add typed resolution and invocation outcomes for not found, restricted, constraint failure, ambiguous group, unsatisfied language, ambiguous variant, invalid arguments, file conflict, and success.
+- [ ] 2.3 Add the internal func host metadata reader and projection layer that parses the selected `.template.config/func.host.json`, joins `symbolInfo` to canonical parameter symbols, and stores the projected definitions on each candidate.
+- [ ] 2.4 Add projection tests for long-name overrides, canonical-name fallback, short-name suppression, hidden parameters, always-visible parameters, hidden templates, malformed host metadata, non-parameter symbols, and absent package metadata.
+
+## 3. Catalog and Constraint Evaluation
+
+- [ ] 3.1 Centralize constraint evaluation so listing and resolution share one fail-closed representation of eligible, restricted, not-evaluated, and failed constraints.
+- [ ] 3.2 Update `Templater.ListAsync` to return every installed template with current-context eligibility and diagnostics.
+- [ ] 3.3 Update the extension bundle constraint to read exact bundle identity and version from the renamed func host parameters.
```
ahmedmuhsin: Three situations that need different answers. No bundle installed at all. Older bundle installed but host.json already permits the newer one. And host.json range excluding what the template needs.

Only the last is a real incompatibility. The other two are just missing installs, and the constraint can't tell them apart because it only sees the resolved version, which is the highest installed version satisfying the range rather than the range itself.

So two asks.

Can we expose the declared range too, something like `func:bundle-range`?

And instead of a dead end, can `func new` offer to install and continue? Today they'd have to abandon the command, run `func workload install`, restart `func new`, and find their template again. A prompt keeps them in place. It's also a middle ground between v4 fetching bundles behind your back and never installing at all, with today's error as the non-interactive fallback.

The bundle seems like the safest candidate for that. First-party, version bounded by the declared range, nothing touched outside the workloads home. Not comparable to running post-action scripts from a third-party template package.

Would mean rebuilding the `Templater` after, since the context snapshot is immutable, but it's command-scoped so probably cheap.

docs/template-engine-integration/design.md
```diff
@@ -0,0 +1,318 @@
...
+**Goals:**
+
+- Give commands one func-owned entry point for engine bootstrapping, catalog access, constraint-aware resolution, and creation.
+- Prevent commands from depending on TemplateEngine implementation types.
+- Guarantee that discovery, constraints, host bindings, parameter metadata, and creation observe one immutable command context.
+- Make invalid selection states explicit and impossible to invoke.
+- Preserve diagnostics throughout progressive candidate narrowing.
+- Keep the resolved template responsible for its own invocation.
+
+**Non-Goals:**
```
ahmedmuhsin, line 20: Not for this change to solve, raising it here so it's threaded. Neither `azure-samples-template-pipeline` nor `func-init-quickstarts` mentions azd, and azd is the advertised path for these repos today. A few things I'd want settled before either gets specified, and it might be worth a non-goal here pointing at where that lands.

I used remote-mcp-functions-dotnet as a representative sample. Two function apps under `src/`, both declared as azd services in `azure.yaml`, a solution file, a root `Directory.Build.props`, and an `infra/` folder.

- Multi-project quickstarts have no single project to extract. Pulling out one subtree loses the build props it depends on and drops half the sample. Are these out of scope for a first pass, or does `func init` grow solution scaffolding? Worth deciding early, a lot of the interesting samples look like this.

- If `func init --quickstart` gives code without infra, the user lands on a project that doesn't match the repo's own README telling them to run `azd up`. Do project templates carry `azure.yaml` and `infra/` so the two compose, or are func quickstarts local-only with a pointer to azd for deployment? Either works, the CLI just needs to say which.

- The case that justifies func quickstarts alongside azd is local-first with no subscription, plus item templates that graft a pattern into an existing project. azd can't do the second at all. Worth stating that scope so the two don't read as competing answers.

- `azure.yaml` already carries `metadata.template` with a name and version, and there's the awesome-azd gallery. `func-new-search` would add NuGet identity plus its own manifest. Second registry over the same repos, or can discovery read what's already there?

- Is the plan a reusable workflow repos opt into with a small config file, with `template.json` generated from it? Hand-maintaining a template beside the sample code will drift. Generating at release time means it can't.

jviau: @ahmedmuhsin this looks like it is all around azd, quickstart, and sample repos. Those are not captured in this spec. This will be better discussed in the relevant spec PRs.

docs/template-engine-integration/design.md
jviau: Templates declare their own constraints. The expectation is that _project_ templates would not define a bundle constraint. Only _item_ templates will define a bundle constraint.

no bundle context -> `func init` -> bundle context now available -> `func new` flow is preserved here.

docs/template-engine-integration/tasks.md
jviau: I need to remove this task because the bundle constraint is not in this PR / spec.

These questions will be better served in the bundle constraint spec PR when it is created.

src/Abstractions/Bundles/ExtensionBundleResolution.cs
jviau: I used the same terminology we had for project context: `NotResolved`, which is what is used when there is no project available to resolve. This is the same: there is nothing for us to resolve. I keep this separate from `null` to differentiate "resolution never ran" and "resolution ran, unable to determine bundle".

satvu (approved):
Partial implementations seem like enough to get started with and docs are clear on tasks. Besides removing bundles related tasks (saving it for its own spec) looks good.

jviau:
> Besides removing bundles related tasks (saving it for its own spec) looks good.

@satvu can you clarify what you mean by this? Is there changes you would like to see to this PR still?

satvu:
> > Besides removing bundles related tasks (saving it for its own spec) looks good.
> 
> sarah (@satvu) can you clarify what you mean by this? Is there changes you would like to see to this PR still?

I meant this comment: https://github.com/Azure/azure-functions-core-tools/pull/5531#discussion_r3771097440 but I see the conversation is resovled?

docs/template-engine-integration/tasks.md
jviau: I agree these states need distinct outcomes, including an interactive install-and-resume path with the existing non-interactive error fallback. I removed the bundle-constraint task from this spec because the declared range and package-install orchestration belong together in the dedicated bundle constraint follow-up, rather than expanding this TemplateEngine foundation PR.

PR #5535, opened 2026-08-07
docs(templates): specify func new execution

jviau:
## Summary

- Define positional template grammar for `func new`, including template identifiers and forwarded template arguments.
- Restrict resolution to item templates and specify command-owned candidate selection.
- Require strict two-stage parsing: select a template first, then parse arguments against that template's schema.
- Define ambiguity handling, interactive prompting, and deterministic `--non-interactive` failures.
- Specify `--path` output targeting and dry-run behavior.

## Stack layer

This top layer is specification-only. The old engine-layer resolver and language-tiebreak implementation from source commits `6c92068b`, `d085e892`, and `0fd50821` was intentionally excluded because it conflicts with the command-owned staged parsing and candidate selection model. There are no compatible `func new` implementation changes in the source slice.

## Validation

- `git diff --check`


docs/func-new-execution/design.md
```diff
@@ -0,0 +1,254 @@
...
+
+Optional parameters are left to template defaults. Invalid explicit values fail immediately rather than being replaced through a prompt. Hidden required values without defaults surface as template-authoring diagnostics because the user has no advertised way to supply them.
+
+`--non-interactive` and `IInteractionService.IsInteractive == false` use the same refusal path. Diagnostics enumerate candidate identities or all missing required aliases so automation can correct the invocation in one pass.
+
+Existing selection rendering can be reused where it accepts func-owned candidate models, but prompting must not depend on legacy template workload or provider types.
+
+### `--path` is the template execution directory
+
+`--path` has one concrete filesystem meaning: the directory passed to TemplateEngine as its working and output directory.
```
satvu, line 186: Can `--path` be confusing here? It seems to mean the working/output directory but `--path` is unclear on its own.

jviau: It is used to specify the root directory for the template context. It is allegory to `dotnet new --output`. I wanted to use `--output|-o`, but we already use that for controlling the textual output format of commands (like specifying to output JSON from the command results). I used `--path` as that is a property we already have.

Usage will be:

`func new httptrigger --path my_folder/` --> will use `my_folder` as the root path for generating the http trigger template.

Any suggestions on how I can improve the wording here for clarity?

satvu: Looks like v4 doesn't use `--output` in this way so perhaps we still have time to change it (as this is preview). Could we potentially change how we currently use `--output` to `--format` and make this `--output`?

jviau: We could, but we would need to make that change to the entire CLI. `--output` is used for JSON formatting across many commands. We would want to align them all on either `--output` or `--format`.

I did discuss this with @fabiocav and decided to not start all that work. If we want to do that lets create a separate issue to track renaming and keep this design as is.

satvu (approved):
Filed this: https://github.com/Azure/azure-functions-core-tools/issues/5568 to keep track of convo for next month.

PR #5537, opened 2026-08-07
Replace Install-Dotnet with UseDotNet@2 to fix CFS violations

satvu:
## Summary

Replaces custom `Install-Dotnet` PowerShell calls with the 1ES-approved `UseDotNet@2` pipeline task to fix CFS violations.

## Problem

The `pipelineUtilities.psm1` `Install-Dotnet` function downloads `dotnet-install` scripts from `raw.githubusercontent.com` and then installs SDKs from `dotnetcli.azureedge.net`. 

## Fix

Replace the custom script invocations with `UseDotNet@2`, which is a 1ES-recognized task that uses approved infrastructure to install .NET SDKs. Network calls happen through the ADO agent's approved channels.

### Files changed
- `eng/ci/templates/official/jobs/build-test.yml` — 1 occurrence
- `eng/ci/templates/public/jobs/build-test-public.yml` — 2 occurrences (Windows build + Linux test jobs)

## About the hardcoded versions

The SDK versions (`3.1.415`, `6.0.417`, `8.0.204`) are **not arbitrary** — they match the existing `$DotnetSDKVersionRequirements` table in `pipelineUtilities.psm1` which previously controlled what `Install-Dotnet` installed. This ensures behavior parity with the old approach.

| SDK | Version | Reason |
|-----|---------|--------|
| .NET 3.1 | `3.1.415` | Required by `Microsoft.ManifestTool.dll` |
| .NET 6.0 | `6.0.417` | Build/test target framework |
| .NET 8.0 | `8.0.204` | Build/test target framework |

> **Note:** `UseDotNet@2` also supports wildcard versions (e.g. `6.0.x`) to auto-pick the latest patch. The hardcoded approach was chosen for safety/parity with the existing configuration. This can be revisited in a follow-up.

## Not included

- .NET 2.1 and 9.0 (also in the psm1 table) are not installed here — they were not being installed by the previous pipeline steps either if already present on the agent. Can be added if needed.
- The `pipelineUtilities.psm1` file itself is left unchanged; the `Install-Dotnet` function still exists but is no longer called from any pipeline.


eng/ci/templates/official/jobs/build-test.yml
```diff
@@ -57,10 +57,20 @@ jobs:
     inputs:
       versionSpec: '18.x'
 
-  - pwsh: |
-      Import-Module ".\pipelineUtilities.psm1" -Force
-      Install-Dotnet
-    displayName: 'Install .NET 6.0 and 3.1'
+  - task: UseDotNet@2
```
jviau: recommend extracting to a common template (`install-dotnet.yml` or `setup-deps.yml`)

jviau (approved):

eng/ci/templates/official/jobs/build-test.yml
satvu: Done — extracted to \ng/ci/templates/steps/install-dotnet.yml\ and referenced from all three call sites (official build, public build, public Linux test). Also removed the CFS-violating \dotnet-install.sh\ download from \install-linux-tools.yml\ since the shared template now handles .NET SDK installation via \UseDotNet@2\ on all platforms.

eng/ci/templates/public/jobs/build-test-public.yml
```diff
@@ -78,10 +76,7 @@ jobs:
     os: 'linux'
   steps:
   - template: ../../steps/install-linux-tools.yml
-  - pwsh: |
-      Import-Module ".\pipelineUtilities.psm1" -Force
```
kshyju, line 82: Can we remove this `.\pipelineUtilities.psm1` file now? Anything else using it?

eng/ci/templates/steps/install-dotnet.yml
```diff
@@ -0,0 +1,15 @@
+steps:
+- task: UseDotNet@2
+  displayName: 'Install .NET 3.1'
+  inputs:
+    version: '3.1.415'
```
kshyju, line 5: Double checking. Do we still use this version of .NET?

eng/ci/templates/public/jobs/build-test-public.yml
satvu: Good call — deleted it. No remaining references after the switch to UseDotNet@2.

eng/ci/templates/steps/install-dotnet.yml
kshyju: Saw the PR description. I wonder Manifest tool has been updated to use a latest stable version. worth checking (as a follow up item)

satvu: Yes — \uild/Build.csproj\ still targets \
etcoreapp3.1\, and the psm1 noted it's required by \Microsoft.ManifestTool.dll\. Might be worth a follow-up to migrate that project off 3.1 since it's EOL, but removing it here would break the build.

eng/ci/templates/public/jobs/build-test-public.yml
satvu: removed it - nothing else is using it.

kshyju (approved):

RohitRanjanMS (approved):

jviau (approved):

PR #4462, opened 2025-06-06
Func pack fix

gavin-aguiar:
### Issue describing the changes in this PR
This PR removes the dependency on 'distlib` from the `Func pack` process.  Replaced with native python libraries. 

 Tests Performed:
 
 Ran: `func pack --output test`
 
 
image

  
resolves #issue_for_this_pr

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

PR #4347, opened 2025-03-31
Merging `feature/foundation` branch into main with directory restructure

aishwaryabh:
### Issue describing the changes in this PR

resolves #4301 

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #5544, opened 2026-08-11
Remove external network dependencies from build process

satvu:
## Summary

Fixes network access violations in the build process by replicating approaches from `main` to eliminate external downloads.

There isn't a complete replication since main does not have a `build` folder w/ this approach (refactored and deleted in #4347, #4464, #4511). 

## Violations Fixed

| Source | Domain | Fix |
|--------|--------|-----|
| `AddDistLib()` | `github.com` (vsajip/distlib) | ✅ Removed entirely — replaced with native Python `Wheel`/`ScriptMaker` classes (matches main PR #4462) |
| `AddTemplatesNupkgs()` | `www.nuget.org` | ✅ Switched to `nuget restore` with `packages.config` + repo's `NuGet.Config` — Azure Artifacts credential provider handles auth (matches main's `download-templates.ps1`) |
| `TestSignedArtifacts` | `functionsbay.blob.core.windows.net` (sigcheck) | ✅ Removed entirely — ESRP pipeline's `SigntoolVerify` already handles signature verification |
| `AddTemplatesJson()` | `cdn.functions.azure.com` | Allowed domain (main still uses it via `Invoke-WebRequest`) |

### `build/BuildSteps.cs`
- Removed `AddDistLib()` method
- Removed `TestSignedArtifacts()` and `GetUnsignedBinaries()` methods
- Replaced `AddTemplatesNupkgs()` with `nuget restore` + `packages.config` approach
- Added `DownloadIfNotExists` helper for remaining templates JSON download

### `build/Program.cs`
- Removed `AddDistLib` and `TestSignedArtifacts` from build pipeline

### `build/Settings.cs`
- Removed `DistLibVersion`, `DistLibUrl`, NuGet URL properties, `NuGetFeedBaseUrl`
- Removed `SigcheckDownloadURL`, `SkipSigcheckTest`
- Kept `TemplatesJsonZip` (allowed domain, configurable via `TEMPLATES_JSON_ZIP_URL`)

### `eng/ci/templates/official/jobs/build-test.yml`
- Removed `TestSignedArtifacts --signTest` pipeline step (redundant with ESRP `SigntoolVerify`)

### `tools/python/packapp/__main__.py`
- Replaced `import distlib.scripts` / `import distlib.wheel` with native `Wheel` and `ScriptMaker` classes
- Updated `ensure_wheel` and `build_independent_wheel` to use `sys.executable -m pip`
- Fixed ABI tag (`cp{version}m` → `cp{version}`)

## Approach
Follows the same patterns used on `main` branch (PRs #4462, #4511) adapted for the `in-proc` branch which still uses the `build/` C# project.


build/BuildSteps.cs
```diff
@@ -376,56 +389,21 @@ public static void TestSignedArtifacts()
...
-                throw new Exception("Something went wrong while testing for signed packages. There must be a few unsigned allowed binaries");
-            }
+            // Use PowerShell Get-AuthenticodeSignature instead of external sigcheck.exe
+            var script = $@"Get-ChildItem -Path '{targetDir}' -Recurse -File | " +
+                "Where-Object { $_.Extension -match '\\.(dll|exe)$' } | " +
+                "ForEach-Object { $sig = Get-AuthenticodeSignature $_.FullName; if ($sig.Status -ne 'Valid') { $_.FullName } }";
 
-            // The first element is simply the column heading
-            unSignedPackages = unSignedPackages.Skip(1).ToList();
+            var output = Shell.GetOutput("powershell", $"-NoProfile -NonInteractive -Command \"{script}\"", ignoreExitCode: true);
```
jviau: Taking a dependency on powershell here seems odd. Is there a .NET API we can use directly?

satvu: Yes replaced with one.

tools/python/packapp/__main__.py
jviau, line 1: This is shipping production code right? This is a significant change here, I am quite concerned about possible regressions. Especially across different OS's, this has been a problematic space in the past.

Can you explain a bit more about why this file is changing so significantly? What was the CFS violation here before that is being addressed?

satvu: The changes are supposed to match this change in main: https://github.com/Azure/azure-functions-core-tools/pull/4462/changes

The violation is:

The build step `AddDistLib()` downloaded `distlib` (a Python packaging library) from `github.com/vsajip/distlib` and bundled it into `tools/python/packapp/distlib/` for each target runtime. The` __main__.py script` then imported `distlib.scripts` and `distlib.wheel` at runtime to install Python wheel packages into virtual environments when packaging Azure Functions Python apps.

jviau: Ah so this is a port of a change we have had in `main` for a while? That alleviates my concern then.

jviau:
Do we need to include #4853 in this PR as well?

satvu:
> Do we need to include #4853 in this PR as well?

Yes good catch - that change should be the only one. The other change w/ these files was a directory move.

jviau (approved):

RohitRanjanMS (approved):

PR #5549, opened 2026-08-12
Onboard dependency restores to CFS for vnext branch

satvu:
## Summary

Onboards the `vnext` branch to Compliant Feed Sources (CFS), matching the approach already applied to `main` (#5532) and `in-proc`.

### `NuGet.Config`
- **Replaced** direct `api.nuget.org` reference with the ADO `upstream-public` feed (`https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json`).
- **Removed** unused feeds that vnext doesn't consume: `infra`, `AzureFunctions`, `AzureFunctionsTempStaging`.
- **Added** `AzureFunctionsPreRelease` feed for pre-release worker/host packages.
- **Added `packageSourceMapping`** to pin packages to their expected sources:
  - `upstream-public` → catch-all (`*`) for public NuGet packages
  - `AzureFunctionsRelease` / `AzureFunctionsPreRelease` → workers + host
  - `Microsoft.Azure.Functions.PowerShellWorker` → PowerShell worker packages

### `eng/ci/templates/steps/replace-nuget-feed.yml`
- Updated the script to replace `upstream-public` → internal `upstream` feed (instead of the now-removed `nuget.org` entry).
- Added logic to propagate the source rename into `<packageSourceMapping>` so restore still works after the swap.

## What this does NOT address

Runtime network calls (CDN manifest checks, GitHub API update checks, PowerShell Gallery lookups) are product behaviors exercised at runtime, not build-time dependency acquisition. They are out of scope for CFS compliance and are not changed here.

## Validation

- ✅ `dotnet restore` — all packages resolve from the new feed configuration
- ✅ `dotnet build` — 0 warnings, 0 errors


jviau (commented):
We can remove the `replace-nuget-feed.yml` altogether and rely on only `upstream-public` plus `infra`.

NuGet.Config
```diff
@@ -2,11 +2,54 @@
...
+    <packageSource key="Microsoft.Azure.Functions.PowerShellWorker">
+      <package pattern="Microsoft.Azure.Functions.PowerShellWorker.*" />
+    </packageSource>
+    <packageSource key="AzureFunctions">
+      <package pattern="Microsoft.Azure.AppService.Middleware" />
+      <package pattern="Microsoft.Azure.AppService.Middleware.Functions" />
+      <package pattern="Microsoft.Azure.AppService.Middleware.Modules" />
+      <package pattern="Microsoft.Azure.AppService.Middleware.NetCore" />
+    </packageSource>
+    <packageSource key="AzureFunctionsTempStaging">
```
jviau, line 42: Do we need the temp staging feed? None of these packages look relevant to vnext

NuGet.Config
```diff
@@ -2,11 +2,54 @@
...
+      <package pattern="Microsoft.Azure.Functions.NodeJsWorker" />
+      <package pattern="Microsoft.Azure.Functions.PythonWorker" />
+      <package pattern="Microsoft.Azure.WebJobs.Script" />
+      <package pattern="Microsoft.Azure.WebJobs.Script.Grpc" />
+      <package pattern="Microsoft.Azure.WebJobs.Script.WebHost" />
+    </packageSource>
+    <packageSource key="Microsoft.Azure.Functions.PowerShellWorker">
+      <package pattern="Microsoft.Azure.Functions.PowerShellWorker.*" />
+    </packageSource>
+    <packageSource key="AzureFunctions">
```
jviau, line 36: Do we need this feed? None of these packages look relevant to vnext.

NuGet.Config
```diff
@@ -2,11 +2,54 @@
...
+      <package pattern="Microsoft.Azure.AppService.Proxy.Client" />
+      <package pattern="Microsoft.Azure.AppService.Proxy.Common" />
+      <package pattern="Microsoft.Azure.AppService.Proxy.Runtime" />
+      <package pattern="Microsoft.Azure.DurableTask.AzureStorage.Internal" />
+      <package pattern="Microsoft.Azure.DurableTask.Core.Internal" />
+      <package pattern="Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption" />
+      <package pattern="Microsoft.Azure.WebSites.DataProtection" />
+    </packageSource>
+    <packageSource key="Infra">
+      <package pattern="Microsoft.Azure.Functions.DotNetIsolatedNativeHost" />
```
jviau, line 52: Does vnext use the dotnet isolated host? Is this needed?

jviau (commented):
We should review the feeds here. Ultimately, I'd like to get to a point where we only have `upstream-public` and `infra`, which infra containing only the packages we don't publish publicly (like worker packages)

NuGet.Config
satvu: These are all transitive dependencies of `Microsoft.Azure.WebJobs.Script.WebHost` (4.1048.200), pulled in by the Host workload. None are directly referenced in any csproj, but NuGet needs them for restore. I confirmed with `dotnet list src/Workloads/Host/Workloads.Host.csproj package --include-transitive`:

- `Microsoft.Azure.AppService.Proxy.Client` → 2.3.20240307.67
- `Microsoft.Azure.DurableTask.AzureStorage.Internal` (transitive)
- `Microsoft.Azure.DurableTask.Core.Internal` (transitive)
- `Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption` → 1.0.5
- `Microsoft.Azure.WebSites.DataProtection` → 2.1.91-alpha

If these are upstreamed into `infra`, we could consolidate. Otherwise they're needed for restore to succeed.

NuGet.Config
satvu: Same situation — these are transitive deps of `Microsoft.Azure.WebJobs.Script.WebHost`:

- `Microsoft.Azure.AppService.Middleware` → 1.5.8
- `Microsoft.Azure.AppService.Middleware.Functions` → 1.5.8
- `Microsoft.Azure.AppService.Middleware.Modules` → 1.5.8
- `Microsoft.Azure.AppService.Middleware.NetCore` → 1.5.8

Not directly referenced in vnext, but pulled transitively by the Host workload. Without this mapping, restore fails with NU1101 because packageSourceMapping blocks them from being resolved via `upstream-public`.

Happy to consolidate into `infra` if we upstream these there.

NuGet.Config
satvu: Yes — `Microsoft.Azure.Functions.DotNetIsolatedNativeHost` (1.0.14) is a transitive dependency of `Microsoft.Azure.WebJobs.Script.WebHost`. The Host workload pulls it in. Without this mapping, restore fails with NU1101.

If `infra` is the right home for all these internal packages, I can consolidate everything under `upstream-public` + `infra` and drop the other feeds entirely.

satvu:
Re: consolidating to `upstream-public` + `infra` and removing `replace-nuget-feed.yml`:

Agreed that's the right end state. The blockers right now are that the Host workload's transitive deps (`Microsoft.Azure.WebJobs.Script.WebHost` → Middleware, Proxy, DurableTask, DotNetIsolatedNativeHost, DataProtection, Grpc) live across AzureFunctions, AzureFunctionsTempStaging, AzureFunctionsRelease, and Infra feeds.

If we upstream all of those into `infra`, I can:
1. Drop AzureFunctions, AzureFunctionsTempStaging, AzureFunctionsRelease, AzureFunctionsPreRelease, and the PowerShellWorker feeds
2. Map everything internal under `infra` with a `Microsoft.Azure.*` pattern
3. Remove `replace-nuget-feed.yml` entirely (no internal swap needed)

Would leave us with just:
```xml
<packageSources>
  <clear />
  <add key="upstream-public" value="...upstream-public..." />
  <add key="infra" value="...infra..." />
</packageSources>
<packageSourceMapping>
  <packageSource key="upstream-public">
    <package pattern="*" />
  </packageSource>
  <packageSource key="infra">
    <package pattern="Microsoft.Azure.*" />
  </packageSource>
</packageSourceMapping>
```

NuGet.Config
fabiocav: It's not used by the CLI. This is likely brough down as a transitive dependency coming from the host workload.

fabiocav: Our comments overlapped, @satvu . Yes, it's a transitive of the host. We can remove this for the CLI, but shouldn't be part of this PR.

jviau:
Okay we can at least remove the replace-nuget-feed.yml at least and let internal CI still use upstream-public

jviau (approved):

PR #3701, opened 2024-06-12
New Pipelines

khkh-ms:
### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

* [ ] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

PR #5559, opened 2026-08-17
Fix pip environment variable passthrough in linux-deb build

satvu:
## Summary

Explicitly pass `PIP_INDEX_URL` through the `env:` block in the 'Build DEB package' Bash@3 task so it survives shell profile sourcing (`bashEnvValue: '~/.profile'`).

## Changes

- `eng/ci/templates/official/jobs/linux-deb-build-pack.yml`: Added `PIP_INDEX_URL: $(PIP_INDEX_URL)` to the existing `env:` block.

## Context

The `PipAuthenticate@1` task sets `PIP_INDEX_URL` as a pipeline variable, but the Bash task sources `~/.profile` which can reset environment variables. Mapping it explicitly in `env:` ensures it's available inside the script regardless of profile contents.

### DIFF
--- a/eng/ci/templates/official/jobs/linux-deb-build-pack.yml
+++ b/eng/ci/templates/official/jobs/linux-deb-build-pack.yml
@@ -89,6 +89,7 @@ jobs:
     env:
       linuxBuildNumber: $(DefaultArtifactVersion)
       consolidatedBuildId: $(ConsolidatedBuildId)
+      PIP_INDEX_URL: $(PIP_INDEX_URL)
 
   - task: Bash@3
     displayName: "Validate DEB architectures"


eng/ci/templates/official/jobs/linux-deb-build-pack.yml
```diff
@@ -89,6 +89,7 @@ jobs:
     env:
       linuxBuildNumber: $(DefaultArtifactVersion)
       consolidatedBuildId: $(ConsolidatedBuildId)
+      PIP_INDEX_URL: $(PIP_INDEX_URL)
```
jviau, line 92: I don't think this will help. All pipeline variables are already environment variables for all tasks. So this won't set anything that isn't already set.

satvu: Yes, you're right this is already being set so this doesn't change anything about it being overriden - agent thinks the bash profile is the most likely source of the override and it was introduced w/o explanation when the pipeline was first created [here](https://github.com/Azure/azure-functions-core-tools/pull/3701/changes#diff-45fd45cc6ff7058cea2c111cee3bb5a3b97baf3b303d665748ca96e622f1b70c).

satvu:
Converting back to draft while investigating if `venv` call is the issue here.

satvu:
Closing as issue was something else.

PR #5574, opened 2026-08-27
fix(python): inherit pip indexes in build container

jviau:
## Summary

- pass `PIP_INDEX_URL` and `PIP_EXTRA_INDEX_URL` by name to Python native-dependency containers
- preserve authenticated package feed routing without exposing credential-bearing values in Docker command output
- add focused tests for Docker run argument generation

## Motivation

The `core-tools.public` Python E2E test `Pack_Python_BuildNativeDeps_OnWindows_WorksAsExpected` failed under network isolation because the containerized `pip install` contacted `pypi.org` directly. The host had been configured by `PipAuthenticate`, but those index variables were not inherited by the Docker container.

## Validation

`dotnet test test\Cli\Func.UnitTests\Azure.Functions.Cli.UnitTests.csproj --framework net10.0 --no-restore --filter FullyQualifiedName~DockerHelpersTests`


GitPaulo (approved):

brettsam (approved):

kshyju (commented):
## Review Findings

**Overall: Looks good — clean, well-tested change.** A couple of observations:

### 1. `--env` argument quoting (low risk)
In `GetDockerRunArguments`, env var names are interpolated directly into `args.Add($"--env {environmentVariable}")`. Since only bare variable *names* (`PIP_INDEX_URL`, `PIP_EXTRA_INDEX_URL`) are passed by the current caller, this is safe today. But if a future caller passes a value containing spaces or shell metacharacters (e.g. `VAR=hello world`), the argument would be split by the docker CLI. Consider adding each as two separate list entries (`args.Add("--env"); args.Add(environmentVariable);`) — though this depends on how `RunDockerCommand` tokenizes the string. Not blocking.

### 2. `dockerRunSetting` branch doesn't get env vars
When `dockerRunSetting` is non-empty (PythonHelpers.cs line 514), the custom docker run command is used as-is — the `PIP_*` env vars won't be forwarded. This is pre-existing behavior, but worth a note in the PR description since users with custom docker run settings would need to add `--env PIP_INDEX_URL --env PIP_EXTRA_INDEX_URL` themselves.

### 3. Tests are solid ✅
The two unit tests cover the happy path and the no-args path well. The extracted `GetDockerRunArguments` method is a nice improvement for testability.

**No blocking issues found.**

kshyju (approved):
Do we need to update release notes?

kshyju (approved):

PR #5584, opened 2026-09-10
Configure NuGet for in-proc E2E projects

satvu:
## Summary

- configure an authenticated user-level NuGet configuration for temporary E2E projects
- remove repository-specific package source mappings from that user configuration
- stop copying the repository `NuGet.Config` into every E2E working directory

## Context

The in-proc E2E suite runs `func init`, which invokes `dotnet new` in a temporary project directory. Copying the repository CFS configuration into that directory overrides the authenticated user configuration and can cause the nested template update request to hang until the outer `func.exe` command times out.

This backports the targeted NuGet setup used by the `main` branch without changing normal Core Tools behavior or migrating the full E2E framework.


satvu:
Closing because this draft was created with the wrong base/head refs; replacing it with an in-proc-targeted draft.

PR #5585, opened 2026-09-10
Configure NuGet for in-proc E2E projects

satvu:
## Summary

- configure an authenticated user-level NuGet configuration for temporary E2E projects
- remove repository-specific package source mappings from that user configuration
- stop copying the repository `NuGet.Config` into every E2E working directory
- verify CFS template update checks explicitly in an isolated template hive

## Context

The in-proc E2E suite runs `func init`, which invokes `dotnet new` in a temporary project directory. Copying the repository CFS configuration into that directory overrides the authenticated user configuration and can cause the nested template update request to hang until the outer `func.exe` command times out.

This backports the targeted NuGet setup from [`eng/ci/templates/steps/authenticate-e2e-dependencies.yml` on `main`](https://github.com/Azure/azure-functions-core-tools/blob/main/eng/ci/templates/steps/authenticate-e2e-dependencies.yml) without changing normal Core Tools behavior or migrating the full E2E framework. The smoke step installs the Functions project template into a temporary hive and runs `dotnet new update --check-only`, guaranteeing the CFS update-query path is exercised independently of template cache state.


satvu:
Superseded by a focused draft using a CI-only --no-update-check workaround.

PR #5594, opened 2026-09-16
refactor: extract focused setup services

ahmedmuhsin:
Part of #5451. Extracts setup feature/profile resolution, dependency planning and installation without changing fixed mappings or RID-pointer behavior.

Immediate stack #5606 is #5594 followed by #5604. Discovery in #5595 and search enrichment in #5605 remain deferred drafts. Metadata discovery and the issue's class-count wording remain open acceptance items.

Validation at 020b6468: SDK 10.0.401 CI-strict Release build, zero warnings/errors, 2,339 passing tests. [Exact-head CI](https://dev.azure.com/azfunc/ae7e3bf3-d41a-4480-9ac0-b6cf9df9ac24/_build/results?buildId=304953).

PR #5604, opened 2026-09-19
fix: preserve setup automation and lifecycle contracts

ahmedmuhsin:
Fixes setup automation on top of #5594 in stack #5606. Preserves physical NDJSON, suppresses advisories for check/JSON mode, skips the picker for --yes or JSON, protects first-run state in check mode, and observes cancellation through terminal rendering and marker completion. Parser-owned version/help/completion actions retain their dispatch behavior.

Keeps existing dependency-first partial installation. Discovery, custom-template ownership, profile-wide preflight and search enrichment are not included.

Validation at 108abe35: SDK 10.0.401 CI-strict Release build, zero warnings/errors, 2,517 passing tests. The new terminal-boundary matrix changed from 14 failures and 22 passing controls to all 36 passing.


proposed/func-setup-design.md
```diff
@@ -1,28 +1,31 @@
 # `func setup` Design (Draft)
 
-This document describes the proposed `func setup` command for the Azure Functions CLI (`func`). It is a working draft.
```
satvu, line 3: Noting for other reviewers/tracking, most of these changes are to reflect existing behavior that existed before this PR (and the one in stack below it).

src/Func/Commands/Setup/SetupFeatureResolver.cs
```diff
@@ -125,7 +125,7 @@ [.. workerRuntimes.OrderBy(static runtime => runtime, StringComparer.OrdinalIgno
             return [configuredStack.Trim()];
         }
 
-        if (!options.NonInteractive && _interaction.IsInteractive)
+        if (!options.NonInteractive && !options.AssumeYes && options.OutputMode != SetupOutputMode.Json && _interaction.IsInteractive)
```
satvu, line 128: Nonblocking: when `--yes` skips the picker and no project runtime is configured, could we print a plain-output hint explaining that setup defaults to host + extension bundle, with `func setup --yes --features <stack>` for language setup? Otherwise, the final success message may suggest the language environment is ready. Keep JSON output unchanged.

satvu (approved):

satvu (approved):

PR #5614, opened 2026-09-21
Fix RID-specific Host workload package layout

satvu:
## Summary
- pack RID-specific Host payloads under `tools/<rid>` instead of `tools/any`
- bump the Host workload to `4.1048.200-preview.3`
- correct Host packaging guidance and release-note versions

## Release impact
This is an intentional Host workload behavior fix. Publish the logical pointer package and every RID implementation after merge. No CLI, worker, or stack release is required.

## Validation
- `dotnet restore`
- `dotnet test test\Workloads\Host.Tests\Workloads.Host.Tests.csproj -c Release --no-restore` (50 passed)
- release-mode `dotnet pack` with `PackAllRids=true` produced the logical package plus six RID packages without warnings
- verified the win-x64 nupkg contains `tools/win-x64/Azure.Functions.Cli.Workloads.Host.exe` and no `tools/any` payload
- released CLI `5.0.0-preview.3` installed Host `4.1048.200-preview.3` from disposable BaGet, selected the win-x64 implementation, started the Host, and returned HTTP 200 with state `Running`


satvu:
> ## Copilot review overview
> **Previously missed (1)**
> In code that hasn't changed since last review
> 
> Medium severity DeployForDebug still copies Host output to stale tools/any path

I traced through `DeployForDebug`, and it looks like it doesn’t actually copy the Host output to `tools/any` today. It only runs for `FuncCliWorkload`; the Host either has no package type during a normal build or is `FuncCliWorkloadRidPackage` for a RID build. It also has `IncludeBuildOutput=false`, so there wouldn’t be a payload to copy through this path anyway.

**The Host does not use this workflow.** Host local iteration uses `FUNC_HOST_CONTENT_ROOT`, pointing the CLI directly at the freshly built Host output. That is why the review’s `DeployForDebug` concern does not affect this Host packaging fix.

fabiocav (approved):

